PR #37, opened 2019-11-05
Fixes #16: basic feed parser

### LINKED ISSUE #16: Develop a worker to parse contents from the blog's url
We should create a worker that would implement the following features:
- Get contents from particular blog post url
- Parse the contents into the JSON format
- Pass the generated JSON file over

Receives: url
Returns: JSON file

c3ho:
Fixes #16 

Hi, this is a extremely basic feed-parser that will just spit out all the posts from the feed provided to console.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,8 @@
   "dependencies": {
     "bent": "^7.0.2",
     "bull": "^3.11.0",
-    "dotenv": "^8.2.0"
+    "dotenv": "^8.2.0",
+    "feedparser-promised": "^2.0.1"
   },
   "devDependencies": {
     "eslint": "^6.6.0",
--- a/src/feed-parser.js
+++ b/src/feed-parser.js
@@ -1,0 +1,7 @@
+const feedparser = require('feedparser-promised');
+
+// feed: url for the feed
+// Accepts a url feed and parses it, on success will return an array
+module.exports = async function (feed) {
+  return feedparser.parse(feed);
+};


c3ho:
Tested with both types of feed:
Atom 1.0: https://blogname.blogspot.com/feeds/posts/default
RSS 2.0: https://blogname.blogspot.com/feeds/posts/default?alt=rss

I wrote a simple test.js file using the code below which you can use to play around.
```
const feedParser = require('./feed-parser.js');
const feed = feedParser('https://c3ho.blogspot.com/feeds/posts/default/-/open-source');
console.log(feed);
```

At the moment in the feed-parser.js file I put down console.log(items) , but it can also tracks metadata which you can check out with console.log(meta). The list is pretty exhaustive and can be found at the bottom of this document [here](https://www.npmjs.com/package/feedparser)

ImmutableBox:
@c3ho If you need any help, you can delegate some issues to me 👍 .

lucacataldo:
Just a note but some issues can be alleviated and code simplified by using a promise-wrapped version of feedparser found at [feedparser-promised](https://www.npmjs.com/package/feedparser-promised) on npm. Might be worth looking into as it'll do a lot of the work for us here.

humphd:
@lucacataldo, we could use it, but if you look at the code, it's essentially what I'm suggesting above, and then we don't add another dep. https://github.com/alabeduarte/feedparser-promised/blob/master/src/feedParserPromised.js.

@c3ho let's move this PR forward ASAP, as it's blocking the rest of the work we need to do.  Maybe we can discuss in class today.

package.json
```diff
@@ -13,7 +13,16 @@
   },
   "homepage": "https://github.com/Seneca-CDOT/telescope#readme",
   "dependencies": {
+    "addressparser": "^1.0.1",
     "bent": "^7.0.2",
-    "bull": "^3.11.0"
+    "bull": "^3.11.0",
+    "feedparser": "^2.2.9",
+    "lodash.assign": "^4.2.0",
```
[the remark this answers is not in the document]
birtony: I can take this on.

c3ho:
@humphd Since I'm not using request anymore and am using bent now I have to change a bit of the code. Mainly, I'm having an issue piping the data to feedparser(or rather I don't know how). Any tips?

humphd:
@c3ho this needs to be commits on the PR, not a comment.  Please submit this the appropriate way so that we can work through the code together.

RE: bent vs. request, please [see the docs](https://github.com/mikeal/bent#usage), which discuss how to get a stream like request does:

```js
const bent = require('bent')
const getStream = bent('http://site.com')
let stream = await getStream('/json.api')
```

Regarding the way you're doing the `Promise` code, this won't work.  Your function has to return a single `new Promise`:

```js
return new Promise((resolve, reject) => {
 
});
```

The `resolve` and `reject` function args are what you use vs. `Promise.resolve()` and `Promise.reject()`.  If you look at the link I provided above, it shows exactly how to do this https://github.com/alabeduarte/feedparser-promised/blob/master/src/feedParserPromised.js.

package.json
```diff
@@ -13,7 +13,11 @@
   },
   "homepage": "https://github.com/Seneca-CDOT/telescope#readme",
   "dependencies": {
+    "addressparser": "^1.0.1",
     "bent": "^7.0.2",
-    "bull": "^3.11.0"
+    "bull": "^3.11.0",
+    "feedparser": "^2.2.9",
+    "mri": "^1.1.4",
```
humphd: Remove this please

package.json
```diff
@@ -13,7 +13,16 @@
   },
   "homepage": "https://github.com/Seneca-CDOT/telescope#readme",
   "dependencies": {
+    "addressparser": "^1.0.1",
```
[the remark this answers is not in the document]
humphd: This is still outstanding, please remove.

src/feed-parser.js
```diff
@@ -0,0 +1,20 @@
+const FeedParser = require('feedparser');
+const bent = require('bent'); // for fetching the feed
+
+module.exports = async function(feed){
+    const feedparser = new FeedParser();
+    const getStream = bent(feed);
+
+    let stream = await getStream('/json.api');
+    return new Promise((res,rej)=>{
```
humphd: Please use `resolve` and `reject` (full words).

src/feed-parser.js
```diff
@@ -0,0 +1,20 @@
...
+        stream.pipe(feedparser)
+            .on('error', rej)
+            .on('readable', function(){
+                // unable to get individual posts
```
humphd: What does this mean?

src/feed-parser.js
```diff
@@ -0,0 +1,20 @@
...
+                var stream = this;
```
humphd: You have already defined `stream` on line 8, this isn't necessary.

src/feed-parser.js
```diff
@@ -0,0 +1,20 @@
...
+                var meta = this.meta;
+                res(meta);
+            }
+        )
+    });
+}
```
humphd: Newline at the end, please.

package.json
```diff
@@ -13,7 +13,11 @@
...
     "bent": "^7.0.2",
-    "bull": "^3.11.0"
+    "bull": "^3.11.0",
+    "feedparser": "^2.2.9",
+    "mri": "^1.1.4",
+    "readable-stream": "^3.4.0"
```
humphd: This also seems unused.  Make sure you don't add dependencies that you don't use.

src/feed-parser.js
```diff
@@ -0,0 +1,20 @@
...
+
+module.exports = async function(feed){
+    const feedparser = new FeedParser();
+    const getStream = bent(feed);
+
+    let stream = await getStream('/json.api');
+    return new Promise((res,rej)=>{
+        stream.pipe(feedparser)
+            .on('error', rej)
+            .on('readable', function(){
```
humphd: Change this to just be `on('readable', resolve);` and return the entire `item` that's being read.

c3ho:
@Luca Cataldo in the end we ended up using feedparser-promised as you suggested. Thank you!

Issue #32, opened 2019-11-05
Add ESLint to lint all our JS/JSON


humphd:
We should really have a proper automated linting step using [ESLint](https://eslint.org/).  You can see good docs on how to do this at https://eslint.org/docs/user-guide/getting-started, and look at how lots of other projects do their linting by reading their `.eslintrc` file(s).

Until we get automated tests written, this can serve as our `npm test` step, and let us start automating CI in other bugs.

kartik-budhiraja:
We can use [Airbnb Javascript style](https://github.com/airbnb/javascript) guide as starting base and extend it. The eslint config is available through npm package `eslint-config-airbnb-base`.

UltimaBGD:
In the event Airbnb style is something we are interested in, I will quickly put up a pull request with it setup for people to test with.

UltimaBGD:
The pull request has been merged, shall we close this issue or keep it up for further discussion?

humphd:
Close.  Also, we should make sure that people use the "Fixes <issue number>" style in their PRs so GitHub will auto-close issues fixed by PRs.

Thanks for working on this!

PR #38, opened 2019-11-05
Add ESLint with Airbnb style

UltimaBGD:
This is a basic setup of ESLint with the Airbnb style guide as a preset.

This was referenced in #32, and this pull request could serve as a simple test for ESLint and what we think of it.

For now, the main additions and changes consist of some formatting changes to the code that already existed, so the npm test command, which now runs ESLint, passes with no errors. (There are still some warnings.)

The styles and specific rules of the NPM tests can be changed and talked about in other issues, but for now this is a simple base.


.eslintrc.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+    "extends": "airbnb-base"
+  };
```
humphd: Watch your indent here (should match with `m` on line 1), and also include a newline at the end of line 3.

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.1",
   "description": "A tool for tracking blogs in orbit around Seneca's open source involvement",
   "scripts": {
-    "test": "echo \"TODO!\" && exit 1",
+    "test": "eslint --ignore-path .gitignore .",
```
humphd: A couple things here:

1. Create a separate script for eslint, and then call it from test:

    ```
    "test": "npm run eslint"
    ```
2. Let's also add an eslint-fix script, to automatically fix the obvious/automatically correctable issues in code:
    ```
    ```

src/index.js
```diff
@@ -8,42 +8,44 @@ const feedWorker = require('./feed-worker');
...
   }
 
   // Process this text file into a list of URL jobs, and enqueue for download
   const feedJobs = processFeedUrls(lines);
-  enqueueFeedJobs(feedJobs);  
+  enqueueFeedJobs(feedJobs);
 
   // Start working on the queue
   feedWorker.start();
+  //Returning true if no other errors were encountered
```
humphd: I don't understand this addition.  Nothing is consuming this return value, it's just a callback.  If airbnb is requiring this, we may want to override that.

UltimaBGD: It was airbnb requiring it, yes.

UltimaBGD: Oh, actually it's standard ESLint.

Viewable here: https://eslint.org/docs/rules/consistent-return

UltimaBGD: I believe it is throwing an error because of the "return process.exit(-1);"

There is only one path to that return, which is counting as odd to ESLint.

package.json
UltimaBGD: Should be resolved in new version

.eslintrc.js
UltimaBGD: Should be resolved.

src/index.js
UltimaBGD: This is the only thing unresolved, as we now know why it is erroring and I thought could use some further commentary. I can still disable the rule if you wish.

humphd: What if you modify my code so it does this:

```js
...
    process.exit(-1);
    return;
}

...
```

And don't do a return at the end.  Is it happy with that?  The extra return here might be making the linter happy, but the code doesn't make as much sense.  Let me know, and maybe we'll disable this rule.

UltimaBGD: It is happy with that, as it technically returns no value.

UltimaBGD: Adjusted and updated.

src/index.js
```diff
@@ -8,41 +8,42 @@ const feedWorker = require('./feed-worker');
...
 }
 
 /**
  * Adds feed URL jobs to the feed queue for processing
- * @param {Array[Object]} feedJobs - list of feed URL jobs to be processed 
+ * @param {Array[Object]} feedJobs - list of feed URL jobs to be processed
  */
-async function enqueueFeedJobs(feedJobs) {
-  for(let feedJob of feedJobs) {
+function enqueueFeedJobs(feedJobs) {
```
humphd: Does this not also need to be `async`, since we are `await`ing within the loop?

UltimaBGD: Well, the loop is in another function that is async, where it's then awaited. I don't know if each function that uses that function needs to be async though. It does work either way.

UltimaBGD: I think it worked without the async at the top simply because there was nothing after to run asynchronously.

humphd: If you `await` in a function, then you are necessarily in an `async` function, and need to decorate your function with the `async` keyword.  So yes, adding this back would be good.

UltimaBGD: It is added back already.

UltimaBGD:
It was intentional as a form of separation between the requires and other code. I believe it was changed by running and testing the —fix command.

UltimaBGD:
In which case, that is two approving reviews, so I shall merge it.

PR #46, opened 2019-11-05
Fixes #15: Added new file sentiment-analysis.js. Added dependencies

### LINKED ISSUE #15: Implement sentiment analysis of blog posts
Sentiment analysis can be helpful in determining how negative or positive a blog post is. It will be also helpful to flag blogs which use contain negative wording.

I would like to work on this feature.

jatinAroraGit:
Fixes #15

The pull request will add the feature of analyzing sentiment of a blog post with the help of a node package called [sentiment](https://www.npmjs.com/package/sentiment). 
Let me know if there are any issues with it.


humphd:
@jatinAroraGit can you rebase this please?

src/sentiment-analysis.js
```diff
@@ -0,0 +1,19 @@
+/* This file contains the code for analyzing blog posts text to identify the
+ * negative or positve words being used in a post and return a summary of it
+ * along with a score. The file uses a node module called sentiment to implement
+ * the functionality of analyzing text of blogs. The function accepts striped
```
ImmutableBox: Typo at: "The function accepts striped" should be "The function accepts stripped"
Otherwise looks good 👍 , need to rebase to match the current master branch.

jatinAroraGit: Hey @ImmutableBox ,
Resolved this in commit https://github.com/Seneca-CDOT/telescope/pull/46/commits/259b31c11f926423304437065cecd538abd8b11e

src/sentiment-analysis.js
```diff
@@ -0,0 +1,19 @@
+/* This file contains the code for analyzing blog posts text to identify the
+ * negative or positve words being used in a post and return a summary of it
+ * along with a score. The file uses a node module called sentiment to implement
+ * the functionality of analyzing text of blogs. The function accepts stripped
+ * HTML text as parameters i.e text containg no tags, and returns a promise
```
humphd: What does "stripped HTML" mean?  Do you want plain text or HTML?  If the former, I'd say that instead.

jatinAroraGit:
Hey there,
The branch is now up to date with the upstream/master.

src/sentiment-analysis.js
```diff
@@ -0,0 +1,19 @@
...
+ * object whcih contains the result
+*/
+
+var Sentiment = require('sentiment');
```
humphd: Prefer `const` to `var`

src/sentiment-analysis.js
```diff
@@ -0,0 +1,19 @@
...
+var sentiment = new Sentiment();
```
humphd: Prefer `const` here too, and also, let's move this into your function below so we don't reuse this between runs.  If there's any shared state between runs, we'll get into weird bugs.

src/sentiment-analysis.js
```diff
@@ -0,0 +1,19 @@
...
+
+module.exports.startAnalysys = function(blogText){
```
humphd:
```suggestion
module.exports.run = async function(text) {
```

src/sentiment-analysis.js
```diff
@@ -0,0 +1,19 @@
...
+  var result;
```
humphd: This can be simplified bit:

```js
return Promise.resolve(sentiment.analyze(text));
```
Since this call doesn't throw, we just need to wrap it in a call to `resolve`

jatinAroraGit:
Hey @humphd ,
Let me know if https://github.com/Seneca-CDOT/telescope/pull/46/commits/8ef394e1a14fa2d2522d8ed15b9d85de7a604cf3 resolves the comments.

ImmutableBox:
Sorry @jatinAroraGit going to have to rebase because of https://github.com/Seneca-CDOT/telescope/pull/55 was merged. Should only need to fix `package.json` file.

src/sentiment-analysis.js
```diff
@@ -0,0 +1,13 @@
+/* This file contains the code for analyzing blog posts text to identify the
+ * negative or positve words being used in a post and return a summary of it
+ * along with a score. The file uses a node module called sentiment to implement
+ * the functionality of analyzing text of blogs. The function accepts plain
+ * text as a parameter i.e text containg no HTML tags, and returns a promise
+ * object whcih contains the result.
+*/
+
+module.exports.run = async function (text) {
+  const Sentiment = require('sentiment');
```
humphd: This line needs to go above: we only want to `require` this once, not every call to `run()`.  The next line, where you create an instance, is what I wanted in here.

humphd: This is still not fixed.  Move line 10 out of this function, and above like you had it before.

Issue #51, opened 2019-11-06
add catch after exports.start in feed-worker.js


robertbegna:
I would like to add a catch after this function so that any errors can be dealt with

Reza-Rajabi:
Would you like to take a look at [this](https://javascript.info/async-await) and decide about how would you like to approach this issue?

vitokhangnguyen:
I can work on this issue and at the same time, switch all Promise-based then() catch() to async/await. Does that sound good?

Reza-Rajabi:
@robertbegna I am closing this issue since the related pull request is closed.
Please reopen it again if this is a mistake from my side.

PR #52, opened 2019-11-06
Issue 51  added a simple catch to the end of the exports.start function to handle some errors

robertbegna:


robertbegna:
yeah i forgot i installed it to address another issue i thought i might tackle, but then i chose to this instead. But yeah I've removed the dependency

robertbegna:
hi @Cadesh i've rebased it so could you take some time to review. Thanks!

birtony:
@robertbegna if this PR fixes #51, could you, please, update the title and the description accordingly?
- The _title_ and the _description_ should probably start with `Fixes #51` 
- `Fixes` is not recognized as a keyword in the _title_, however, it is convenient for other contributors to see what is the PR about right in [the list of all PRs](https://github.com/Seneca-CDOT/telescope/pulls)
- Mentioning `Fixes` in the _description_ will allow us to automatically close the related issue once the PR is merged
- For reference, check https://github.com/Seneca-CDOT/telescope/pull/66, as an example. When we go to https://github.com/Seneca-CDOT/telescope/issues/20 we can see that it `May be fixed by #66`

package-lock.json
```diff
@@ -4815,9 +4762,9 @@
       "dev": true
     },
     "rimraf": {
-      "version": "2.7.1",
-      "resolved": "https://registry.npmjs.org/rimraf/-/rimraf-2.7.1.tgz",
-      "integrity": "sha512-uWjbaKIK3T1OSVptzX7Nl6PvQ3qAGtKEtVRjRuazjfL3Bx5eI409VZSqgND+4UNnmzLVdPj9FqFJNPqBZFve4w==",
+      "version": "2.6.3",
```
Reza-Rajabi: has been overrides by a lower version

src/feed-worker.js
```diff
@@ -7,8 +7,10 @@ exports.start = function () {
   // Start processing jobs from the feed queue...
   feedQueue.process(async (job) => {
     const { url } = job.data;
+
     console.log(`Processing job - ${url}`);
     // For now, just get the feed data and dump to the console
     return request(url).then((data) => console.log(`${data} + '\n\n'`)).catch((err) => { throw err; });
   });
```
Reza-Rajabi: It looks like the last version has been overridden once because the original version was this:
```javascript
return request(url).then((data) => console.log(`${data}\n\n`));
```
and actually there is nothing wrong with it. So I suggest to not change this function. If you want to catch from async function you need to do in try and catch block rather than a promise like catch.

Issue #42, opened 2019-11-05
Add testing infrastructure


humphd:
We're starting to get some code in the repo, and it's time to think about how we'll test all the features we add.

I'd like to see us get an initial test in place, so we can figure out our infrastructure, and let others add more tests in future PRs.

Let's use [Mocha](https://mochajs.org/) or [Jest](https://jestjs.io/) to automate our tests.  At first, we need a simple `npm test` run that runs one test against something like the feed parser function, or something else really basic: call into a function and it expects to get back a given result.

ImmutableBox:
I would like to try my hand at this. After some research/testing I decide to use [Jest](https://jestjs.io/). My reasoning: 
- [Jest ](https://jestjs.io/)setup was easier to setup.
- [Jest ](https://jestjs.io/)sent a warning message about problems with async operations while [Mocha](https://mochajs.org/) did not.

[Mocha](https://mochajs.org/) does seem to be more flexible, but the setup takes some time. Maybe in later development there can be a switch, but for a starting point ill be using [Jest ](https://jestjs.io/).

ImmutableBox:
Merged with https://github.com/Seneca-CDOT/telescope/pull/55.

PR #55, opened 2019-11-06
Adding testing infrastructure

ImmutableBox:
This pull request sets up the testing infrastructure using [Jest](https://jestjs.io/) .

To run the test : `npm test`
I only created one simple test as a starting point. The test does check if the Redis server is open.

**NOTE**: `npm test` will most likely not end correctly with error:
"Jest did not exit one second after the test run has completed. This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue" 
 
(CTRL + C to exit test.)

This is because the Redis server is still online, closing the server after testing would fix the issue.
If there is a way to handle opening and closing Redis inside the project, this problem won't happen anymore <- CREATE NEW ISSUE TO SOLVE THIS

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
     "eslint": "eslint --ignore-path .gitignore .",
     "eslint-fix": "eslint --fix --ignore-path .gitignore .",
     "test": "npm run eslint",
+    "jest": "jest",
     "start": "node src"
   },
   "repository": "Seneca-CDOT/telescope",
@@ -21,6 +22,7 @@
   "devDependencies": {
     "eslint": "^6.6.0",
     "eslint-config-airbnb-base": "^14.0.0",
-    "eslint-plugin-import": "^2.18.2"
+    "eslint-plugin-import": "^2.18.2",
+    "jest": "^24.9.0"
   }
 }
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,3 +1,6 @@
 module.exports = {
   "extends": "airbnb-base",
+  "env": {
+    "jest": true
+  }
 };
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,108 @@
         "@babel/highlight": "^7.0.0"
       }
     },
+    "@babel/core": {
+      "version": "7.7.2",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.7.2.tgz",
+      "integrity": "sha512-eeD7VEZKfhK1KUXGiyPFettgF3m513f8FoBSWiQ1xTvl1RAopLs42Wp9+Ze911I6H0N9lNqJMDgoZT7gHsipeQ==",
+      "dev": true,
+      "requires": {
+        "@babel/code-frame": "^7.5.5",
+        "@babel/generator": "^7.7.2",
+        "@babel/helpers": "^7.7.0",
+        "@babel/parser": "^7.7.2",
+        "@babel/template": "^7.7.0",
+        "@babel/traverse": "^7.7.2",
+        "@babel/types": "^7.7.2",
+        "convert-source-map": "^1.7.0",
+        "debug": "^4.1.0",
+        "json5": "^2.1.0",
+        "lodash": "^4.17.13",
+        "resolve": "^1.3.2",
+        "semver": "^5.4.1",
+        "source-map": "^0.5.0"
+      },
+      "dependencies": {
+        "semver": {
+          "version": "5.7.1",
+          "resolved": "https://registry.npmjs.org/semver/-/semver-5.7.1.tgz",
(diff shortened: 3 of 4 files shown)


cindyledev:
I ran this test on my machine and it worked great. My first test failed because I was trying to open redis in a port that was already occupied. I killed the instance using the port and tried again and the second test was successful as expected.
Annotation 2019-11-06 153302

cindyledev:
I just realized there's already a npm test script in the package.json file, be careful when you merge that you don't delete the test script that's already there

ImmutableBox:
> 
> 
> I just realized there's already a npm test script in the package.json file, be careful when you merge that you don't delete the test script that's already there

Yup. Thanks, Ill rebase this pull request later today when I have time.

ImmutableBox:
@UltimaBGD I had to merge your ESLint pull request with my Jest testing infrastructure pull request. It would be great if you can look it over and give an approval. I changed `npm test` to call Jest instead of ESLint. ESLint can still be called with `npm run eslint` and `npm run eslint-fix`.

humphd:
@ImmutableBox what do you mean by "I had to merge with your eslint pr"?  You don't want to do that. If you need to test, that's fine locally, but you should rebase on it when it gets landed.

humphd:
To confirm: that PR has landed, so you can just rebase on `master`, don't merge.

ImmutableBox:
> 
> 
> To confirm: that PR has landed, so you can just rebase on `master`, don't merge.

Yes, I rebase to master. I then used `git merge master` with the branch I was working on and fixed merging conflicts on `package.json` and 'package-lock.json'. When I ran `npm run eslint-fix` it seemed to fix the formatting for all the other files.

This is what my `git log` looks like:
image

Where `Adding testing infrastructure #42` is my first commit, `Add ESLint with Airbnb style` is what's currently on master, and the commit I just uploaded.

If this is wrong I can do it again.

humphd:
OK, so this is wrong.  You've done a merge, and we don't want to merge while on a PR branch, only rebase.  You want to update your commits so they are in line with the latest `master`, but not merge with it.  The merge only happens when the PR gets approved and landed.

We can fix this together in class tomorrow if you want, or you can do it yourself.  Basically, you want to rebase your `8b5731bf61ce781088a83ebcd522638527824d87` commit on `master`, and fix the conflicts to produce something similar to what your merge does.  Then you need to `git push origin issue42 -f`.

UltimaBGD:
You should be able to run Jest and ESlint in the same command with the "&&" operators, is there a reason why ESlint had to be taken out of the test?

ImmutableBox:
> 
> 
> You should be able to run Jest and ESlint in the same command with the "&&" operators, is there a reason why ESlint had to be taken out of the test?

No, not really just following this [blog](https://medium.com/asos-techblog/javascript-quality-control-with-jest-and-eslint-b888eeb4b0c4) he had it separated. Ill add it back in.

humphd:
@UltimaBGD the `&&` operator won't work on all shells (e.g., many Windows shells).  There are few work arounds for this:

1. we can use `pre` and `post` scripts to run things before/after other steps, https://docs.npmjs.com/misc/scripts
2. we could use https://www.npmjs.com/package/npm-run-all to allow for a set of steps

In general, we have to be careful that our scripts work on Windows, since so many common tricks like this will fail :(

UltimaBGD:
Okay, I think a pre script would make sense for this. How do we want to go about adding that in? Get this merged as it is and then I make another PR with the prescript?

humphd:
Given that nothing is depending on these tests yet, or even eslint for that matter, let's get this rebased (package* files need updating) and then do more PRs on top.

UltimaBGD:
Alright, I will keep my eye on it and work on that after.

ImmutableBox:
> 
> 
> Given that nothing is depending on these tests yet, or even eslint for that matter, let's get this rebased (package* files need updating) and then do more PRs on top.

I think I got it down.

`git log`:
image

I think I still have to make another commit or rebase, since `Jest` doesn't work anymore.

humphd:
Yes, that looks right, good work.  Just do whatever you have to in order to get Jest to work again, and then do `git commit --amend --no-edit` and update your current commit, then `git push origin issue42 -f` when you are ready to have us review.

ImmutableBox:
OK. Jest is now working with `npm test`, ill create a new issue for running both ESLint and Jest together with the comments provided.

package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "eslint": "eslint --ignore-path .gitignore .",
     "eslint-fix": "eslint --fix --ignore-path .gitignore .",
-    "test": "npm run eslint",
+    "test": "jest",
```
humphd: Leave the `test` script alone, and instead add one for `jest`:

```json
"test": "npm run eslint",
"jest": "jest",
```

In the new bug you file we can combine these via a call to `"pretest": "npm run eslint"` and change `test` to run `jest`

ImmutableBox: Fixed in https://github.com/Seneca-CDOT/telescope/commit/c3d8a6a55d1ba26c4180f3a2ac7e48247062f97c.

PR #66, opened 2019-11-07
Fixes #20: Added basic summarize functionality

### LINKED ISSUE #20: Implement feature to generate summary of the blog post

ThomasLuu00:
Fixes #20

Added a summarizer that uses [gensim's TextRank Summarizer](https://radimrehurek.com/gensim/summarization/summariser.html) in python. 

 I included code that allows it to be called from the node.js framework just to test it and to make it easier for whoever wants to call the function.  

It isn't good, but it's a starting point.  I will continue working on a document summary model.


birtony:
Fixes #20

summarizer/summarize.js
```diff
@@ -0,0 +1,8 @@
+const { SummarizerManager } = require('node-summarizer');
+
+module.exports = async function (text, numSentences = 3) {
+  const Summarizer = new SummarizerManager(text, numSentences);
```
humphd: `const summarizer = ...`

Use lower case to start a variable name, only use a capital for a Constructor function name.

summarizer/summarize.test.js
```diff
@@ -0,0 +1,16 @@
+const fs = require('fs');
+const summarize = require('./summarize');
+
+describe('summarize tests...', () => {
```
humphd: We need to convert this to Jest format (i.e., `test()` vs. `describe()`)

summarizer/summarize.test.js
```diff
@@ -0,0 +1,16 @@
...
+  const text = fs.readFileSync('./summarizer/text.txt', 'utf-8');
```
humphd: This should go in a `before` test setup call.  You can also drop the `readFileSync` and just use `readFile`, since we can use async calls in a test setup function.

summarizer/summarize.test.js
```diff
@@ -0,0 +1,16 @@
...
+  var data;
+
+  it('summarize returns string', async () => {
+    data = await summarize(text, 5);
+    expect(typeof data).toBe('string');
```
humphd: Should we also test for the exact summary?

ThomasLuu00: I think it's a good idea, but I'm am not certain I have the best approach to it. I think I could compare how similar the machine summary is to a human summary and have it pass if it meets a threshold, but this requires a good human summary for each blog input.

summarizer/summarize.test.js
ThomasLuu00: I am unable to next tests in Jest, I use describe to define a scope just in case I wish to add other types of test cases into the file.  It isn't really needed at the moment so I could just remove describe,.

summarizer/summarize.test.js
humphd: What I mean is: run your code once, get the summary it produces, see if you think it makes sense.  If it does, hard code that into your test, so we know in the future if it breaks for this given blog post text.  This is really a regression test, to guard against future problems.

ThomasLuu00: Alright, I added the test.

.stylelintrc.js
```diff
@@ -6,7 +6,8 @@ module.exports = {
     "LICENSE",
     "*.*",
     "images/*.*",
-    "!*.css"
+    "!*.css",
+    "summarizer/*.txt"
```
humphd: This change makes me think we need to file a new issue to adjust how our `stylelint` works, such that it only looks at files in a particular directory, maybe `src/frontend/styles/*` or something, and then we don't collide with random files like this.

summarizer/summarize.test.js
```diff
@@ -0,0 +1,23 @@
+const fs = require('fs').promises;
```
humphd: Let's move this into `test/` where the other tests are.

.stylelintrc.js
ThomasLuu00: Posted as issue #162.

ThomasLuu00: I closed the issue as duplicate. Did not see the issue you posted.

humphd: As you know, this is being fixed in #151, so why don't you reset your changes to this file, so you don't have merge conflicts with that:

```
git checkout master .stylelintrc.js
```

And you can go back to what it was.  It will fail Travis, but we can ignore.  By the way, you might want to add another review to #151 so we can land it.

test/summarize.test.js
```diff
@@ -0,0 +1,23 @@
...
+const summarize = require('../summarizer/summarize');
+
+
+describe('summarize tests...', () => {
+  let inputData;
+  let predicted;
+  let expected;
+
+  beforeAll(async () => {
+    inputData = await fs.readFile('./summarizer/input.txt', 'utf-8');
```
humphd: Move these files into `test/`, and/or just inline the contents of the files into this test file so we don't even have to read from disk.

test/summarize.test.js
```diff
@@ -0,0 +1,23 @@
...
+    expected = await fs.readFile('./summarizer/expected.txt', 'utf-8');
+  });
+
+  test('summarize returns expected value', async () => {
+    predicted = await summarize(inputData, 5);
+    return expect(predicted).toBe(expected);
+  });
+
+  test('summarize returns string', async () => expect(typeof predicted).toBe('string'));
```
humphd: Tests can be run in different order, so don't count on `predicted` to exist here.  I'd re-calculate this in each test so you know it's good.  Try not to have `test()` functions depend on shared state.  It can lead to random failures.

test/summarize.test.js
```diff
@@ -0,0 +1,23 @@
...
+
+  test('summarize returns correct number of lines', () => expect(predicted.split('.').length).toBe(6));
```
humphd: What does `predicted` look like that you have to call split?  I don't follow the logic here.

ThomasLuu00: 'Predicted' should be returning 5 sentences since I pass 5 sentences as a parameter for summarize. So I split 'predicted' in order to get the number of sentences. The module I use requires that a number of sentences be specified to produce a summary.

humphd: Ah, I see.  Defining a sentence as "ends with a `.`" could be problematic, if a sentence includes something like "I'm working on a file named foo.js in my course."  Probably doing `/\. {1,2}/` for "period followed by 1 or 2 spaces" would be more accurate.  That still leaves out sentences that end with a `;`.  Anyway, this gets into the weed.  If you want to tighten up your `split()` to deal with the periods too, that's fine. Otherwise, we'll go with what you have here.

src/summarize.js
```diff
@@ -0,0 +1,8 @@
+const { SummarizerManager } = require('node-summarizer');
+
+module.exports = async function (text, numSentences = 3) {
+  const summarizer = new SummarizerManager(text, numSentences);
+  return summarizer.getSummaryByRank().then(
+    (summaryObject) => Promise.resolve(summaryObject.summary),
```
humphd: `(summaryObject) => summaryObject.summary` is all you need here

humphd:
Looks like your regex isn't passing this test anymore.  I think it's fine if you want to revert to what you had, so we can land this and improve it later.

ThomasLuu00:
I'm not sure why, but the TextRank algorithm in node-summarizer throws out inconsistent results.  Sometimes it fails the test but once its rerun it might pass, or just fail again.

ThomasLuu00:
So from what I read on the [node-summarizer](https://github.com/SwapnikKatkoori/node-summarizer#desc) repo,  looks like TextRank might not always throw out the same result, but it should still be generating the correct number of sentences each time.

humphd:
Fascinating.  OK, let's adjust the tests to match the expectations of the library, and include some comments in your tests/code that indicate what your research has uncovered about how it varies with each run.

PR #68, opened 2019-11-07
Added some rules to '.eslintrc.js'

manekenpix:
Added some rules to `.eslintrc.js` that I think would be useful.

- [`no-console`](https://eslint.org/docs/rules/no-console)
- [`no-param-reassign`](https://eslint.org/docs/rules/no-param-reassign)
- [`func-names`](https://eslint.org/docs/rules/func-names)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,6 +6,6 @@ module.exports = {
   rules: {
     "no-console": "off",
     "no-param-reassign": ["error", { props: false }],
-    "func-names": 0,
+    "func-names": "off",
   }
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,11 @@
 module.exports = {
   "extends": "airbnb-base",
   "env": {
-    "jest": true
+    "jest": true,
+  },
+  rules: {
+    "no-console": "off",
+    "no-param-reassign": ["error", { props: false }],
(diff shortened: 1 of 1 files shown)


.eslintrc.js
```diff
@@ -1,6 +1,11 @@
 module.exports = {
   "extends": "airbnb-base",
   "env": {
-    "jest": true
+    "jest": true,
+  },
+  rules: {
+    "no-console": "off",
```
humphd: Question: why `"off"` vs. `0`, which you use below?

Also, since this is a `.js` file, we can include comments, so would you mind adding a comment above each of these new rules explaining what you're doing here?  That way others can discover what this is all about.

manekenpix: For `no-console`, I used what eslint uses in their examples [here](https://eslint.org/docs/rules/no-console), and when I was testing `func-names` lint complained and wanted me to use 0. I can test if `no-console` works with `0` or if `func-names` works with `off` to have more consistency.

Also, I didn't add any comments because all the info to read about these changes is included in the description of the PR. Should I still add it to the js file?

manekenpix: I changed it and now they all use `off`.

manekenpix:
I added some comments to `eslintrc.js`for clarification.

PR #49, opened 2019-11-05
add error catch in readFile function

aprilllllllll:
HI everyone~
I Just add error catching in the readFile function for the enqueueFeedJobs function call.
Pls review and let me know what you think.
Thank you.

PR #76, opened 2019-11-07
Fixed linting errors

manekenpix:
After merging #49 a lot of linting errors were introduced.
This fixes those errors.
Let's make sure we rebase and run the test before committing and/or submitting PR


aprilllllllll:
Might because I change in the file in Github after having conflict.
Thank you!

Issue #82, opened 2019-11-07
Creating a Test that ensure's the email ability is working correctly


neilong31:
Creates testing for feature in #25 

I will be using nodemailer and ethereal to test that the we are able to send emails out.

Similar to [https://nodemailer.com/smtp/testing/](https://nodemailer.com/smtp/testing/)

RyanWils:
Hi Neil, 
I will submit a PR for a basic email sending function so you can start on your tests.
Sorry I miss clicked on the close issue button.

Issue #60, opened 2019-11-06
Start using .env to configure our system


humphd:
In keeping with the [12 Factor App](https://12factor.net) approach, we need to move toward using environment variables so we can have a proper [config](https://12factor.net/config).

Let's use `dotenv` for this.  Here's a good discussion on how to set this up: https://www.twilio.com/blog/2017/08/working-with-environment-variables-in-node-js.html

Once we have the env system in place, we can start enabling/disabling parts of the system or features based on flags in our environment.  It will also make it easier to configure things like Redis in dev, CI, or deployment environments.

The first part of this should be to get an initial config file in place that we can add to over time.

s-arika:
@humphd Could I try this one out?

humphd:
@s-arika it's all yours.  Let us know how you make out with your research, and any questions you might have.

hansal7014:
This would be useful for my issue too (#35 ). We can use an environment variable to set the log level.

s-arika:
@humphd  If I create an .env.sample file and have the developer fills in the values for the keys in the .env.sample file and rename it as .env file is that ok? Or should I just create both an .env.sample file and .env file and put the .env file in .gitignore.

s-arika:
@hansal7014 If you know the variables that you'll need, I could add them if you'd like

manekenpix:
>  If I create an .env.sample file and have the developer fills in the values for the keys in the .env.sample file and rename it as .env file

I'd go with that option. Also, maybe we could call it .env.example?

s-arika:
Yes sorry got confused, even if I create a .env file and ignore it, we'll still need the .env.example

SeanPrashad:
Fantastic work on this @s-arika - I was wondering if we can close this issue now? 🤔

s-arika:
Yup, thanks @SeanPrashad !

PR #93, opened 2019-11-07
Fix #25: Implement basic email sender

### LINKED ISSUE #25: Implement ability to email admins/users
This issue is a part of the [MVP Features](https://github.com/Seneca-CDOT/telescope/blob/master/docs/overview.md#mvp-features).

It should send an email to an admin or the user if:

- A post is deemed inappropriate
- Any major errors are logged (ex. issues downloading feeds)
- If it crashed

If there any other times an email should be sent out please suggest them as well.

RyanWils:
Contributes to issue #25 

I created a basic email sender using Nodemailer. 
To use the function, private information(email/pass,recipients) must be filled once the .env variables are set up. 

refer to issue #82 for the tests.

I have also left a comment at the top of the file on how to use this file. 
1. import the file (const sendEmail = require('./email-sender);) 
2. send email (sendEmail.sendMessage("Put your error message in here");)

### DIFF
--- a/env.example
+++ b/env.example
@@ -6,3 +6,12 @@ PORT=8080
 
 # REDIS_URL specifies Redis server info
 REDIS_URL=redis://127.0.0.1:6379
+
+# NODEMAILER_USERNAME is sender's username credential	
+NODEMAILER_USERNAME=	
+
+# NODEMAILER_PASSWORD is sender's password credential	
+NODEMAILER_PASSWORD=
+
+# NODEMAILER_SERVER is sender's server that they are using (Example: Gmail)
+NODEMAILER_SERVER=	
--- a/env.example
+++ b/env.example
@@ -13,5 +13,5 @@ NODEMAILER_USERNAME=
 # NODEMAILER_PASSWORD is sender's password credential	
 NODEMAILER_PASSWORD=
 
-# NODEMAILER_SERVER is sender's server that they are using (Example: Gmail)
+# NODEMAILER_SERVER is sender's server that they are using (Example: smtp.example.com)
 NODEMAILER_SERVER=	
--- a/src/email-sender.js
+++ b/src/email-sender.js
@@ -47,11 +47,11 @@ exports.sendMessage = async function (receipiants, subjectMessage, message) {
     };
 
     // Send the email with the email content
-    transporter.sendMail(mailOptions, (err, info) => {
+    transporter.sendMail(mailOptions, (err, info,) => {
       if (err) {
         reject(err); // Send promise.reject if an error occurs
       } else {
-        resolve(info); // Send promise.resolve if an error occurs
+        resolve(info.accepted); // Send promise.resolve if an error occurs
       }
     });
   });
--- a/src/email-sender.js
+++ b/src/email-sender.js
@@ -14,16 +14,28 @@ const nodemailer = require('nodemailer');
 
 exports.sendMessage = async function (receipiants, subjectMessage, message) {
   return new Promise((resolve, reject) => {
-    // Credientials to send an email from
-    const transporter = nodemailer.createTransport({
+    const transporter = nodemailer.createTransport('SMTP', {
       // Email Server (.env variable must be used refer to the Contribution.md)
-      service: process.env.NODEMAILER_SERVER,
+      host: process.env.NODEMAILER_SERVER,
+      port: 25,
+      secure: false,
       auth: {
         // Email Name (.env variables must be used refer to the Contribution.md)
         user: process.env.NODEMAILER_USERNAME,
         // Email Pass (.env variables must be used refer to the Contribution.md)
         pass: process.env.NODEMAILER_PASSWORD,
-      },
+      }
+    });
+
+    // verify connection configuration
+    transporter.verify(function (error, success) {
+      // If error then print to console
+      if (error) {
+        console.log(error);
+      // else print a ready message
+      } else {
+        console.log("Server is ready to take our messages");
+      }
     });
 
     // Email Content
--- a/src/email-sender.js
+++ b/src/email-sender.js
@@ -1,39 +1,44 @@
 const nodemailer = require('nodemailer');
 
 /*
-*                             HOW TO USE                                 *
-* Import this file - const sendEmail = require('./email-sender);         *
-* send email - sendEmail.sendMessage("Put your error message in here");  *
+                             HOW TO USE                                 
+ Import this file - const sendEmail = require('./email-sender);         
+ send email - sendEmail.sendMessage('emailName@hotmail.com','Seneca had an error', '<p>An Error had occured</p>');  
+
+ Parameters examples 
+ receipiants - 'emailName@hotmail.com,emailName2@hotmail.com'
+ subjectMessage - 'Seneca Telescope had an error'
+ message - '<p>An Error has occurred <br/>Error Message:</p>'
 */
 
-exports.sendMessage = async function (errorMessage) {
-  // Recipients list of users to send the email too
-  // receipiants is commas separated (emailName@hotmail.com,emailName2@hotmail.com)
-  const receipiants = '';
-
+exports.sendMessage = async function (receipiants, subjectMessage, message) {
+  return new Promise((resolve, reject) => {
+ 
   // Credientials to send an email from
   const transporter = nodemailer.createTransport({
-    service: 'gmail',
+    service: process.env.NODEMAILER_SERVER, //Email Server (.env variable must be used refer to the Contribution.md)
     auth: {
-      user: '', // Email Name (.env variables must be used refer to issue#60)
-      pass: '', // Email Pass (.env variables must be used refer to issue#60)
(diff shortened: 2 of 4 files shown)


humphd:
@RyanWils you can't share username/password pairs on GitHub like this.  Please get rid of this account.

src/email-sender.js
```diff
@@ -0,0 +1,38 @@
...
+
+    //Recipients list of users to send the email too
+    var receipiants = 'rwilson31@myseneca.ca,david.humphrey@senecacollege.ca';
+
+    //Credientials to send an email from 
+    var transporter = nodemailer.createTransport({
+        service: 'gmail',
+        auth: {
+            user: 'seneca.telescope@gmail.com',
+            pass: 'Password12345%^'
```
humphd: This is a no-go.  Username/Password pairs need to get read in via environment variables on `process.env`

RyanWils: Okay thank you, Sorry. It was a test email, I have deleted the account. 
Will change the code ASAP.

RyanWils:
I apologies for putting private information on the previous commit. 
I have removed the personal information in the newest commit and also deleted that test email I was using.  
It seems like there is an issue open for the .env variables - issue #60
I have left the variables blank for when this issue is completed.

humphd:
I wonder if we should use something like https://sendgrid.com/ for the mail transport layer.  Creating an account in GMail or the like isn't sustainable when it's shared among so many users.  There might be other services too, but this is one that came to mind, since we could use it long term for free.

RyanWils:
Thank you for the suggestion, it seems like a good choice (40,000 emails for the first 30 days, then 100/day forever. should be plenty). Ya you make a good point, I think its best to switch from Nodemailer

humphd:
@RyanWils I'm not sure if it's what we want or not, but something for you to look at.  I think you could use it *with* nodemailer, or use their node.js API.  Needs research.

humphd:
https://www.sendinblue.com/ is another one, there are a bunch.

humphd:
Using one of these services, for devs who want to hack on the email portion of the code, they could sign up for their own API key, put it in their `.env` and then they could work on that part of the code.  In production, we'd use a different one.

RyanWils:
Ya that would be for the best. For whichever service I pick, I will make sure to document the set up steps.

RyanWils:
Do we want to send an actual email to the users?
I have tested sendgrid and they dont actually send an email, but store it in their activity feed.
image
What are we looking for?

RyanWils:
I messed up on the titles of my commits for the Travis CI fixes as it references other topics. 
I have tried to reword my comments with git rebase -i master, but I cannot as the head is detached.
Anyone have any experience with this?

c3ho:
@RyanWils I'm not sure if [this ](https://stackoverflow.com/questions/10228760/fix-a-git-detached-head/17045215)may help, but it sounds like you went back to some previous commits and changed from there.

RyanWils:
Hi Calvin thanks for the solution, but sadly that solution did not work as my master is up to date. I created this pull request 6 days ago, so a lot of stuff got merged and even though I pulled into my master and rebased my branch into master it still says its detached. 

Screenshots of the issue 
image

says this for all the commits
image

src/email-sender.js
```diff
@@ -9,9 +9,9 @@ const nodemailer = require('nodemailer');
 exports.sendMessage = async function (errorMessage) {
   // Recipients list of users to send the email too
   // receipiants is commas separated (emailName@hotmail.com,emailName2@hotmail.com)
-  var receipiants = '';
+  const receipiants = '';
```
neilong31: Hey @RyanWils ! Something that I've noticed is that we have a recipients const, but we never have a way to update who the recipients are in the scope of this function. Maybe we can add a way to pass sendMessage the recipients in a array of strings?

src/email-sender.js
```diff
@@ -0,0 +1,39 @@
...
+  const mailOptions = {
+    from: '', // Email Name
+    to: receipiants,
+    subject: 'Seneca Telescope had an error', // Subject Line
+    html: `<p>An Error has occurred <br/>Error Message: ${errorMessage} </p>`, // Message Body
+  };
+
+  // Send the email with the email content
+  transporter.sendMail(mailOptions, (err, info) => {
+    if (err) {
```
neilong31: Hey @RyanWils ! Maybe we could return and err if it transporter.sendMail() doesn't work and a true or something else if it does. This would make it easier to test for future development. What do you think?

src/email-sender.js
RyanWils: Hi Neil I was thinking about the same changes to make to the send message function, but i think that should be implemented down the road because it would be a hassle to create an array of string every time you want to call this function. We can implement an optional parameters down the road once we start using it.

src/email-sender.js
RyanWils: Ya as you highlighted in your suggestion, the transporter sends the email and will print an error if it unsuccessful/print the information if it does not have an error. From my knowledge this is not reliable because just like the other frameworks to send emails, Nodemailer just send a message to the outlook server and if its received successfully by the outlook server it returns true. Even if the email does not send to the user, as long as it is received by the outlook server it will return true. This is where you tests come in, maybe we can find something that returns false if it cannot find an email.

src/email-sender.js
```diff
@@ -0,0 +1,39 @@
+const nodemailer = require('nodemailer');
+
+/*
+*                             HOW TO USE                                 *
+* Import this file - const sendEmail = require('./email-sender);         *
+* send email - sendEmail.sendMessage("Put your error message in here");  *
+*/
+
+exports.sendMessage = async function (errorMessage) {
```
humphd: We need to add a way to specify a few things, at the very least:

1. To
1. Subject
1. Body

I would put these in as params to the function.

src/email-sender.js
```diff
@@ -0,0 +1,39 @@
...
+  // Recipients list of users to send the email too
+  // receipiants is commas separated (emailName@hotmail.com,emailName2@hotmail.com)
+  const receipiants = '';
+
+  // Credientials to send an email from
+  const transporter = nodemailer.createTransport({
+    service: 'gmail',
+    auth: {
+      user: '', // Email Name (.env variables must be used refer to issue#60)
+      pass: '', // Email Pass (.env variables must be used refer to issue#60)
```
humphd: Let's do `SMTP_SERVER`, `SMTP_USERNAME`, `SMTP_PASSWORD` as environment variables in `env.sample`, and you can read them from `process.env`

src/email-sender.js
```diff
@@ -0,0 +1,39 @@
...
+    },
+  });
+
+  // Email Content
+  const mailOptions = {
```
humphd: Let's allow users to pass this in to your function directly, and we can validate that the required bits are here or error.

src/email-sender.js
humphd: Another thing you need to do, is properly wrap the callback style of nodemailer into a Promise:

```js
// This needs to return a Promise
async function doSomething() {
  return new Promise((resolve, reject) => {
    ...
    sendMail(options, function(err, info) {
        if(err) reject(err);
        else resolve(info);
    });
  });
}
```

or you can do it using [util.promisify()](https://nodejs.org/dist/latest-v8.x/docs/api/util.html#util_util_promisify_original)

env.example
```diff
@@ -6,3 +6,12 @@ PORT=8080
...
 REDIS_URL=redis://127.0.0.1:6379
+
+# NODEMAILER_USERNAME is sender's username credential	
+NODEMAILER_USERNAME=	
+
+# NODEMAILER_PASSWORD is sender's password credential	
+NODEMAILER_PASSWORD=
+
+# NODEMAILER_SERVER is sender's server that they are using (Example: Gmail)
+NODEMAILER_SERVER=
```
humphd: This is wrong for a server, see https://nodemailer.com/smtp/

src/email-sender.js
```diff
@@ -0,0 +1,58 @@
...
+      subject: subjectMessage, // Subject Line
+      html: message, // Message Body
+    };
+
+    // Send the email with the email content
+    transporter.sendMail(mailOptions, (err, info) => {
+      if (err) {
+        reject(err); // Send promise.reject if an error occurs
+      } else {
+        resolve(info); // Send promise.resolve if an error occurs
```
neilong31: One quick thing, could we resolve with info.accepted? That way with our test we can just test if it resolves with the correct recipients?

RyanWils: ya for sure, let me include it

RyanWils: would you like me to use info.rejected as well for the reject or is the error enough?

neilong31: Ah yes, could you please add that as well? After that I think it should be all good!

neilong31: No, I think it's good actually! It should return the error.

RyanWils: okay anything to change, if there isn't we should be good to merge. 
If there isn't anything please approve of these changes, thanks.

PR #97, opened 2019-11-07
- CI/CD Travis (#43)

### LINKED ISSUE #43: Set up CI/CD pipeline at Travis CI
Set up the base pipeline as per [comment](https://github.com/Seneca-CDOT/telescope/issues/6#issuecomment-549646158)  on Travis

kartik-budhiraja:
Fixes #43, CI/CD pipeline for Travis.

Edit: The pipeline is working as expected, we need to fix our tests because they are not exiting clean.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: node_js
 node_js:
   - "stable"
-install: npm install
 services:
   - redis-server
-script: npm test
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+language: node_js
+node_js:
+  - "stable"
+install: npm install
+services:
+  - redis-server
+script: npm test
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -0,0 +1,7 @@
+language: node_js
+node_js:
+  - "stable"
+install: npm install
```
humphd: You can remove this line, it's the default for node.js projects

.travis.yml
```diff
@@ -0,0 +1,7 @@
...
+services:
+  - redis-server
+script: npm test
```
humphd: Same with this, it's not necessary to define it unless your tests do something other than `npm test`

test/feed-queue.test.js
```diff
@@ -14,4 +14,6 @@ const feedQueue = require('../src/feed-queue');
  * after testing would fix the issue. Maybe we can handle opening and closing
  * Redis inside this project <- CREATE NEW ISSUE TO SOLVE THIS PROBLEM.
 */
-test('Testing feed-queue.js', () => feedQueue.add('http://ajhooper.blogspot.com/feeds/posts/default'));
+test('Testing feed-queue.js', () => {
+  feedQueue.add('http://ajhooper.blogspot.com/feeds/posts/default');
```
humphd: The failure on this test is something we can deal with in another PR.  This test isn't really correct as it is right now, so the failure isn't a problem.

PR #100, opened 2019-11-08
Fixes #88 - Write data to xml file instead of logging to console

### LINKED ISSUE #88: Write current data from worker to XML file
It might be helpful for development / hacking something together to test if we have the data written to an xml file instead of printed to the console.

lucacataldo:
As others are working on development related to parsing & processing of the data we're fetching, I thought it might be a good idea to be writing the data we've got to an xml file. This will allow us to visualize and format the xml data to determine what we're actually getting back in a meaningful manor.

No dependencies added other than the built in "fs" module, also added the tempData.xml file to .gitignore.

Related: #88


humphd:
I don't understand the purpose of this PR.

lucacataldo:
looking back it looks like the feature wasn't quite necessary as I misunderstood another PR so I've closed this PR

PR #101, opened 2019-11-08
Fix #70: added dependencies markdown

### LINKED ISSUE #70: Add a .md file to discuss the modules/dependencies being used
Since we will be using NodeJS, it will be appropriate to have a discussion page to add or remove modules/dependencies. It would also help others to understand why a particular module is used and how it will help the project.

robertbegna:
Hello, I've made another pull request regarding adding a dependencies mark down file. Previous pull request had too many prior unrelated commits so I closed that previous request, and updated my forked repo before adding the changes.

### DIFF
--- a/Dependencies.md
+++ b/Dependencies.md
@@ -1,0 +1,1 @@
+This markdown file will serve as a discussion page to add or remove modules to the Telescope project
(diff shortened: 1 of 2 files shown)


humphd:
I don't understand the purpose of this.  Dependencies are "documented" in https://github.com/Seneca-CDOT/telescope/blob/master/package.json#L19-L27, and we can include architectural discussions in the readme or contributing or code comments.  Having to keep this in sync with package.json is going to be a pain.

PR #105, opened 2019-11-08
Add .env File

s-arika:
This PR references #60 

Added:
- .env.example file.  So far it has a few environment variables for PORT, NODE_ENV and login for nodemailer
- config file. Creates a module with all the variables added and it maps and exports them. Now all files can import the config.js module and use it to access our environment variables using 
const { port } = require('../config/config'); also wouldn't need to write process.env before the variable


.env.example
```diff
@@ -0,0 +1,4 @@
+NODE_ENV=development
+PORT=8080
+NODEMAILER_USERNAME=
+NODEMAILER_PASSWORD=
```
manekenpix: Next line missing

s-arika: added line

.env.example
```diff
@@ -0,0 +1,4 @@
+NODE_ENV=development
```
humphd: Above each line, please add a comment

```
# NODE_ENV should be one of "development" or "production"
NODE_ENV=development

PORT=8080

...
```

This example file can serve as both a template, and documentation.

config/config.js
```diff
@@ -0,0 +1,10 @@
+const dotenv = require('dotenv');
+
+dotenv.config();
+
+module.exports = {
```
humphd: I wonder about this vs. just letting people use `process.env` everywhere.  If we go with this approach, then every file has to `require` this file.  If we leave it all on `process.env.*` we don't.

I'd suggest we remove this and export nothing from `config.js`, and just make sure we `require('./config')` early in the code.

s-arika: ok, do you want me to add 
```
const dotenv = require('dotenv');
dotenv.config();
```
 at the very beginning of index.js for now ?

.env.example
s-arika: - renamed .env.example to env.example
- added instructions in contributing.md
- added comments in env.example

docs/CONTRIBUTING.md
```diff
@@ -28,6 +28,9 @@ An easier solution would be to use Docker.
 
 **Setup**
 1. Navigate to the root directory of telescope.
+1. Create a new file and name it ".env".
+1. Copy the contents of "env.example" into ".env".
```
humphd: Combine these first 2 lines into one: `1. Copy env.example to .env to create a new environment configuration`

env.example
```diff
@@ -0,0 +1,11 @@
...
+NODE_ENV=development
+
+# PORT is the port used by the web server
+PORT=8080
+
+# NODEMAILER_USERNAME is sender's username credential
+NODEMAILER_USERNAME=
+
+# NODEMAILER_PASSWORD is sender's password credential
+NODEMAILER_PASSWORD=
```
humphd: Let's remove the NODE_EMAILER_* config until we need it (e.g., future PR can add it if necessary)

s-arika: removed NODE_EMAILER_* config

docs/CONTRIBUTING.md
s-arika: Combined first 2 lines into one

Issue #75, opened 2019-11-07
Extract plain text form html


humphd:
We are building a bunch of pieces of our pipeline that will give us HTML, that is, a blog post from a user's feed.  What we also need for later parts of the analysis pipeline is the ability to work on plain text vs. html.

Let's add a parsing function that works like this:

- given some HTML as an argument
- parse the HTML using https://github.com/jsdom/jsdom into a DOM
- `querySelector()` for the element we want (maybe all of `document.body`?)
- once we have that `elem` we can obtain it's text using `elem.innerText` and return that from the function

We also need tests for this, which can happen in another issue/PR.  @PavanKKamra has expressed some interest in this, and he might be glad to have some others work with him on it.

PavanKKamra:
Hello David, 

I would like to attempt this issue. 

Thank you, 
Pavan K.

humphd:
OK @PavanKKamra, go for it.  Please ask questions in here, and keep us updated on your progress.

To get started, I would do the following:

- being a new branch `issue-75`
- add JSDOM using `npm install --save jsdom`
- add a new file to `src` named `text-parser.js` or something
- in your new file, create a function with this basic shape:

```js
const jsdom = require("jsdom");

const { JSDOM } = jsdom;

exports.run = async function (html) {
  // put your code here.
}
```

Reza-Rajabi:
@humphd Do we need to close #16 now?

Reza-Rajabi:
@humphd It looks similar to #16 to me.

humphd:
@Reza-Rajabi it's a portion of what happens after #16 I think.  We can make #16 be specifically about just parsing from url to xml.

humphd:
In #83, an initial implementation of this is being landed, so the tests can pass.  We can probably improve this a bit by changing the JSDOM parsing to use [`fragment`](https://github.com/jsdom/jsdom#fragment) parsing, and simplify the code a bit.

humphd:
If we can't get this to happen with JSDOM, we may need to switch to a full browser via Puppeteer, see https://github.com/GoogleChrome/puppeteer/issues/3260#issuecomment-422081505.

humphd:
NOTE for later: we can also use [`page.setContent(html)`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagesetcontenthtml-options) to render a snippet of HTML in a real DOM, before calling into it to get `innerText`.

Issue #106, opened 2019-11-08
ESLint Errors in dialogue.js


s-arika:
There are some lint errors in dialogue.js that needs to be corrected.

manekenpix:
This issue was resolved in #107

PR #109, opened 2019-11-08
Fixes #83: test case for text-parser

### LINKED ISSUE #83: Test for issue #75
A simple test to check if the html can be converted to text plain once #75  is merge

Test are written in [Jest](https://jestjs.io/) 
Command to run test: **npm test**

Test is saved in **test** repo

giatuongtran9:
Adding test case for parsing html to text, reference to issue #75 

@PavanKKamra, apparently JSOM does not support innetText, based on this [Issue](https://github.com/jsdom/jsdom/issues/1245). I have tested it and innerHTML and textContent works fine but innerText will return 'undefined'

### DIFF
--- a/src/text-parser.js
+++ b/src/text-parser.js
@@ -1,0 +1,12 @@
+const jsdom = require('jsdom');
+
+const { JSDOM } = jsdom;
+
+// Simple function to convert html to text
+// Issue #75: https://github.com/Seneca-CDOT/telescope/issues/75
+exports.run = async function (html) {
+  const dom = new JSDOM(html);
+  const result = dom.window.document.querySelector('body').textContent;
+
+  return Promise.resolve(result);
+};
--- a/test/text-parser.test.js
+++ b/test/text-parser.test.js
@@ -1,0 +1,17 @@
+const textParser = require('../src/text-parser');
+
+
+/**
+ * textParser.run() will convert the html to text plain
+ * Function should only extract what in the body
+ */
+test('Testing text-parser', async () => {
+  const result = await textParser.run('<!DOCTYPE html><p>Hello World</p>');
+  expect(result).toBe('Hello World');
+});
+
+// Issue: Don't work with new line
+test('Testing text-parser with new line', async () => {
+  const result = await textParser.run('<!DOCTYPE html><html><head><title>OSD600</title></head><body style = "text-align:center;"><h1>Seneca</h1>\n<div>OpenSource Telescope</div></body></html>');
+  expect(result).toBe('Seneca\nOpenSource Telescope');
+});
--- a/package-lock.json
+++ b/package-lock.json
@@ -4076,10 +4076,10 @@
         "standard-as-callback": "^2.0.1"
       }
     },
-     "ipaddr.js": {
-      "version": "1.9.0",
-      "resolved": "https://registry.npmjs.org/ipaddr.js/-/ipaddr.js-1.9.0.tgz",
-      "integrity": "sha512-M4Sjn6N/+O6/IXSJseKqHoFc+5FdGJ22sXqnjTpdZweHK64MzEPAyQZyEU3R/KRv2GLoa7nNtg/C2Ev6m7z+eA=="
+    "ip-regex": {
+      "version": "2.1.0",
+      "resolved": "https://registry.npmjs.org/ip-regex/-/ip-regex-2.1.0.tgz",
+      "integrity": "sha1-+ni/XS5pE8kRzp+BnuUUa7bYROk="
     },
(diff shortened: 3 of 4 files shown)


giatuongtran9:
Hi @humphd , I have change the test to return a Promise by async await.

I have noticed that **dialogue.js** has some errors whenever I run the test 'npm run test'
Screen Shot 2019-11-08 at 7 35 31 PM

humphd:
@giatuongtran9 there is a PR to fix that in #106 if you want to review.

humphd:
@giatuongtran9 it looks like you've merged and pulled in a bunch of unrelated changes.  Also, there are merge conflict markers in `test/text-parser.test.js`.

NOTE: you want to rebase but not merge with master.

giatuongtran9:
@humphd I made some mistakes at the beginning and try to rebase to master but somehow pulled all the new files. The test fail because we don't have the **text-parser.js** yet

humphd:
OK, we'll have to fix this git history.  I can help you when we see each other at the start of the week, or you can try and do it yourself between now and then.

As far as the failure, tell Jest to `skip` these tests until we have an implementation: https://jestjs.io/docs/en/api.html#testskipname-fn

giatuongtran9:
I will try to fix this but I think I will need your help. skip cannot be used here because the test need the run method from text-parser.js so we can comment out the test or do this for now:
```
expect('Hello World').toBe('Hello Word');
```

humphd:
OK @giatuongtran9, how about this.  Since @PavanKKamra doesn't seem to have started his work yet, why don't you add `text-parser.js` and have it export a function at least.  In the code, you can leave a comment to the Issue that will implement it in #75. Then @PavanKKamra can work from there.

@PavanKKamra how is your work coming?  It sounds like we're just about ready for your code here...

giatuongtran9:
@humphd  I have added simple function for text-parser and @PavanKKamra can continue from there

src/text-parser.js
```diff
@@ -0,0 +1,13 @@
...
+const { JSDOM } = jsdom;
+
+
+// Simple function to convert html to text
+// Issue #75
+exports.run = async function (html) {
+  const dom = new JSDOM(html);
+  const result = dom.window.document.querySelector('body').textContent;
+
+  return result;
```
humphd: `return Promise.resolve(result);` so it gives back a `Promise`.

src/text-parser.js
```diff
@@ -0,0 +1,13 @@
+const jsdom = require('jsdom');
+
+const { JSDOM } = jsdom;
+
+
```
humphd: Remove this extra blank line, please.

test/text-parser.test.js
```diff
@@ -4,16 +4,13 @@ const textParser = require('../src/text-parser');
 /**
  * textParser.run() will convert the html to text plain
  * Function should only extract what in the body
- * Fuction should only extract what in the body
  */
 test('Testing text-parser', async () => {
```
humphd: Do you not need to `skip` this one too?

src/text-parser.js
```diff
@@ -0,0 +1,13 @@
...
+// Simple function to convert html to text
+// Issue #75
```
humphd: Replace this with a full GitHub URL to issue 75, please.

test/text-parser.test.js
```diff
@@ -10,7 +10,8 @@ test('Testing text-parser', async () => {
   expect(result).toBe('Hello World');
 });
 
-test.skip('Testing text-parser with new line', async () => {
-  const result = await textParser.run('<!DOCTYPE html><html><head><title>OSD600</title></head><body style = "text-align:center;"><h1>Seneca</h1><div style="white-space: pre">OpenSource Telescope</div></body></html>');
+// Issue: Don't work with new line
```
humphd: What does this mean?

giatuongtran9: my function in text-parser.js won't convert the new line and I don't know why. I will fix it later

humphd: I suspect it's due to limitations in JSDOM, which doesn't have any layout engine, and things like newlines depend on having one.  I left some notes in #75 on how we might overcome this later.

humphd:
@giatuongtran9 you need to regenerate `package-lock.json` like so:

```
npm install
git add package-lock.json
```

giatuongtran9:
@humphd Thank you and @cindyledev please review it again

giatuongtran9:
@cindyledev please review again, if you left it here, other stuff will get merged and this will forever get conflict and never get merge
@humphd sorry for asking to review it again so many time

PR #65, opened 2019-11-07
Adds Docker files as a milestone for issue-4

Reza-Rajabi:
**_What is this pull request:_**
This will add Dockerfile and .dockerignore to the project. So that we can run the project without installing node packages like so:

```bash
$ docker-compose up --build
```

and more importantly, we can ship the Docker image to Kubernetes later on.

PR #113, opened 2019-11-09
Adds Docker files

humphd:
I think this was meant to land in `master`, was originally #65.  I've created a new PR so we can move it to `master`.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,13 @@
+**/node_modules/
+**/node_modules_linux/
+**/docs/
+**/Dockerfile
+**/.dockerignore
+**/.editorconfig
+**/.eslintrc.js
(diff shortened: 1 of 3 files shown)


Reza-Rajabi:
@humphd I have created issue-4 under repo assuming that the Kubernetes will take us about 3 milestones. Based on this assumption the pull request has been made for this branch and not the master branch. I am not sure if that was the best choice though. 
Maybe it is better to land it on the master so that we could see if there are any problems.

humphd:
Because taking this code into `master` is fairly low-risk (we aren't depending on it in production), I'd say it's a good idea to introduce it early, so we can test, improve, and get it ready to ship.  So my vote is: merge.

Issue #117, opened 2019-11-09
Invalid CSS selector in planet.css


humphd:
```css
/* Custom h1 to ensure major posts are separated */
h1c{
	font-size: 2em;
	margin: .67em 0;
	border-bottom: 15px solid var(--nav-font-color);
	display: inline-block;
	font-weight: 700;
	text-align: center;
}
```

`h1c` should probably be `h1`.  Someone might want to add an issue/pr to include [stylelint](https://stylelint.io/) like we have for eslint, to catch these types of bugs automatically.

cc @ODAVING, this came in with 0786e2cc it looks like.

humphd:
Filed #118 for stylelint.

humphd:
UPDATE: the work in #125 caught this, which is a good sign moving forward.

ODAVING:
@humphd Sorry, didn't see this issue until now. h1c is a different custom class. I didn't want to alter h1 as it causes a bug in the "participants page" heading.

humphd:
If it's a custom class, please prefix it with `.`

ODAVING:
Custom tag. my mistake

humphd:
HTML5 doesn't allow for custom element definition like this.  Please use standard HTML/CSS, and create a class.

PR #125, opened 2019-11-09
Fixes #118: added stylelint

### LINKED ISSUE #118: Add stylelint to our automated testing
Some people are beginning to experiment with CSS, and it would be good to add linting for it, to catch typos and other bugs (#117 is an example that we could catch via automation).

Let's add [stylelint](https://stylelint.io/) similar to eslint, so it can be run automatically as part of `npm test`.

cindyledev:
Added style lint to catch invalid css styles

How to run script:
`npm run stylelint`
Annotation 2019-11-09 150129

This PR successfully picks up the `h1c` invalid css selector in planet.css 
Fixes #118

### DIFF
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -1,0 +1,43 @@
+"use strict";
+
+module.exports = {
+  "ignoreFiles": [
+    "Dockerfile",
+    "LICENSE",
+    "*.*",
+    "images/*.*",
+    "!*.css"
+  ],
+  rules: {
+    "at-rule-no-unknown": true,
+    "block-no-empty": true,
+    "color-no-invalid-hex": true,
+    "comment-no-empty": true,
+    "declaration-block-no-duplicate-properties": [
+      true,
+      {
+        ignore: ["consecutive-duplicates-with-different-values"]
+      }
+    ],
+    "declaration-block-no-shorthand-property-overrides": true,
+    "font-family-no-duplicate-names": true,
+    "font-family-no-missing-generic-family-keyword": true,
+    "function-calc-no-invalid": true,
+    "function-calc-no-unspaced-operator": true,
+    "function-linear-gradient-no-nonstandard-direction": true,
+    "keyframe-declaration-no-important": true,
+    "media-feature-name-no-unknown": true,
+    "no-descending-specificity": true,
+    "no-duplicate-at-import-rules": true,
+    "no-duplicate-selectors": true,
+    "no-empty-source": true,
+    "no-extra-semicolons": true,
+    "no-invalid-double-slash-comments": true,
+    "property-no-unknown": true,
+    "selector-pseudo-class-no-unknown": true,
+    "selector-pseudo-element-no-unknown": true,
+    "selector-type-no-unknown": true,
+    "string-no-newline": true,
+    "unit-no-unknown": true
+  }
+};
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -5,6 +5,7 @@
   "scripts": {
     "eslint": "eslint --ignore-path .gitignore .",
     "eslint-fix": "eslint --fix --ignore-path .gitignore .",
+    "stylelint": "stylelint --ignore-path .gitignore .",
```
humphd: We need to hook this into our `test` script, or people won't run it, and it won't happen in CI automatically.

There are a few ways to do it.  Here's one way, but adding a `lint` and `postlint` script (NOTE: I also changed the `pretest` to use lint vs. eslint):

```
"stylelint": "stylelint --ignore-path .gitignore .",
"lint": "npm run eslint",
"postlint": "npm run stylelint",
"pretest": "npm run lint",
```

Another way would be to run parallel scripts using `npm-run-all` like they do in https://github.com/stylelint/stylelint/blob/master/package.json.

Either way is probably fine.  Our goal is for `npm test` to do this:

- eslint
- stylelint
- jest

Once you do this, we should start failing Travis due to our stylesheet having errors.

cindyledev:
I've hooked the stylelint script to the test script. Please review my changes. 

Adding npm-run-all can be filed as another issue.

humphd:
```
planet.css
 207:1  ✖  Unexpected unknown type selector "h1c"   selector-type-no-unknown
```

This is great, [it would have caught](https://travis-ci.org/Seneca-CDOT/telescope/builds/609836000#L234) the issue in #117.

.stylelintrc.js
```diff
@@ -0,0 +1,43 @@
...
+module.exports = {
+  "ignoreFiles": [
+    "Dockerfile",
+    "LICENSE",
+    "*.*",
+    "images/*.*",
+    "!*.css"
+  ],
+  rules: {
+    "at-rule-no-unknown": true,
```
humphd: Question: why not just use the standard config of rules in https://github.com/stylelint/stylelint-config-standard?  How did you arrive at this set of rules?

package.json
```diff
@@ -27,6 +30,8 @@
     "eslint": "^6.6.0",
     "eslint-config-airbnb-base": "^14.0.0",
     "eslint-plugin-import": "^2.18.2",
-    "jest": "^24.9.0"
+    "jest": "^24.9.0",
+    "stylelint": "^11.1.1",
+    "stylelint-webpack-plugin": "^1.0.3"
```
humphd: Why this dep?

.stylelintrc.js
cindyledev: I used https://github.com/stylelint/stylelint-config-recommended , should I use the standard config instead of recommended?

humphd: Instead of manually defining these rules, you can install the recommended:

```
npm install stylelint-config-recommended --save-dev
```

and then use

```
  "extends": "stylelint-config-recommended"
```

And you don't/won't have to specify all these.  You'd only use these `rules` if you wanted to override something, which we might.  This is very similar to [what we're doing in eslint](https://github.com/Seneca-CDOT/telescope/blob/master/.eslintrc.js#L2) with the [Airbnb ruleset](https://www.npmjs.com/package/eslint-config-airbnb)

package.json
cindyledev: Sorry, this was a mistake. I thought we need both stylelint and stylelint-webpack-plugin for stylelint to work. I'll remove it

humphd:
I'm excited to see this land, as it will help us catch bugs faster in review for subsequent PRs.

PR #135, opened 2019-11-10
Update CONTRIBUTING.md to specify starting Redis

### LINKED ISSUE #134: Update CONTRIBUTING.md to specify starting Redis
My Node version is `12.12.0` and my operating system is Mac OS Catalina `10.15.1 (19B88)`.

As a new individual to telescope, I was trying to setup the project using the steps listed under `Setup` found [here](https://github.com/Seneca-CDOT/telescope/blob/master/docs/CONTRIBUTING.md#environment-setup).

However, an error occurred after step 5 (`5. Run npm test`) which states that the `able to add a job to the feed queue` test failed due to `connect ECONNREFUSED 127.0.0.1:6379`.

After grepping the repo for the port `6379`, I realized that I had not started the Redis container using ~`docker-compose up -d`~ `docker-compose up -d redis`. It would be great to include that command in our `CONTRIBUTING.md` docs before `Run npm install`.

SeanPrashad:
Fixes #134

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -30,7 +30,7 @@ An easier solution would be to use Docker.
 1. Navigate to the root directory of telescope.
 1. Copy env.example to .env to create a new environment configuration.
 1. Replace default key values in .env with credentials.
-1. Begin the Redis container using `docker-compose up -d`
+1. Start the docker container for Redis using `docker-compose up -d redis`
 1. Run `npm install`.
 1. Run `npm test`
 1. IF eslint detect some issues run `npm run eslint-fix` before manually fixing the issue (Will save you time :smile:) and then run `npm test` again.
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -30,6 +30,7 @@ An easier solution would be to use Docker.
 1. Navigate to the root directory of telescope.
 1. Copy env.example to .env to create a new environment configuration.
 1. Replace default key values in .env with credentials.
+1. Begin the Redis container using `docker-compose up -d`
 1. Run `npm install`.
 1. Run `npm test`
 1. IF eslint detect some issues run `npm run eslint-fix` before manually fixing the issue (Will save you time :smile:) and then run `npm test` again.
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -31,6 +31,7 @@ An easier solution would be to use Docker.
 1. Copy env.example to .env to create a new environment configuration.
 1. Replace default key values in .env with credentials.
 1. Start the docker container for Redis using `docker-compose up -d redis`
+    - To stop the docker container for Redis, run `docker-compose stop redis`
 1. Run `npm install`.
 1. Run `npm test`
 1. IF eslint detect some issues run `npm run eslint-fix` before manually fixing the issue (Will save you time :smile:) and then run `npm test` again.


docs/CONTRIBUTING.md
```diff
@@ -29,7 +29,8 @@ An easier solution would be to use Docker.
 **Setup**
 1. Navigate to the root directory of telescope.
 1. Copy env.example to .env to create a new environment configuration.
-1. Replace default key values in .env with credentials. 
+1. Replace default key values in .env with credentials.
+1. Begin the Redis container using `docker-compose up -d`
```
manekenpix: This actually initializes both Redis and telescope. For local testing, it's better to just execute `redis-server` in a terminal and then following the next steps (`npm install`, `npm test`...)

SeanPrashad: Nice catch!! Running `docker ps` showed both the container for telescope and redis running - I think we'd only need redis running for the tests.

I'm not sure I follow your suggestion to run the command, `redis-server`. From my understanding, the docker container for Redis has been made available such that a contributor doesn't need to have Redis installed on their machine - they can run the container, whether they're on Windows/Mac/Linux or some other variant, and this will remove the dependency of having Redis installed locally.

Please let me know if anything is unclear! 😀

manekenpix: The idea was to run everything as a [multi-container Docker application](https://docs.docker.com/compose/), but now that you mentioned it, I think what you suggested is a good idea: available such that contributors don't need to have Redis installed on their machine.

For that, we can do this.
- First time a contributor runs a docker container for redis:
`docker run --name redis-telescope -p 6379:6379 redis`
This creates a container for redis named redis-telescope and exposes it on port 6379, the one we need for telescope.

- Contributor wants to run the same container again:
`docker start -a redis-telescope`
This runs the same container, using the option `-a` (foreground running) so the contributor can see the output.
For running it in detached mode, don't use `-a`.

For removing the container redis container: `docker container rm redis-telescope`.

@PagerDuty test it and let me know what you think.

CC: @Reza-Rajabi What do you think about all this?

SeanPrashad: What you're saying is making total sense! I have a few questions:

* I think we might want to reconsider using `docker run --name redis-telescope -p 6379:6379 redis` over `docker-compose up -d redis`. The Redis container metadata could possibly change in the future and having it specified in the docker-compose file ensures that we have a single source of truth - what do you think? 🤔

* I didn't see those commands you suggested for spinning up and taking down the Redis container anywhere in the `CONTRIBUTING` file nor the `README`. I want to try and avoid increasing the scope/complexity in this patch but I think it would definitely be worth a discussion for another PR!

I think to produce a frictionless experience for newcomers so they don't get frustrated that they can't even setup the project, we should keep it simple. See below for the entire overview of what I'm suggesting:

image

manekenpix: Oh! My bad, you're right, going with `docker-compose up -d redis` is definitely better. Just one thing, if you want to recommend running in detached mode `-d`, maybe you could mention how to stop it. Other than that, I think it's good to go.

docs/CONTRIBUTING.md
```diff
@@ -29,7 +29,8 @@ An easier solution would be to use Docker.
 **Setup**
```
Reza-Rajabi: What if we breakdown this section to two or three parts:
- Using npm to install dependencies and have Redis installed
- Have Docker installed and just use 'docker-compose up --build' command (temporary *)
- Optionally suggesting a mix of two, for example, npm to install dependencies and Redis from Docker image (as it is in your pull request)

*Our Docker files for development and debug environments are coming by @wajeehsheikh .

manekenpix: The way @PagerDuty is suggesting doesn't need Redis installed on your machine, that's why I think it's the best option. Running `docker-compose up -d redis` is enough. If anything, we should remove the `Redis set up` and just add instructions to install `docker`.

Reza-Rajabi: That't true. 
```
Optionally suggesting a mix of two, for example, npm to install dependencies and 
{ use } Redis from Docker image (as it is in your pull request)
```

I have missed "{ use }" in this 👆sentence.

Reza-Rajabi:
@SeanPrashad @manekenpix 
Let me elaborate my comment in here:

**Set up**

- Navigate to the root directory of telescope. 
- Copy env.example to .env to create a new environment configuration.
- Replace default key values in .env with credentials. 

Then: 

- Option 1 (You have Redis installed):
       - Run `npm install`.
       - Run `npm test`

- Option 2 (You have Docker installed):
       - Run `docker-compose up --build`

- Option 3 (You have Docker installed):
       - Start the docker container for Redis using `docker-compose up -d redis`
       - Run `npm install`.
       - Run `npm test`

IF eslint detect some issues run `npm run eslint-fix` before manually fixing the issue (Will save you time :smile:) and then run `npm test` again.

Reza-Rajabi:
Something like that

manekenpix:
I like @Reza-Rajabi of having both possibilities just in case. @PagerDuty, what do you think?

SeanPrashad:
Some awesome points made by the both of you @manekenpix & @Reza-Rajabi! 👏🏽

For better visibility by everyone else, I definitely think that we should open a new issue, maybe labeled something like `Update CONTRIBUTING.md to have generalized setup steps`, and copy-pasta [this comment](https://github.com/Seneca-CDOT/telescope/pull/135#issuecomment-553558817) alongside the need to add the Docker installation links for [Windows](https://docs.docker.com/docker-for-windows/install/), [Mac](https://docs.docker.com/docker-for-mac/install/), and Linux (_which will be trickier than the former two_).

What do you two think? 🤔

SeanPrashad:
Some awesome points made by the both of you @manekenpix & @Reza-Rajabi! 👏🏽

For better visibility by everyone else and to keep the scope to what #134 wanted, I strongly suggest that we open a new issue! It could maybe be labeled something like `Update CONTRIBUTING.md to have generalized setup steps` with the body text copied from [this comment](https://github.com/Seneca-CDOT/telescope/pull/135#issuecomment-553558817) alongside the need to add the Docker installation links for [Windows](https://docs.docker.com/docker-for-windows/install/), [Mac](https://docs.docker.com/docker-for-mac/install/), and Linux (_which will be trickier than the former two_).

After the above is done, this patch should be good to go (_I added instructions on how to stop the running Redis container_). What do you two think? 🤔

Reza-Rajabi:
> Some awesome points made by the both of you @manekenpix & @Reza-Rajabi! 👏🏽
> 
> For better visibility by everyone else and to keep the scope to what #134 wanted, I strongly suggest that we open a new issue! It could maybe be labeled something like `Update CONTRIBUTING.md to have generalized setup steps` with the body text copied from [this comment](https://github.com/Seneca-CDOT/telescope/pull/135#issuecomment-553558817) alongside the need to add the Docker installation links for [Windows](https://docs.docker.com/docker-for-windows/install/), [Mac](https://docs.docker.com/docker-for-mac/install/), and Linux (_which will be trickier than the former two_).
> 
> After the above is done, this patch should be good to go (_I added instructions on how to stop the running Redis container_). What do you two think? 🤔
> 
> **Edit:** I think it'd be great to loop in @cindyledev to help out with the new issue! I'd be able to help mentor/pair on the issue in my spare time

@SeanPrashad  looks good then.

PR #136, opened 2019-11-10
Fix #59: Added feed list parser

### LINKED ISSUE #59: Automatically fetch and parse complete feed list from wiki
At the moment we're using a hack to read feed URLs from a text file.  This is fine for our initial efforts, but we should really be pulling those feed URLs from https://wiki.cdot.senecacollege.ca/wiki/Planet_CDOT_Feed_List.

Looking at the markup in that page, we need to grab the contents of the single `pre` element on the page:

```html
<pre>
################# Failing Feeds Commented Out [Start] #################

#Feed excluded due to getaddrinfo ENOTFOUND s-aleinikov.blog.ca s-aleinikov.blog.ca:80
#[http://s-aleinikov.blog.ca/feed/atom/posts/]
#name=Sergey Aleinikov
...
</pre>
```

We then need to lightly process this:
- ignore lines beginning with `#` (comments)
- read URLs out of the `[url]` format (surrounded in square brackets)
- read the person's name from `name=User Name`

Later we might decide to store this in a db or something, but let's begin by pulling this down from the cdot wiki.

cryolis:


cryolis:
#59

humphd:
Please don't merge master into your PRs.  You want to rebase onto master.

Also, you have a missing comma in package.json

src/parser.js
```diff
@@ -0,0 +1,40 @@
+const bent = require('bent');
+const jsdom = require('jsdom');
+
+const request = bent('string');
+const url = 'https://wiki.cdot.senecacollege.ca/wiki/Planet_CDOT_Feed_List';
```
humphd: This should get moved into our `env.sample` file and be read from `process.env` using the `src/config` module.

src/parser.js
```diff
@@ -0,0 +1,40 @@
...
+const { JSDOM } = jsdom;
+const nameTest = /^name/i;
```
humphd: Move all these variables into the scope in which they are defined.  No need for module level variables here.

src/parser.js
```diff
@@ -0,0 +1,40 @@
...
+let lines = '';
+let feed = '';
+let feedList = '';
+let tempLine = '';
+
+/**
+* Get feed data from url
+* process it to remove unnecessary information like "name" and square brackets
+*/
+request(url)
```
humphd: Let's wrap this in a function we can call.  In fact, let's split it into 2 functions.  One function can download the data from the wiki and parse out the `<pre>`; the second function can accept the text inside it and do the processing. This should make it easier to write tests for each.

src/parser.js
```diff
@@ -0,0 +1,40 @@
...
+  .then((data) => {
+    const dom = new JSDOM(data);
+    feedList = dom.window.document.querySelector('pre').textContent;
+    lines = feedList.split(/\r\n|\r|\n/);
+
+    lines.forEach((element) => {
+      if (!linkTest.test(element)) {
+        if (element.startsWith('[')) {
+          tempLine = element.replace(/\[href="|\[/, '');
```
humphd: I wonder if we should just change the wiki to fix that one `[href=...` example you found, so it matches the correct format?

humphd: Also, special cases like this deserve a comment above to explain why you're bothering with this extra work to replace here.

src/parser.js
```diff
@@ -0,0 +1,40 @@
...
+          tempLine = tempLine.replace(']', '');
+          feed += `${tempLine}\n`;
+        }
+        if (nameTest.test(element)) {
+          tempLine = element.replace(/^name = |^name= |^name=/i, '');
```
humphd: Why not use a capture group here, and one or more whitespace?  Something like:

```
/^\s*name\s*=\s*([\w ]+)/
```
Then you an work with the first capture group to get the name.

src/parser.js
```diff
@@ -0,0 +1,40 @@
...
+          feed += `${tempLine}\n`;
+        }
+      }
+    });
+    // eslint-disable-next-line no-console
+    console.log(feed);
```
humphd: Let's make this into a set of tests.  If you don't want to do them here, you can get someone else to work on it in a new PR.

cryolis: I'l remove that and write all the data into a file so Nick can use it for his issue

src/parser.js
cryolis: I don't see these files you mentioned, is anyone still working on them?

src/parser.js
humphd: No, don't do anything in files.  We will use Redis/Bull as our message bus, not the filesystem.  For now, just get this exposed as a set of functions, and next week we'll hook it all together.

src/parser.js
humphd: See the work in #145 by @hansal7014 or what @miggs125 is doing in #141.  See https://github.com/Seneca-CDOT/telescope/blob/master/env.example and https://github.com/Seneca-CDOT/telescope/blob/master/src/config.js.

src/parser.js
cryolis: I removed href from wiki and added some more comments to functions

src/parser.js
```diff
@@ -0,0 +1,59 @@
+require('dotenv').config();
```
humphd: You can just require the `src/config` file here instead, it already does this.

humphd: Let's rename this file `src/parser.js` to something less generic.  How about `src/wiki-feed-parser.js`?  We have a lot of "parsers" now, so it's confusing.

src/parser.js
```diff
@@ -0,0 +1,59 @@
...
+      }
+      console.log(feed);
+      return feed;
+    });
+  }).catch((err) => { throw err; });
+
+  return parsedData;
+}
+
+parseData();
```
humphd: Instead of calling this function at the end, please `export` your two functions from this module.  Then we can write tests for it, and also hook it into other parts of the program.

src/parser.js
```diff
@@ -0,0 +1,59 @@
...
+const { JSDOM } = jsdom;
+
+/*
+* getData() returns Promise { <pending> }
+* It gets the data from 'pre' tag in a provided url
+* Splits the data into lines so its easier to process as a string array
+* That data is then returned as a Promise
+*/
+function getData() {
+  const list = request(process.env.FEED_URL)
```
humphd: You can just do `return request(process.env.FEED_URL)` here, no need for `const list`.

src/parser.js
```diff
@@ -0,0 +1,59 @@
...
+    .then((data) => {
+      const dom = new JSDOM(data);
+      let feedList = '';
```
humphd: Combine this line with the next (there's no need for the empty string declaration, since you throw it away immediately):

```js
let feedList = dom.window.document.querySelector('pre').textContent;
```

src/parser.js
```diff
@@ -0,0 +1,59 @@
...
+* That data is then returned as a Promise
+*/
+function parseData() {
+  const nameCheck = /^name/i;
+  const commentCheck = /^#/;
+
+  let feed = '';
+  let line = '';
+
+  const parsedData = getData().then((data) => {
```
humphd: `return getData().then((data) => {` is all you need

src/parser.js
```diff
@@ -0,0 +1,59 @@
...
+          line = element.replace(/([\[\]']+)/g, '');
+          feed += `${line}\n`;
+        }
+        if (nameCheck.test(element)) {
+          line = element.replace(/^\s*name\s*=\s*/, '');
+          feed += `${line}\n`;
+        }
+      }
+      console.log(feed);
+      return feed;
```
humphd: We need to think about this flow a bit.  You're in a `then()` resolving your data, and want to parse it into items that can be returned.  Within this, you're in a `forEach` and you're returning within the callback to the `forEach`, which doesn't really do anything, since nothing will receive what you return.

Your `then()` should probably have a return before it exists, so whatever code consumes this Promise can do something with the processed data.

cryolis:
I did, but then it didn't show up in this PR, When I create a new branch and try to push new commit I have to create a separate PR for that.  I was only able to push the  changes to issue#59 branch on the upstream.

humphd:
OK, it's not perfect in terms of git history, but let's go with it and fix it up later in follow-up issues/PRs.  Thanks for this.

@mskuybeda, you can build on top of this now.

Issue #40, opened 2019-11-05
Implement pagination for front-end


ODAVING:
The current planet CDOT is long and arduous to read every post. I'm thinking of implementing a pagination in JS to use for the web-page. Feel free to add more suggestions with regards to this.

humphd:
@eekbatani, @ODAVING can we close this now that the pagination.js file has landed in tree?  If YES, please close this.

Issue #96, opened 2019-11-07
Combine EditConfig, ESLint, and Prettier to manage code style and formatting


vitokhangnguyen:
A noticeable example of this is the size of code indentation: ESLint is expecting 2 spaces while EditConfig is specifying 4.

I can uniform these miscellaneous things.

ragnarokatz:
@vitokhangnguyen you should be able to set the parameters in the eslint config file. please refer to this link: https://stackoverflow.com/questions/48902050/eslint-airbnb-with-4-spaces-for-indent

manekenpix:
@vitokhangnguyen I think the best solution would be getting rid of EditConfig. It seems ESLint is the one we're going to be using, so I don't see the need of having EditorConfig.

vitokhangnguyen:
@ragnarokatz Thank you for the link. I will fix the bug accordingly.

@manekenpix After some research, I find that there is still value to keep both. Although ESLint checks the code for quality and rule compliance, it does not adjust the configuration of the code editor. This can lead to situations that the contributor has to see a lot of errors/warnings during writing code before they run a command to lint the code or fix them manually.

manekenpix:
@vitokhangnguyen how about using [prettier](https://www.npmjs.com/package/prettier) then? It's an npm package and it [integrates with linter](https://prettier.io/docs/en/integrating-with-linters.html) quite smoothly. Would you think it'd be a better option?

vitokhangnguyen:
@manekenpix Funny that you mention it. During my reading, I came across this article: https://blog.theodo.com/2019/08/why-you-should-use-eslint-prettier-and-editorconfig-together/
I think that using a combination of these 3 is a good solution too.

I can go ahead and configure for these 3 to work together smoothly.

jerryshueh:
Which style are we going with? This conflict is causing eslint testing to fail, we should decide on a standard soon. There's also other default rules such as `noplusplus` which shouldn't affect the readability of the code. Unless we want to enforce rules like that as a standard going forward.

vitokhangnguyen:
We are  going with most of the rules of [Airbnb Coding Styke ](https://github.com/airbnb/javascript) with some changes, I suggest, such as indentation size of 4 spaces in stead of 2,

humphd:
By the way, my goal this week is to get all of EditorConfig, eslint, and Prettier integrated in our repo.  They can, and should!, live happily together.

humphd:
[...]

PR #137, opened 2019-11-10
added pagination to feed for #40

eekbatani:
I paginated the blog feed as per issue #40.

The current configuration shows three posts per page. This limit can be easily changed by changing the value of the "limitPerPage" variable at the top of pagination.js 

Even though our current blog posts are static placeholders, the pagination logic is ready to handle a dynamic feed without the need of any additional work.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,6 +2,7 @@ module.exports = {
   "extends": "airbnb-base",
   "env": {
     "jest": true,
+    "browser": true,
   },
   rules: {
     /**
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -5,16 +5,14 @@ module.exports = {
     "browser": true,
     "jquery": true
   },
-  "plugins": [
-    "dollar-sign",
-    "jquery"
-],
+
   rules: {
     /**
      * Disallow the use of console
      * https://eslint.org/docs/rules/no-console
      */
     "no-console": "off",
+
 
     /**
      * Disallow Reassignment of Function Parameters
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,7 +3,12 @@ module.exports = {
   "env": {
     "jest": true,
     "browser": true,
+    "jquery": true
   },
+  "plugins": [
+    "dollar-sign",
+    "jquery"
+],
   rules: {
     /**
      * Disallow the use of console
@@ -23,7 +28,5 @@ module.exports = {
      */
     "func-names": "off",
 
-  
-    'no-plusplus': 'off'
   }
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -22,5 +22,8 @@ module.exports = {
      * https://eslint.org/docs/rules/func-names
      */
     "func-names": "off",
+
+  
+    'no-plusplus': 'off'
   }
 };
--- a/index.html
+++ b/index.html
@@ -1,255 +1,455 @@
-<!doctype html>
 <html lang="en">
-
-<head>
-    <title>Planet CDOT</title>
-    <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
-    <meta name="generator" content="http://intertwingly.net/code/venus/">
-    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
-    <link rel="stylesheet" href="planet.css" type="text/css">
-    <!-- <link rel="alternate" href="http://zenit.senecac.on.ca/~chris.tyler/planet/atom.xml" title=""
-            type="application/atom+xml"> -->
-    <!-- CSS Bootstrap 4.0 -->
-
-    <!-- Font Awesome -->
-    <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.8.2/css/all.css">
-    <!-- Bootstrap core CSS -->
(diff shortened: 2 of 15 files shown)


humphd:
@eekbatani this is failing eslint.  Please fix all the issues you see when you run `npm run eslint`

.eslintrc.js
```diff
@@ -3,7 +3,12 @@ module.exports = {
   "env": {
     "jest": true,
     "browser": true,
+    "jquery": true
   },
+  "plugins": [
+    "dollar-sign",
```
humphd: I'm not familiar with this plugin.  Why have you added it, and what does it offer over the `jquery` plugin you use below?

pagination.js
```diff
@@ -1,3 +1,4 @@
+/*eslint-env jquery*/
```
humphd: This shouldn't be necessary if you define it in the `env` in `.eslintrc.js`, which you do.  Or am I missing something?

humphd:
You also need to rebase this and update re-run `npm install` to fix `package-lock.json`

pagination.js
eekbatani: My apologies, you're I was having some issues with the command earlier on my end because my NodeJs was out of date. I added those lines experimentally , I'll proceed to push a cleaner version.

cindyledev:
What's happening with that index.html file? There are so many lines being removed and added because of incorrect indentation.

humphd:
One thing to note: in #96 we're going to start using [Prettier](https://prettier.io/), which will make this problem go away.

Having said that, @cindyledev makes a great point: we should only touch lines associated with a fix, so we don't mess up git history for the remaining code.

eekbatani:
> What's happening with that index.html file? There are so many lines being removed and added because of incorrect indentation.

I had to add more sample blog posts to demonstrate the pagination functionality. Each post is a few paragraphs therefore many lines look like their being added.

.eslintrc.js
```diff
@@ -27,5 +25,14 @@ module.exports = {
      * https://eslint.org/docs/rules/func-names
      */
     "func-names": "off",
+
+    "max-len": ["error", 256, 2, {
```
humphd: Why alter these rules?  Move your comments above the line they affect vs. at the end.  Lines this long are not good.

.eslintrc.js
```diff
@@ -27,5 +25,14 @@ module.exports = {
...
     "func-names": "off",
+
+    "max-len": ["error", 256, 2, {
+            ignoreUrls: true,
+            ignoreComments: false,
+            ignoreRegExpLiterals: true,
+            ignoreStrings: false,
+            ignoreTemplateLiterals: false,
+          }],
+    'no-plusplus': 'off'
```
humphd: Why this change?

eekbatani: The "no-plusplus" rule by default disables unary operators.  The unary operator is used several times in the pagination logic, and is a more efficient way to write the code as it reduces the number of lines needed. I didn't see why it should be omitted, so I set it to 'off'.

As for the rest, I will refactor the code to abide by the default rules and remove those conditions.

.eslintrc.js
eekbatani: Good point, I'll focus on changing the code instead of the rules.

.eslintrc.js
humphd: Often these things are disabled due to bugs that don't seem obvious, but people find through experience.  You can read a bit about why they disabled it in https://github.com/airbnb/javascript/issues/540.

In modern JS, using for-loops like you're doing with a lot of `i++` isn't really necessary anymore.  It's more bug prone than using for-of, `forEach`, `map`, etc.  I'd rewrite your code to not need to use an index counter at all.  That's the point of this rule.

eekbatani: Okay fair enough I'll make the changes.

index.html
```diff
@@ -1,255 +1,455 @@
...
-
-    <!-- Font Awesome -->
-    <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.8.2/css/all.css">
-    <!-- Bootstrap core CSS -->
-    <link href="https://cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/4.3.1/css/bootstrap.min.css" rel="stylesheet">
-    <!-- Material Design Bootstrap -->
-    <link href="https://cdnjs.cloudflare.com/ajax/libs/mdbootstrap/4.8.10/css/mdb.min.css" rel="stylesheet">
-
-
-</head>
```
cindyledev: This doesn't needed to be removed. I see that you added the exact same lines after so it won't effect the results but this code was originally added by two other contributors. Let's not alter the git history.

eekbatani: I did not remove or move those lines (4 to 22) that you are referring to. They are still present. Only their alignment has been corrected when I made my own additions to the file as per standard HTML conventions.  As you said, this has no effect on the actual output, nor has it changed any of the work of the original contributors.
image

humphd:
Agree, we can allow syntax changes we won't later, since we're about to reformat all the code in #139 soon.

humphd:
A side note: one thing that this PR does that I *don't* want you to do is `Merge branch 'master' into addPagination`.  Use rebase instead when you need to update.

Looks like this has enough reviews to land, so if you're done, go ahead and merge it and we'll iterate in follow-ups.

Issue #120, opened 2019-11-09
Add coverage reporting to Jest


humphd:
Having tests is great, but only if they test the code we have: all functions, branches within those functions (e.g., if/else), etc.  It can be hard to know if you are testing everything in your code without getting detailed reports.

One tool we can use is a test coverage report.  [Jest will automatically](https://jestjs.io/docs/en/cli#coverage) do this if we configure it.  Someone should enable code coverage so we can see if we're missing some part of a file.

jpjjulie:
Hi, I would like to work on this issue.

miggs125:
FYI this can be done by configuring jest by adding the following to the package.json file:
```json
"jest": {
    "collectCoverage": true,
    "coverageReporters": ["json", "html"],
    "testEnvironment": "node",
    "coveragePathIgnorePatterns": [
      "/node_modules/"
    ],
  }
```
Here we are telling jest to collect coverage, the type of format, ignore coverage for node_modules and the test environment (node). You can also opt to configure jest in a ```jest.config.js ``` file instead. Depending on the amount of configuration you'd like to include, the latter might make more sense and would make things easier to maintain.

You can find more information on configuring jest [here](https://jestjs.io/docs/en/configuration). If you need any help feel free to ping me here on on slack and I'll be happy to help :)

jpjjulie:
Thank you very much @miggs125

Issue #35, opened 2019-11-05
Add logging support to our backend


humphd:
Let's do proper logging for everything in the back-end using [Pino](http://getpino.io/#/).  Ideally, I'd like to see us logging things for regular operations, errors, but also offer debug logs to help developers see what's going on.

We can roll this out in stages, but it would be good to add initial logging support with at least INFO or ERROR levels.

hansal7014:
Hi @humphd,
I would like to work on this issue.

humphd:
Awesome @hansal7014, go for it!

PR #141, opened 2019-11-11
Fix #63: Add an Express Web Server

### LINKED ISSUE #63: Create HTTP server
We need to create a web server for telescope. We should add what has been mentioned in #62 to this.

miggs125:
fixes #63. This adds nodemon for debug builds as well as a started web server file with a health check. We can discuss additional endpoints here and I can add them in as we discuss them. I've added minimist so that the server can be started on a different port if needed. This can be used like so:  

```bash
 node server.js --port=4040  # starts server at port 4040 
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
     "pretest": "npm run lint",
     "test": "npm run jest",
     "jest": "jest",
-    "start": "node src"
+    "start": "node src",
+    "debug-server": "nodemon server.js"
   },
   "repository": "Seneca-CDOT/telescope",
   "license": "BSD-2-Clause",
@@ -23,6 +24,8 @@
     "bent": "^7.0.2",
     "bull": "^3.11.0",
     "dotenv": "^8.2.0",
+    "express": "^4.17.1",
+    "express-healthcheck": "^0.1.0",
     "feedparser-promised": "^2.0.1",
     "sentiment": "^5.0.2"
   },
@@ -31,7 +34,9 @@
     "eslint-config-airbnb-base": "^14.0.0",
     "eslint-plugin-import": "^2.18.2",
     "jest": "^24.9.0",
+    "nodemon": "^1.19.4",
     "stylelint": "^11.1.1",
-    "stylelint-config-recommended": "^3.0.0"
+    "stylelint-config-recommended": "^3.0.0",
+    "supertest": "^4.0.2"
   }
 }
--- a/test/server.test.js
+++ b/test/server.test.js
@@ -1,0 +1,18 @@
+const request = require('supertest');
+const app = require('../src/backend/web/app');
+
+describe('Health Check', () => {
+  it('should return a JSON object with property "uptime"', async () => {
+    const res = await request(app).get('/health');
+    expect(res.statusCode).toEqual(200);
+    expect(res.body).toHaveProperty('uptime');
+  });
+});
+
+describe('GET "/" endpoint', () => {
+  it('should return status 200 and content-type of text/html', async () => {
+    const res = await request(app).get('/');
+    expect(res.status).toEqual(200);
+    expect(res.get('Content-type')).toContain('text/html');
+  });
+});
--- a/package-lock.json
+++ b/package-lock.json
@@ -567,6 +567,21 @@
       "integrity": "sha512-2scffjvioEmNz0OyDSLGWDfKCVwaKc6l9Pm9kOIREU13ClXZvHpg/nRL5xyjSSSLhOnXqft2HpsAzNEEA8cFFg==",
       "dev": true
     },
+    "abbrev": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
+      "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==",
+      "dev": true
+    },
+    "accepts": {
+      "version": "1.3.7",
+      "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.7.tgz",
+      "integrity": "sha512-Il80Qs2WjYlJIBNzNkK6KYqlVMTbZLXgHx2oT0pU/fjRHyEp+PEfEPY0R3WCwAGVOtauxh1hOxNgIf5bv7dQpA==",
+      "requires": {
+        "mime-types": "~2.1.24",
+        "negotiator": "0.6.2"
+      }
+    },
     "acorn": {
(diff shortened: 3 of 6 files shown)


miggs125:
Oh wow I totally missed that. Will fix tonight!

package.json
```diff
@@ -23,14 +24,19 @@
     "bent": "^7.0.2",
     "bull": "^3.11.0",
     "dotenv": "^8.2.0",
+    "express": "^4.17.1",
+    "express-healthcheck": "^0.1.0",
     "feedparser-promised": "^2.0.1",
+    "minimist": "^1.2.0",
```
humphd: Let's remove `minimist` and `opn`.  The latter isn't used at all, and the former isn't needed, since we won't be starting the server directly from the command line (see comments below).

package.json
```diff
@@ -11,7 +11,8 @@
     "pretest": "npm run lint",
     "test": "npm run jest",
     "jest": "jest",
-    "start": "node src"
+    "start": "node src",
+    "debug": "nodemon server.js"
```
humphd: I think `debug` is too vague, since we are potentially going to have multiple entry points/apps running at the same time.  Should this be `web` and `web-debug` or `server` and `server-debug`?

Also, you've removed `src/index.js` (by accident I hope?), and we should think about integrating the code in there with the web server startup.  See my comment below on this.

server.js
```diff
@@ -0,0 +1,24 @@
+require('./src/config');
+const express = require('express');
+const argv = require('minimist');
+const path = require('path');
+const healthcheck = require('express-healthcheck');
+
+const args = argv(process.argv.slice(2));
+const HTTP_PORT = process.env.PORT || args.port || 3000;
```
humphd: Let's not allow passing args in like this. We have an `env` system in place, let's just use that.  If people want to override it on the command line, they can with `PORT=12345 npm start`.

server.js
```diff
@@ -0,0 +1,24 @@
...
+const server = express();
```
humphd: Let's follow best practices and break the definition of express app away from the web server, see https://github.com/goldbergyoni/nodebestpractices/blob/master/sections/projectstructre/separateexpress.md.  This will make testing a lot easier.

server.js
```diff
@@ -0,0 +1,24 @@
...
+server.use('/health', healthcheck());
+
+// *********************  ROUTES  *****************************
+
+server.get('/', (req, res) => {
+  res.sendFile(path.join(__dirname, 'index.html'));
+});
+
+server.listen(HTTP_PORT, () => {
+  console.log(`Telescope listening on port ${HTTP_PORT}`);
```
humphd: When #35 lands, we can switch to use proper logging here.  cc @hansal7014

src/index.js
```diff
@@ -1,52 +0,0 @@
-/**
```
humphd: Please restore this file.

server.js
```diff
@@ -0,0 +1,24 @@
...
+const healthcheck = require('express-healthcheck');
+
+const args = argv(process.argv.slice(2));
+const HTTP_PORT = process.env.PORT || args.port || 3000;
+const server = express();
+
+
+// ********************* MIDDLEWARE **************************
+
+server.use('/health', healthcheck());
```
humphd: This would be pretty easy to write a test for, see https://github.com/goldbergyoni/nodebestpractices/blob/master/sections/projectstructre/separateexpress.md#example-test-your-api-in-process-using-supertest-popular-testing-package

server.js
```diff
@@ -0,0 +1,24 @@
...
+
+// *********************  ROUTES  *****************************
+
+server.get('/', (req, res) => {
```
humphd: This should also have a basic test

Grommers00:
any update on this?

miggs125:
@Grommers00 I updated changes to reflect the ongoing discussion. Waiting for reviews.

package.json
humphd: This is still outstanding...

server.js
```diff
@@ -0,0 +1,16 @@
...
+const app = require('./app.js');
+
+const HTTP_PORT = process.env.PORT || 3000;
+
+const server = app.listen(HTTP_PORT, () => {
+  console.log(`Telescope listening on port ${HTTP_PORT}`);
+});
+
+process.on('unhandledRejection', (err) => {
+  console.log('UNHANDLED REJECTION:  Shutting down...');
```
humphd: Once #142 lands, we can move all this into proper logs.  For now, this is good.

app.js
```diff
@@ -0,0 +1,12 @@
+const express = require('express');
+const healthcheck = require('express-healthcheck');
+
+const router = require('./src/routes/index');
+
+const app = express();
+
+app.use('/health', healthcheck());
```
humphd: Question: why here and not in your router?

app.js
```diff
@@ -0,0 +1,12 @@
+const express = require('express');
+const healthcheck = require('express-healthcheck');
+
+const router = require('./src/routes/index');
```
humphd: ```const router = require('./src/routes');``` (no need for `index`

package.json
```diff
@@ -31,7 +34,15 @@
     "eslint-config-airbnb-base": "^14.0.0",
     "eslint-plugin-import": "^2.18.2",
     "jest": "^24.9.0",
+    "nodemon": "^1.19.4",
     "stylelint": "^11.1.1",
-    "stylelint-config-recommended": "^3.0.0"
+    "stylelint-config-recommended": "^3.0.0",
+    "supertest": "^4.0.2"
+  },
+  "jest": {
```
humphd: This seems unrelated to your changes.  We have an issue on #120, can you comment there, and they can add this?

test/server.test.js
```diff
@@ -0,0 +1,10 @@
+const request = require('supertest');
+const app = require('../app');
+
+describe('Health Check', () => {
+  it('should return a JSON object with property "uptime"', async () => {
+    const res = await request(app).get('/health');
+    expect(res.statusCode).toEqual(200);
+    expect(res.body).toHaveProperty('uptime');
+  });
+});
```
humphd: Let's add one more test that `/` returns a 200 and a content-type of text/html...

app.js
miggs125: I added it here because it is middleware so I figured it belonged in app.js but it also makes sense to have it in the router. If it makes more sense to have it in router I will move it there.

package.json
miggs125: yup! I'll remove it from here and comment in #120

miggs125:
@humphd I've made the changes requested. I will make changes to where the /health middleware will be included pending the outcome of our /health middlware [conversation](https://github.com/Seneca-CDOT/telescope/pull/141#discussion_r345855477). Once this is resolved, I will rebase all of the commits into something clearer and comprehensive commit message.

Issue #108, opened 2019-11-08
Hook new feed parser into feed worker


robertbegna:
Hello, 
My previous requests weren't able to get merged, and I am unsure of how to contribute to this project. Does anyone have any issues they could recommend for me to try?

humphd:
I'm going to morph this into a bug for @robertbegna.  We are about to land #37, which adds url to xml feed parsing.  We need to hook that into our feed worker.  We can do something really basic to start, and I've got a model in https://github.com/Seneca-CDOT/telescope/pull/37#pullrequestreview-314522718.

jayson528:
i spoke with the prof in class and i was assigned to this issue

PR #143, opened 2019-11-11
Fixes #22: Inactive Blog Filter

### LINKED ISSUE #22: Implement a way to filter active and inactive bloggers
There is a long list of bloggers but a lot of them do not blog anymore. We should have a way to show currently active bloggers and hide the ghosts.

jerryshueh:
Addresses #22. Had to perform some workarounds on this one as we do not yet have an official feed-worker implemented. That issue has been assigned to @robertbegna in issue #108, and I didn't want to take away from any work he might want to do. In order to test my implementation, I had to create a new, altered feed-worker in _/test_.

A _feeds-redlist.json_ was created in _root_, which contains filtered feeds. Since they are not permanently banned or anything, I didn't want to call it a "blacklist". Each item contains an `url` attribute for the feed, and a `lastUpdate` attribute (ISO 8601 datetime string) for the blog's last known activity, according to the feed data. A 0 value means the blog has a dead feed.

New module created in _/src/inactive-blog-filter.js_, which performs most of the logic:
* First is option `check()`, which takes a feed URL and a callback function. It will check a provided feed url against the redlist file and return the result (true/false) to the callback, which can then perform whatever task. 
* Second is `update()`, which performs a complete sweep of all feeds on the feedlist to check if they are inactive or not. Then it re-writes the redlist file to reflect the latest changes. This may become an issue if the feedlist grows, so for now, we'll say we can run this periodically and not live with each feed update. The reason I chose to do this is because, again, we do not have a proper feed-worker to accumulate data, nor a proper database to search. Once we move forward, this module can be updated to perform specific searches on stored blogs.

### DIFF
--- a/env.example
+++ b/env.example
@@ -7,11 +7,15 @@ PORT=8080
 # REDIS_URL specifies Redis server info
 REDIS_URL=redis://127.0.0.1:6379
 
-# NODEMAILER_USERNAME is sender's username credential	
-NODEMAILER_USERNAME=	
+# NODEMAILER_USERNAME is sender's username credential
+NODEMAILER_USERNAME=
 
-# NODEMAILER_PASSWORD is sender's password credential	
+# NODEMAILER_PASSWORD is sender's password credential
 NODEMAILER_PASSWORD=
 
 # NODEMAILER_SERVER is sender's server that they are using (Example: smtp.example.com)
-NODEMAILER_SERVER=	
+NODEMAILER_SERVER=
+
+# BLOG_INACTIVE_TIME is the period (days) of inactivity
+# before a blog will be considered redlisted
+BLOG_INACTIVE_TIME=360
--- a/feeds-redlist.json
+++ b/feeds-redlist.json
@@ -1,0 +1,30 @@
+[
+  {
+    "url": "http://ajhooper.blogspot.com/feeds/posts/default",
+    "lastUpdate": "2008-10-18T17:22:32.366Z"
+  },
+  {
+    "url": "http://KrazyDre.blogspot.com/feeds/posts/default?alt=rss",
+    "lastUpdate": null
+  },
+  {
+    "url": "http://nadavid.blogspot.com/feeds/posts/default",
+    "lastUpdate": "2008-09-07T22:12:00.544Z"
+  },
+  {
+    "url": "http://ljubomirgorscak.blogspot.com/feeds/posts/default",
+    "lastUpdate": null
+  },
+  {
+    "url": "http://dcucereavii.blogspot.com/feeds/posts/default?alt=rss",
+    "lastUpdate": "2009-11-17T00:03:00.000Z"
+  },
+  {
+    "url": "http://gkrilov.blogspot.com/feeds/posts/default",
+    "lastUpdate": "2009-09-19T04:17:33.156Z"
+  },
+  {
+    "url": "http://nashutzu.blogspot.com/feeds/posts/default",
+    "lastUpdate": null
+  }
+]
--- a/src/inactive-blog-filter.js
+++ b/src/inactive-blog-filter.js
@@ -1,0 +1,143 @@
+/**
+ * Returns whether or not the blog should be marked as inactive
+ * Criteria for filtering (in milliseconds) is based on filters.json under blog.inactive
+ */
+require('./config.js');
+const fs = require('fs');
+const feedParser = require('./feed-parser');
+
+/**
+ * Condition for passing redlist some() check
+ * @param {string} feedUrl - url of the feed to check against redlist
+ * @param {Object} redItem - object containing a redlist entry
+ * @returns {boolean} - bool value
+ */
+function isRedlisted(feedUrl, redItem) {
+  return (feedUrl === redItem.url);
+}
+
+/**
+ * Simple function that takes the difference between current date and post date
+ * If difference is greater than threshold, the blog is considered inactive
+ * @param {Date} postDate - javascript date object of the post being checked
+ * @returns {number} - date difference in milliseconds
+ */
+function dateDiff(postDate) {
+  const currentDate = new Date();
+  return currentDate - postDate;
+}
+
+/**
+* Callback for redlist check
+* @callback checkCallback
+* @param {boolean} result - true/false whether or not a feed url is redlisted
+*/
+
+/**
+ * Checks if feed url is redlisted
+ * @param {string} feedUrl - url of the feed to check against redlist
+ * @param {checkCallback} callback - a callback that runs after the check
+ */
+function check(feedUrl, callback) {
+  // Read redlist file
+  fs.readFile('feeds-redlist.json', 'utf-8', (err, redListRaw) => {
+    if (err) {
+      // Error reading file
+      callback(err, false);
+      return;
+    }
+
+    if (redListRaw.length === 0) {
+      // File is empty
+      callback(undefined, false);
+      return;
+    }
+
+    // Concat to array no matter what
+    const redList = [].concat(JSON.parse(redListRaw));
+
+    callback(null, redList.some(isRedlisted.bind(null, feedUrl)));
+  });
+}
+
+/*
+ * Performs a separate sweep of all feeds to see which ones are inactive,
+ * then updates last post date in feeds-redlist.json
+ *
+ * Integration with system can be improved once feed-worker is augmented to
+ * pass feed data more intuitively
+ *
+ * Due to amount of operations, this can be run periodically instead of with every feed update
+ */
+function update() {
+  // Read the feeds list file
+  fs.readFile('feeds.txt', 'utf8', (err, lines) => {
+    if (err) {
+      console.error('unable to read initial list of feeds, cannot update', err.message);
+      return;
+    }
+
+    // Divide the file into separate lines
+    const feedUrlList = lines
+      .split(/\r?\n/)
+      // Basic filtering to remove any ines that don't look like a feed URL
+      .filter((line) => line.startsWith('http'))
+      // Convert this into an Object of the form expected by our queue
+      .map((url) => ({ url }));
+
+    const redlistUpdate = [];
+    let linesRead = 0;
+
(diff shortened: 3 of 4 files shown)


src/inactive-blog-filter.js
```diff
@@ -0,0 +1,143 @@
...
+*/
+
+/**
+ * Checks if feed url is redlisted
+ * @param {string} feedUrl - url of the feed to check against redlist
+ * @param {checkCallback} callback - a callback that runs after the check
+ */
+function check(feedUrl, callback) {
+  // Read redlist file
+  fs.readFile('feeds-redlist.json', (err, redListRaw) => {
```
humphd: A tip: in node you can just use `require('./relative/path/to/json/file.json')` to read and parse JSON.  Wrap it in a try/catch to deal with parsing errors.

src/inactive-blog-filter.js
```diff
@@ -0,0 +1,143 @@
...
+
+      let recentPostDate = new Date();
+
+      // In case of invalid/ dead feeds
+      if (typeof feed[0] !== 'undefined') {
+        recentPostDate = new Date(feed[0].date);
+
+        // Check if the blog is inactive
+        // !!!  Older than 1 year = inactive, but we can change this or
+        // !!!  parameterize it in a config somewhere
```
humphd: We have a config in `.env`.  You need to add something to the `env.sample` file declaring this with a named environment variable, and read it in from `process.env`

jerryshueh: Oops, didn't notice this! I'll address it asap.

humphd: You should [watch the repo](https://help.github.com/en/github/receiving-notifications-about-activity-on-github/watching-and-unwatching-repositories) so you can follow what's happening.

src/inactive-blog-filter.js
jerryshueh: Neat trick, I'll make the adjustments!

jerryshueh: So it seems non-global requires go against airbnb as I'm getting an eslint error. Is there any way around this, or should I revert back to the standard method with readFile?

jerryshueh:
Made the changes to env.example so the filter module now calls process.env to retrieve variables. Non-global requires go against airnbn and fails eslint, so I've reverted back to traditional readFile for now.

env.example
```diff
@@ -3,3 +3,7 @@ NODE_ENV=development
 
 # PORT is the port used by the web server
 PORT=8080
+
+# INACTIVETIME is the period (millisecond) of inactivity
```
humphd: Can we do this in some unit other than `ms` and you calculate your `ms` based on that?  I'd suggest weeks, months, or days are more understandable here.

env.example
```diff
@@ -3,3 +3,7 @@ NODE_ENV=development
...
+# before a blog will be redlisted
+INACTIVETIME=31556952000
```
humphd: `BLOG_INTERACTIVE_TIME`

feeds-redlist.json
```diff
@@ -0,0 +1,30 @@
+[
+  {
+    "url": "http://nashutzu.blogspot.com/feeds/posts/default",
+    "lastUpdate": 0
```
humphd: Sometimes you use a `Number` and others a Date serialized to a `String`.  I'd suggest `null` if you mean "not set".

src/inactive-blog-filter.js
```diff
@@ -0,0 +1,145 @@
+/**
+ * Returns whether or not the blog should be marked as inactive
+ * Criteria for filtering (in milliseconds) is based on filters.json under blog.inactive
+ */
+const fs = require('fs');
+const dotenv = require('dotenv');
```
humphd: Use our `config` module from `src/config`, which will do this once.

src/inactive-blog-filter.js
humphd: OK, leave it then.

src/inactive-blog-filter.js
```diff
@@ -0,0 +1,145 @@
...
+ * @param {string} feedUrl - url of the feed to check against redlist
+ * @param {checkCallback} callback - a callback that runs after the check
+ */
+function check(feedUrl, callback) {
+  // Read redlist file
+  fs.readFile('feeds-redlist.json', (err, redListRaw) => {
+    if (err) {
+      // Error reading file
+      console.warn('problem reading redlist file:', err.message);
+      callback(false);
```
humphd: In node, you always propagate the error back as the first arg.  In this case, `callback(err)`

src/inactive-blog-filter.js
```diff
@@ -0,0 +1,145 @@
...
+      return;
+    }
+
+    if (redListRaw.length === 0) {
+      // File is empty
+      console.warn('redlist file is empty, no filters applied\n\tsee: feeds-redlist.json');
+      callback(false);
```
humphd: `callback()` which will use `undefined` for the error arg.

src/inactive-blog-filter.js
```diff
@@ -0,0 +1,145 @@
...
+      return;
+    }
+
+    // Concat to array no matter what
+    const redList = [].concat(JSON.parse(redListRaw));
+
+    callback(redList.some(isRedlisted.bind(null, feedUrl)));
```
humphd: `callback(null, redList.some(isRedlisted.bind(null, feedUrl)));` first arg is error, should be null here.

src/inactive-blog-filter.js
```diff
@@ -0,0 +1,145 @@
...
+*/
+
+/**
+ * Checks if feed url is redlisted
+ * @param {string} feedUrl - url of the feed to check against redlist
+ * @param {checkCallback} callback - a callback that runs after the check
+ */
+function check(feedUrl, callback) {
+  // Read redlist file
+  fs.readFile('feeds-redlist.json', (err, redListRaw) => {
```
humphd: `fs.readFile('feeds-redlist.json', 'utf8', (err, redListRaw) => {` for a text file

src/inactive-blog-filter.js
```diff
@@ -0,0 +1,145 @@
...
+/*
+ * Performs a separate sweep of all feeds to see which ones are inactive,
+ * then updates last post date in feeds-redlist.json
+ *
+ * Integration with system can be improved once feed-worker is augmented to
+ * pass feed data more intuitively
+ *
+ * Due to amount of operations, this can be run periodically instead of with every feed update
+ */
+function update() {
```
humphd: The `update()` function is async, should we return a promise and only `resolve()` when it's done below?

test/feed-worker-filtertest.js
```diff
@@ -0,0 +1,14 @@
+const feedQueue = require('../src/feed-queue');
```
humphd: This isn't really a test.  Did you want to actually write some tests using Jest for this code and add to `test/` or file an Issue for someone to do it later?

wajeehsheikh: @humphd @jerryshueh 

I can create a tester for this.

env.example
jerryshueh: Let's do days. I think that's a denomination that's easy enough to understand.

env.example
jerryshueh: I'll address this, but before I do, did you mean `BLOG_INACTIVE_TIME`?

humphd: Yeah, sorry, `BLOG_INACTIVE_TIME`, good catch.

feeds-redlist.json
jerryshueh: Good point! I'll have it write a null for unresponsive/ dead feeds.

src/inactive-blog-filter.js
jerryshueh: Oh, that's right! I'll fix this.

test/feed-worker-filtertest.js
jerryshueh: Looks like it can be a good issue for people who want to jump in on making test. Let's post this formally after this PR gets passed.

src/inactive-blog-filter.js
jerryshueh: This will be addressed in the next PR push.

src/inactive-blog-filter.js
jerryshueh: I will address this for the next PR push.

src/inactive-blog-filter.js
jerryshueh: I will address this for the next PR push.

src/inactive-blog-filter.js
jerryshueh: `update()` itself is not async, although it does use `fs.readfile`, but I chain directly off it with an anonymous callback. If we would like to make `update()` async, perhaps we could frame it as a separate issue?

src/inactive-blog-filter.js
jerryshueh: I will address this in the next PR push.

jerryshueh:
Pushed new changes to PR: 
* The blog filter module now correctly requires _./config.js_ to load env variables. 
* `check()` now uses error-first callback. 
* Redlist uses `null` to denote "dead' feeds.
* `env.INACTIVETIME` changed to `env.BLOG_INACTIVE_TIME` and represents time in days instead of milliseconds

jerryshueh:
Looks like some changes have been made since I rebased, so now there's a conflict with env.example. It will be fixed with the next rebase before the merge, if everything else checks out.

src/inactive-blog-filter.js
UltimaBGD: I believe you mention you addressed this in another push. If so, this should be resolved?

jerryshueh: @UltimaBGD Yes, I was waiting on the community to see if anyone had any inputs. Since it seems to be okay, I'll resolve this conversation and post an issue when this PR gets successfully merged.

jerryshueh:
Sorry, I had to rebase this PR again as there were several merges during the time I addressed the last fix. If I could get another set of reviews that would be great.

PR #149, opened 2019-11-11
Fix 148 added page view counter for the main page

robertbegna:
Hello all,
So this pull request serves as possible solution to implementing a feature that tracks the number of visitors to the CDOT Planet.
I added a simple function that tracks visit to the home page based on the localStorage in the browser for now, 
I also thought it could be a good idea to track visits to individual blog, that way a blog's popularity could be determined from the list. However, doing this requires storing view counts to a database, and that hasn't been uploaded to master as far as I can tell, so I guess it's a feature that can be impelmented down the line.

### DIFF
--- a/index.html
+++ b/index.html
@@ -12,6 +12,8 @@
 
 <body>
     <h1>Planet CDOT</h1>
+    <div id="CounterVisitor">
+    </div>
 
     <div class="daygroup">
         <h2>October 30, 2019</h2>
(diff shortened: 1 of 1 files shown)


humphd:
This is only going to show you how many times you've visited this page from your own browser, though.  When we have a full page later we can add Google Analytics or the like.

c3ho:
Hi, I know this is a closed PR, but I'd like to reference each of the blogging sites we use should have an API to retrieve views for a specific post such as [this](https://developers.google.com/blogger/docs/3.0/reference/pageViews/get) one for blogger.

PR #152, opened 2019-11-12
Fix #151 - Confine CSS to its own directory and update stylelint

### LINKED ISSUE #151: Confine CSS to its own directory, update stylelint
In https://github.com/Seneca-CDOT/telescope/pull/66#discussion_r344993179 we ran into an issue where our `stylelint` checks are looking at files that aren't relevant.  Right now, we're [checking every file that *isn't* being ignored](https://github.com/Seneca-CDOT/telescope/blob/master/package.json#L8).  Let's flip this around and *only* check files in certain paths.

To do this, let's move our frontend code to `src/web/*` or something, and within that create `src/web/styles/*`.  We can move our stylesheet(s) there, and tell `stylelint` to only look at that directory for now vs. our entire tree.

lozinska:
.css file moved to it's own directory src/style and stylelint updated in order to check only directory with front-end


package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "eslint": "eslint --ignore-path .gitignore .",
     "eslint-fix": "eslint --fix --ignore-path .gitignore .",
-    "stylelint": "stylelint --ignore-path .gitignore .",
+    "stylelint": "stylelint src/style/*.css",
```
humphd: Can you also remove all the `ignore` portion of the stylelintrc file and still have everything work?  I suspect so, but please try and update this if so.

src/style/planet.css
```diff
@@ -0,0 +1,241 @@
+/* Regular theme starts */
```
humphd: Did you `git mv` this file?  It should show it as a rename vs. a new file being created.

package.json
lozinska: I will remove the ignore portion in stylelintrc file.

src/style/planet.css
lozinska: I didn't do git mv and now I understand why I saw that new file was created...What should I do now? Reverse changes that was done?

humphd:
```
git rm src/style/planet.css
mkdir -p src/frontend/style
git mv planet.css src/frontend/style
```

NOTE: I've changed this to include a `src/frontend` parent dir to separate the frontend code from the server, parsers, etc.

lozinska: Thank you @humphd

lozinska:
I created directory frontend directory where I create style and moved planet.css in there, also I removed ignore part from stylelintrc file

lozinska:
Hi @humphd you mentioned that I have failures. But I can't spot where is the problem. Moreover I don't  have mentioned about .png files

humphd:
I have the wrong PR number.  There is one PR open right now that is failing CSS stylelint due to png files.  It shouldn't be looking at these files, which your PR here will fix.  I was saying, "This PR will fix that other problem."

You don't have any failures, this is good to go.  We just need another reviewer!  See if you can find one so we can get this merged.

lozinska:
Ok @humphd, I understand now. Thank you for your comments

humphd:
landed!

PR #157, opened 2019-11-12
Fix #119 - update eslint with environments currently used

### LINKED ISSUE #119: Update eslint env for all environments we are using
In `.eslintrc.js`, we currently define our environments like so:

```js
"env": {
    "jest": true,
    "browser": true,
},
```

This is good, but we should add a few more from the [list of possible env](https://eslint.org/docs/user-guide/configuring#specifying-environments), specifically:

- `node` - for node.js
- `es2017` - for ECMAScript 2017, which has `async`/`await`

RyanMarzec:
updated the eslint env to include the current environments we are using.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,7 +3,9 @@ module.exports = {
   "env": {
     "jest": true,
     "browser": true,
-    "jquery": true
+    "jquery": true,
+    "node": true,
+    "es2017": true
   },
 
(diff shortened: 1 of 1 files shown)


humphd:
Updated title so this will autoclose the Issue.  @RyanMarzec you have to use "Fix #number" or fixes or closes, etc.

PR #158, opened 2019-11-12
Fixes 148: Added a Page Hit Counter Function

robertbegna:
In this pull request, I implemented a simple Hit Counter that can store the number of visitors to main CDOT page to a Redis Database. The port number can be changed to the one selected by those making the database.
Aside from tracking visitors to the main page, I thought it could also be great to track the number of visitors to individual blogs. Perhaps this function can be updated to address that feature as well.

### DIFF
--- a/counter.js
+++ b/counter.js
@@ -1,0 +1,8 @@
+var db = require("redis-client").createClient();
+require("http").createServer(function(request, response) {
+    db.incr("count",function(err, reply) {
+        response.writeHead(200, { "Content-Type": "text/plain" });
+        response.write(reply.toString());
+        response.end();
+    });
+}).listen(8181);


miggs125:
@robertbegna Also, I found the **exact** code from this [gist](https://gist.github.com/fbrnc/19c3ce69bbc5b4083b5a3653109f2148) with a simple google search. If you are going to use someone else's code in another project, you should make sure that the licences are compatible AND give credit where it is due.

c3ho:
on top of what @miggs125 has mentioned, please make sure the PR passes the CI tests as people will not review them otherwise.

humphd:
I'm going to close this for now, and we can think about analytics later in the front end with Google Analytics or the like.  Robert has another bug to work on as it is.

PR #142, opened 2019-11-11
Fixes #35 - Add logging support to our backend.

hansal7014:
### What changes are being made? (feature)
This pull request adds a **logger.js** module that exports a logger instance to be used in other modules to implement logging for important events. Implementing the logger is also documented in the **logging.md** file.

### Why are these changes necessary? Link any related issues
Adds logging support to our backend using Pino as requested in issue #35 .

PR #123, opened 2019-11-09
Fix #74: Validate blog URLs.

Cadesh:
I tested some ways to validate the URLs, the one that worked better was with valid-url [(https://www.npmjs.com/package/valid-url)](url)

These are the changes to the code:
1) Added a new file utils.js with a function to validade URLs. 
2) Added a file utils.test.js with a test for the new function.

PR #102, opened 2019-11-08
Added Extracting of URLs from blog feed function

smilegodly:
I wrote some code that is meant to get us started with extracting URL's from blog feeds . 
This code takes in a blog feed and gives back all the URL's in the feed.
All these URL's will later be categorized by what type of URL, is it a  pull request, a issue, etc..
Fixes: #53

PR #166, opened 2019-11-13
Fix #94: add function for counting domains

### LINKED ISSUE #94: Implement blog post count by domain
It is good to add a function to count the number of blog posts from each domain, such as Github, Wordpress, Medium, etc.

KarlaChen:
Added domain-analysis.js module for counting the number of blog posts from different domains. The function returns an array of json object with domain name and count properties.


src/domain-analysis.js
```diff
@@ -0,0 +1,44 @@
...
+    }
+    return domainCounter;
+};
+
+function countDomain(url){
+    
+}
+
+function findDomain(url){
+    var userDomain = url.replace('http://','').replace('https://','').split('/')[0];
```
humphd: It's really hard to do this for all cases with a URL (i.e., operate on strings).  Instead, use node's built in [`url` module and its `parse` method](https://nodejs.org/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost).  Here's how it works:

```js
const url = require('url');
url.parse('https://github.com/Seneca-CDOT/telescope/pull/166/files');
/* gives you an object like this:
Url {
  protocol: 'https:',
  slashes: true,
  auth: null,
  host: 'github.com',
  port: null,
  hostname: 'github.com',
  hash: null,
  search: null,
  query: null,
  pathname: '/Seneca-CDOT/telescope/pull/166/files',
  path: '/Seneca-CDOT/telescope/pull/166/files',
  href: 'https://github.com/Seneca-CDOT/telescope/pull/166/files' 
}
*/
```

src/domain-analysis.js
```diff
@@ -0,0 +1,44 @@
+/* This file contains the code for analyzing number of blog posts from different domains.
+ * The module receives an array of blog urls and returns an array of json object in format
+ * {name: domainName, count: number of frequency}
+*/
+
+
+module.exports.countDomain = function (url) {
+    var domainCounter = []
+
+    for (let i = 0; i < url.length; i++ ){
```
humphd: This code is really similar to what @smilegodly is doing in #102.  I think you should probably build on what he has, and use the list of URLs his code returns to do the count, then we can store that in Redis later.

src/domain-analysis.js
```diff
@@ -0,0 +1,44 @@
...
+                domainObject.count = 1;
+                domainCounter.push(domainObject);
+            }
+        }
+    }
+    return domainCounter;
+};
+
+function countDomain(url){
+
```
humphd: Why is this empty function here?

src/domain-analysis.js
KarlaChen: @humphd  Got it, I will fix it today.

haichuan0424: @humphd 
We are trying to count the number of blog posts from each domain, not counting number of url domains inside each post, so it should be different from #102. We can simply read from an array of urls and use url.parse to get the hostname and push or add count an object like {name: domainName, count: number}  to a domainList array and return this list.

src/domain-analysis.js
```diff
@@ -0,0 +1,50 @@
...
+/* createDomainObject() will create an object if it's not exist */
+function createDomainObject(domainName) {
+  return { name: domainName, count: 1 };
+}
+
+/* it will return object array: {name: domainName, count: numberOfDomain} */
+module.exports.countDomain = function (feedUrls) {
+  const domainSummary = [];
+  if (feedUrls.length !== 0) {
```
miggs125: is this check necessary? Will an emtpy array (feedUrls) ever be passed to this function?

src/domain-analysis.js
```diff
@@ -0,0 +1,50 @@
...
+    const { hostname } = urls.parse(feedUrls[0].url);
+    domainSummary.push(
+      createDomainObject(getDomain(hostname)),
+    );
+    const newFeedUrls = feedUrls.slice(1);
+    newFeedUrls.forEach(({ url }) => {
+      const newDomainName = getDomain(urls.parse(url).hostname);
```
miggs125: maybe rename ```newDomainName``` to ```currentDomainName``` or something more appropriate

src/domain-analysis.js
```diff
@@ -0,0 +1,50 @@
...
+
+/* it will return object array: {name: domainName, count: numberOfDomain} */
+module.exports.countDomain = function (feedUrls) {
+  const domainSummary = [];
+  if (feedUrls.length !== 0) {
+    const { hostname } = urls.parse(feedUrls[0].url);
+    domainSummary.push(
+      createDomainObject(getDomain(hostname)),
+    );
+    const newFeedUrls = feedUrls.slice(1);
```
miggs125: rename ```newFeedsUrls``` to something more meaningful, also a comment might be useful here indicating why you are slicing the array

src/domain-analysis.js
```diff
@@ -0,0 +1,50 @@
...
+  return domain;
+}
+
+/* createDomainObject() will create an object if it's not exist */
+function createDomainObject(domainName) {
+  return { name: domainName, count: 1 };
+}
+
+/* it will return object array: {name: domainName, count: numberOfDomain} */
+module.exports.countDomain = function (feedUrls) {
```
miggs125: "Domain" is very vague as this could apply to any domain name. Since you are trying to count how many people are using certain blogging services, perhaps a name that reflects this is more appropriate

src/index.js
```diff
@@ -5,6 +5,7 @@
 const fs = require('fs');
 const feedQueue = require('./feed-queue');
 const feedWorker = require('./feed-worker');
+const domainAnalysis = require('./domain-analysis');
```
miggs125: domainAnalysis is too vague. Change this variable name.

src/domain-analysis.js
```diff
@@ -0,0 +1,50 @@
+/* This file contains the code for analyzing number of blog posts from different domains.
```
miggs125: the name domain-analysis.js is not specific enough to what you want to do here. Rename the file.

src/domain-analysis.js
KarlaChen: That's true, it won't pass an empty array to this function.

src/domain-analysis.js
KarlaChen: Can I just remove the newFeedUrls , and do like this 
``` 
  feedUrls.slice(1).forEach(({ url }) => {
    const currentDomainName = getDomain(urls.parse(url).hostname);
    const domainIndex = domainSummary.findIndex((domain) => domain.name === currentDomainName);
    if (domainIndex === -1) {
      domainSummary.push(
        createDomainObject(currentDomainName),
      );
    } else {
      domainSummary[domainIndex].count += 1;
    }
  });```

Comment (miggs125) [reply src/domain-analysis.js]: Yeah that works!

File: src/index.js [hunk 10]
```diff
@@ -46,6 +47,7 @@ fs.readFile('feeds.txt', 'utf8', (err, lines) => {
 
   enqueueFeedJobs(feedJobs).then(() => {
     feedWorker.start();
+    console.log(counterBlogDomain.blogDomainCounter(feedJobs));
```
miggs125: #142

src/count-blog-domain.js
```diff
@@ -0,0 +1,52 @@
+/* This file contains the code for counting the blog posts from different domains.
+ * The module receives an array of blog urls and returns an array of json object in format
+ * {name: domainName, count: number of frequency}
+*/
+
+//  import url library
+const urls = require('url');
```
humphd: drop the `s` from `urls`, it should be `url`

src/count-blog-domain.js
```diff
@@ -0,0 +1,52 @@
...
+
+/* createDomainObject() will create an object if it's not exist */
+function createDomainObject(domainName) {
+  return { name: domainName, count: 1 };
+}
+
+/* it will return object array: {name: domainName, count: numberOfDomain} */
+module.exports.blogDomainCounter = function (feedUrls) {
+  const domainSummary = [];
+  const { hostname } = urls.parse(feedUrls[0].url);
```
humphd: `url.parse()` can throw, if it gets an invalid URL.  You should wrap this in a `try...catch`, and deal with the err case too.

humphd: Question: why do this to the first item in the list outside the loop?  Why not just do it the same for all of them inside the loop?  If you must do this, add a comment explaining why, since it is not the normal way one would do it.

src/index.js
KarlaChen: ok, I will remove it from here.

src/count-blog-domain.js
KarlaChen: @humphd  for ```const { hostname } = urls.parse(feedUrls[0].url);```
When the domainSummary[] is empty, the first object must be pushed into the array, it doesn't need to check is it exist or not. If we only call blogDomainCounter() function once, it's fine to push it in the loop. But we need to keep the counter number update, that mean the function is not only call one time. If we put this function inside the loop and call it  thousand time, the first object has to be checked thousand time. That's not necessary.

humphd: OK, just add a comment explaining this above this line.

src/count-blog-domain.js
KarlaChen: @humphd I tried to drop the ```s``` from ```const urls = require('url');```
it will have error in  ```feedUrls.slice(1).forEach(({ url }) => {``` cause they are using the same name. The ```{url}``` using in ```feedUrls.slice(1).forEach(({ url }) => {``` is the object property name so I cannot change this one.

humphd: In your loop below, rename it to `feedUrl` instead of `url` to better match with the context of the code.

src/count-blog-domain.js
KarlaChen: I saw someone did the validate blogs url in pr #123 already, that mean the urls I got should be validated already. so do I still need to catch the error?

humphd: Anything that *can* throw will eventually *throw*.  You always need to wrap code the can crash the process so it can't.  What if the other code fails to catch one URL (e.g., it has a bug?).  What if someone refactors the code and removes the url check before you get these?  You have to program defensively.

KarlaChen: Got you. I will add it.

src/count-blog-domain.js
```diff
@@ -0,0 +1,61 @@
...
+
+/* it will return object array: {name: domainName, count: numberOfDomain} */
+module.exports.blogDomainCounter = function (feedUrls) {
+  const domainSummary = [];
+  /* When the domainSummary[] is empty,the first object must be pushed into the array,
+  * it doesn''t need to check is it exist or not.
+  */
+
+
+  const { hostname } = url.parse(feedUrls[0].url);
```
humphd: What if `feedUrls` is an empty array?  I think you need a check above to make sure that there is at least 1 element in it, or this will also crash.

src/count-blog-domain.js
```diff
@@ -0,0 +1,61 @@
...
+      const domainIndex = domainSummary.findIndex((domain) => domain.name === currentDomainName);
+      if (domainIndex === -1) {
+        domainSummary.push(
+          createDomainObject(currentDomainName),
+        );
+      } else {
+        domainSummary[domainIndex].count += 1;
+      }
+    } catch (err) {
+      console.log(err);
```
humphd: Use the new logging module that was just landed.  See the logging.md file in docs.

src/count-blog-domain.js
KarlaChen: But we won't  pass an empty array to the function, right? If there is no feedUrls, we don't need to count it.

humphd: Again, you assume things.  That's not how computers work though: they do what we tell them; and if the calling code has a bug, you end up crashing here.  In general, anything that can go wrong, will eventually go wrong.  Write your code assuming the worst.

PR #170, opened 2019-11-13
Fixes 168: Created frontend folder and moved specified files to the folder

robertbegna:
Hello,
I moved the specified files to new folder "frontend" within the src directory. Added index.html, planet.css, and pagination.js to this folder. Index.html's link to its css styling indicates the css file should still be in same directory as it, so I moved both files to maintain same file structure. I do believe that another issue indicated that css styling should be put in a separate folder, but I refrained addressing that in this pull request to facilitate merging. 
I believe I moved all the front end related files from the src directory, please let me know if any other files need to be moved.


robertbegna:
@humphd alright, so the for the sake of simplicity, I'll move feeds.txt back to its original directory. Then, I guess another issue will need to be filed to update the call sites

PR #171, opened 2019-11-13
Fixes #144 - Added the Redis URL to env.example and pass it when making Bull queue

### LINKED ISSUE #144: Specify redis url in our config
Split off from #112, this deals with specifying server info about redis in our env, and reading that in when we create a queue.

We should move the Redis URL into our .env and when we make Bull Queues, pass this URL on via the constructor. With #105, we can do this. Doing so will make it easier to switch between development and production environments:

```js
const queue = new Queue('name-of-queue', redisUrl);
```

zufishanali:
Fixes #144 

Changes made:
- Added the Redis URL to `env.example` 
- Pass the URL when making Bull queue via the constructor in `src/feed-queue.js`

### DIFF
--- a/env.example
+++ b/env.example
@@ -3,3 +3,6 @@ NODE_ENV=development
 
 # PORT is the port used by the web server
 PORT=8080
+
+# REDIS_URL specifies Redis server info
+REDIS_URL='redis://127.0.0.1:6379'
--- a/env.example
+++ b/env.example
@@ -5,4 +5,4 @@ NODE_ENV=development
 PORT=8080
 
 # REDIS_URL specifies Redis server info
-REDIS_URL='redis://127.0.0.1:6379'
+REDIS_URL=redis://127.0.0.1:6379
--- a/src/feed-queue.js
+++ b/src/feed-queue.js
@@ -1,7 +1,6 @@
-const config = require('./config');
 const Bull = require('bull');
+require('./config');
 
-const redisUrl = process.env.REDIS_URL;
-const queue = new Bull('feed-queue', redisUrl);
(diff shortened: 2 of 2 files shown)


env.example
```diff
@@ -3,3 +3,6 @@ NODE_ENV=development
 
 # PORT is the port used by the web server
 PORT=8080
+
+# REDIS_URL specifies Redis server info
+REDIS_URL='redis://127.0.0.1:6379'
```
s-arika: Let's remove the quotes so that it'll be REDIS_URL=redis://127.0.0.1:6379 and add a new line after

src/feed-queue.js
```diff
@@ -1,5 +1,7 @@
+const config = require('./config');
```
s-arika: Put this after bull require and remove const config. Lines 1 and 2 will look like 
const Bull = require('bull');
require('./config');

src/feed-queue.js
```diff
@@ -1,5 +1,7 @@
...
 const Bull = require('bull');
 
-const queue = new Bull('feed-queue');
+const redisUrl = process.env.REDIS_URL;
```
s-arika: this line can be removed and replace line 5 with 
const queue = new Bull('feed-queue', process.env.REDIS_URL);

env.example
zufishanali: Removed the quotes an added new line.

src/feed-queue.js
zufishanali: Removed `const config = ` and moved it to be below bull require.

src/feed-queue.js
zufishanali: Combined the lines to appear as requested.

PR #176, opened 2019-11-13
Fix#95 Add test for counting domains from blog posts

haichuan0424:
Test module domain-analysis.js countDomain function. Input is an array of blog url objects and expect the output to be an array of object with format {name: domainName, count: number}.

### DIFF
--- a/test/count-blog-domain.test.js
+++ b/test/count-blog-domain.test.js
@@ -1,0 +1,30 @@
+const countBlogDomain = require('../src/count-blog-domain');
+
+const feedUrls = [
+  { url: 'https://neilong31.wordpress.com/feed/' },
+  { url: 'https://haichuan0424.wordpress.com/feed/' },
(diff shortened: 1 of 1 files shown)


humphd:
Failing tests:

```
 FAIL  test/count-blog-domain.test.js
  ● empty list of Urls

    TypeError: Cannot read property 'url' of undefined

      39 | 
      40 | 
    > 41 |   const { hostname } = url.parse(feedUrls[0].url);
         |                                              ^
      42 |   domainSummary.push(
      43 |     createDomainObject(getDomain(hostname)),
      44 |   );

      at Object.blogDomainCounter (src/count-blog-domain.js:41:46)
      at Object.<anonymous> (test/count-blog-domain.test.js:29:26)
```

@KarlaChen, remember you told me it would always be an array with something in it?  It isn't :)

Issue #86, opened 2019-11-07
SAML2 - Middleware


Grommers00:
Please add the middleware that we can start using to start building out our SAML2 system

Do some research to make sure these are compatible before pushing and merging.

neilong31:
@Grommers00 I'd like to work on this issue!

Grommers00:
Cool, I assigned you!

vitokhangnguyen:
Issue #63 when solved can have great benefit for this issue

PR #178, opened 2019-11-13
Added login and logout routes

neilong31:
Helps with #86 by adding routes that with OneLogin

What is added is:

1. login route
1. oauth callback route (We can change the redirection to index.htm
1. logout route


miggs125:
Additionally, any endpoint to be implemented needs a test. See #141 to see how these can be implemented.

neilong31:
Oh okay thanks! I think I see my mistake. I'll try and post my fix tonight.

humphd:
@neilong31 it looks like you merged this in without getting review. I see @miggs125 has comments above that don't seem to be addressed.

I'm not clear what happened here, but I'd like to have it fixed in a follow-up PR.  Please make one to deal with the outstanding comments, and let's not merge our own code without review approvals.

Issue #74, opened 2019-11-07
Validate blog URLs


Cadesh:
Currently the feeder do not validate the blog's URLs. Students will add and modify blogs URLs and there must be a way to check if these new URLs are valid and reachable.  In this issue the objective is to check for valid format of URLs only.

Cadesh:
The idea is to create a utils.js file to hold and export the method to validate the URLs. This utils.js file will have more functions in the future. 
There are several possibilities to validate a URL:
1) is-url package: [https://www.npmjs.com/package/is-url](url)
2) validurl package [https://www.npmjs.com/package/valid-url](url)
3) with NodeJS url module [https://nodejs.org/docs/latest/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost](url)
4) use a regex and code it yourself.

humphd:
4. is out.  URLs are really tricky to get right in a regex.

PR #179, opened 2019-11-14
Closes #54: Fetch data from GitHub API

### LINKED ISSUE #54: Use Github API to fetch data related to URLs from a blog post.
After extracting URLs from a blog post (issue #53), use the [Github API](https://developer.github.com/v3/) (if it's a Github URL) to fetch related data like author, description, date of creation, avatars, labels, assignees, projects...

Library that could be used for this: [github-url-parser](https://www.npmjs.com/package/parse-github-url).

manekenpix:
Closes #54 
This takes a valid Github URL and returns fetched data related to it from the [GitHub API](https://developer.github.com/v3/).
At the moment, it supports URLs for:
- Users: https://github.com/{myUser}
- Repositories: https://github.com/{myUser}/{repository}
- Issues https://github.com/{myUser}/{repository}/issues/{issue#}
- Pull Requests https://github.com/{myUser}/{repository}/pull/{pullRequest#}

Every request to the GitHub API returns a big chunk of data, and I chose the properties that I thought could be interesting for Telescope. This can be changed, please feel free to suggest other properties.
To see full chucks of data fetched from the GitHub API:
- User [Seneca-CDOT](https://api.github.com/users/Seneca-CDOT)
- Repository [Telescope](https://api.github.com/repos/Seneca-CDOT/telescope)
- Issue [Telescope's issue #1](https://api.github.com/repos/Seneca-CDOT/telescope/issues/1)
- Pull Request [Telescope's PR #1](https://api.github.com/repos/Seneca-CDOT/telescope/pulls/1)

What returns:
- User:
```js
{
  user: String,
  avatarURL: String,
  name: String,
  company: String,
  blog: String,
  email: String,
  bio: String,
};
```

- Repository:
```js
{
 avatarURL: String,
 description: String,
  license: {
    key: String,
    name: String,
    spdx_id: String,
    url: String,
    node_id: String,
  },
  openIssues: Number,
  forks: Number,
  createdAt: String,
  language: String,
};
```

- Issue and Pull Request:
```js
{
  login: String,
  avatarURL: String,
  body: String,
  createdAt: String,
  branch: String,
  repo: String,
  };
```

I wanted to use [bent](https://www.npmjs.com/package/bent), but unfortunately it only likes `200` status codes, and the GitHub API does a lot redirections (status code `301`), so I couldn't get it to work

I also included tests for the 4 valid types of URLs.

### DIFF
--- a/env.example
+++ b/env.example
@@ -29,3 +29,6 @@ NODEMAILER_SERVER=
 # BLOG_INACTIVE_TIME is the period (days) of inactivity
 # before a blog will be considered redlisted
 BLOG_INACTIVE_TIME=360
+
+# GITHUB_TOKEN is used when fetching data from the GitHub API
+GITHUB_TOKEN=
--- a/package.json
+++ b/package.json
@@ -36,10 +36,12 @@
     "feedparser-promised": "^2.0.1",
     "pino": "^5.13.6",
     "pino-pretty": "^3.3.0",
-    "node-summarizer": "^1.0.7",
     "valid-url": "^1.0.9",
-    "jsdom": "^15.2.1",
     "nodemailer": "^6.3.1",
+    "jsdom": "^15.2.1",
+    "node-fetch": "^2.6.0",
+    "node-summarizer": "^1.0.7",
+    "parse-github-url": "^1.0.2",
     "sentiment": "^5.0.2"
   },
   "devDependencies": {
--- a/src/github-url.js
+++ b/src/github-url.js
@@ -1,0 +1,134 @@
+require('./config');
+const parseGithubUrl = require('parse-github-url');
+const fetch = require('node-fetch');
+
+const githubAPI = 'http://api.github.com';
+
+
+const isValidGithubUrl = (ghUrl) => ghUrl.protocol === 'https:'
+                                  && ghUrl.host === 'github.com'
+                                  && (ghUrl.branch !== 'issues'
+                                    || ghUrl.branch !== 'pull'
+                                    || ghUrl.branch !== 'master');
+
+/**
+ * @description Takes a url, checks if it's a valid github url
+ * and returns JSON with fetched data related to the url from Github's V3 API.
+ * At the moment, it accepts URLs of repositories, issues and pull requests.
+ * @link https://developer.github.com/v3/
+ * @param {string} url Url that will be processed.
+ * @return An object with all the fetched data.
+ */
+exports.getGithubUrlData = async (incomingUrl) => {
+  const ghUrl = parseGithubUrl(incomingUrl);
+
+  if (!isValidGithubUrl(ghUrl)) {
+    throw new Error('Invalid GitHub url');
+  }
+
+  let subUrl = '';
+
+  /**
+   * If repo is null, it's a user URL.
+   * The request needs to be format this way:
+   * GET /users/:user
+   */
+  if (ghUrl.repo === null) subUrl = `/users/${ghUrl.path}`;
+
+  /**
+   * If branch is master and there's a repo,
+   * it's repo URL:
+   * GET /repos/:user/:repo
+   */
+  else if (ghUrl.branch === 'master') subUrl = `/repos/${ghUrl.repo}`;
+
+  /**
+   * Otherwise it's a {pull request, issue} URL:
+   * GET /repos/:user/:repo/{issues, pulls}
+   */
+  else subUrl = ghUrl.branch === 'pull' ? `/repos/${ghUrl.repo}/pulls` : `/repos/${ghUrl.repo}/issues`;
+
+  /**
+   * Add the {issue,pull request} number at the end of the request:
+   * GET /repos/:owner/:repo/{issues, pulls}/:{issue, pull_number}
+   */
+  subUrl += (ghUrl.filepath !== null ? `/${ghUrl.filepath}` : '');
+
+  /**
+   * Add GITHUB_TOKEN if in process.env
+   */
+  subUrl += typeof process.env.GITHUB_TOKEN !== 'undefined' ? `?access_token=${process.env.GITHUB_TOKEN}` : '';
+
+  return fetch(`${githubAPI}${subUrl}`)
+    .then((res) => res.json())
+    .then((data) => {
+      let fetchedData;
+
+      /**
+       * Format the fetched data in a specifi way depending if the url
+       * is for a repo, user or for a {pull request, issue}
+       */
+      // User
+      if (ghUrl.repo === null) {
+        const {
+          login: user,
+          avatar_url: avatarURL,
+          name,
+          company,
+          blog,
+          email,
+          bio,
+        } = data;
+
+        fetchedData = {
+          user,
+          avatarURL,
+          name,
+          company,
+          blog,
+          email,
+          bio,
+        };
+
+      // Repo
+      } else if (ghUrl.branch === 'master') {
+        const {
+          owner: { avatar_url: avatarURL },
+          description,
+          license,
+          open_issues: openIssues,
+          forks,
+          created_at: createdAt,
+          language,
+        } = data;
+
+        fetchedData = {
+          avatarURL,
+          description,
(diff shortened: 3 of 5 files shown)


package.json
```diff
@@ -34,6 +34,8 @@
     "express": "^4.17.1",
     "express-healthcheck": "^0.1.0",
     "feedparser-promised": "^2.0.1",
+    "node-fetch": "^2.6.0",
```
humphd: We're using [bent](https://github.com/mikeal/bent#readme) elsewhere, can you use it too?

src/github-url.js
```diff
@@ -0,0 +1,131 @@
+const parseGithubUrl = require('parse-github-url');
+const fetch = require('node-fetch');
+
+const githubAPI = 'http://api.github.com';
+
+
+const isValidGithubUrl = (ghUrl) => ghUrl.protocol === 'https:'
```
humphd: We're starting to get a bunch of "URL" specific things we want to do, @Cadesh has a PR up for one, @smilegodly has another, @KarlaChen has a third.  Later on, let's think about how we can combine some of them into common modules.  For now, duplication is OK.

humphd: Probably a good idea to document that this is not a URL string, but a parsed URL Object you expect.

src/github-url.js
```diff
@@ -0,0 +1,131 @@
...
+                                  && ghUrl.host === 'github.com'
+                                  && (ghUrl.branch !== 'issues'
+                                    || ghUrl.branch !== 'pull'
+                                    || ghUrl.branch !== 'master');
+
+/**
+ * @description Takes a url, checks if it's a valid github url
+ * and returns JSON with fetched data related to the url from Github's V3 API.
+ * At the moment, it accepts urls of repositories, issues andpull requests.
+ * @link https://developer.github.com/v3/
```
humphd: Question: why not use GitHub's JS wrapper for this API, https://octokit.github.io/rest.js/?  It has a lot of convenience things already done for you.  We could file a follow-up issue later to refactor to use it.

humphd:
Beautiful tests, too, by the way!

package.json
manekenpix: I left a comment in the description:
> I wanted to use bent, but unfortunately it only likes `200` status codes, and the GitHub API does a lot redirections (status code `301`), so I couldn't get it to work.

src/github-url.js
manekenpix: I thought about waiting until @Cadesh's PR gets merged so I can use his tools to validate URLs, but I don't know when it's going to happen so I decided to duplicate. I'll change that once #74 gets merged.

src/github-url.js
manekenpix: I wasn't aware of it at all! I'll file an issue to refactor this using octokit/rest.js once it gets merged.

manekenpix:
I added using a [GitHub token](https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line) included in the `.env` file.

package.json
humphd: Ah, good to know.  Sorry I missed your comment.  Let's stay on `node-fetch`, and consider file an issue to do after this that replaces all uses of `bent` with `node-fetch`.  I'd rather not have multiple deps for doing requests.

env.example
```diff
@@ -6,3 +6,6 @@ PORT=8080
 
 # REDIS_URL specifies Redis server info
 REDIS_URL=redis://127.0.0.1:6379
+
+#GITHUB_TOKEN is used when fetching data from the GitHub API
+GITHUB_TOKEN=123456789
```
humphd: Leave the value blank, and add a link in your comment to docs on how to get one of these and what it is.  Also, add a space after the `#` in your comment above.

src/github-url.js
humphd: Sounds good, just add a comment above this re: url object vs string.

src/github-url.js
```diff
@@ -0,0 +1,137 @@
...
+          body,
+          createdAt,
+          branch: ghUrl.branch,
+          repo: ghUrl.name,
+        };
+      }
+      return fetchedData;
+    })
+    .catch((err) => {
+      throw new Error(err);
```
humphd: If `err` is already an `Error`, why re-wrap it?  Also, if you're only `catch()`'ing this to `throw` it, you don't need the `catch()`

test/github.test.js
```diff
@@ -0,0 +1,86 @@
+const gh = require('../src/github-url');
+
+/**
+ * Test to validate fetching data when passing a user URL.
+ * It uses the Seneca-CDOT URL
+ */
+test('test fetching data for a valid user URL', async () => {
+  const validUserUrl = 'https://github.com/Seneca-CDOT';
```
humphd: Testing against the live API has the potential to fail (network down, API down, throttling, etc).  Often people will mock network APIs in tests, and give back simulated responses.  Here's an example you can look at https://github.com/humphd/learn-travis/blob/master/test/app.mocked.test.js.  In a follow up, we can file an issue to use mock responses here.

src/github-url.js
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * @description Takes a url, checks if it's a valid github url
+ * and returns JSON with fetched data related to the url from Github's V3 API.
+ * At the moment, it accepts URLs of repositories, issues and pull requests.
+ * @link https://developer.github.com/v3/
+ * @param {string} url Url that will be processed.
+ * @return An object with all the fetched data.
+ */
+exports.getGithubUrlData = async (incomingUrl) => {
+  const ghUrl = parseGithubUrl(incomingUrl);
```
Reza-Rajabi: in here, you might want to wait for the `parseGithubUrl` to get the job done, and then check `isValidGithubUrl`.
something like:
```
const ghUrl = await parseGithubUrl(incomingUrl);
```

Reza-Rajabi: Very well done. Very nice, clean and properly commented code. I have learned always from your code.

miggs125: @Reza-Rajabi from the [parse-github-url ](https://github.com/jonschlinkert/parse-github-url/blob/master/index.js) documentation, it doesn't look like the function returns a promise which means that we don't need to use ```await``` as this is meant to "wait" for promises to be resolved before continuing with execution.

src/github-url.js
miggs125: Also, If we are to use ```async/await``` throughout the code-base perhaps we should create a handler so that we don't have ```try/catch``` everywhere and centralize error handling. This can be in a separate issue though.

miggs125:
I think I accidently dismissed everyone's review :/

Reza-Rajabi:
@miggs125 sorry for my late response. I have just arrived home, but I have got your message by email.
I appreciate your detail-oriented consideration by looking into the imported library.
I think it doesn't matter if it returns a promise or not, and actually `await` implicitly will convert it to `Promis.resolve(whatever expression is)` . To me, it is safer and good practice to wrap the imported function to a promise, unless you say it's a micro process and doesn't worth a wast of an embedded function call (calling await keyword). If this is the case, I would say: I don't know which one is better. Thank you.

Reza-Rajabi:
@manekenpix I was thinking about suggesting a change for that big "return" at the end to something else, and I am writing it here to not forget it, and also you might find it useful for your follow up changes:
changing this:
``` javascript
return fetch(`${githubAPI}${subUrl}`)
     .then((res) => res.json())
     .then((data) => {
               let fetchedData;
                   // ...
       return fetchedData;
       });
```

to this:
```javascript
const response = await fetch(`${githubAPI}${subUrl}`);
const data = await response.json();
let fetchedData;
//...
return fetchedData;
```

manekenpix:
Yes, @miggs125 suggested that [here](https://github.com/Seneca-CDOT/telescope/pull/179#discussion_r346490390), I should probably have done it that way to maintain consistency. That part will probably change some time soon, keep an eye on whoever does it so you can recommend that change.
Thank you

Issue #12, opened 2019-10-31
Mockups for user interface


Reza-Rajabi:
We have discussed in [issue #3](https://github.com/Seneca-CDOT/telescope/issues/3) about our frontend technology. But how our final frontend looks like? This is important that we think about it for the number of reasons:
- Knowing how our UI looks like helps us to identify our app's features and our project's scope.
- Having a UI design gives us an idea about who can work on what in terms of our front-end development (We can modularize our front-end development)
- The design itself could be somebody's task if they are interested in and if it is acceptable for the purpose of this project.

ODAVING:
If no one is working on the mockup, I can use Adobe XD to create a mock up of what it can look like. @humphd @Reza-Rajabi Do you guys prefer the mockup? or should I just go ahead and create an HTML document issue #3

lucacataldo:
@ODAVING I don't mind taking the reigns on the design & prototyping or helping you out. Want me to start out by making a shared XD document? That'll allow everyone to contribute ideas and whatnot to the design.

ODAVING:
If this is acceptable then I’d like that! Otherwise we should get started
with the static html. Awaiting David’s response to this.

lucacataldo:
Sounds good! I'll get started on a general mockup with XD regardless that way the actual creation of the HTML will be easier later on. Will make a link available later on today for review & editing.

Reza-Rajabi:
@ODAVING @lucacataldo I made this issue more as a discussion, so feel free to start working on this as you wish.

lucacataldo:
Here's the link to review & comment on the design, if anyone would like write / edit access to the file please contact me.

https://xd.adobe.com/view/f1e160d0-2c49-4700-4ca6-2f2f660dfb6f-87d0/

KarlaChen:
@lucacataldo For the participants page, can we add a quick search bar then people can easy to find the people name.

ODAVING:
This is a good place to get inputs while we build the static page

lucacataldo:
@KarlaChen Good idea Karla, I've added a "quick and dirty" search bar design to the participants modal

KarlaChen:
@lucacataldo I was thinking use anchorScroll or search bar which one is better, eg: There is A,B, C on the side, when you click A, it will go to the section A and show all the name start from A.

lucacataldo:
[...]

PR #180, opened 2019-11-14
Fixes #153 - New front-end structure & design

### LINKED ISSUE #153: Base html template and style doesn't render properly and needs work
@ODAVING and everyone who worked together in #12 & the frontend slack channel did a great job of creating a base HTML for us to work with from the template but it needs some work and cleaning up to get it ready for proper templating. I'd like to rework the structure of the HTML to make it more understandable and easier to feed content into. Additionally, while doing this, some of the CSS will need to be modified to make it more scalable and responsive.

lucacataldo:
### Fixes #153 

This pull request introduces a new front end design and structure that builds on what @ODAVING had created using the old Planet structure and the design mockups created in #12 . The new design is responsive and uses CSS grid to allow for easier maintenance down the line. It also features a cleaner HTML structure and includes an easily template-able <article> tag as the basis for each blog post. A small UX tweak I added was to sticky the "Title" section of each article to the top while scrolling since a major gripe of mine when using the old Planet was that it was difficult to know which article you were reading when scrolling through a wall of text.

Obviously this UI is still feature-light but I hope that it is a starting point that we can use going forward as we add to it. Eg: Pagination, search, post stats, etc.

You can view the live demo [here](https://lucacataldo.github.io/telescope-demo/)

Screenshot:
image

In this PR I also removed some old, unused files like the planet.css & images folder that was found in the root dir. I also used a temporary logo.svg I created until the logo PR gets merged / resolved. 

Some other


src/frontend/style/main.css
```diff
@@ -0,0 +1,241 @@
...
+    article section {
+        margin-right: 0px;
+        margin-left: 0px;
+    }
+
+    .content {
+        padding-right: 20px;
+        padding-left: 20px;
+    }
+}
```
manekenpix: Newline missing.

lucacataldo: Fixed. Thanks!

PR #185, opened 2019-11-14
fix #174:Moved files to frontend and changed paths

### LINKED ISSUE #174: Move src/dialogue.js to src/frontend
We have a stray frontend file in `src/` but we're moving all of those files to `src/frontend/*`.  We need to move `src/dialogue.js` into `src/frontend` and also update all the files in `src/frontend` so their paths work relative to their new home (i.e., use relative paths from `index.html` so they can all find each other).

wajeehsheikh:
1. Moved `src/dialogue.js` into `src/frontend` 
2. Changed paths for `dialogue.js `in `index.html`


cindyledev:
I don't know if this should be a separate issue or if it could be done here but I see there are two planet.css files. One located in telescope/src/frontend/ and another in telescope/src/frontend/style/

wajeehsheikh:
> I don't know if this should be a separate issue or if it could be done here but I see there are two planet.css files. One located in `telescope/src/frontend/` and another in `telescope/src/frontend/style/`
> 
> Edit: nvm this was already pointed out by @zufishanali in the issue itself. Please address the planet.css file in this PR.

So the above issue was included in my original commit and push but @lucacataldo changed the structure of the frontend #180 to fix issue #153. So I had to revert my commit + push and stick to just moving dialoge.js file.

humphd:
@wajeehsheikh why do these files show as DIFFs and not rename/moves?  How did you "move" these files?

wajeehsheikh:
> @wajeehsheikh why do these files show as DIFFs and not rename/moves? How did you "move" these files?

I originally used mv for the dialoge.js file but git did not track it. Then realized I had to use git mv but git gave me an option to track the move by doing `git add src/frontend/dialogue.js`

Also, my index.html had a conflict with merged pr #184. So I had to fix this conflict on my side.

cindyledev:
> > I don't know if this should be a separate issue or if it could be done here but I see there are two planet.css files. One located in `telescope/src/frontend/` and another in `telescope/src/frontend/style/`
> > Edit: nvm this was already pointed out by @zufishanali in the issue itself. Please address the planet.css file in this PR.
> 
> So the above issue was included in my original commit and push but @lucacataldo changed the structure of the frontend #180 to fix issue #153. So I had to revert my commit + push and stick to just moving dialoge.js file.

Okay I see what happened here, I was looking at your previous commits but now I see planet.css isn't even a thing anymore after #180 . Thanks for the clarification.

PR #186, opened 2019-11-14
Fix #75: Added Fragmentation to HTML to text Parser

### LINKED ISSUE #75: Extract plain text form html

PavanKKamra:
Issue - #75

### DIFF
--- a/src/text-parser.js
+++ b/src/text-parser.js
@@ -5,8 +5,7 @@ const { JSDOM } = jsdom;
 // Simple function to convert html to text
 // Issue #75: https://github.com/Seneca-CDOT/telescope/issues/75
 exports.run = async function (html) {
-  const dom = new JSDOM(html);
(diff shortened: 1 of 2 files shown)


UltimaBGD:
This has two approvals and can be merged, I will do so.

PR #192, opened 2019-11-14
Fixes #156 Add test cases for feedParser

### LINKED ISSUE #156: Create testing for feedparser
For the recent PR #37. There are no test cases available. Assigning issue to myself with help from @evlnyng.

c3ho:
I think the only test case missing here is one for testing an IP address, as feedParser will actually try to connect to an IP address instead and get `ECONNREFUSED (IP here)` message instead for the test case. Perhaps @evlnyng could provide some insight if she is available.

### DIFF
--- a/test/feed-parser.test.js
+++ b/test/feed-parser.test.js
@@ -1,0 +1,27 @@
+const feedParser = require('../src/feed-parser');
+
+/**
+ * feedParser(feed) async function returning an object if the feed URI is correct and there are
+ * things to parse. Invalid URI or non-feed URI should throw an error.
+*/
+test('passing a valid Atom feed URI and getting title of first post', async () => {
+  const data = await feedParser('https://c3ho.blogspot.com/feeds/posts/default/-/open-source?alt=rss');
+  expect(data[data.length - 1].title).toBe('First Post!');
+});
+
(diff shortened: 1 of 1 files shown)


c3ho:
Pretty sure I added a new line to the end of the file, but don't know if its showing up or not.

test/feed-parser.test.js
```diff
@@ -0,0 +1,27 @@
+const feedParser = require('../src/feed-parser');
+
+/**
+ * feedParser(feed) async function returning an object if the feed URI is correct and there are
+ * things to parse. Invalid URI or non-feed URI should throw an error.
+*/
+test('passing a valid Atom feed URI and getting title of first post', async () => {
+  const data = await feedParser('https://c3ho.blogspot.com/feeds/posts/default/-/open-source?alt=rss');
```
humphd: We can do this in the short term, but this kind of test is going to fail when you remove your blog at some point in the future.  A better option is to use a mock network request here, using something like https://www.npmjs.com/package/jest-fetch-mock.  We'd simulate getting a feed from the network, and return a prebaked XML document you can parse.

humphd: Please file another issue to mock this out later.

c3ho: Filed issue under #210

PR #78, opened 2019-11-07
Fixes #17: Implemented basic word counter & test

birtony:
Fixes #17: This function utilizes [wordcount](https://www.npmjs.com/package/wordcount) as suggested by @humphd in https://github.com/Seneca-CDOT/telescope/issues/17#issuecomment-549649707

Issue #17, opened 2019-10-31
Implement word count feature for the contents of the blog post


birtony:

humphd:
Maybe we can use https://www.npmjs.com/package/wordcount?

Reza-Rajabi:
This seems to have conflict with #7 and #19 .

humphd:
@Reza-Rajabi I don't think they are in conflict, but rather represent parallel steps along an "analysis" pipeline.  We just have to chain them all together.

PR #193, opened 2019-11-14
Fix #21: Spam and empty post detection

### LINKED ISSUE #21: Create a spam checker for blog posts

jordanhui19:
Issue #21 

Created function in check-for-spam.js to check for spam using the following rules:
-blog post must have title
-body of blog post must have at least 20 words in plain text (content within html tags such as <a> or  do not count)
-blog post cannot be in all caps

The function takes an article returned by [feedparser-promised](https://www.npmjs.com/package/feedparser-promised) and checks the title and description properties. If the article does not fit all the rules, it returns true. If all the rules are followed, false is returned

Also created tests for the function in check-for-spam.test.js using a blog with test posts: https://telescopetestblog.blogspot.com/


c3ho:
Hey, I was the one who worked on feedparser-promised and will be happy to review this one later on tonight.

jordanhui19:
Thanks.
If you have any questions feel free to ask.

src/check-for-spam.js
```diff
@@ -0,0 +1,41 @@
...
+
+  function getAllLetters(str) { // Counts number of alphabetical characters in string
+    return str.replace(/[^A-Za-z]/g, '').length;
+  }
+
+  // The main body of the blog post with all HTML tags removed
+  const noTagsDesc = removeTags(content.description);
+
+  // If the title is empty, post is considered spam
+  if (content.title != null) {
```
miggs125: change to 
```js
if(content.hasOwnProperty(title)){...}
```
checkout the [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty) for details.

src/check-for-spam.js
```diff
@@ -0,0 +1,41 @@
+// content refers to the contents of the blog post (article) sent by feedparser
+module.exports = function (content) {
+  // Minimum words set to 20, can be changed
+  const MINWORDS = 20;
```
miggs125: Excellent use of constants. I like this. I would suggest creating a file with all the constants definitions and then importing from there. This would be one of them. For now this is good though. If you'd like to work on another issue you can create one for this and self-assign

src/check-for-spam.js
```diff
@@ -0,0 +1,41 @@
...
+
+  // Functions
+  function removeTags(str) { // Removes HTML tags from a string, leaving only plaintext
```
humphd: Use [text-parser.js](src/text-parser.js) instead

src/check-for-spam.js
```diff
@@ -0,0 +1,41 @@
...
+    if (str != null && str !== '') {
+      return str.toString().replace(/<[^>]*>/g, '');
+    }
+    return '';
+  }
+
+  function getWordCount(str) { // Counts number of words in string
```
humphd: We have another PR doing word count, see #17, #78.  You'll want to consume their data here.

src/check-for-spam.js
jordanhui19: I can't use that cause content always has the property title, its just that if the title of the blog is blank, content.title it returns null.

src/check-for-spam.js
jordanhui19: text-parser.js being async was causing a lot of issues with both my code and the test scripts that I could not figure out how to work around, so I decided to leave them out, as I already had a working script.

src/check-for-spam.js
jordanhui19: That feature appears to have issues and hasn't been merged yet so I left my own word counter in for now. I can implement it once it is finished.

src/check-for-spam.js
humphd: You should make your code expect to be async as well, and you'll have no issues.  Do you need help doing that?  If so, happy to guide you a bit.

jordanhui19:
@humphd I added in the spam-filter node for spam detection based on a trained machine, on top of the checks I already have in place. 
Spam-filter does not raise a flag if there is no text in a post and I figure that posts that are in all caps can be considered spam, so I left the checks in to handle those cases

humphd:
OK, let's start with this and improve it as we go.  I'm about to go teach for a few hours, get some other folks to start reviewing this, and I'll look again later.

src/check-for-spam.js
jordanhui19: I tried making my function async as well as using await in the tests, but in tests it keeps returning undefined.

humphd: Did you return a Promise?  An `async` function needs to return a `Promise`.  If you `await` something, it will use that `Promise`, but if not, you need to return a `Promise`, of if your code is purely synchronous, you can use `return Promise.resolve(yourReturnValue)`

jordanhui19: Alright I changed spam-checker to async and implemented text-parser

src/check-for-spam.js
jordanhui19: I'm going to leave MINWORDS in the compenent file for now, as I feel like we don't need an entirely separate file for one constant.
That being said, if there are a lot more constants in other people's components, making a separate file would be a good idea.

c3ho:
@miggs125 @humphd are you guys able to get any output from the spam-checker? I passed in a bit of the data from the parsed feed like so
```
const t = await spamChecker(returnedFeed[returnedFeed.length - 1].description)
```
and keep getting undefined

jordanhui19:
I had to use .then to get the results properly. For some reason async wasn't working

c3ho:
```
const check = require('./spam-checker');
const feedParser = require('./feed-parser');

async function parseFeed(url) {
  feedParser(url)
    .then((data) => {
      check(data[data.length - 1].description);
    })
    .then((result) => {
      console.log(result);
    });
}

parseFeed('https://c3ho.blogspot.com/feeds/posts/default/-/open-source?alt=rss').then((x) => {
  console.log(x);
});
```

This should technically work then right?

jordanhui19:
Pass it data[data.length - 1], not just data[data.length - 1].description

c3ho:
image

jordanhui19:
oh wait for feedParser you need to use feedParser.parse(url)

c3ho:
don't think so, as I'm calling the async function provided by feed-parser.js it should automatically parse it.

```
const feedparser = require('feedparser-promised');

module.exports = async function (feed) {
  return feedparser.parse(feed);
};
```

package.json
```diff
@@ -45,6 +45,8 @@
     "pino": "^5.13.6",
     "pino-pretty": "^3.3.0",
     "sentiment": "^5.0.2",
+    "spam-filter": "^1.1.1",
```
humphd: Don't hand edit this file for dependencies.  Prefer `npm install --save package` so it puts it in the right place (npm will alphabetize it).

jordanhui19:
I'm not sure then.
In the test file I required the feed-parser node itself and used feedParser.parse(url) and it worked.

package.json
jordanhui19: Github was saying there was a conflict in the package.json, so I edited it to include all the dependencies

humphd:
From Slack:

> @c3ho `await textParser.run(content.description).then((noTagsDesc) => {` is odd code
> if you await, you expect to get back a result, not a promise
> so it should either be: `const text = await textParser.run(content.description);`
> or `return textParser.run(content.description).then(...)`

jordanhui19:
Alright I fixed it.
@c3ho 's code should work now

src/spam-checker.js
```diff
@@ -0,0 +1,48 @@
+const filter = require('spam-filter')('naiveBayes');
+const textParser = require('./text-parser');
+// content refers to the contents of the blog post (article) sent by feedparser
+module.exports = async function (content) {
+  // Minimum words set to 20, can be changed
+  const MINWORDS = 5;
+
+  // Functions
+
+  /* function removeTags(str) { // Removes HTML tags from a string, leaving only plaintext
```
humphd: Can we dump this commented out code?  You can add it back later in another PR

c3ho:
I think I'll leave this last part here just for future reference. Are we going to be passing the individual arrays  from feedParser such as 
```
const parsedURL = feedParser(feedURL);
const parsedText = spamCheck(parsedURL[0])??
```

Other than that, I think it looks good now.

src/spam-checker.js
humphd: I'll do it and rebase, so we can get this landed.

jordanhui19:
yeah you'd pass it one element from the array generated by feed-parser, meaning one blog post at a time

src/spam-checker.js
jordanhui19: Sorry, I just got home.
Thanks for removing it

humphd:
OK, it's merged!  Any updates you do now should happen in a new PR off `master`.

package.json
jordanhui19: I'm getting the same conflicts again.
Is there any way to resolve them without manually editing package.json?

jordanhui19: Never mind, the PR got merged while I was typing that

humphd:
Fixing these conflicts is tricky.  My advice is:

- don't hand edit `package.json` to add deps, use `npm install --save package` or `npm install --save-dev package`
- remove your `node_modules` and `package-lock.json` and re-run `npm install` to generate both fresh again
- re-run `npm install` one more time to make sure `package-lock.json` is good

Anyway, it's all good, I fixed it before I merged.

c3ho:
Last nitpick thing although it is merged, we should check for title spam before calling text-parser since content.title doesn't rely on the text-parser and we can save an operation there. No point in parsing the whole description of the body only to have the content.title empty.

humphd:
@c3ho good catch, file a follow-up issue, and reference this PR.

PR #194, opened 2019-11-14
Fix #181: Adding Windows and MacOS support to CircleCI config.yml file

### LINKED ISSUE #181: Run CircelCI on windows and mac environment
The current config file only runs CI on a linux environment. We should add support for running the integrations on windows and mac environments.

robertbegna:
After reading the documentation, and also using this video as a guide
https://www.youtube.com/watch?v=3V84yEz6HwA&t=185s, I added three separate jobs build on one type of environment (MacOS, Linux, Windows). These jobs can run in parallel if we use workflows and then list each job under the workflow

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ jobs:
           name: Execute tests
           command: npm test
 
-  build-Macos:
+  build-macos:
     macos:
       xcode: "9.0"
     working_directory: ~/telescope
@@ -33,7 +33,7 @@ jobs:
           name: Execute tests
           command: npm test
   
-  build-Windows:
+  build-windows:
     working_directory: ~/telescope
     executor:
       name: win/vs2019
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,6 +19,9 @@ jobs:
   buildwindows:
     executor: win/vs2019
     working_directory: ~/telescope
+    docker:
+      - image: circleci/node:lts
+      - image: circleci/redis:latest
     steps:
       - checkout
       - run: Write-Host 'Hello, Windows'
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,6 @@
 version: 2 # use version 2.0 of CircleCI
+orbs:
+  win: circleci/windows@1.0.0
 jobs: # a basic unit of work in a run
   build:
     working_directory: ~/telescope
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,22 +1,18 @@
 version: 2.1
 orbs:
-     win: circleci/windows@1.0.0 
-  jobs:
-    build:
-      docker: 
-        - image: circleci/node:4.8.2 # the primary container, where your job's commands are run
-      steps:
-        - checkout # check out the code in the project directory
-        - run: echo "hello world" # run the `echo` command
-    buildWindows:
-    executor: win/vs2019
+  win: circleci/windows@1.0.0
+
+jobs:
+  build:
+    working_directory: ~/telescope
+    docker:
+      - image: circleci/node:lts
+      - image: circleci/redis:latest 
     steps:
       - checkout
-      - run: Write-Host 'Hello, Windows
-
-
-workflows:
-  my_workflow:
-    jobs:
-      -build
-      -buildWindows
+      - run:
+          name: Install dependencies
+          command: npm ci
+      - run:
+          name: Execute tests
+          command: npm test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,6 +16,8 @@ jobs: # a basic unit of work in a run
 
   buildmacos:
     working_directory: ~/telescope
+    macos:
+      xcode: "10.0"
     docker:
       - image: circleci/node:lts
       - image: circleci/redis:latest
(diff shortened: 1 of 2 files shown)


humphd:
Hey @robertbegna, I see you're working on this.  I notice that CircleCI has a lot of sample configs from various projects available at https://circleci.com/docs/2.0/example-configs/ which you might want to look at too.  Some of them do parallel builds on different platforms.

.circleci/config.yml
```diff
@@ -13,3 +16,44 @@ jobs:
       - run:
           name: Execute tests
           command: npm test
+
+  build-Macos:
```
humphd: Try `build-macos` all lower case

.circleci/config.yml
```diff
@@ -13,3 +16,44 @@ jobs:
...
+    steps:
+      - checkout
+      - run:
+          name: Install dependencies
+          command: npm ci
+      - run:
+          name: Execute tests
+          command: npm test
+  
+  build-Windows:
```
humphd: Same here, `build-windows`

robertbegna:
@sukhbeersingh can you offer any advice on how to resolve these errors?

humphd:
@robertbegna how is this going?  It looks like you're trying lots of things, but it might be useful to discuss your progress in here.  What have you learned?  How close are you?  Should we do more research?  What's the right path forward?

robertbegna:
@humphd 
a lot of these commits were just errors with syntax and me trying to the get the hang the of syntax, in addition to trying other things I found in the docs.
So far now, I've managed to make three parallel builds one for linux, macos, and windows. I keep running into crashes when I try to add more than one container to a build. Like the linux one  uses the docker container with the image, and i'm not sure if the other two builds need it, since they have their own executors. Also for that MacOS build, it seems to fail because we need to select a MacOS plan, so Im not sure how to approach that (or if we even need it). In addition the tests in the linux build failed in that Windows build so for now i just have steps that console/write "Hello, Windows" and it works for now. Hopefully this can be enough for this PR.

humphd:
This is great @robertbegna!  It sounds to me like you're making progress, and just need more time.  I'd say let's keep you on this fix, and know that it's coming, but doesn't have to merge tonight.  Keep researching and figuring things out.  It will be great to have this working in full.

humphd:
Thanks for the update.  It helps to know where things are at.

robertbegna:
@humphd so will you add a macos plan so i can see if the macos build is successful or not, or should I just stick to focusing on linux and windows?

humphd:
Let's leave macOS out of this, for cost reasons, and focus on the other two.

sukhbeersingh:
@robertbegna Quick advice, you should probably sign up with CircleCI and add your forked repo to the jobs. Once you do that, you can test your changes in your own branch before you actually make a Pull Request. Gives you more flexibility and you woudn't have to depend on CDOT's CircleCI account. If you need help setting this up, let me know.

sukhbeersingh:
@robertbegna So i pulled your code and rebased it on current master. This is what I get

circle

@humphd I remember you had to fix something similar for .travis.yml using a hack to set all line endings to not be crlf. We still have this issue even after prettier. Is [this](https://stackoverflow.com/a/53769213) a solution?

How should we move ahead from here?

humphd:
cc @Manan311, @vitokhangnguyen, who are also working on related things, or might have some knowledge of how to deal with this.

Here's what I understand.

1. We have a [kludge in place to alter the way that `git` does a `checkout` in Travis](https://github.com/Seneca-CDOT/telescope/blob/master/.travis.yml#L20-L25).  I think @Manan311 did this, and I'm not sure where he found the code.  It works, but it shouldn't be necessary.  Basically, it alters the way that files are checked out into the working directory for Travis such that Windows line endings (e.g., `CRLF`) is *never* used, and everything is normalized to Unix line endings (`LF`).
1. We have recently added Prettier to the repo, and run it over all the files.  Our [Prettier config specifies using `LF` line endings](https://github.com/Seneca-CDOT/telescope/blob/master/.prettierrc#L4).
1. Our eslint uses airbnb's config, and I suspect it is setting a `linebreak-style` rule.  It's possible to [override this](https://eslint.org/docs/rules/linebreak-style#using-this-rule-with-version-control-systems)
1. We have `.editorconfig` ([using `LF`](https://github.com/Seneca-CDOT/telescope/blob/master/.editorconfig#L11)) and soon `.vscode` (#256) configuration hints.
1. Finally, our developers use a mix of Windows, Linux, and macOS to do development, so we have to be able to support everything.  Welcome to the real world!

Here's my proposal to try and fix this:

1. Create a `.gitattributes` file in the root of our project, which sets `* text=auto` and `eol=lf`.  You can see how the Prettier project does it [here](https://github.com/prettier/prettier/blob/master/.gitattributes).
1. Try removing the lines we have that I listed above for changing the `crlf` setting in git
1. Try pushing to git and building on Travis, and see what all 3 platforms do.

This may or may not solve it, but I think it will.

humphd:
@sukhbeersingh we talked briefly in class about you adding the `.gitattributes` file as discussed above, and also updating our eslint config to stop erroring on line ending style.  Are you going to file that, or do I need to?

humphd:
@robertbegna looks like this has stalled.  Are you still planning to work on it, or should we close for now?

humphd:
I'm going to close this for now, and if we want to restart it, let's open a new PR.

PR #195, opened 2019-11-14
Fix #108: integrate feed parser with feed worker

### LINKED ISSUE #108: Hook new feed parser into feed worker

jayson528:
updated function to return an array of "processed" post objects.

### DIFF
--- a/src/feed-worker.js
+++ b/src/feed-worker.js
@@ -19,7 +19,8 @@ exports.start = function () {
         };
         processedPosts.push(processedPost);
       });
+      // We can pass these objects into another queue, For now just printing to the console.
+      console.log(processedPosts);
     }
-    return processedPosts;
   });
 };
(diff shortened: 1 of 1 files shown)


src/feed-worker.js
```diff
@@ -1,14 +1,25 @@
...
+          author: post.author,
+          date: post.date,
+          title: post.title,
+          description: post.description,
+          postURL: post.link,
+        };
+        processedPosts.push(processedPost);
+      });
+    }
+    return processedPosts;
```
humphd: You can't really `return` anything here.  You're deep inside an async call that's working on processing URLs.  What we'll really need to do is feed these into another queue for analysis.

Can you just log something to the console instead?  I know it's minor, but it will help other people see that it's working.

PR #198, opened 2019-11-15
Added tests for email-sender.js

### LINKED ISSUE #82: Creating a Test that ensure's the email ability is working correctly

neilong31:
Fixes #82 

Currently adds two testers for email-sender.js, which uses nodemailer to send out emails.

I still have to add the code for the test recipients to come in as a .env variable.

What has been tested:

1. Testing that sendMessage is being called with the right parameters
1. That the resolve that sendMessage returns gives us the recipient that we passed to it, since it should resolve the recipients that have been accepted


env.example
```diff
@@ -22,3 +22,6 @@ NODEMAILER_SERVER=
 # BLOG_INACTIVE_TIME is the period (days) of inactivity
 # before a blog will be considered redlisted
 BLOG_INACTIVE_TIME=360
+
+# NODEMAILER_TESTRECIPIENT is a email that will receive email messages from telescope	
+NODEMAILER_TESTRECIPIENT=adell.crist@ethereal.email
```
manekenpix: I'm not sure about hardcoding the fake recipient into `env.example`. I think you can just add it to the test.

src/backend/web/app.js
```diff
@@ -9,4 +9,5 @@ app.use('/health', healthcheck());
 
 app.use('/', router);
 
+
```
manekenpix: Why the newline here?

env.example
neilong31: Yeah I totally understand and was thinking the same thing, but I saw idea put up [here](https://github.com/Seneca-CDOT/telescope/pull/93). I was trying to keep it consistent with the codebase. I can still change it, what do you think?

src/backend/web/app.js
neilong31: Yeah I messed up and removed the newline. It was an issue from a different branch.

env.example
RyanWils: I added the username, password and server to the .env file because it was confidential information. You may want to check what you can do with that link. Cause I don't think stuff like emails needs to be in the .env file as recipients is hard cored in the send message function

RyanWils:
I think it's a good base to start with. Once thing I noticed us that you should add comments to your code so people can use it after you finish creating it. I think you merged improperly as well. The proper commands if you want to pull to the latest version is git pull upstream master when your on the master branch then go to your branch and git rebase master

env.example
neilong31: Okay, I've removed it from the .env file. Please take a look

neilong31:
Yeah I know, I messed up. After I pulled git pulled upstream master, I git rebased master, but I forgot to git checkout into my branch.

env.example
neilong31: I used test recipient on the .env file because sendMessage needs a recipient to be passed for it to be tested with the expected output. Also, the reason I put it in the .env file is so anyone who is pulls down the codebase and wants to test with different email addresses can go directly to .env and not have to search the codebase to look where to find it. I could instead just remove the email and leave NODEMAILER_TESTRECIPIENT to be empty?

RyanWils:
Everyone looks good, maybe its better to group the .env variables together with the other nodemailer variables. You might want to put some instructions for creating test email (ethereal email). Is it not possible to fix the merges from the master? Its going to say you created/ touched alot of different files.

RyanWils:
it seems like const transporter = nodemailer.createTransport('SMTP', {  is an old format which is failing on the Travis Ci. 
You can remove the 'SMTP to make it const transporter = nodemailer.createTransport({
This works the same way, was just trying to be more explicit

neilong31:
@RyanWils Sounds good! I saw the PR you created. If the test is good as is for now do mind approving the PR? That way we can have the test when we test that new PR.

RyanWils:
@neilong31 The bug fix was pushed, is it possible to grab the latest changes and see if this function has any errors?

RyanWils:
Okay thats fine, It cannot connect cause their no server yet. Its not possible to fix the merges from main stream? Dave himself said to tell the others that we shouldn't do this.

test/email-sender.test.js
```diff
@@ -0,0 +1,20 @@
...
+  const testSubjectMessage = 'Test';
+  const testHTML = '<h1>Hello World</h1>';
+  const testRecipient = process.env.NODEMAILER_TESTRECIPIENT;
+  const testReturnValue = await emailsender.sendMessage(
+    testRecipient,
+    testSubjectMessage,
+    testHTML,
+  );
+  // Tests if the expected resolve is correct
+  expect(testReturnValue).resolves.toBe([testRecipient]);
```
varshannagarajan: This tests checks to see if it works given a good email, but we should also add a test with bad data to be more through. I will approve this PR but you should either add one later or create another issue for it.

test/email-sender.test.js
```diff
@@ -5,7 +5,7 @@ const emailsender = require('../src/email-sender');
 // and checks if it's resolve returns what is to be expected,
 // which an array of the testRecipient since it returns
 // an array of addresses it sent the email too
-test('Tests if sendMessage resolves with expected info', async () => {
+test.skip('Tests if sendMessage resolves with expected info', async () => {
```
KarlaChen: why this test need to skip?

neilong31: It will currently fail since we have no connection to the server, but the issue before has been fixed with #227

PR #200, opened 2019-11-15
Fix #62: added bull board visualization route

### LINKED ISSUE #62: Add UI dashboard support for Bull queues
We're using Bull to manage our work queues, so let's also add a UI so we can see what's happening.  https://github.com/vcapretz/bull-board will give us the ability to add a route to a web server to show what's happening in our queues:

Eventually this should be behind an admin login, but for now, we can expose it.  It will probably make debugging easier too.

varshannagarajan:
Added bull board to visualize the queues. Accessible from the router /admin/queues.

image


src/backend/web/routes/index.js
```diff
@@ -7,4 +7,6 @@ router.get('/', (req, res) => {
   res.sendFile(path.join(__dirname, '../../../frontend/index.html'));
 });
 
+router.use('/admin', require('./admin'));
```
miggs125: let's keep all the imports in one location at the top of the file.

src/backend/web/routes/admin.js
```diff
@@ -0,0 +1,8 @@
+const express = require('express');
+const { UI } = require('bull-board');
+
+const router = express.Router();
+
+router.use('/queues', UI);
```
miggs125: I'm not sure I understand what is going on here. Are you trying to add middleware to the router or the server? If it's the server you're after, the file you should be modifying is app.js. Check out that file and see how the middleware is implemented for the /health endpoint

src/backend/web/routes/admin.js
```diff
@@ -0,0 +1,8 @@
+const express = require('express');
+const { UI } = require('bull-board');
```
miggs125: this should be in app.js

src/backend/web/routes/admin.js
```diff
@@ -0,0 +1,8 @@
+const express = require('express');
```
miggs125: There is no need for this. We are already importing this in app.js

src/backend/web/routes/admin.js
```diff
@@ -0,0 +1,8 @@
...
+const { UI } = require('bull-board');
+
+const router = express.Router();
```
miggs125: not needed

src/backend/web/routes/admin.js
```diff
@@ -0,0 +1,8 @@
...
+
+router.use('/queues', UI);
+
+module.exports = router;
```
miggs125: not needed

src/index.js
```diff
@@ -6,6 +6,9 @@ const fs = require('fs');
 const feedQueue = require('./feed-queue');
 const feedWorker = require('./feed-worker');
 
+// Start the web server
+require('./backend/web/server');
```
miggs125: We are not starting the server here. This isn't needed.

humphd: For this PR they do need to start it.  The Queues and Express need to exist in the same context, so that Bull Queues are in scope.  I told them to do this here, just to get it working.  We'll refactor next week when we get all this code landed and integrate.

src/backend/web/routes/admin.js
humphd: The Bull UI adds a route to an existing express server, yes.  It gives you this dashboard based on queues you are running.

src/index.js
miggs125: sounds good to me

src/backend/web/routes/index.js
varshannagarajan: I do not believe this can go above line 4 where we declare router.

miggs125: that's okay, after router its fine. Move the require there, store it in a variable and replace the on this line with said variable

src/backend/web/routes/admin.js
varshannagarajan: I do believe this is all currently needed to make the routes work. I followed this guide (https://www.youtube.com/watch?v=paNikhYqdz0&t=1s) and checked the express documentation about express routing (https://expressjs.com/en/guide/routing.html) and their example at the very bottom of the screen also does it in this way.

I also resolved some of the other comments made in this file since they were all the same.

miggs125: @varshannagarajan  yup you're right! I stand corrected :)

src/backend/web/routes/index.js
vitokhangnguyen: I think that for the purpose of using Router and modularize our routes, doing `require` on other Router is a good practice since we will knnow exactly which file is associated to which parent route.

miggs125: @vitokhangnguyen  The idea is to keep all declarations at the top of the code so they are easy to find. Additionally, you can provide a useful and meaningful name to the import which will be more descriptive than the name of the file. Storing the require in a variable also allows us to change that name later on.  Also, image that your file is located in a folder structure such as:
```/foo/bar/baz/finally.js```

instead of writing ```router.use('/some/route/', require('./foo/bar/baz/finally.js')``` we can write
```router.use('/some/route/', finally)``` which is significantly cleaner.

I'd be happy to read any documentation stating otherwise, but I've yet to find an example in the [Express Router](https://expressjs.com/en/guide/routing.html) documentation and definitely not what is being stated in the [NodeJS best practices repository](https://github.com/goldbergyoni/nodebestpractices#-38-require-modules-first-not-inside-functions).

vitokhangnguyen: I think that it is easier to find the file if it is imported near where it is used, especially when it is guaranteed to be used only once (the file is small and a Route file does not usually has more than 1 export). Besides, I do not find what meaning a variable name can transmit that a filename cannot. In fact, having extra variables in the code file make it harder to trace the code.

I say this comes down to coding style and the author has his or her owned. I prefer to see what Router linked to what path so I can just jump there when I need to inspect the code, rather than having to trace the variable around the code (I might not know that I can find it at the top) and then the path.

miggs125: Again, the idea is to make this codebase easy to maintain. Additionally, your export could be an object with many methods, given your preference, I'd have to require and call the method all in the same line. You could say that we can make an exception in this case and put it at the top since the import is an object but then you'd be mixing 2 styles and making things more confusing for someone later on. Importing at the top is standard practice. It's what was taught in web222 and web322 and good practice as I've shown in the linked documentation above. You may be right it comes down to style, but we are aiming for consistency here and it's not about preference. This is the main reason we have a linter. We are not coding for ourselves but for the maintainers and future contributors to the codebase long after we are gone. If you fail to see how variable names can provide meaning and clarity to your code, then maybe read [this](https://medium.com/mindorks/meaningful-names-a-dimension-of-writing-clean-code-fdae1ae4f0b1) article.  I also dont want to have a file named some like: someRidiculouslyLongFileNameJustBecauseIDontWantToUseMeaningfulVariableNames.js. I will happily change my mind if you can link to credible documentation stating importing at the top is not both good and standard practice.

miggs125: Let's just get this code in and clean up later

humphd:
@varshannagarajan the `package-lock.json` file has drifted out of date, can you please update with `npm install` and push another commit?  We can merge as soon as you do, and we should do it now before other things change here.

varshannagarajan:
> @varshannagarajan the `package-lock.json` file has drifted out of date, can you please update with `npm install` and push another commit? We can merge as soon as you do, and we should do it now before other things change here.

I resolved the conflict on Github since it looked like a minor issue. I hope that's ok.

PR #201, opened 2019-11-15
Fix #71: use Redis to store parsed feed URLs and names

### LINKED ISSUE #71: Create relational database and its interface to store all information from parsed feeds
I would like to implement the MySql to store data from parsed feeds and overall information about posts. I think we will also need to have an interface for the server so back end scripts can exchange data. The data about that posts also needs to be properly organized  and structured. As a result, there should be a js script which gets infromation from MySql and format it into JSON so other back end scripts can use it. 

If you want to jump in and help me with MySql organization, feel free to join discussion.

mskuybeda:
@humphd  here is storage file modified according to your request with hash map sets returning only json object on read. The key of 1000 was set directly in redis db. Let me know if there is other way to do it or if this way is fine. 

PS: unit test will be done soon (before deadline)


src/storage.js
```diff
@@ -0,0 +1,34 @@
+//get config
```
humphd: Remove this

src/storage.js
```diff
@@ -0,0 +1,34 @@
...
+require('./config');
+//get redis constant searches redis globally
```
humphd: Remove this

src/storage.js
```diff
@@ -0,0 +1,34 @@
+//get config
+require('./config');
+//get redis constant searches redis globally 
+
+const Redis = require('ioredis'); //chnage to ioredis
```
humphd: Remove comment at the end

src/storage.js
```diff
@@ -0,0 +1,34 @@
...
+//connecting client
+const client = new Redis(process.env.REDIS_URL);
+
+//result from require, displays externally
+module.exports = {
+    addFeed: function(name, url, f) {
```
humphd: `ioredis` supports Promises, so let's make this async and return a Promise instead.

src/storage.js
```diff
@@ -0,0 +1,34 @@
...
+      const feed = { 'name': name, 'url': url }; //insted of json store id to a hash
```
humphd: `const feed = { name, url };`

src/storage.js
```diff
@@ -0,0 +1,34 @@
...
+      client.pipeline().incr('user_id').get('user_id', function(err, id) {
```
humphd: Does `incr('user_id')` work if it's never been set?  Also, should this be named `feed_id`?  It's not a user.

src/storage.js
```diff
@@ -0,0 +1,34 @@
...
+          if (err)
+            f(err, id);
+          else
+            client.hset('feeds', id, JSON.stringify(feed), f); //transform json object into string and save to redis feeds
```
humphd: Why not use `hmset` and then you can pass an Object vs. having to stringify to JSON?

humphd: See https://github.com/luin/ioredis/blob/90b1aecb4cc9f340aacb0dd54b200819092a4597/test/functional/transaction.ts#L136-L156 for an example of this.

humphd:
Also, you have a lot of lint errors, see https://travis-ci.org/Seneca-CDOT/telescope/builds/612154819#L229.  Make sure you run `npm run test` before you submit again, and fix them all.

src/storage.js
mskuybeda: I have not considered using hmset because it is deprecated as stated here. https://redis.io/commands/hmset

However, upon your recommendation I have tried it and here is the result it was giving me. 
image

Therefore I will stick with stringify

humphd: OK, `hset` it is.  But I still don't understand why you don't just set multiple key/value pairs as specified in https://redis.io/commands/hset.  Adding extra cost of stringify/parse seems unnecessary if it will do it for us automatically.

src/storage.js
mskuybeda: I was playing with it for couple of hours now. Could you show me a an example of how it suppose to work with promises?

humphd: From the ioredis README.md:

```js
// `exec` also returns a Promise:
var promise = redis
  .pipeline()
  .set("foo", "bar")
  .get("foo")
  .exec();
promise.then(function(result) {
  // result === [[null, 'OK'], [null, 'bar']]
});
```

mskuybeda:
@humphd I am trying to create a test file, however, I cannot connect to the redis-mock. I am using this code and it results in the following error 
"console.error node_modules/bull/lib/queue.js:526
      { Error: read ECONNRESET
          at TCP.onStreamRead (internal/stream_base_commons.js:111:27) errno: 'ECONNRESET', code: 'ECONNRESET', syscall: 'read' }"
`
const redis = require('redis-mock');

const client = redis.createClient('redis://127.0.0.1:6379');

const storage = require('../src/storage');

// insert new without hashes
test('', async () => {
  // expect(process.env.redis).toBe();
  const feed = { name: 'James Smith', url: 'http://seneca.co/jsmith' };
  let addedFeed;
  await storage.addFeed(feed.name, feed.url);
  await client.get('1000', (err, value) => {
    addedFeed = JSON.parse(value);
    expect(value).toBe(feed);
  });
  client.end();
  expect(addedFeed).toBe(feed);
});
`

humphd:
@mskuybeda you can use Travis to test this with your PRs, because Redis works there.

humphd:
How's this going, @mskuybeda?

mskuybeda:
@humphd well, it is going. Still stuck on test and trying to see what you meant by stringify/parse, since I stores obj.toString

humphd:
If you have to use stringify/parse to get this initial work finished sooner, so be it.  Later on we can migrate to try and use multiple JS key/value args.  It's not critical in this first PR.

mskuybeda:
Okay, on it. However the issue I am stuck on is that I do not get any vlaues from the test (the code I have sent before)

humphd:
I got this working for you, see my PR to your fork in https://github.com/mskuybeda/telescope/pull/1.

src/storage.js
```diff
@@ -0,0 +1,26 @@
...
+    // "If the key does not exist, it is set to 0 before performing the operation"
+    // https://redis.io/commands/INCR
+    const feedId = await redis.incr(FEED_ID);
+    await redis
+      .multi()
+      // Using hmset() until hset() fully supports multiple fields:
+      // https://github.com/stipsan/ioredis-mock/issues/345
+      // https://github.com/luin/ioredis/issues/551
+      .hmset(feedId, 'name', name, 'url', url)
+      .sadd('feeds', feedId)
```
humphd: `'feeds'` should be `FEEDS` here.

mskuybeda:
@humphd Thanks for your assistance. I added two more test for storage, first to check return of feed id in getFeeds after inserting and second to check increment

env.example
```diff
@@ -7,6 +7,9 @@ PORT=8080
 # REDIS_URL specifies Redis server info
 REDIS_URL=redis://127.0.0.1:6379
 
+# MOCK_REDIS=1 for testing
```
humphd: "MOCK_REDIS=1 will use an in-memory, mock Redis instance. Useful for testing."

package.json
```diff
@@ -35,11 +35,15 @@
...
+    "valid-url": "^1.0.9",
     "jsdom": "^15.2.1",
     "nodemailer": "^6.3.1",
-    "sentiment": "^5.0.2"
+    "sentiment": "^5.0.2",
+    "ioredis": "^4.14.1",
+    "ioredis-mock": "^4.18.2"
   },
   "devDependencies": {
+    "cross-env": "^6.0.3",
```
humphd: I added this, but then decided against using it in this PR.  Can you remove it?

src/storage.js
humphd: This got resolved in GitHub, but the code is still wrong.  Prefer the const `FEEDS` as the first argument here.

humphd:
I filed #258 to build on the `MOCK_REDIS` changes we did in here.

humphd:
@mskuybeda this added 160 commits unrelated to your PR.  That won't work.  Do you need help?

mskuybeda:
@humphd Yes, that is quite confusing now

humphd:
Out of curiosity, so I can help other students who are making the same mistake, what did you just do?  Which commands did you run when you tried to do this?

mskuybeda:
These ones that you told me to
"
git checkout master
git pull upstream master
git checkout issue-71
git rebase master
Fix your conflicts
rm -fr node_modules package-lock.json
npm install to update the package-lock.json
"

after that I used git pull, fixed conflicts and pushed

humphd:
It's the `git pull` at the end that did it.  Notice that wasn't in my set of commands I gave you.  When you `pull` you are doing a `fetch` and then `merge`, and this is why you ended up with all those commits merged into your PR.

I'm fixing this now locally, I'll merge in a minute...

humphd:
OK, merged!  This is on `master` now, so we can keep iterating there.  I'll close this.

mskuybeda:
great thanks

PR #207, opened 2019-11-15
Fix #115: added test cases for sentiment analysis

### LINKED ISSUE #115: Create Test for sentiment analysis of a blog.
Need to write test for code in https://github.com/Seneca-CDOT/telescope/blob/master/src/sentiment-analysis.js file to check its functionality.

I will take up this issue but will need help as I have not written tests from scratch. Also, other pull request are welcomed for fixing this issue.

dbeigi:
This pull request is to address #115 . I wrote two test cases that will test if sentiment analysis functionality will a) return the unexpected object b) return the expected object.

@jatinAroraGit would love to see your feedback on the following tests.

### DIFF
--- a/test/sentiment-analysis.test.js
+++ b/test/sentiment-analysis.test.js
@@ -1,0 +1,37 @@
+
+const sentimentAnalysis = require('../src/sentiment-analysis');
+
+const expected = {
+  score: 1,
+  comparative: 0.1111111111111111,
+  calculation: [{ allergic: -2 }, { love: 3 }],
+  tokens: [
+    'i',
+    'love',
+    'cats',
+    'but',
+    'i',
+    'am',
+    'allergic',
+    'to',
+    'them',
+  ],
+  words: [
+    'allergic',
+    'love',
+  ],
+  positive: [
+    'love',
+  ],
+  negative: [
+    'allergic',
+  ],
+};
+
+test('test if sample text does not return the correct object', () => sentimentAnalysis.run('I like dogs, but my girlfriend is afraid of them').then((data) => {
+  expect(data).toEqual(expect.not.objectContaining(expected));
+}));
+
+test('test if sample text returns the correct object', () => sentimentAnalysis.run('I love cats, but I am allergic to them.').then((data) => {
+  expect(data).toEqual(expect.objectContaining(expected));
+}));
--- a/test/sentiment-analysis.test.js
+++ b/test/sentiment-analysis.test.js
@@ -28,10 +28,12 @@ const expected = {
   ],
 };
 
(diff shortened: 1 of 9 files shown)


test/sentiment-analysis.test.js
```diff
@@ -0,0 +1,37 @@
...
+    'love',
+  ],
+  negative: [
+    'allergic',
+  ],
+};
+
+test('test if sample text does not return the correct object', () => sentimentAnalysis.run('I like dogs, but my girlfriend is afraid of them').then((data) => {
+  expect(data).toEqual(expect.not.objectContaining(expected));
+}));
```
manekenpix: Just a suggestion:
```js
test('test if sample text does not return the correct object', async () => {
  const data = await sentimentAnalysis.run('I like dogs, but my girlfriend is afraid of them');
  expect(data).toEqual(expect.not.objectContaining(expected));
});
```

test/sentiment-analysis.test.js
```diff
@@ -0,0 +1,37 @@
...
+
+test('test if sample text returns the correct object', () => sentimentAnalysis.run('I love cats, but I am allergic to them.').then((data) => {
+  expect(data).toEqual(expect.objectContaining(expected));
+}));
```
manekenpix: Same here
```js
test('test if sample text returns the correct object', async () => {
  const data = await sentimentAnalysis.run('I love cats, but I am allergic to them.');
  expect(data).toEqual(expect.objectContaining(expected));
});
```

dbeigi: applied the requested change.

test/sentiment-analysis.test.js
dbeigi: applied the requested change.

test/sentiment-analysis.test.js
```diff
@@ -0,0 +1,39 @@
...
+    'love',
+  ],
+  negative: [
+    'allergic',
+  ],
+};
+
+test('test if sample text does not return the correct object', async () => {
+  const data = await sentimentAnalysis.run('I like dogs, but my girlfriend is afraid of them');
+  expect(data).toEqual(expect.not.objectContaining(expected));
```
humphd: Why do you not just compare `data` to `expected` with a deep equals here?  What does `.toEqual(expect.not.objectContaining(expected))` mean?  Maybe you're right, but this seems odd.

dbeigi: @humphd From what I have read, `.toEqual` should be used for objects. Regards to `expect.not.objectContaining(expected)` , I am using this to check if the contents(properties) of the object do not match what we are expecting. This was suggested by @jatinAroraGit and after reading the docs I agreed to his suggestion.

humphd: Interesting, OK, sounds fine.  Thanks for educating me.

Issue #197, opened 2019-11-14
add test case for issue #22


dbeigi:
Need test cases for this issue #22

jerryshueh:
Thanks @dbeigi for opening this issue. To clarify, the inactive blog filter needs tests to verify it's actually returning properly for `update()` and `check()` functions. 

`check()` should return `false` if an item is **not** filtered (aka not listed in the redlist JSON), and `true` if it **is** redlisted.

`update()` parses the entire feeds list (from the text file, for now) using our feed-parser and checks whether or not the blog has been inactive for a certain amount of time (value is in .env). There are 3 scenarios:
* Feed is active (last post is within threshold)
* Feed is inactive (last post is older than threshold)
* Feed is dead (no posts or produces undefined data for whatever reason)
After running `update()`, the redlist JSON should be updated. We may need to create dummy blogs or alter the function to take dummy feed data to reproduce these conditions.

Silvyre:
Looking into this.

Silvyre:
Testing for `check()` was merged in #211.

As @jerryshueh mentioned, in order to produce proper testing for `update()`, we'll need to set up some testing conditions (e.g. a mock version of `feed-redlist.json`), or alter `update()`. If anyone has any ideas or suggestions regarding this, feel welcome to post them!

jerryshueh:
We _could_ pass `update()` the parsed JSON rather than the module reading the redlist itself. It was one of the things I debated, but in the end I decided to have the module handle everything because I wanted to blackbox it at the time. Also, since the parsing of feeds is asynchronous, the final JSON will not be the same each time, even if the URLs we pass are the same, because it all depends on the order the sync parsing happens. So rather than relying on the order of items in the file, we can only test whether a feed exists in the redlist at all. `check()` would be useful here, provided that it first passes successfully.

A method would be to change `update()` to receive the path to a JSON, with it taking the location of the operational redlist file by default (or better yet, take it from the .env). That way, if you _do_ pass it a filepath to some mock redlist, it will read from there instead. If you don't, it will default to the operational redlist JSON.

Just my two cents. If someone has more experience in developing for viable tests, please chime in!

ImmutableBox:
@Silvyre Hi, I also want to work on this issue. I just want to create a few tests for the `dateDiff` function to improve the coverage of `inactive-blog-filter`:

[...]

PR #211, opened 2019-11-15
Added tests for inactive blog checking (#197)

Silvyre:
This pull request adds testing for the `check()` function defined in [inactive-blog-filter.js](https://github.com/Seneca-CDOT/telescope/blob/master/src/inactive-blog-filter.js#L41).

This pull request does *not*, however, cover testing for the `update()` function. As noted in https://github.com/Seneca-CDOT/telescope/issues/197#issuecomment-554145637, the creation of some testing environment, or some alteration of the `update()` function needs to be done to facilitate proper testing. I plan to continue looking into this, and plan to make a separate pull request when I come up with a solution.

### DIFF
--- a/test/inactive-blog-filter.test.js
+++ b/test/inactive-blog-filter.test.js
@@ -1,0 +1,23 @@
+const fs = require('fs');
+const inactiveFilter = require('../src/inactive-blog-filter');
+
+const redlist = JSON.parse(fs.readFileSync('../feeds-redlist.json'));
+
+describe('Redlisted feed checking', () => {
+  it('should return false for bad feed URLs', async () => {
+    const badURLs = [redlist[0].slice(0, -1), '', undefined];
+    badURLs.forEach((url) => {
+      inactiveFilter.check(url, (callback, result) => {
+        expect(result).toBe(false);
+      });
+    });
+  });
+
+  it('should return true for all feed URLs in feeds-redlist.json', async () => {
+    redlist.forEach((feed) => {
+      inactiveFilter.check(feed.url, (callback, result) => {
+        expect(result).toBe(true);
+      });
+    });
+  });
+});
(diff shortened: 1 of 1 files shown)


test/inactive-blog-filter.test.js
```diff
@@ -0,0 +1,23 @@
+const fs = require('fs');
+const inactiveFilter = require('../src/inactive-blog-filter');
+
+const redlist = JSON.parse(fs.readFileSync('feeds-redlist.json', 'utf8'));
```
humphd: Change this to `const redlist = require('feeds-redlist.json');`

test/inactive-blog-filter.test.js
```diff
@@ -0,0 +1,23 @@
...
+
+describe('Redlisted feed checking', () => {
+  it('should return false for bad feed URLs', async () => {
```
humphd: You can remove `async` here, everything you're doing is sync

test/inactive-blog-filter.test.js
```diff
@@ -0,0 +1,23 @@
...
+    const badURLs = [redlist[0].url.slice(0, -1), '', undefined];
+    badURLs.forEach((url) => {
+      inactiveFilter.check(url, (callback, result) => {
+        expect(result).toBe(false);
+      });
+    });
+  });
+
+  it('should return true for all feed URLs in feeds-redlist.json', async () => {
```
humphd: Same thing here, drop `async`

test/inactive-blog-filter.test.js
Silvyre: Thank you for the review!
(Resolved by b8019ac.)

test/inactive-blog-filter.test.js
Silvyre: (Resolved by b8019ac.)

test/inactive-blog-filter.test.js
Silvyre: (Resolved by b8019ac.)

Issue #112, opened 2019-11-09
Meta Issue for Best practices research on using Bull and Redis


humphd:
We're using Redis and Bull to implement our backend queue pipeline.  Most of our development team has never used either before, so this bug is a starting point for documenting things *other* established projects do when they use these two technologies together.  To start I've been going through these:

- https://github.com/sourcegraph/sourcegraph
- https://github.com/coralproject/talk
-  https://github.com/zeit/node-file-trace
-  https://github.com/salesforce/refocus
-  https://github.com/OriginProtocol/origin
-  https://github.com/syuilo/misskey
-  https://github.com/Requarks/wiki
-  https://github.com/Chocobozzz/PeerTube
-  https://github.com/GetStream/Winds
-  https://github.com/withspectrum/spectrum
-  https://github.com/outline/outline
-  https://github.com/LearningLocker/learninglocker
-  https://github.com/ParabolInc/action

The goal of this bug is to highlight things we could also do.  People should file specific Issues and do Pull Requests to add/fix things in this list (i.e., this bug isn't about fixing them directly).  For people looking for things to work on, this list has some ideas, and if you dig through the code above, you'll find more.

## Redis

1. ~~Move the Redis URL into our `.env` and when we make Bull Queues, pass this URL on via the constructor.  With #105, we can do this.  Doing so will make it easier to switch between development and production environments:~~ happening in #144
1. We could take this one further and also add username, password, etc like [this code does](https://github.com/syuilo/misskey/blob/873444c3c6f7e3b9ecb6cce569d8cdb3fa032012/src/queue/index.ts#L17-L25).  See also [this code](https://github.com/syuilo/misskey/blob/65d359b57ae94ab9f53f995082dde645ebf7bf11/src/db/redis.ts) which manages the Redis client instance using it.
1. Wrapping Redis calls and exposing them as JS operations is an interesting idea, see [this code](https://github.com/Chocobozzz/PeerTube/blob/2b4dd7e26d93c2d9bef4f365cb03c511eff4ca8f/server/lib/redis.ts) especially lines 210 onward.  If we use [ioredis](https://github.com/luin/ioredis), as I've suggested in #71, it already has promise support.
[...]

PR #217, opened 2019-11-15
Fix #173 - Queue Shutdown

### LINKED ISSUE #173: Implement Queue Shutdown
As mentioned in Issue #112, We need to add a "graceful shutdown" for each queue, in the case that the process is being killed (by OS, user initiated via SIGHUP, etc).  The [docs on this with Bull are here](https://github.com/OptimalBits/bull/blob/develop/REFERENCE.md#queueclose), and you can see it happen in [this file](https://github.com/Chocobozzz/PeerTube/blob/2b4dd7e26d93c2d9bef4f365cb03c511eff4ca8f/server/lib/job-queue/job-queue.ts#L101-L106)

izhuravlev:
This PR fixes issue #173.

**I would like to explain what I have done.** 
In [the code example](https://github.com/Chocobozzz/PeerTube/blob/2b4dd7e26d93c2d9bef4f365cb03c511eff4ca8f/server/lib/job-queue/job-queue.ts#L101-L106), provided in issue #112 by @humphd, JobQueue is _a class_, and `terminate()` function is implemented as a member of this class - hence, it can be called from every place where `JobQueue` is used.
The main aim of the `terminate()` function is to close _a number of queues_, which are implemented in the `JobQueue class`. Implementing the same logic for us requires me to rewrite the structure of the `feed-queue.js`, making it a class.

**Another thing that I wanted to mention:**
I have made a little research on the "Killing" the process, and, as stated [here](https://stackoverflow.com/a/4479816):

 _"There is no way you can execute any code in your application when it is being Killed by Operating System or User. That is why it's called Killing."_

The `close()` function for the _graceful shutdown_, as I understood, is just a destructor for the queue, and can be called in the other files like `feed-worker.js` when there is no more need for the Queue to be used. 

**Although**, for the sake of error handling, logging and _graceful termination_ **I added** the try-catch block with the logging on the `feed-worker.js` file. 

I am open to any suggestions and will be happy to correct my work if I am wrong.

### DIFF
--- a/src/backend/index.js
+++ b/src/backend/index.js
@@ -3,13 +3,47 @@
  */
 
 const fs = require('fs');
+<<<<<<< HEAD:src/backend/index.js
 const { logger } = require('./utils/logger');
 const feedQueue = require('./feed/queue');
 const feedWorker = require('./feed/worker');
+=======
+const feedQueue = require('./feed-queue');
+const feedWorker = require('./feed-worker');
+const parentLogger = require('../src/lib/logger');
+
+const log = parentLogger.child({ module: 'module-name' });
+>>>>>>> Issue-173: Remade Server Graceful Shutdown:src/index.js
 
 const log = logger.child({ module: 'basic-queue' });
 // Start the web server
+<<<<<<< HEAD:src/backend/index.js
 require('./web/server');
+=======
+const server = require('./backend/web/server');
+
+/**
+ * Stops the Redis Queue, closes all connections gracefuly
+ */
+function shutDown() {
+  log.info('Received kill signal, shutting down gracefully');
+
+  // Close the Redis Queue
+  feedQueue.close();
+
+  // Close server and terminate all connections
+  server.close(() => {
+    log.info('Closed out remaining connections');
+    process.exit(0);
+  });
+
+  // Force shutting down
+  setTimeout(() => {
+    log.error('Could not close connections in time, forcefully shutting down');
+    process.exit(1);
+  }, 10000);
+}
+>>>>>>> Issue-173: Remade Server Graceful Shutdown:src/index.js
 
 /**
  * Process a string into a list of Objects, each with a feed URL
@@ -55,6 +89,9 @@ fs.readFile('feeds.txt', 'utf8', (err, lines) => {
     process.exit(-1);
     return;
   }
+
+  process.on('SIGTERM', shutDown);
+  process.on('SIGINT', shutDown);
 
   // Process this text file into a list of URL jobs, and enqueue for download
   const feedJobs = processFeedUrls(lines);
--- a/src/backend/index.js
+++ b/src/backend/index.js
@@ -3,23 +3,14 @@
  */
 
 const fs = require('fs');
-<<<<<<< HEAD:src/backend/index.js
-const { logger } = require('./utils/logger');
-const feedQueue = require('./feed/queue');
-const feedWorker = require('./feed/worker');
-=======
 const feedQueue = require('./feed-queue');
 const feedWorker = require('./feed-worker');
 const parentLogger = require('../src/lib/logger');
 
 const log = parentLogger.child({ module: 'module-name' });
->>>>>>> Issue-173: Remade Server Graceful Shutdown:src/index.js
 
 const log = logger.child({ module: 'basic-queue' });
 // Start the web server
-<<<<<<< HEAD:src/backend/index.js
-require('./web/server');
-=======
 const server = require('./backend/web/server');
 
 /**
@@ -43,7 +34,6 @@ function shutDown() {
     process.exit(1);
   }, 10000);
 }
->>>>>>> Issue-173: Remade Server Graceful Shutdown:src/index.js
 
 /**
  * Process a string into a list of Objects, each with a feed URL
--- a/src/backend/index.js
+++ b/src/backend/index.js
@@ -5,9 +5,9 @@
 const fs = require('fs');
 const feedQueue = require('./feed/queue');
 const feedWorker = require('./feed/worker');
-const parentLogger = require('./utils/logger');
+const { logger } = require('./utils/logger');
 
-const log = parentLogger.child({ module: 'main' });
+const log = logger.child({ module: 'main' });
 
 // Start the web server
 const server = require('./web/server');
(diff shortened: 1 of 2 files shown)


izhuravlev:
Just mentioned the stupidest "123" commit - I am sorry for this, this was done by VS code functionality which never worked for me before.

izhuravlev:
@humphd how do I fix that?

humphd:
Since there is so little code in this PR, I would suggest you do this:

```
git checkout master
git pull upstream master
git checkout -b old-issue-173 issue-173
git checkout -B issue-173 master
...re-do your changes in the correct files...
git add files...
git commit -m "Fix #173: add proper shutdown code for queue"
git push origin issue-173 -f
```

izhuravlev:
@humphd Sure do

izhuravlev:
I get an `npm test` error which I suppose is not related to my issue - I am not changing anything in either `feed-queue.js` or `inactive-blog-filters.js`. 
Should I ignore that? 
Screenshot: 
image

src/feed-worker.js
```diff
@@ -1,26 +1,35 @@
...
-          postURL: post.link,
-        };
-        processedPosts.push(processedPost);
-      });
-      // We can pass these objects into another queue, For now just printing to the console.
-      console.log(processedPosts);
-    }
-  });
+  // for the sake of error handling and gracefull shutdown
+  try {
```
humphd: I don't think wrapping this much code in a try catch is the right move.  We'd be better to do this in something like https://nodejs.org/api/process.html#process_event_uncaughtexception

Reza-Rajabi:
@izhuravlev I did a small research. [Here](https://hackernoon.com/graceful-shutdown-in-nodejs-2f8f59d1c357) you might find the steps that you wanna take: 
```
           Handle process kill signal
           Stop new requests from client
           Close all data process
           Exit from process
```

and [here](https://stackoverflow.com/questions/43003870/how-do-i-shut-down-my-express-server-gracefully-when-its-process-is-killed) is an example by code. Notice how they defined the shutdown function. Also we do need to take care of both `SIGTERM` and `SIGINT` like in the mentioned example. [Here](https://nodejs.org/api/process.html#process_signal_events) is why. So you can add something like: 

```javascript
// from stackoverflow: 
// https://stackoverflow.com/questions/43003870/how-do-i-shut-down-my-express-server-gracefully-when-its-process-is-killed
process.on('SIGTERM', shutDown);
process.on('SIGINT', shutDown);

function shutDown() {
    console.log('Received kill signal, shutting down gracefully');
    
    // We also need to close our Redis queue
    feedQueue.close();

    // continue from stackoverflow
    server.close(() => {
        console.log('Closed out remaining connections');
        process.exit(0);
    });

    setTimeout(() => {
        console.error('Could not close connections in time, forcefully shutting down');
        process.exit(1);
    }, 10000);
}
```

This way we make sure any undesired interruption gets handled.
I think you can add this pice in anywhere outside the function that you currently wrapped in try {} catch {}, but I am not sure if this module is the best candidate to have this code on.
I didn't try this code and I just tried to provide something that might be useful for the last minutes.

izhuravlev:
Thank you very much @Reza-Rajabi for your research!

izhuravlev:
In this commit, I have returned feed-worker.js to its initial state, created a shutDown function in the index.js, provided `SIGTERM` and `SIGINT` support inside the fs.readFile function and added logging.

src/index.js
```diff
@@ -5,9 +5,34 @@
 const fs = require('fs');
 const feedQueue = require('./feed-queue');
 const feedWorker = require('./feed-worker');
+const parentLogger = require('../src/lib/logger');
+
+const log = parentLogger.child({ module: 'module-name' });
```
humphd: `'module-name'` needs to be a name.  In this case, you could use something like `'main'`.

src/index.js
```diff
@@ -46,6 +71,9 @@ fs.readFile('feeds.txt', 'utf8', (err, lines) => {
     return;
   }
 
+  process.on('SIGTERM', shutDown);
+  process.on('SIGINT', shutDown);
```
humphd: Move these out of this function, to like 36 above.

miggs125:
Looks good aside from the changes @humphd requested.

humphd:
How is this coming?  Didn't see you on Thursday to find out, we should try and get this landed soon.

src/index.js
izhuravlev: Sure

izhuravlev:
Wanted to rebase my fixing commit on the commit ae756fd, but I don't think that it worked out as it should.

src/index.js
```diff
@@ -5,9 +5,37 @@
...
+
+  // Force shutting down
+  setTimeout(() => {
+    log.error('Could not close connections in time, forcefully shutting down');
+    process.exit(1);
+  }, 10000);
+}
+
+process.on('SIGTERM', shutDown);
+process.on('SIGINT', shutDown);
```
humphd: I wonder if we should just move all the "shutdown" logic in here.  In `src/backend/web/server.js` we have:

```js
process.on('unhandledRejection', err => {
  console.log('UNHANDLED REJECTION:  Shutting down...');
  console.log(err.name, err.message);
  server.close(() => {
    process.exit(1);
  });
});
```

Should we move that block of code here?  Then it's all contained in one place.

src/index.js
```diff
@@ -5,9 +5,37 @@
...
+const server = require('./backend/web/server');
+
+/**
+ * Stops the Redis Queue, closes all connections gracefuly
+ */
+function shutDown() {
+  log.info('Received kill signal, shutting down gracefully');
+
+  // Close the Redis Queue
+  feedQueue.close();
```
humphd: I think we should probably `await` this, or see if you can pass a callback.  At the very least, we should log errors that this produces.

src/index.js
izhuravlev: It looks for me as `process.on('unhandledRejection',...` is more related to any case of error handling (like, for all of the errors), while `SIGINT` and `SIGTERM` are more related to Shutdown - so they could stay separated.

But I believe none of the logic will be broken if we move it.

src/index.js
izhuravlev: @humphd Will using `await feedQueue.close()` be enough?

src/index.js
Reza-Rajabi: Maybe we could export `shutdown` function and call it from other modules like:
``` 
process.on('unhandledRejection', importedModule.shutDown(error));
```
that requires slightly updating the shutDown to log the error ...

src/index.js
Reza-Rajabi: That's so tricky. If we want to do it `await`, maybe we want also `setTimeout` before `await feedQueue.close();`

humphd: Here's my suggestion:

```js
setTimeout() => {
    log.error('Could not close connections in time, forcefully shutting down');
    process.exit(1);
}, 10000);

// Try to shut down 
Promise.all([
  // Shutdown the queue
  feedQueue.close().catch(err => log.error('Unable to close feed queue gracefully', err),
  // Shutdown the web server - we could also use util.promisify if you want
  new Promise((resolve, reject) => server.close(err => {
    if(err) {
      log.error('Unable to close web server gracefully', err);
      reject(err);
    } else {
      resolve();
    }
   })
])
.then(() => process.exit(0))
.catch(err => log.error(error));
```

humphd: I noticed this while looking up something for your code: https://expressjs.com/en/advanced/healthcheck-graceful-shutdown.html.

We could do one of those in a later PR too.  I don't want to expand this PR forever, but it's neat to see.

humphd:
I've tested this locally, made a few tweaks, and merged this into master.  I wanted to make sure it got in so we didn't push it back yet again.  Thanks for working on it.

PR #234, opened 2019-11-16
fix #91: created a login page for SAML testing

### LINKED ISSUE #91: Create a dummy HTML site for SAML2 Testing
Need a dummy SAML2 site for SAML2 testing this can be done with simple HTML and javascript and is used for testing purposes only.

agarcia-caicedo:
Created a simple html login page for SAML testing purposes

### DIFF
--- a/src/login/login.html
+++ b/src/login/login.html
@@ -1,0 +1,26 @@
+<form>
+  
+    <div class="container">
+      <!--For SAML to work an username is required, in this case SAML will connect with Seneca,
+      so the username in this case will be the Seneca provided email-->
+      <label for="uname"><b>Email</b></label>
+      <input type="text" placeholder="Enter Email" name="ename" required>
+      <br>
+      <br>
+      <!--Once SAML is implimented, the password should be the one for the Seneca email-->
+      <label for="psw"><b>Password</b></label>
+      <input type="password" placeholder="Enter Password" name="psw" required>
+      <br>
+      <br>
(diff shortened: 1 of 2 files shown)


Grommers00:
Looks good - very simple clean code,  hopefully  we can eventually add in more to use this as the base for our tests regarding saml2.  This may need to be updated later it looks like to reflect seneca's redirect page.

src/login/login.html
```diff
@@ -0,0 +1,23 @@
+<form>
+  
+    <div class="container">
```
neilong31: Maybe we can add a h1 tag that tells the user to login or something?

src/login/login.html
```diff
@@ -0,0 +1,23 @@
+<form>
+  
+    <div class="container">
+      <label for="uname"><b>Email</b></label>
+      <input type="text" placeholder="Enter Email" name="ename" required>
```
neilong31: Could we also add some comments explaining what things do for future students?

neilong31:
It looks good tho and shouldn't break anything :)

src/login/login.html
agarcia-caicedo: I added some comments, thanks for the feedback

src/login/login.html
agarcia-caicedo: As of right now, the login page is for testing purposes only, so I feel the tag would superfluous for this

PR #235, opened 2019-11-16
Fix #233 - add grid system

### LINKED ISSUE #233: Add grid system to index.html
It seems like there is no grid system for our mockups. I think it is a good idea to add a grid system.

cyh0968:
I added [normalize.css](https://necolas.github.io/normalize.css/) grid.css, and I added box-sizing: 

border-box since it makes easier to control. 
Using normalize makes browsers render more consistently.

### DIFF
--- a/src/frontend/index.html
+++ b/src/frontend/index.html
@@ -7,6 +7,8 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Telescope | Tracking Blogs in Orbit</title>
     <!-- CSS -->
+    <link rel="stylesheet" href="https://yarnpkg.com/en/package/normalize.css">
+    <link rel="stylesheet" href="./style/grid.css">
     <link rel="stylesheet" href="./style/main.css">
     <!-- JavaScript  -->
     <script src="./dialogue.js"></script>
--- a/src/frontend/style/grid.css
+++ b/src/frontend/style/grid.css
@@ -1,0 +1,662 @@
+/*
+The code is from http://www.responsivegridsystem.com/ under Creative Commons Attribution 3.0 License.
+https://creativecommons.org/licenses/by/3.0/
+*/
+
+.section {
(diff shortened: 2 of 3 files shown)


src/frontend/style/grid.css
```diff
@@ -0,0 +1,689 @@
+
+/*  SECTIONS  ============================================================================= */
```
humphd: Where is this code from?  Did you write all this from scratch?  If not, we need a license.

src/frontend/index.html
```diff
@@ -7,19 +7,23 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Telescope | Tracking Blogs in Orbit</title>
     <!-- CSS -->
+    <link rel="stylesheet" href="./style\normalize.css">
```
humphd: Can we not use CDN links?

src/frontend/style/grid.css
cyh0968: I'm not sure about the license. The code is from http://www.responsivegridsystem.com/
image 

Under the [Creative Commons Attribution 3.0 License](https://creativecommons.org/licenses/by/3.0/). If this license may cause an issue, I will replace the code with the bootstrap CDN link.

Unfortunately, the responsive grid system does not support the CDN link.

src/frontend/index.html
cyh0968: I updated it to the CDN link.

humphd:
Why not just use CSS Grid?  It's all baked into browsers now: https://css-tricks.com/snippets/css/complete-guide-grid/

humphd:
@cyh0968 I'm going to close this for now, and get you to do a new PR on just adding the `normalize.css`.  Do that as a CDN link in the frontend index file, and do it in a new branch off current master.  Let's leave the grid out until we have code that needs it, and then, try to use more modern options.

PR #237, opened 2019-11-16
fixed #64 added OPML generator

### LINKED ISSUE #64: Add OPML endpoint export for feed list
In #59 we're adding the ability to ingest our current feed list from the wiki.  In #63 we're adding a web server.  It might be nice to connect both of these and offer a way to get the feed list *out* of our app as an [OPML](https://en.wikipedia.org/wiki/OPML) file.  OPML seems to be the standard way of sharing a feed list these days.

What I envision is adding a route that allows for obtaining an XML file in OMPL format with all our student feeds included.  We might be able to use something like https://www.npmjs.com/package/@kohlmannj/opml-generator to do it.

jrdnlx:
Added a js file that has an opml generator to generate as an opml file to fix issue #64 
I had to export the objArray array is wiki-parser.js to access the rss urls and the processedPosts array to access the posts url and posts title. 

I am not sure if this is correctly so please correct me if i did this incorrectly.

### DIFF
--- a/src/backend/web/routes/opml.js
+++ b/src/backend/web/routes/opml.js
@@ -1,0 +1,28 @@
+const express = require('express');
+const opml = require('opml-generator');
+
+const router = express.Router();
+
+const header = {
+  title: 'OPML Feeds',
+  dateCreated: new Date(2020, 2, 9),
+  ownerName: 'Jordan',
+};
+
+const outlines = [
+  {
+    text: 'txt',
+    title: 'My Open Source Experience',
+    type: 'rss',
+    xmlUrl: 'https://jrdnlxopensource.blogspot.com/feeds/posts/default/-/open-source',
+    htmlUrl: 'https://jrdnlxopensource.blogspot.com/',
+  },
+];
+
+router.get('/', (req, res) => {
+  // call the opml() function here, and return it on res, using the correct content-type of " text/x-opml"
+  res.setHeader('Content-type', 'text/x-opml');
+  res.send(opml(header, outlines));
+});
+
+module.exports = router;
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "node-fetch": "2.6.0",
     "node-summarizer": "1.0.7",
     "nodemailer": "6.3.1",
+    "opml-generator": "1.1.1",
     "parse-github-url": "1.0.2",
     "passport": "0.4.0",
     "passport-openid": "0.4.0",
--- a/src/backend/web/routes/index.js
+++ b/src/backend/web/routes/index.js
@@ -1,11 +1,13 @@
 const express = require('express');
 const path = require('path');
 const admin = require('./admin');
+const opml = require('./opml');
 
 const router = express.Router();
 
 router.use(express.static(path.join(__dirname, '../../../frontend')));
 
 router.use('/admin', admin);
+router.use('/opml', opml);
 
 module.exports = router;


Grommers00:
As soon as this passes the ci/riccleCI and travis build, I'll take a look, but looks like you need to run it through an ESL lint before pushing.

humphd:
@jrdnlx this is getting further out of date.  We should figure out how to get it moving forward and merged.  Can we discuss on Thursday in the lab?

jrdnlx:
Yes sorry, I would love to, or maybe your office hours

humphd:
This is still outstanding, can we get it finished soon?

jrdnlx:
@humphd its failed a check that is in the dependency that has been provided. Thoughts?
 
image

src/opml-generator.js
```diff
@@ -0,0 +1,17 @@
...
+
+const outlines = {
+  text: 'txt',
+  title: 'My Open Source Experience',
+  type: 'rss',
+  xmlUrl: 'https://jrdnlxopensource.blogspot.com/feeds/posts/default/-/open-source',
+  htmlUrl: 'https://jrdnlxopensource.blogspot.com/',
+};
+
+module.exports = opml(header, outlines);
```
humphd: `outlines` is expected to be an Array vs. an Object, see https://www.npmjs.com/package/opml-generator#usage

src/backend/web/routes/opml.js
```diff
@@ -0,0 +1,13 @@
+const { XMLHttpRequest } = require('xmlhttprequest');
+const express = require('express');
+const opml = require('../../../opml-generator.js');
+
+const router = express.Router();
+
+const xhr = new XMLHttpRequest();
```
humphd: This section can be swapped out for a route, something like this:

```js
router.get('/opml', (req, res) => {
  // call the opml() function here, and return it on res, using the correct content-type of " text/x-opml"
});
```

src/backend/web/routes/opml.js
```diff
@@ -0,0 +1,13 @@
...
+xhr.open('GET', opml);
+xhr.responseType = 'document';
+xhr.send();
+
+module.exports = xhr;
```
humphd: You can get rid of this.

src/opml-generator.js
```diff
@@ -0,0 +1,19 @@
+const opml = require('opml-generator');
```
humphd: I would move all this code into your opml.js route.  I don't think there's much value breaking it out here.

src/backend/web/routes/opml.js
```diff
@@ -0,0 +1,28 @@
...
+    type: 'rss',
+    xmlUrl: 'https://jrdnlxopensource.blogspot.com/feeds/posts/default/-/open-source',
+    htmlUrl: 'https://jrdnlxopensource.blogspot.com/',
+  },
+];
+
+router.get('/opml', (req, res) => {
+  // call the opml() function here, and return it on res, using the correct content-type of " text/x-opml"
+  opml(header, outlines);
+  res.setHeader('Content-type', 'text/x-opml');
```
birtony: Right now the return of `opml` is not used anywhere. You need to send it back with the `res`. I assume, you could use [res.send](https://sailsjs.com/documentation/reference/response-res/res-send) method for that.

src/backend/web/routes/opml.js
```diff
@@ -0,0 +1,28 @@
...
+  ownerName: 'Jordan',
+};
+
+const outlines = [
+  {
+    text: 'txt',
+    title: 'My Open Source Experience',
+    type: 'rss',
+    xmlUrl: 'https://jrdnlxopensource.blogspot.com/feeds/posts/default/-/open-source',
+    htmlUrl: 'https://jrdnlxopensource.blogspot.com/',
```
birtony: According to the [documentation](https://github.com/azu/opml-generator#usage), it seems that `xmlUrl` and `htmlUrl` should be used with quotes:
```js
'xmlUrl': 'https://jrdnlxopensource.blogspot.com/feeds/posts/default/-/open-source',
'htmlUrl': 'https://jrdnlxopensource.blogspot.com/',
```

jrdnlx:
@humphd I think I fixed it, I ran it locally and it works. Please double check

Thanks

PR #252, opened 2019-11-16
Fix #145 Adding logging to events in the feed queue

### LINKED ISSUE #145: Add logging to events in the feed queue
As suggested by @humphd in #130.

evlnyng:
Fixes issue #145, using our new logger to log events that are emitted from our feed queue. 

This is pretty much just the code found [here](https://github.com/OriginProtocol/origin/blob/e9881d687241fdbf557239a90eeac12d96792d62/infra/cron/src/scheduler.js#L22-L73), which is under the MIT license. 

One thing I am concerned about is that it posts to the console during the tests (see below). I am looking into this, but if anyone has any advice or knows how to fix this, I'd love your help! 

Screen Shot 2019-11-16 at 4 51 21 PM

Additionally, there are some events (`cleaned` and `stalled`, for example) that aren't emitted when running `npm start` at the moment. I am researching on how to write tests for them.

### DIFF
--- a/src/feed-queue.js
+++ b/src/feed-queue.js
@@ -1,11 +1,72 @@
 const Bull = require('bull');
 const { setQueues } = require('bull-board');
+const parentLogger = require('../utils/logger');
+
+const log = parentLogger.child({ module: 'feed-queue' });
 
 require('./config');
 
-const queue = new Bull('feed-queue', process.env.REDIS_URL || 'redis://127.0.0.1:6379');
+/**
+ * Function to log all events in the feed queue
+ */
+function watch(feedQueue) {
+  feedQueue
+    .on('error', (error) => {
+      // An error occured
+      log.error(`Queue ${feedQueue.name} error:`, error, ' Stack: ', error.stack);
+    })
+    .on('waiting', (jobID) => {
+      // A job is waiting for the next idling worker
+      log.info(`Job ${jobID} is waiting.`);
+    })
+    .on('active', (job) => {
+      // A job has started (use jobPromise.cancel() to abort it)
+      log.info(`Job ${job.id} is active`);
+    })
+    .on('stalled', (job) => {
+      // A job was marked as stalled. This is useful for debugging
+      // which workers are crashing or pausing the event loop
+      log.info(`Job ${job.id} has stalled.`);
+    })
+    .on('progress', (job, progress) => {
+      // A job's progress was updated
+      log.info(`Job ${job.id} progress:`, progress);
+    })
+    .on('completed', (job) => {
+      // A job has been completed
+      log.info(`Job ${job.id} completed.`);
+    })
+    .on('failed', (job, err) => {
+      // A job failed with an error
+      log.error(`Job ${job.id} failed:`, err);
+    })
+    .on('paused', (job) => {
+      // The queue was paused
+      log.info(`Queue ${feedQueue.name} resumed. ID:`, job.id);
+    })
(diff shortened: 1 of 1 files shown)


src/feed-queue.js
```diff
@@ -1,11 +1,72 @@
...
+      // (the last item in the queue was returned by a worker)
+      log.info(`Queue ${feedQueue.name} was drained.`);
+    })
+    .on('removed', (job) => {
+      log.info(`Job ${job.id} was removed.`);
+    });
+}
 
 // For visualizing queues using bull board
+const queue = new Bull('feed-queue', process.env.REDIS_URL);
```
manekenpix: Before this PR, this line was like this:
```js
const queue = new Bull('feed-queue', process.env.REDIS_URL || 'redis://127.0.0.1:6379');
```

Is this intentional?

evlnyng: Hey, nope! Something strange was happening with my VSCode, and I didn't catch that - I'll fix it now.

humphd:
I'll file a separate issue on the logs-in-tests solution.

humphd:
Filed #255 for the logging issue.

humphd:
I also notice that Travis is failing all 3 builds:

```
Ran all test suites.

{"level":30,"time":1573944613057,"pid":6089,"hostname":"travis-job-ebc4a21f-0e09-479e-8d21-c4f521d7e550","module":"feed-queue","msg":"Job 1 is waiting.","v":1}

Jest did not exit one second after the test run has completed.

This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

The build has been terminated
```

This looks like the redis connection not being closed properly in our tests, which I thought we fixed.  Is this failure specific to this change, or was it happening before?

humphd:
I think this change is causing it, since your [last PR that landed on `master` was all green](https://travis-ci.org/Seneca-CDOT/telescope/builds/612835397?utm_medium=notification&utm_source=github_status).

humphd:
I suspect these event listeners are leaking, and holding open the queue instance.  We're might need to wrap the [`close()` method](https://github.com/OptimalBits/bull/blob/develop/lib/queue.js#L533) and manually remove these event listeners.

Looking at the [`Bull code`](https://github.com/OptimalBits/bull/blob/develop/lib/queue.js#L322), the `Queue` object is an `EventEmitter`, which means it has [`.removeAllListeners('event-name')`](https://nodejs.org/api/events.html#events_emitter_removealllisteners_eventname).

Let's sort this out before we land this, and break CI.

humphd:
It might actually be the [Bull Board visualization route](https://github.com/Seneca-CDOT/telescope/blob/master/src/backend/web/routes/admin.js) we added, which is getting started by `index.js`.  The call to [setQueues()](https://github.com/Seneca-CDOT/telescope/blob/master/src/feed-queue.js#L9) actually [returns the references to the queues it creates](https://github.com/vcapretz/bull-board/blob/master/index.js#L41), and we aren't using it.  We might need to close them somehow...

cc @varshannagarajan, @vitokhangnguyen , who I think worked on this code originally?

humphd:
Continuing this work in #288.

PR #139, opened 2019-11-11
Fix issue #96: Combine EditConfig, ESLint, and Prettier to manage code style and formatting

vitokhangnguyen:
This PR solves issue #96 

# Changes Summary
- Prettier added as a dev-dependency and will handle code styling issues from now on.
- Adjust ESLint/Prettier rules to adapt with EditCofnig's rule
- Add a few rules to avoid potential bugs/typos

PR #253, opened 2019-11-16
Add Husky to Telescope

manekenpix:
This adds [`huskyjs`](https://github.com/typicode/husky#husky) to Telescope.

Huskyjs is a npm package that allows custom scripts to be ran against a repository before or after git events take place. For Telescope, it'll run `lint`, included in [`package.json`](https://github.com/Seneca-CDOT/telescope/blob/master/package.json), right before a collaborator tries to commit. If it fails, the changes will not be committed.

More info about huskyjs [here](https://dev.to/chriss/git-hooks-and-huskyjs-13gb).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -552,6 +552,12 @@
       "resolved": "https://registry.npmjs.org/@types/node/-/node-8.10.59.tgz",
       "integrity": "sha512-8RkBivJrDCyPpBXhVZcjh7cQxVBSmRk9QM7hOketZzp6Tg79c0N8kkpAIito9bnJ3HCVCHVYz+KHTEbfQNfeVQ=="
     },
+    "@types/normalize-package-data": {
+      "version": "2.4.0",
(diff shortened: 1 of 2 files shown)


manekenpix:
@humphd I just took a look at #139 and it seems @vitokhangnguyen hasn't included husky yet. I can mention it to him, or I can wait until he merges his PR and then use mine with his changes so husky will use prettier instead of eslint (which is what it does right now). What do you recommend?

humphd:
Let's build on his, and maybe layer in some other commit hooks to do things after we start using Prettier.  Keep this open, and rebase when his happens?

manekenpix:
Ok, I'll wait for his.

humphd:
#282 is in, if you want to iterate on it now.

manekenpix:
@humphd He already added Husky:

> Make npm run prettier script as a precommit hook using Husky

I'll close this one.

Issue #256, opened 2019-11-17
Add a .vscode folder with VSCode Workspace recommendations


humphd:
Many of us use VSCode to do our development.  One of the features it allows is creating a set of [project-wide settings for all users who use VSCode](https://code.visualstudio.com/docs/getstarted/settings).  For example, you can see how [Microsoft does this for VSCode itself](https://github.com/microsoft/vscode/tree/master/.vscode).  A Workspace can include things like:

- common editor settings (e.g., project indent amounts, newline type, etc)
- [recommended extensions](https://code.visualstudio.com/docs/editor/extension-gallery#_workspace-recommended-extensions) for this project (for us that would be eslint, prettier, EditorConfig, etc.)
- launch commands to be run from within VSCode's debugger

The [docs for Workspace Settings](https://vscode.readthedocs.io/en/latest/getstarted/settings/) have lots of ideas.

Let's see if we can help our developers by adding a `.vscode/` dir with Workspace settings.

vitokhangnguyen:
This issue when solved would bring many benefits to #96 where we will have ESLint, Prettier and EditorConfig working

manekenpix:
Closed with #285

PR #285, opened 2019-11-19
Fix issue #256: Add .vsccode folder

vitokhangnguyen:
This PR addresses issue #256 

I added a basic `.vscode` folder with some code formatting config, extension recommendation (eslint, prettier and editorconfig) and a script to run debugging in VSCode.

I am open to more idea on what to add.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,8 @@
+{
+  "recommendations": [
+    "editorconfig.editorconfig",
+    "dbaeumer.vscode-eslint",
+    "esbenp.prettier-vscode",
+    "streetsidesoftware.code-spell-checker"
+  ]
+}
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,12 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Launch Program",
+      "program": "${workspaceFolder}/src/index.js",
+      "skipFiles": ["<node_internals>/**"]
+    }
+  ]
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,9 @@
+{
+  "editor.insertSpaces": true,
+  "editor.tabSize": 2,
+  "editor.detectIndentation": false,
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "editor.formatOnSave": true,
+  "files.eol": "\n",
+  "files.insertFinalNewline": true
+}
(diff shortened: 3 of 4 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+  "editor.insertSpaces": true,
+  "editor.tabSize": 2,
+  "editor.detectIndentation": false,
+  "files.eol": "\n",
+  "files.insertFinalNewline": true
+}
```
manekenpix: How about adding something like this:
```json
// Set the default
"editor.formatOnSave": false,
```

Or maybe doing it for a specific language:
```json
"[javascript]": {
    "editor.formatOnSave": true
}
```

More [here](https://github.com/prettier/prettier-vscode)

.vscode/extensions.json
```diff
@@ -0,0 +1,7 @@
+{
+  "recommendations": [
+    "editorconfig.editorconfig",
+    "dbaeumer.vscode-eslint",
+    "esbenp.prettier-vscode"
+  ]
```
humphd: Can you add "Code Spell Checker" too? `"streetsidesoftware.code-spell-checker"` to help people catch spelling mistakes in comments, variable names, etc.  It's great!

.vscode/launch.json
```diff
@@ -0,0 +1,12 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Launch Program",
+      "program": "${workspaceFolder}/src/index.js",
+      "skipFiles": ["<node_internals>/**"]
```
humphd: In a follow-up PR, it would be cool to add various launch.json options to debug the code too.

.vscode/settings.json
humphd: Yeah, that might be nice, so it happens while you're in the tool without having to run `eslint -fix` or prettier

vitokhangnguyen: Yea, this is a good idea. I will have that in.
I was wondering between leaving the ability of free-styling to the developer until they make a commit (we already have a precommit hook) or enforcing it every time they save a file, too.

.vscode/extensions.json
vitokhangnguyen: Definitely!

.vscode/launch.json
vitokhangnguyen: I can file an issue for this

vitokhangnguyen:
@humphd Most of the settings will be automatic except...

1. Recommended extensions. There will be a pop-up from VSCode for this but we might have to click on it and install them manually.
1. Launch config. Maybe we can add a small doc about this part. Currently, there is nothing huge about this.

So if we agree to write any, how do you want me to put it?

Issue #258, opened 2019-11-17
Allow tests to use mock Redis with Bull Queue


humphd:
In #201 we're adding the initial support for Redis in our code.  Part of that change includes adding support for the `MOCK_REDIS` environment variable.  When `MOCK_REDIS=1` is set, we have code to swap our Redis JS provider from [ioredis](https://github.com/luin/ioredis) to [ioredis-mock](https://github.com/luin/ioredis), an in memory replacement that simulates talking to a Redis database server.

So far, we only use it in our storage code, and the Bull `Queue` we have for the feed queue isn't using it.  Bull allows you to [supply your own Redis client instances to a Queue](https://github.com/OptimalBits/bull/blob/28a2b9aa444d028fc5192c9bbdc9bb5811e77b08/PATTERNS.md#reusing-redis-connections).

Using this, we can modify [feed-queue.js](src/feed-queue.js] to use the `lib/redis` we're adding in #201, and pass the 3 Redis connections a `Queue` needs. Doing so will let our tests run without a Redis server in development, or on CI.

PR #288, opened 2019-11-20
Fix #252, #258, #299: add logging to feed queue, refactor lib/redis, lib/queue

humphd:
I'm testing a theory on the test failures in #252.  To do it, I made a fix for #258, and merged/updated the work in #252.  I want to see what this does on Travis.

### DIFF
--- a/src/feed-queue.js
+++ b/src/feed-queue.js
@@ -1,9 +1,8 @@
-const Bull = require('bull');
 const { setQueues } = require('bull-board');
+const { createQueue } = require('./lib/queue');
 
-require('./config');
-
-const queue = new Bull('feed-queue', process.env.REDIS_URL || 'redis://127.0.0.1:6379');
+// Create a Bull Redis Queue
+const queue = createQueue('feed-queue');
 
 // For visualizing queues using bull board
 setQueues(queue);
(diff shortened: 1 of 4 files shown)


humphd:
OK, this worked.  Here's what I've done:

1. Rewritten `src/lib/redis.js` to expose the ability to create a new Redis (or MockRedis) instance vs. only exposing a pre-made instance.
1. Updated `src/storage.js` to use this change
1. Merged the work in https://github.com/Seneca-CDOT/telescope/pull/252 by @evlnyng to add logging to the feed queue
1. Refactored the feed queue to put most of the logic of a queue in `src/lib/queue.js`, where I manage the redis connections according to the [Bull docs](https://github.com/OptimalBits/bull/blob/28a2b9aa444d028fc5192c9bbdc9bb5811e77b08/PATTERNS.md#reusing-redis-connections).  We need this so that we can have a Bull queue use the fake redis connections, and not timeout in tests on the real redis server.

To test this, you can do a few things:

```
$ LOG_LEVEL=debug npm test
...
{"level":20,"time":1574216938808,"pid":6817,"hostname":"MacBook-Pro","msg":"Shimming MockRedis info() method","v":1}
```

or run the server:

```
npm start
```

@evlnyng I'm going to close your PR, since your work is in this one now.  Please take a look at my changes and see if it's OK.

PR #291, opened 2019-11-20
Fixes 272: Initial set up of webhint to Telescope

robertbegna:
I've done the initial set up of webhint to the project following the user guide https://webhint.io/docs/user-guide/ and also the other links provided in the issue. For setting up webhint for local dev, I assumed the port being used was 3000, assuming we're using react for that. Please let me know if that should be changed, and anything else that should be addressed


.hintrc
```diff
@@ -0,0 +1,5 @@
+{
+    "extends": [
```
humphd: We're using 2-space indents (this looks like 4)

.hintrc
```diff
@@ -0,0 +1,5 @@
...
+        "all"
+    ]
+}
```
humphd: It's always good to add a newline at the end of the last line in every file.

package.json
```diff
@@ -24,7 +24,9 @@
     "test": "npm run jest",
     "jest": "cross-env MOCK_REDIS=1 jest",
     "start": "node src",
-    "debug-server": "nodemon server.js"
+    "debug-server": "nodemon server.js",
+    "webhint": "hint http://localhost:3000",
```
humphd: Because `webhint` requires the server to be running, you should make sure it's up first.  We'll need to add [npm-run-all](https://www.npmjs.com/package/npm-run-all) to do this in a cross platform way (i.e., vs using `&&`):

```
npm install --save-dev npm-run-all
```

Then you can do:

```
"webhint": "cross-env PORT=3000 npm-run-all -r -p start hint",
```

So this defines a `hint` and `webhint` script.  The `webhint` is what we'd use to run things, and it defines the server port to use as 3000, then runs two things in parallel: our server (`start`) and also `hint`

humphd: https://webhint.io/docs/user-guide/development-flow-integration/local-server/ discuss this.

package.json
```diff
@@ -24,7 +24,9 @@
...
+    "test-hint": "hint ./test/test_files"
```
humphd: This can be removed.

.circleci/config.yml
```diff
@@ -13,3 +13,5 @@ jobs:
       - run:
           name: Execute tests
           command: npm test
+      - run: npm install
+      - run: npm run test-hint
```
humphd: I don't know if we should do this on CI or not.  I lean towards 'no' and let people run it locally.

.travis.yml
```diff
@@ -40,3 +40,6 @@ before_install:
       redis-server --service-start
       redis-cli info
     fi
+script:
+  - npm install
+  - npm run test-hint
```
humphd: Same thing here.  Let's let people run it manually on local dev machines.

robertbegna:
So after installing brew and redis on my mac, it appears to working?
Running npm run webhint will generate a hint-report directory, in which there is an index.html file which generates a report for all the errors on the front end

robertbegna:
added the hint-rc folder to the .gitignore file so it won't cause further builds to crash

humphd:
I've rebased, extended this, and merged in d36593d7b5785912933f1aeac250eb02dc9c98db.

PR #293, opened 2019-11-20
Fix #280 Move jest-fetch-mock to devDependencies

### LINKED ISSUE #280: Move `jest-fetch-mock` from `dependencies` to `devDependencies` in `package.json`
The `dependencies` needs to have only production-level dependencies. So we need to move `jest-fetch-mock` from `dependencies` to `devDependencies` in the `package.json` file

manekenpix:
Closes #280 

With this, `jest-fetch-mock` is now part of `devDependencies`.

Also, I moved `setupJest.js` to a lib folder.


miggs125:
LGTM

PR #295, opened 2019-11-20
Remove pacakge-lock.json, switch to exact versions in package.json

humphd:
This PR changes our tree to use [exact npm versions](https://docs.npmjs.com/misc/config#save-exact) instead of a `package-lock.json` file, and removes the existing `package-lock.json` file from git, as well as all reference to it.  I've added it to `.gitignore` so it won't go in again.

I'm suggesting this change to try and help devs working on the project avoid merge conflicts and constant churn in `package-lock.json`.  This is a [best practice](https://devcenter.heroku.com/articles/node-best-practices#use-a-smart-npmrc) used by many projects, and I *hope* it will alleviate some of the headaches people are having doing rebases.

NOTE: I've also removed the `loglevel=silent` line from our `.npmrc` file, because it has been masking some errors on CI when installs fail.


humphd:
cc @sukhbeersingh, @robertbegna: looks like Circle CI isn't happy with me for removing the lock file.  Is there some way to clear its cache and/or let it operate without a lock file?

humphd:
OK, fixed by switching from `npm ci` to `npm install`.  The former requires a lock file, the latter does not, see https://docs.npmjs.com/cli/ci.html#description.

humphd:
Everything is green, so this should be good for review now.

humphd:
Let's see how this goes!  Fingers crossed...

PR #297, opened 2019-11-21
Fix #172: implement queue failed feeds behaviour

### LINKED ISSUE #172: Implement Queue Failed Feeds/Downloads Behaviour
As described by @humphd in Issue #112:
"One of our design goals is to have failed downloads of feeds get automatically retried some number of times, but to apply an exponential backoff so we don't just go into an endless loop of failure. [This code shows one way to do this via options to the Bull queue.](https://github.com/syuilo/misskey/blob/873444c3c6f7e3b9ecb6cce569d8cdb3fa032012/src/queue/index.ts#L87-L96)"

ragnarokatz:
Reference issue #172.

**Status update:**
Ready for review.

Solution: extracted worker callback function out, changed it to using promises instead of done, made it much easier to cope with unit testing.

**Historical conversations:** (outdated, still good for reference)

WIP Update:
I need help with writing the test cases to validate the `feed-worker.js` module. Currently it is not working.

Contrary to its name, the function `queue.process` does not process anything immediately upon its call, **it's actually registering a callback function for the queue that is only called everytime a job is placed in the queue**. See bull's official documentation here: https://github.com/OptimalBits/bull/blob/develop/REFERENCE.md#queueprocess.

Since code inside the callback function gets called at a later time, I need to have a different way of testing their functionalities rather than just having a straight up function call. I've consulted bull library's own set of test cases, namely this one: https://github.com/OptimalBits/bull/blob/develop/test/test_sandboxed_process.js to help with setting up my own.

Basically the current format is roughly this:
```
feedWorker.start(); // calls feedQueue.process() and hook up callback function

feedQueue.on('completed', (job, value) => { // hook up callback to test the success branch of the callback function
  expect(job.data).toEqual(job1.data);
  expect(value[0].title).toEqual('First Post!');
  done();
});

feedQueue.on('error', error => { // hook up callback to test the failure branch of the callback function
  expect(error.message.startsWith('Failed to parse url: '));
  done();
});

feedQueue.add(job1); // adds a job into the queue, triggering the callback
```

However, the tests seem to unable to reach `done()` inside the callback handlers, and it seems that `feedQueue.on('completed')` and `feedQueue.on('error')` never get called. As a result, the tests just hang and jest doesn't exit cleanly.

### DIFF
--- a/src/feed-worker.js
+++ b/src/feed-worker.js
@@ -2,15 +2,14 @@ const feedQueue = require('./feed-queue');
 const feedParser = require('./feed-parser');
 const extractUrls = require('./extract-urls');
 
-exports.start = function() {
-  // Start processing jobs from the feed queue...
-  feedQueue.process(async job => {
-    const { url } = job.data;
+exports.workerCallback = async function(job) {
+  const { url } = job.data;
+  try {
     const posts = await feedParser(url);
     const processedPosts = [];
     if (posts.length > 0) {
       posts.forEach(post => {
-        // We can extract any other infromation from the post that we need here.
+        // We can extract any other information from the post that we need here.
         const processedPost = {
           author: post.author,
           date: post.date,
@@ -23,6 +22,15 @@ exports.start = function() {
       });
       // We can pass these objects into another queue, For now just printing to the console.
       console.log(processedPosts);
+      return Promise.resolve(processedPosts);
     }
-  });
+    return Promise.reject(new Error(`Failed to extract posts from url: ${url}`));
+  } catch (err) {
+    return Promise.reject(err);
+  }
+};
+
+exports.start = function() {
+  // Start processing jobs from the feed queue...
+  feedQueue.process(this.workerCallback);
 };
--- a/src/index.js
+++ b/src/index.js
@@ -32,7 +32,15 @@ function processFeedUrls(lines) {
 async function enqueueFeedJobs(feedJobs) {
   feedJobs.forEach(async feedJob => {
     console.log(`Enqueuing Job - ${feedJob.url}`);
-    await feedQueue.add(feedJob);
+    await feedQueue.add(feedJob, {
+      attempts: 8,
+      backoff: {
+        type: 'exponential',
+        delay: 60 * 1000,
+      },
+      removeOnComplete: true,
+      removeOnFail: true,
+    });
   });
 }
 
--- a/test/feed-worker.test.js
+++ b/test/feed-worker.test.js
@@ -1,0 +1,41 @@
+const feedWorker = require('../src/feed-worker');
+
+describe('testing the worker callback function', () => {
+  const job1 = {
+    data: { url: 'https://c3ho.blogspot.com/feeds/posts/default/-/open-source?alt=rss' },
+  };
+  const job2 = { data: { url: 'c3ho.blogspot.com/feeds/posts/default/-/open-source?alt=rss' } };
+  const job3 = { data: { url: 'https://c3ho.blogspot.com/feeds/posts/default/-/open-source' } };
+  const job4 = { data: { url: 'c3ho.blogspot.com/feeds/posts/default/-/open-source' } };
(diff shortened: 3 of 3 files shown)


ODAVING:
@ragnarokatz squash the commits so it passes the CI test

test/feed-worker.test.js
```diff
@@ -0,0 +1,48 @@
...
+// which will block jest from completing.  After all tests are done, close it.
+afterAll(() => feedQueue.close());
+
+describe('testing the worker callback function', () => {
+  feedWorker.start();
+
+  const job1 = { data: 'https://c3ho.blogspot.com/feeds/posts/default/-/open-source?alt=rss' };
+  const job2 = { data: 'c3ho.blogspot.com/feeds/posts/default/-/open-source?alt=rss' };
+  const job3 = { data: 'https://c3ho.blogspot.com/feeds/posts/default/-/open-source' };
+  // const job4 = { data: 'c3ho.blogspot.com/feeds/posts/default/-/open-source' };
```
ODAVING: Is there a reason why these are still here? if they're not going to be used, maybe remove them?

humphd:
My fix in #288 should fix the timeout issue.  Waiting on review.

humphd:
OK, #288 is in.  If you rebase on that, it should help.

ragnarokatz:
@humphd thanks. I will try and let you know.

test/feed-worker.test.js
ragnarokatz: They are going to be used. It's just that even the first three tests are not passing right now, so they are temporarily commented out.

ragnarokatz:
@humphd It seems that the callback attached to `queue.process` doesn't get called when a new job is added into queue using `queue.add`. I see that `npm test` uses `cross-env MOCK_REDIS=1 jest "./test/feed-worker.test.js"` to run the tests, could it be that `bull` is not capable of reproducing the producer/consumer process due to MockRedis?

ragnarokatz:
After a bit of research, https://github.com/OptimalBits/bull/issues/962 pretty much answers the above question. `bull` only depends on `ioredis`, therefore neither `redis-mock, fakeredis nor ioredis-mock` are feature complete from `bull`. I'll have to come up with some other way to test this.

humphd:
@ragnarokatz good job digging into this.  Let's discuss tomorrow and see if we can figure out another plan.

We might need to use real Redis to test the portion that you care about, mock for the rest.

ragnarokatz:
@ODAVING @humphd @manekenpix @izhuravlev This PR is now ready for review. Please take a look when you have time, thanks.

src/feed-worker.js
```diff
@@ -23,6 +22,15 @@ exports.start = function() {
       });
       // We can pass these objects into another queue, For now just printing to the console.
       console.log(processedPosts);
+      return Promise.resolve(processedPosts);
```
humphd: NOTE: Because you already `await` on line 8 above, and because this is an `async` function, you can skip the `Promise.resolve()` and `Promise.reject()` throughout this function.  An `async` function implicitly returns a `Promise`.  What you've done here will work, but is unnecessary, and the `Promises` will get unwrapped to expose your result as intended.

If you were in a function, and wanted to convert a synchronous call to be a `Promise`, you could do this.  However, if you're going to use both `Promise.resolve()` and `Promise.reject()`, you might was well `return new Promise((resolve, reject) => { /* your code here */ })`.

We should probably clean this up later in another PR, since this has been merged.

src/index.js
```diff
@@ -32,7 +32,15 @@ function processFeedUrls(lines) {
 async function enqueueFeedJobs(feedJobs) {
   feedJobs.forEach(async feedJob => {
     console.log(`Enqueuing Job - ${feedJob.url}`);
-    await feedQueue.add(feedJob);
+    await feedQueue.add(feedJob, {
+      attempts: 8,
```
humphd: These magic numbers (`8`, `60 * 1000`) should probably go in the env.  Would be good in another PR.

Issue #221, opened 2019-11-15
Uncaught error in dialogue.js


jatinAroraGit:
Type: Bug. 
Area: Front-End

**Description**:
There is an error originating from  dialogue.js file.

**How to reproduce**:
Make sure you are running the latest copy of the project.

1. Open index.html
2. In Chrome , open Developer settings and go to console tab

You will see the following error:

![image](https://user-images.githubusercontent.com/28763265/68974214-2c802e00-07be-11ea-862c-f20b20316410.png)

By clicking on the file name you can see what line in the file is throwing the error.

![image](https://user-images.githubusercontent.com/28763265/68974087-df9c5780-07bd-11ea-9e33-37774ec9f96f.png)

humphd:
Looks like a DOM timing bug on startup, wrap this in a `DOMContentLoaded` event.

jatinAroraGit:
Hey @humphd ,
I did what you recommended but it still throws the error.
Looks like there are lot of things wrong in this file.
We do not have any button with the id="participants"
and neither any span element with that class.
I ran the in-built debugger in chrome. It only hits the button call if I add the id in the participants button in index.html.
Rest I am not sure what this file was for but needs some updating.

humphd:
See https://github.com/Seneca-CDOT/telescope/blob/master/src/frontend/index.html#L35, the `<button>` needs an `id="participants"` for this selector to work.

ODAVING:
@jatinAroraGit during iterations of the FE the dialogue portion of index.html had gotten removed. The code for that is linked below. Please make sure that when participants button gets clicked it'll open a new animated window, where you can either close it with the 'x' mark on the top right or get rid of it if you were to click outside the dialogue box area. That was the original intention of the dialogue.js

Hope this guides you: 

index.html
`

    <div id="dialogue" class="dBox normalTheme">
        
        <div class="dBox-content normalTheme">
            <h1>Participants
                <span class="dBox-close">&bigotimes;</span></h1>
            <div class="row normalTheme" id="srchDiv">
                <div class="col">
                    <form action="#" class="srchBox" style="margin: auto; max-width: 270px;">
                        <input type="text" placeholder="Search..." name="srchBtn">
                        <button type="submit">Find</button>
                    </form>
                </div>
            </div>
            <br>
            <div class="container-fluid normalTheme">
                <div class="row">
                    <div class="col-md-2" id="dBox-text">
                        TEST CONTENT
                    </div>
[...]

PR #301, opened 2019-11-21
This will fix the uncaught errors in dialogue.js and wil fix Issue #221

jatinAroraGit:
In file dialogue.js:
Wrapped the whole code in DOMContentLoaded event.
Removed code related to element span , which is not being used anywhere in index.html.

Moved the call to script below the index.html file where all other scripts are.


jatinAroraGit:
Hey @humphd , 
Did I did rebasing wrong ?
By the looks of the tree I think I did.

humphd:
Closing in favour of #308

PR #302, opened 2019-11-21
Search bar added to Navbar instead of participants

### LINKED ISSUE #242: Add search bar for the mock-ups
It seems like that the telescope has its search bar to search the participant. However, it is uncomfortable to find the search bar. I think we should expose the search bar on the top always in mobile and desktop environments. For instance, the GitHub site has a search bar on the top, so we don't have to look for it. With one click, we can search whatever we want.

ODAVING:
fixes  #242 
Added the search button as well as its animation to HTML/CSS. The button will transition between two states, as pictured below 
**FROM:**
issue242_a
**TO:**
issue242_b

### DIFF
--- a/src/frontend/style/main.css
+++ b/src/frontend/style/main.css
@@ -19,6 +19,7 @@ body {
   --nav-btn-bgcolor: #fdfdfd;
   --content-font-color: #000000;
   --font-main: 'Montserrat', Arial, sans-serif;
+  --srch-after: #da291c;
 }
 
 /* MAIN GRID */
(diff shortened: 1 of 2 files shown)


src/frontend/style/main.css
```diff
@@ -1,3 +1,6 @@
+/* Font Used for Search Icon */
+@import url('https://fonts.googleapis.com/css?family=Inconsolata:700');
```
lucacataldo: Maybe keep font dependencies to a minimum? Not only for a more consistent design but also to avoid loading a bunch (although this one is only ~60KB, but it adds up).

ODAVING: changed it to the main font we use in HTML.

Issue #310, opened 2019-11-21
Removing unnecessary console message in tests


evlnyng:
<img width="401" alt="Screen Shot 2019-11-21 at 11 58 02 AM" src="https://user-images.githubusercontent.com/34118418/69359329-68037800-0c56-11ea-9699-c8de181e81d3.png">

Pictured above.

humphd:
Duplicate of #244, closing both now that this is merged.

PR #311, opened 2019-11-21
Fix issue #310 - Remove unnecessary console message

evlnyng:
Fixes issue #310. 

Before:
Screen Shot 2019-11-21 at 11 58 02 AM

After:
Screen Shot 2019-11-21 at 12 02 27 PM


miggs125:
thanks!

PR #320, opened 2019-11-21
Connected index.html and login.html

agarcia-caicedo:
Added connectivity with the main page and login, this allows the login project to be better integrated with the main project.


varshannagarajan:
Looks straight forward, but I do think when we are making changes to the html we should include screen shots of what the changes look like.

Issue #318, opened 2019-11-21
add fixtures.js in test folder


ragnarokatz:
In a similar format used by `bull`'s own test suites, I would like us to implement our own set of fixtures to be used for testing. They can be easily used to mock things, as well as reduce repetition in the code.

Bull's own testing fixtures: https://github.com/OptimalBits/bull/tree/develop/test/fixtures

In the immediate time frame, I would like the following code from `test/feed-parser.test.js` to be moved into the proposed `test/fixtures.js`:
```
const nock = require('nock');

...

nock('https://test321.blogspot.com')
    .get('/feeds/posts/default/-/open-source')
    .reply(
      200,
      '<?xml version="1.0" encoding="UTF-8" ?><rss version="2.0"><channel><title>W3Schools Home Page</title><link>https://www.w3schools.com</link><description>Free web building tutorials</description><item><title>RSS Tutorial</title><link>https://www.w3schools.com/xml/xml_rss.asp</link><description>New RSS tutorial on W3Schools</description></item><item><title>XML Tutorial</title><link>https://www.w3schools.com/xml</link><description>New XML tutorial on W3Schools</description></item></channel></rss>'
    );

...

nock('https://test321.blogspot.com')
    .get('/feeds/posts/default/-/open-source?alt=rss')
    .reply(
      200,
      '<?xml version="1.0" encoding="UTF-8" ?><rss version="2.0"><channel></channel></rss>'
    );
```

These code appear repeated in the file, and there's potential for them to be used elsewhere as well (in the upcoming `feed-worker.test.js` file)

jerryshueh:
I don't see any self-assignees or PRs on this issue, so I assume it is up for grabs? If so, I'd like to try my hand at this seeing as I haven't had a hand in anything test-related. This might be a good start :smiley:.

ragnarokatz:
@jerryshueh Yes it is up for grabs. Feel free to work on it.

humphd:
Closed in #321

PR #321, opened 2019-11-21
Addresses #318 - Add fixtures.js

jerryshueh:
Added new `fixtures.js` module for reusable code in testing. For now, as @ragnarokatz suggested, it contains some `nock` calls that were previously used in `feed-parser.test.js`, but could have been reused. I also created constants for the test blog URIs (atom/rss) as well as the valid feed return body, exposing them through the module. This should help begin our development of more robust tests by introducing fixtures that reproduce specific test environments. Once again, thanks to @ragnarokatz for providing the suggestion! #318

### DIFF
--- a/test/feed-parser.test.js
+++ b/test/feed-parser.test.js
@@ -1,4 +1,5 @@
 const nock = require('nock');
+const fixtures = require('./fixtures');
 const feedParser = require('../src/feed-parser');
 
 /**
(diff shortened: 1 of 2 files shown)


ragnarokatz:
@humphd Yes, I'm planning on using these functions for the upcoming `feed-worker.test.js` file, where I would like to pass in the last two URL test cases from `feed-parser.test.js` to check the results.

PR #324, opened 2019-11-22
Fix #323: Disable linebreak style and add gitattributes

### LINKED ISSUE #323: Add .gitattributes file and disable linebreak-style
Currently, when windows users run npm test against the latest changes prettier complains about line endings. Same for our CI services. See #194. This issue can be solved for the time being by disabling the `linebreak-style` rule from `.eslintrc.js`.

sukhbeersingh:
This PR disallows eslint to enforce consistent line endings. This is being done to support our windows users, who are currently not able to pass linting tests. To make sure any code that is added now has consistent line endings, 'lf' in our case, I have also added a `.gitattributes` file. You can read more about it [here](https://git-scm.com/docs/gitattributes). 
Specifically about [text](https://git-scm.com/docs/gitattributes#_code_text_code) and [eol](https://git-scm.com/docs/gitattributes#_code_eol_code).

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -28,5 +28,11 @@ module.exports = {
      * https://eslint.org/docs/rules/func-names
      */
     'func-names': 'off',
+
+    /**
+     * Disallow enforcement of consistent linebreak style
+     * https://eslint.org/docs/rules/func-names
+     */
+    'linebreak-style': 'off',
   },
 };
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - 'node'
+  - lts/*
 services:
   - redis-server
 os:


.eslintrc.js
```diff
@@ -28,5 +28,11 @@ module.exports = {
      * https://eslint.org/docs/rules/func-names
      */
     'func-names': 'off',
+
+    /**
+     * Disallow enforcement of consistent linebreak style
+     * https://eslint.org/docs/rules/func-names
+     */
+    'linebreak-style': 0,
```
manekenpix: Just a very tiny nitpick. For consistency, can you make it "off" instead of 0?

sukhbeersingh: Thanks for pointing that out. I totally missed that.

manekenpix:
Oh, you have to rebase to fix the logger problem.

sukhbeersingh:
Rebased. Successful on Circle but Travis doesn't like it. Doesn't show anything descriptive.

manekenpix:
@sukhbeersingh it doesn't seem to be related to your fix, right?

sukhbeersingh:
Not as far as I think.

humphd:
Have we seen this failure before?  It's really odd.

I do note that it's passing eslint across all platforms, which is good.

manekenpix:
@humphd it seems to be feedParser-test, but I just ran it locally and it passed, so I'm not sure about what it could be. Maybe with the PR you just submitted it'll fix it?

humphd:
It looks like it started failing here https://travis-ci.org/Seneca-CDOT/telescope/builds/615287253

humphd:
Screen Shot 2019-11-21 at 10 35 12 PM

Build on the left worked (https://travis-ci.org/Seneca-CDOT/telescope/jobs/615216407).  Build on the right failed (https://travis-ci.org/Seneca-CDOT/telescope/jobs/615287254).  I notice the version of node went from 13.1.0 to 13.2.0.

Maybe that's breaking us?

@sukhbeersingh can you try this again by specifying a node.js version of `lts/*` to use the LTS version vs. latest unstable in the Travis CI file?, see https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#specifying-nodejs-versions

sukhbeersingh:
So this is kind of weird and may be a false flag, but I went to work on another branch, came back to this branch `issue323` and the fix doesnt work anymore. I get the same error message from prettier about `Delete CR`. I am doing all this on windows. Do I need to add something of [this](https://stackoverflow.com/a/53769213) sort to the prettier config?

humphd:
What is the error you get from Prettier?

sukhbeersingh:
The exact same as in the picture I posted yesterday https://github.com/Seneca-CDOT/telescope/pull/194#issuecomment-556027611

manekenpix:
@humphd your PR for more tests uses node 13.2.0 and it failed. My version at home (passing the tests) 12.13. Maybe that's the reason?

sukhbeersingh:
You were right guys @humphd @manekenpix . It was the latest version of node causing problems.

humphd:
Nice!

For this eol issue, I don't understand why we're still seeing it.  Silly question, does `'linebreak-style': 0,` work where `'linebreak-style': 'off',` doesn't, maybe?

Beyond that, I don't get why it would error for you.

sukhbeersingh:
No , setting it to `off` or `0` doesnt make a difference.

humphd:
From the git docs, `Note that setting core.autocrlf to true or input overrides core.eol (see the definitions of those options in git-config(1)).`  So maybe your local settings are overriding this change?

humphd:
See https://mirrors.edge.kernel.org/pub/software/scm/git/docs/gitattributes.html

sukhbeersingh:
> core.autocrlf

Indeed, it was set to true, I changed it to input now. Still not effect, but I think thats what the problem was. Should we put it in docs to inform windows users to set their config to input?

humphd:
From https://prettier.io/docs/en/options.html#end-of-line, they say it should be `false`

humphd:
I think we should take this PR on master.  It will make things better, but clearly not 100%.  Let's follow up in later PRs.

Issue #257, opened 2019-11-17
sort URLs by type


smilegodly:
Take the URLs from the array of URLs and sort them by type...
An example would be sorting all Github URL's to be in their own array to later be used by Github API.

Cadesh:
Can I work on this please?

manekenpix:
Thank you for the idea, we might come back to this sometime in the future. I'm closing this issue now and we can open a new one if/when we decide to add this.

PR #331, opened 2019-11-23
added sort URL by type function

smilegodly:
wrote a function that takes in the array of URLs, a search term, and returns an array of only the searched for URLs.

fixes https://github.com/Seneca-CDOT/telescope/issues/257

### DIFF
--- a/src/feed-worker.js
+++ b/src/feed-worker.js
@@ -1,6 +1,7 @@
 const feedQueue = require('./feed-queue');
 const feedParser = require('./feed-parser');
 const extractUrls = require('./extract-urls');
+const sortUrls = require('./sort-urls');
 
 exports.start = function() {
   // Start processing jobs from the feed queue...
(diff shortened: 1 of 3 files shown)


src/feed-worker.js
```diff
@@ -19,7 +20,9 @@ exports.start = function() {
           postURL: post.link,
         };
         processedPosts.push(processedPost);
-        extractUrls.extract(post.description);
+
+        const arrayOfUrls = extractUrls.extract(post.description);
+        sortUrls.sortUrlsByType(arrayOfUrls, 'https://github.com');
```
humphd: We already have code that does this in https://github.com/Seneca-CDOT/telescope/blob/master/src/github-url.js

cindyledev:
So what's happening here? Is this stale now?

manekenpix:
Thank you for your contribution, we might come back to this sometime in the future. I'm closing this PR now and we can submit a new one if/when we decide to add this.

PR #332, opened 2019-11-23
Fix #260 implement addPost and getPost method

### LINKED ISSUE #260: Allow a Post to get stored in Redis
In #201 we added [storage.js](src/storage.js) and some initial [tests](test/storage.test.js).  We need to extend that to allow for a Post to be saved and returned from Redis.  We can (eventually) connect this with the work happening in #259 to add a `Post` type, accepting and returning `Post` objects.

We need `addPost()` and `getPost()` methods to be added to the `storage` module.  The code can be similar to what we have now for `addFeed()` and `getFeed()`.  In Redis, a `Post` can be a Hash, with all the different properties in the `Post` object corresponding to fields in the Hash--like the feed does with `name` and `url`.

Each `Post` in Redis will need a unique key.  Let's start by using a `Post`'s `guid` property, and see if that works well enough.

We'll also want to create a `posts` Sorted Set in Redis to hold all the `guid` keys for the posts that we store.  Adding an item to the `posts` set should include the `published` time as a score. In other words, we'll sort the posts in our set according to the date they were written, so it's easy to get them out in the order we want.

See https://redis.io/topics/data-types and other docs on how the Hash and Sorted Set data types work in Redis, and https://github.com/luin/ioredis for docs on how to call them from JavaScript.

KarlaChen:
Added addPost() function using guid as unique key for each post and store posts as sorted set by published date.
Added getPost() by guid and getPosts() by range of dates.
Added test for testing above mentioned functions.

### DIFF
--- a/src/storage.js
+++ b/src/storage.js
@@ -1,4 +1,4 @@
-const { redis } = require('../lib/redis');
+const { redis } = require('./backend/lib/redis');
 
 // Redis Keys
 const FEED_ID = 'feed_id';
--- a/src/storage.js
+++ b/src/storage.js
@@ -51,13 +51,13 @@ module.exports = {
         'site',
         post.site
       )
-      .zadd(POSTS, new Date(post.published).getTime(), post.guid) // sort set by published date as scores
+      .zadd(POSTS, post.published.getTime(), post.guid) // sort set by published date as scores
       .exec();
   },
 
   getPosts: (startDate, endDate) =>
     // Get all posts between start and end dates in the sorted set
-    redis.zrangebyscore(POSTS, new Date(startDate).getTime(), new Date(endDate).getTime()),
+    redis.zrangebyscore(POSTS, startDate.getTime(), endDate.getTime()),
 
   getPost: guid => redis.hgetall(guid),
 };
--- a/test/storage.test.js
+++ b/test/storage.test.js
@@ -14,4 +14,50 @@ describe('Tests for storage', () => {
     const result = await storage.getFeed(feedId);
     expect(result).toEqual(feed2);
   });
+
+  const post = {
+    guid: 'tag:blogger.com,1999:blog-7100164112302197371.post-522285656016053350',
+    author: 'Neil David',
+    title: 'My First Blog',
+    link: 'http://nadavid.blogspot.com/2008/09/my-first-blog.html',
+    content:
+      "I have never done this before, so let's make this short. This post is just a test on how my blog post will look like.",
+    text: 'post one text',
+    updated: '2009-09-07T22:23:00.544Z',
+    published: '2009-09-07T22:20:00.000Z',
+    url: 'http://seneca.co/jsmith',
+    site: 'wordpress.com',
+  };
+
+  const post2 = {
+    guid: '2tag:blogger.com,1999:blog-7100164112302197371.post-522285656016053350',
+    author: 'Neil David2',
+    title: 'My First Blog2',
+    link: 'http://nadavid2.blogspot.com/2008/09/my-first-blog.html',
+    content:
+      "I have never done this before, so let's make this short. This post is just a test on how my blog post will look like.",
+    text: 'post one text2',
+    updated: '2008-09-07T22:12:00.544Z',
+    published: '2008-09-07T22:09:00.000Z',
+    url: 'http://seneca.co/jsmith',
+    site: 'wordpress.com',
+  };
+
+  it('should allow retrieving a post by guid after inserting', async () => {
+    await storage.addPost(post);
+    const result = await storage.getPost(post.guid);
+    expect(result.link).toEqual(post.link);
+  });
+
+  it('get all posts returns corrent number of posts', async () => {
+    await storage.addPost(post2);
+    const result = await storage.getPosts(post2.published, post.published);
+    expect(result.length).toEqual(2);
+  });
+
+  it('get all posts returns sorted posts by date', async () => {
+    const result = await storage.getPosts(post2.published, post.published);
+    expect(result[0]).toEqual(post2.guid);
+    expect(result[1]).toEqual(post.guid);
+  });
 });
--- a/test/storage.test.js
+++ b/test/storage.test.js
@@ -23,8 +23,8 @@ describe('Tests for storage', () => {
     content:
       "I have never done this before, so let's make this short. This post is just a test on how my blog post will look like.",
     text: 'post one text',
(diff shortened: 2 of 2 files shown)


src/storage.js
```diff
@@ -23,4 +26,41 @@ module.exports = {
...
 
   getFeed: feedID => redis.hgetall(feedID),
+
+  //  addPost
+  addPost: async post => {
+    const postId = await redis.incr(POST_ID);
+    await redis
+      .multi()
+      .hmset(
+        postId,
```
humphd: Let's try using `post.guid` here, and not bother with `postId` or `POST_ID` above.  I think the `guid` should be unique enough to use here, but we can experiment.

src/storage.js
```diff
@@ -4,8 +4,11 @@ const { redis } = require('./lib/redis');
 const FEED_ID = 'feed_id';
 const FEEDS = 'feeds';
 
+const POST_ID = 'post_id';
+const POSTS = 'posts';
+
 module.exports = {
-  addFeed: async (name, url) => {
+  addFeed: async (author, url) => {
```
humphd: Why this change?  It's not wrong, but it seems unrelated to this fix.

src/storage.js
```diff
@@ -4,8 +4,11 @@ const { redis } = require('./lib/redis');
 const FEED_ID = 'feed_id';
 const FEEDS = 'feeds';
 
+const POST_ID = 'post_id';
```
humphd: I bet we can get rid of this, see my comment below.

src/storage.js
```diff
@@ -23,4 +26,41 @@ module.exports = {
   getFeeds: () => redis.smembers(FEEDS),
 
   getFeed: feedID => redis.hgetall(feedID),
+
+  //  addPost
+  addPost: async post => {
+    const postId = await redis.incr(POST_ID);
```
humphd: Let's drop this.

src/storage.js
```diff
@@ -23,4 +26,41 @@ module.exports = {
   getFeeds: () => redis.smembers(FEEDS),
 
   getFeed: feedID => redis.hgetall(feedID),
+
+  //  addPost
```
humphd: This comment isn't adding anything, let's remove it.

src/storage.js
```diff
@@ -23,4 +26,41 @@ module.exports = {
...
+  addPost: async post => {
+    const postId = await redis.incr(POST_ID);
+    await redis
+      .multi()
+      .hmset(
+        postId,
+        'guid',
```
humphd: If we store the post by `guid`, we probably don't need to store it in the hash, this can be removed.

src/storage.js
```diff
@@ -23,4 +26,41 @@ module.exports = {
...
+        'updated',
+        post.updated,
+        'published',
+        post.published,
+        'url',
+        post.url,
+        'site',
+        post.site
+      )
+      .zadd(POSTS, Date.parse(post.published), postId)
```
humphd: I think that we can assume that `post.published` will already be a date vs. a date string.  Also, you can use `post.guid` instead of `postId`.

src/storage.js
```diff
@@ -23,4 +26,41 @@ module.exports = {
...
+      .exec();
+    return postId;
```
humphd: I don't think we need to return anything, since we'll already know the key for this post, namely, the `post.guid`

src/storage.js
```diff
@@ -23,4 +26,41 @@ module.exports = {
...
+  },
+
+  getPosts: () => redis.zrangebyscore(POSTS, 0, +Infinity),
```
humphd: This is great.  Either in this PR, or in a follow-up, we can support paging here, by allowing the caller to pass arguments to specify which elements in the set to get.  That way they can request a portion vs. everything.

src/storage.js
haichuan0424: @humphd When we doing the test, we found weird error that the name property got passed into the post object. After we changed this name to match the property name inside the post object, the error disappeared.

humphd: If that's true, we should probably file a bug with ioredis or something.  For now, leave it as `author`, but please show me when I see you next time.

src/storage.js
haichuan0424: @humphd In order to sort the set, the score value must be numeric. The Date.parse function will transfer the date string to an integer millisecond value based on how far ahead to a base time.

humphd: Agree.  I'm saying, assume you have a `Date` and just call `getTime()` to get the `ms`:

```
> new Date()
2019-11-23T02:53:01.273Z
> new Date().getTime()
1574477586738
```

src/storage.js
haichuan0424: Sure, will show you next week.

src/storage.js
haichuan0424: OK, we can try this.

src/storage.js
KarlaChen: Here is the error if we keep using "name"
image

humphd: I pulled your branch, and figured out the issue.  It's not `name` vs. `author`.  The problem is that you are re-using keys in Redis.  As I mentioned in my review, we need to use the `post.guid` as the key for a Post Hash, and to put into the sorted set.  You are using two auto-incrementing values, and the feedId and postId are colliding, causing the data to be missing in your tests.

If you implement the changes I requested, this should be fixed.

humphd:
That failure is #343, not related to your changes.

src/storage.js
```diff
@@ -23,4 +25,39 @@ module.exports = {
...
+        'updated',
+        post.updated,
+        'published',
+        post.published,
+        'url',
+        post.url,
+        'site',
+        post.site
+      )
+      .zadd(POSTS, new Date(post.published).getTime(), post.guid) // sort set by published date as scores
```
humphd: I'm still worried about `post.published` here.  I think it should already be a `Date`, and you should just call `getTime()` on it.  cc @jatinAroraGit, who is working on the `Post` class.

src/storage.js
```diff
@@ -23,4 +25,39 @@ module.exports = {
...
+      .exec();
+  },
+
+  getPosts: (startDate, endDate) =>
```
mskuybeda: Not sure about startDate and endDate. You could just pass it as Date parameter.

manekenpix: I agree with @mskuybeda, I think we could pass a Date and get everything between that date and [the current date](https://javascript.info/date#creation).

Issue #167, opened 2019-11-13
Test case for Extracting of URLs from blog feed 


dbeigi:
Goal is to write some test cases for the issue #102

humphd:
@dbeigi I assume you're doing this?  If so, assign yourself.  If not, ignore this.

dbeigi:
Oh I seem to forget to assign myself

dbeigi:
@smilegodly did you want to work on this together?

smilegodly:
@dbeigi 
It's already in process of being written.
Thanks though for the offer.

agarcia-caicedo:
Looks like this was merger, closing.

PR #337, opened 2019-11-24
Added test cases for extract-urls

smilegodly:
Wrote a test case for extract-urls.
I plan to add more test cases for this in the future but I wanted to get this started now.

starts fixing : https://github.com/Seneca-CDOT/telescope/issues/167

### DIFF
--- a/src/backend/utils/extract-urls.js
+++ b/src/backend/utils/extract-urls.js
@@ -1,8 +1,9 @@
-// The goal is to create a function that given a html blog
-// it will parse the URLs from that feed and return an array of urls
-// Later on , you can then organize the types of urls (github etc)...
+/*
+  This function returns an array of urls from the html or
+  returns an empty array if there were no urls in the blog feed data
+*/
 const getUrls = require('get-urls');
-// this function returns an array of urls from the html
+
 exports.extract = function(htmlData) {
   const urls = getUrls(htmlData);
   const arrayOfUrls = Array.from(urls);
--- a/test/extract-urls.test.js
+++ b/test/extract-urls.test.js
@@ -1,0 +1,47 @@
+const getUrls = require('../src/backend/utils/extract-urls');
+
+test('The Array contains the URL from the blog feed data', () => {
+  const htmlData =
+    'Feel free to view all the changes <a href="https://github.com/FreezingMoon/AncientBeast/pull/1555">here</a>!</div>';
+  const urlArray = getUrls.extract(htmlData);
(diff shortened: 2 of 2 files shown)


src/backend/utils/extract-urls.js
```diff
@@ -1,10 +1,14 @@
...
+  this function returns an array of urls from the html or
+  returns null if there were no urls in the blog feed data
+*/
 const getUrls = require('get-urls');
-// this function returns an array of urls from the html
+
 exports.extract = function(htmlData) {
   const urls = getUrls(htmlData);
   const arrayOfUrls = Array.from(urls);
+  if (arrayOfUrls.length === 0) {
```
humphd: I would suggest returning an empty array is better than `null`, as it will mean that callers don't have to do an extra `if` check before trying to iterate.  If there are no elements in the array, it won't cause any problems, since the loops won't run.

test/extract-urls.test.js
```diff
@@ -0,0 +1,45 @@
...
+    'Feel free to view all the changes <a href="https://github.com/FreezingMoon/AncientBeast/pull/1555">here</a>!</div>';
+  const urlArray = getUrls.extract(htmlData);
+  expect(urlArray.length).toBe(1);
+});
+
+test('The url array remains empty after given no URL from the blog feed data', () => {
+  const htmlData =
+    'Feel free to view all the changes <a href="Lorem ipsum dolor sit amet consectetur adipiscing elit">here</a>!</div>';
+  const urlArray = getUrls.extract(htmlData);
+  expect(urlArray[0]).toBe(undefined);
```
humphd: `expect(urlArray.length).toBe(0)`

test/extract-urls.test.js
```diff
@@ -0,0 +1,45 @@
...
+  const res = Array.isArray(urlArray);
+  expect(res).toBe(true);
+});
+
+test('The length of the array matches the number of URLs in the string', () => {
+  const htmlData =
+    'Feel free to view all the changes <a href="https://github.com/FreezingMoon/AncientBeast/pull/1555">here</a>!</div>';
+  const urlArray = getUrls.extract(htmlData);
+  expect(urlArray.length).toBe(1);
+});
```
humphd: `expect(urlArray[0]).toBe('https://github.com/FreezingMoon/AncientBeast/pull/1555');`

humphd:
NOTE: windows failure on travis isn't related to this change, see discussion in #384

Issue #294, opened 2019-11-20
Keep feeds in Redis synchronized with wiki over time


humphd:
This is related to #266, #264.  We are going to be downloading the feed list (names, urls) from the wiki, and placing them in Redis.  We'll need to keep this data in sync:

1. we need to re-run this download code over and over at some interval, every 10 minutes?  It should be configurable in the `env`
1. when we re-download, we have to be careful to remove any feeds/posts from Redis that shouldn't be there.  For example, if we download the wiki feeds the first time, and it includes "David Humphrey" and "https://david.humphrey.com/feed", we'll add that to Redis.  But, on the second download, this feed could be removed (user deleted it).  In this case, we need to clean up Redis to remove the feed and posts (if any) for that user.

I suspect that this will involve altering how we store feeds in Redis so that we can easily look them up by name/url, which we can't do at the moment.

sukhbeersingh:
I'd like to take this up.

sukhbeersingh:
This is potentially a duplicate of #503.

humphd:
This one is different.  In #503 we added a way to loop forever, and re-download/parse feeds.

This issue is about keeping the feeds in Redis up-to-date with what's in the wiki.  The work in #503 won't do anything to remove a feed from Redis if it's removed from the wiki, for example.

humphd:
@Silvyre, I was thinking about this today.  Our wiki feed list is (currently) our "source of truth," and we want to use it as the official list of what we store in our system.  Given this, here are the cases I can think of that we need to deal with:

1. the wiki feed list has 100 feeds in it, redis has zero feeds in it (first run): we need to put all 100 feeds into redis
1. the wiki feed list has 100 feeds in it, redis has the same 100 feeds in it (second, third, ... normal run).  We don't need to add anything to redis
1. the wiki feed list has 101 feeds in it, redis has 100 feeds in it (new feed added to wiki): we need to add the new feed to redis
1. the wiki feed list has 99 feeds in it, redis has 100 feeds in it (feed removed from wiki): we need to remove this feed (and all posts associated with it) from redis
1. the wiki feed has 100 feeds in it, redis has 100 feeds in it, but they don't match (one feed was deleted from the wiki, and and a new one added): we need to delete the one that was deleted, and add the new one.

[...]

PR #341, opened 2019-11-24
Fix #266: inserted wiki parser into feed enque

### LINKED ISSUE #266: Connect wiki-feed-parser to feed queue
We have code to download and parse the Wiki Feed List page in [wiki-feed-parser.js](src/wiki-feed-parser.js).  We should connect this into our [logic for enqueuing feed jobs](https://github.com/Seneca-CDOT/telescope/blob/master/src/index.js#L50).  Most of the code in [src/index.js](src/index.js) related to reading [feeds.txt](feeds.txt) can get deleted now, along with that file.

When this is done, our feed queue should get filled with all the real feeds of our wiki list.

NOTE: #265 is adding tests for the wiki-feed-parser.

cryolis:
I also made a small change in wiki-feed-parser.js


cryolis:
#266

sukhbeersingh:
Hi @cryolis thanks for submitting this PR. My issue #294 depends on your PR so I can start working on it. Can you give a summary of what your PR is doing?

cryolis:
It basically takes the parsed data from wiki with a bunch of urls and sends them to the feed to be processed.

sukhbeersingh:
I fetched your branch and this is what I get.
issue#266

Are you able to run your changes locally?

cryolis:
Ye it runs fine for me. Have you installed redis though?

cryolis:
image
You should be getting something like this when you run my code

cryolis:
Ye I just checked, you get this error when you don't start redis. You either didn't start it or did not install it.

sukhbeersingh:
I tried running a local installation of redis and also ran it through docker(exposing port `6379`). Doesnt run for me. Although, when I switch to the master branch, I am able to run both ways.

cryolis:
Thats strange, when i shut down redis I got exact same error as the one you have. Don't really know why it doesn't run for you then.

cryolis:
Oh I noticed your running it on port 3000, try changing .env file for port 8080. maybe it'l do the trick

cryolis:
The only time Im getting that same error is when Im not running redis. I dont know why this is happening to you

humphd:
I have Redis running, and pulled this branch:

```
$ npm start
> telescope@0.0.1 start /Users/humphd/repos/telescope
> node src

[ 2019-11-26 18:51:47.711 ] INFO  (26940 on brightness.local): Telescope listening on port 3000
[ 2019-11-26 18:51:47.777 ] ERROR (26940 on brightness.local): UNHANDLED REJECTION:  Shutting down...
[ 2019-11-26 18:51:47.778 ] ERROR (26940 on brightness.local): TypeError: Only absolute URLs are supported
```

Which means we have a `Promise` that is `reject()`'ing and not being caught.  I modified the code in `src/backend/web/server.js` that is logging this, to give a full stack:

```
    TypeError: Only absolute URLs are supported
        at getNodeRequestOptions (/Users/humphd/repos/telescope/node_modules/node-fetch/lib/index.js:1299:9)
        at /Users/humphd/repos/telescope/node_modules/node-fetch/lib/index.js:1404:19
        at new Promise (<anonymous>)
        at fetch (/Users/humphd/repos/telescope/node_modules/node-fetch/lib/index.js:1401:9)
        at Object.module.exports.getData (/Users/humphd/repos/telescope/src/wiki-feed-parser.js:14:10)
        at Object.module.exports.parseData (/Users/humphd/repos/telescope/src/wiki-feed-parser.js:37:15)
        at enqueueWikiFeed (/Users/humphd/repos/telescope/src/index.js:17:12)
        at Object.<anonymous> (/Users/humphd/repos/telescope/src/index.js:33:1)
        at Module._compile (internal/modules/cjs/loader.js:959:30)
        at Object.Module._extensions..js (internal/modules/cjs/loader.js:995:10)
```

These are happening in `wiki-feed-parser.js`, and being re-`throw`n vs. dealt with, crashing the server.  The actual error comes from deep in `node-fetch`, see https://github.com/bitinn/node-fetch/blob/bf8b4e8db350ec76dbb9236620f774fcc21b8c12/src/request.js#L193.

The most likely cause of this is that `fetch(process.env.FEED_URL)` is throwing when `FEED_URL` is the empty string.  For @cryolis, it's likely set in the environment; for the rest of us, it likely isn't?

@cryolis we should do a few things here:

1. File a new issue to get full stacks in our logging code when we deal with unhandled rejections.  cc @izhuravlev, who is working on this in #173.
1. Add a check for `process.env.FEED_URL` not being a real URL.  You can use the code that @Cadesh wrote in `src/utils.js` to check if a string is a valid URL.
1. Add some debug/info logging to this code (i.e., `logger.debug()` for things like which URL we are using; `logger.info()` for standard details about what's happening), so we can better debug
1. Don't `catch` and re-`throw` errors.  Either deal with them (e.g., log an error and die, or recover and move on), or let the caller deal with them.

src/index.js
```diff
@@ -2,66 +2,38 @@
...
-    await feedQueue.add(feedJob, {
-      attempts: 8,
-      backoff: {
-        type: 'exponential',
-        delay: 60 * 1000,
-      },
-      removeOnComplete: true,
-      removeOnFail: true,
+async function enqueueWikiFeed() {
+  wikiFeed.parseData().then(data => {
```
humphd: Let's use `await` here, and it will look something like this:

```js
async function enqueueWikiFeed() {
  const data = await wikiFeed.parseData()
  Promise.all(data.map(async feedJob => {
    log.info(`Enqueuing feed job for ${feedJob.url}`);
    await feedQueue.add(feedJob, {
       attempts: 8,
        backoff: {
          type: 'exponential',
          delay: 60 * 1000,
        },
        removeOnComplete: true,
        removeOnFail: true,
    })
  }))
  .catch(err => log.error({ err }, 'Error queuing wiki feeds');
}
```

NOTE: you can't do a `forEach` like this with an `async` function, you have to use `Promise.all` and create `Promises` for each

src/index.js
```diff
@@ -2,66 +2,38 @@
...
-    .catch(error => {
-      console.log(error);
-    });
-});
+enqueueWikiFeed()
+  .then(() => {
+    feedWorker.start();
+  })
+  .catch(error => {
+    console.log(error);
```
humphd: `log.error({ error }, 'Unable to enqueue wiki feeds');`

humphd:
This needs a rebase, too, since it's sat so long.

cryolis:
I just noticed that my "dump.rdb" was included into this PR. It should probable be added into ".gitignore". I can do that with next commit.

src/backend/index.js
```diff
@@ -74,59 +74,32 @@ process.on('uncaughtException', err => {
...
-        type: 'exponential',
-        delay: process.env.FEED_QUEUE_DELAY_MS || 60 * 1000,
-      },
-      removeOnComplete: true,
-      removeOnFail: true,
-    });
-  });
+async function enqueueWikiFeed() {
+  const data = await wikiFeed.parseData();
+  Promise.all(
```
humphd: I think this should be `await Promise.all(...)`.

src/backend/index.js
```diff
@@ -74,59 +74,32 @@ process.on('uncaughtException', err => {
...
+    data.map(async feedJob => {
+      log.info(`Enqueuing feed job for ${feedJob.url}`);
+      await feedQueue.add(feedJob, {
+        attempts: 8,
```
humphd: You are erasing changes that happened in a recent fix to make these environment variables.  Do you need to rebase?

src/backend/index.js
```diff
@@ -74,59 +74,32 @@ process.on('uncaughtException', err => {
...
+        backoff: {
+          type: 'exponential',
+          delay: 60 * 1000,
```
humphd: Same thing here, check what's on `master` right now.

.gitignore
```diff
@@ -30,3 +30,6 @@ yarn-error.log*
 
 # web-hint report
 hint-report/
+
+#Redis dump file
+dump.rdb
```
humphd: I'm confused as to why you're adding this.  Why do we want this included?  You're not using it anywhere.  Also, why would you add this to the `.gitignore` and then add it in your commit?

cryolis: I added it into my previous commit, when I noticed it I realized that we dont need that at all, since it is generated by Redis, so I added that to .gitignore.

humphd: That's fine (i.e., adding to `.gitignore`).  But you also added it to this PR (the file itself):

Screen Shot 2019-11-29 at 9 15 28 PM

src/backend/utils/wiki-feed-parser.js
```diff
@@ -50,7 +50,7 @@ module.exports.parseData = function() {
             feed.push(`${line}`);
             let obj = {
               name: feed[feed.length - 1],
-              link: feed[feed.length - 2],
+              url: feed[feed.length - 2],
```
humphd: Will this change break anything else we have?  It's probably fine, just want to confirm that we don't have any callers who depend on this yet.

cryolis: It shouldn't really, although I noticed that Travis is failing on something, but it looks like the problem is with environment variables, not sure why though.

humphd: Travis failure is #387, unrelated to your changes.  OK, we should be good.  Get another review, and let's merge this!

Issue #4, opened 2019-10-28
Discussion: Kubernetes


Silvyre:
"Running on Seneca's [Kubernetes](https://kubernetes.io/) container cloud" is listed by [our overview document](https://github.com/Seneca-CDOT/telescope/blob/master/docs/overview.md) as one of the features required of Telescope to constitute a minimum viable product (MVP).

This issue exists to:
- aggregate research and useful information for those interested in contributing to this MVP feature. 
- discuss the directions we may take in implementing this feature.

_Importantly, research contributed to this thread will be referenced when we write documentation, which will be used by future contributors to our project. Therefore, by contributing to our research, you are also helping lay the groundwork for our project's documentation!_

Questions that have yet to be answered include:
1. As far as our project is concerned: what is Kubernetes, and what role will it play here?
1. How may we gain access to Seneca's Kubernetes container cloud? Who do we need to talk to?
1. Will the use of Kubernetes potentially affect [other MVP features](https://github.com/Seneca-CDOT/telescope/blob/master/docs/overview.md#mvp-features)?
1. Are there any security concerns that may come up when dealing with this feature?
1. Are there any standards or best practices that we should adhere to?
1. Should we create a 'Kubernetes' channel on the seneca-open-source Slack (e.g. to help facilitate open conversations), or should we keep all discussion on Kubernetes centralized within this GitHub issue?

**Currently, we have these tasks to do:**
  - [ ] Update documentation -------> (we have self documented files, but should make issue to have them on Documents)
  - [x] Add docker files to project
  - [ ] Adding Helm charts -------> (changd to raw YAML files)
  - [x] Deploy on Kubernetes
  - [x] Adding Health checks

c3ho:
@Silvyre Regarding 6. Perhaps we could use Kubernetes related tags for issues?

Reza-Rajabi:
I am also interested to work on this issue.

Reza-Rajabi:
My goal for this week is to make docker files and docker image.
I appreciate if anybody let me know if they have any comment on that.

manekenpix:
@Reza-Rajabi I've worked with docker before, we can work together on that.

Reza-Rajabi:
@manekenpix my pleasure. I am thinking of using the CloudNativeJs template for the Node.js, Express backend. What do you think about it?

Silvyre:
Sounds great! I'm interested in contributing a Makefile.

manekenpix:
[...]

PR #357, opened 2019-11-27
Milestone 2 of #4: Adds Docker file, dev version, with SSO tester image

Reza-Rajabi:
- Milestone 2  of issue #4

- What changes have been made in the included files?
     - The Dockerfile upgraded to a fully functional development version. Now we can use all commands from the `scripts` list under the `package.jason` file as well as: test, jest, debug-server (dang! just noticed the debug-server is not there!), etc. The `nodemon` is available and the container can be self restarted by development changes. Example of using the updated file:  in the command line, 
``` $ script="test" docker-compose up --build ``` will build an test telescope
     - The SSO SAML2 tester module has been included
     - Slightly improvement on `.dockerignore` and `docker-compose` files applied

- The problem that has been found so far:
     - The queues don't get created or handled completely which is traceable on  Bull Dashboard

- What you may expect next?
     - The deploy version of the Docker file to be included in milestone 3 (weekend).
     - To be ready to deploy and close issue #4 by providing .yaml files for Kubernetes set up in milestone 4 (next Wednesday).

- How can you help if you would like to?
     - Any feedback, critique, complain, comment, review and thoughts are definitely helpful.
     - Forgive me for the delay and locking an important task for a long time.
     - The problem mentioned above is persisting in different ways. Any clue to solve it would be appreciated.

### CODE REVIEW

**Comment by manekenpix** [comment]
@Reza-Rajabi Looking pretty good!
I'm testing your changes right now and I'm not sure about what the problem is.
I launched telescope with `script="start" docker-compose up --build` and this is what I see.

Screenshot from 2019-11-27 09-47-17

Port 3000 is exposed (that's why I can access localhost:3000), and redis didn't crash (as you can see in the right side of the image).
Could you provide more details so I can replicate the problem?

**Comment by manekenpix** [comment]
I did some more digging and to address the problem you mention [here](https://github.com/Seneca-CDOT/telescope/issues/4#issuecomment-558002660), I tried this (also removed line 20 `network_mode: host` ):

`docker-compose.yml`
```yml
version: "3"

...

telescope:
   environment:
      - REDIS_URL=redis://redis_telescope

...

redis:
  container_name: redis_telescope
```

This way, all the containers are properly exposed:
Screenshot from 2019-11-27 10-19-49

Also, I think we should change the wild card access for all the containers (`0.0.0.0`):
Screenshot from 2019-11-27 10-28-00

Maybe only allowing access to `telescope` and/or `ssoipd`?

Please test it and let me know what you think

Reza-Rajabi:
@manekenpix awsome 🙌. I can't believe it. The current file is not working for me. I read many articles about it and it seems like it's a problem mostly with mac. I will test your solution and I will tell you the result 🙏.

Reza-Rajabi:
I have tested it on mac, @manekenpix tested on Linux (I guess), and that would be great if we could test it on windows and another mac.

manekenpix:
@Reza-Rajabi I'll test it later on Windows (I have a VM at home running W10).

Grommers00:
Sounds good, let me know the results. I haven't used Docker much so from what I see the code looks good with the changes!

Reza-Rajabi:
> I did some more digging and to address the problem you mention [here](https://github.com/Seneca-CDOT/telescope/issues/4#issuecomment-558002660), I tried this (also removed line 20 `network_mode: host` ):
> 
> `docker-compose.yml`
> 
> ```yaml
> version: "3"
> 
> ...
> 
> telescope:
>    environment:
>       - REDIS_URL=redis://redis_telescope
> 
> ...
> 
> redis:
>   container_name: redis_telescope
> ```
> 
> This way, all the containers are properly exposed:
> Screenshot from 2019-11-27 10-19-49
> 
> Also, I think we should change the wild card access for all the containers (`0.0.0.0`):
> Screenshot from 2019-11-27 10-28-00
> 
> Maybe only allowing access to `telescope` and/or `ssoipd`?
> 
> Please test it and let me know what you think

@manekenpix This works for mac. Awesome! 👏

.dockerignore
```diff
@@ -1,13 +1,20 @@
-**/node_modules/
-**/node_modules_linux/
+**/.circleci/
+**/.vscode/
+**/.idea/
```
humphd: Why include this?  We don't have this.

.dockerignore
```diff
@@ -1,13 +1,20 @@
...
+**/coverage/
 **/docs/
-**/Dockerfile
+**/node_modules/
+**/node_modules_linux/g
```
humphd: Same question here

.dockerignore
```diff
@@ -1,13 +1,20 @@
...
 **/.editorconfig
-**/.eslintrc.js
-**/.gitignore
 **/.git
-**/README.md
+**/.gitattributes
+**/.gitignore
+**/.travis.yml
+**/CODE-OF-CONDUCT.md
+**/Dockerfile
```
humphd: This seems like something you'd want to include, not ignore?

.dockerignore
```diff
@@ -1,13 +1,20 @@
...
 **/LICENSE
```
humphd: I would leave the LICENSE in.

.dockerignore
```diff
@@ -1,13 +1,20 @@
...
-**/.vscode
-**/.idea
+**/README.md
+**/docker-compose.yml
```
humphd: Same thing here, include this, right?

Dockerfile
```diff
@@ -1,27 +1,84 @@
-FROM node:10
```
humphd: Why switch away from a standard `node` image to `ubuntu`?  We don't have binary dependencies, do we?  I'd prefer to have our node app run inside a `NODE:lts` image, since that's what we use in CI.

If I'm missing something and you have to use Ubuntu, let me know.

Dockerfile
```diff
@@ -1,27 +1,84 @@
...
+# Install Linux `apt-utils` package to enhance packages configuration
+# Install Linux `bc` package to support interactive execution of statements
+# Install Linux `inotify-tools` to support nodemon file change tracking
+# Install Linux `curl` package to support transfer data from URL
+RUN apt-get update -y
+RUN apt-get install -y apt-utils
+RUN apt-get install -y bc
+RUN apt-get install -y inotify-tools
+RUN apt-get install -y curl
+RUN echo "Finished installing OS dependencies"
```
humphd: Let's get rid of `nodemon`, we don't need it in production, and if it's requiring us to pull in all sorts of OS deps, it's not what we want here.

Dockerfile
```diff
@@ -1,27 +1,84 @@
...
+
+# Install Node.js from https://github.com/nodesource/distributions
+RUN curl -sL https://deb.nodesource.com/setup_12.x | bash -
+RUN apt-get install -y nodejs
+RUN echo "Installed Node.js Version 12"
```
humphd: Let's prefer the pre-build `NODE:lts` to building.

docker-compose.yml
```diff
@@ -1,13 +1,48 @@
...
+    env_file: .env
+    environment:
+      # a place-holder for the variable passed by CLI and with a default value
+      # the passed or default value is the desired command to run `telescope`
+      script: ${script:-debug-server}
     depends_on:
       - redis
-    ports:
-      - '3000:3000'
+    # PORT MAPPING IN `host` MODE IS WRONG AND MUST BE CHANGED.
```
humphd: If there's a follow-up issue to deal with here, let's file that instead of this.

env.example
```diff
@@ -54,3 +55,7 @@ BLOG_INACTIVE_TIME=360
 
 # GITHUB_TOKEN is used when fetching data from the GitHub API
 GITHUB_TOKEN=
+
+# REDIS_VERSION is the version of the Redis in the docker-compose file
+# Redis most stable version with an image in the Docker Hub; https://redis.io, https://hub.docker.com/_/redis
+REDIS_VERSION=5.0.7
```
humphd: I don't think we need this in the env, it should probably go in the Dockerfile.

.dockerignore
Reza-Rajabi: My IDE is Webstorm (Jetbrain Co.) and it produces this file like .vscode for Visual Studio, I thought maybe some other people are using this. I have the same problem with Git. Everybody always includes .vscode only :)
I will remove this if it's not popular.

humphd: That's fine.  Should we be adding this to our `.gitignore` too then?

.dockerignore
Reza-Rajabi: The `g` leter at the end is a typo. About including the `node_modules_linux`, for the first version of the Docker files I have used [CloudNativeJs](https://cloudnativejs.io) templates as I mentioned in my comments before the first pull request. [here](https://github.com/CloudNativeJS/docker/blob/master/.dockerignore) is the template. I didn't use Linux with Nod.js at all, so my assumption was this is similar to node_modules for Linux system.

.dockerignore
Reza-Rajabi: Since the Docker uses the Dockerfile to build the image, the image doesn't need to carry the plan with itself. So ignoring it makes the image smaller.

.dockerignore
Reza-Rajabi: It makes a lot more sense. Sure.

.dockerignore
humphd: Let's remove it then.  Things we don't need/understand shouldn't go in here, as they are something we can add when we do need them.

.dockerignore
humphd: OK, sounds good.

.dockerignore
Reza-Rajabi: Same answer. The image doesn't need the blueprint anymore when it gets built.

humphd: Based on your comment above, this is probably fine too.

Dockerfile
Reza-Rajabi: When I was updating the Docker file, I checked the Node.js and Docker Hub and I realized that version 12 is a better choice and the only reason that I used version 10 in the first version, was because of the templates that had not been updated. Then I changed it to `From node:12`. When I started building the container, it started installing Linux Debian and I surprised because for the first one it had installed Ubuntu. So, to have a same stack always, regardless of updating the version of Node image, I decided to build the whole container from scratch.

Dockerfile
Reza-Rajabi: 😢 it took me days to figure out how to have container aware of the changes for `nodemon` , and now it's like my child 😃. 
Sure, I will remove everything related to it.

Dockerfile
humphd: If we get rid of `nodemon`, which I suggest below, we shouldn't need to install anything beyond node itself.  It's better if we can use a pre-built image vs. building something on our own, since we don't have to maintain anything that way: they will be the default node images.

Being built on Debian or Ubuntu is really the same thing, Ubuntu is a fork of Debian.  Probably once we get this running, we can switch to use alpine or something even smaller.

Dockerfile
humphd: I'm breaking everyone's heart this week, sorry :(  But on the bright side, think of all the amazing learning you've done! 

I think dumping this is the better solution, since it's not really needed, it just got added when the web server went in.

Dockerfile
Reza-Rajabi: After my other comment, if you still think using the `node` image is better, I will change it back.

humphd: If possible, `NODE:lts` would be the best, yes.

docker-compose.yml
Reza-Rajabi: It gets solved actually by @manekenpix

humphd: OK, I wondered, but since you'd left this in, I wasn't sure.  We can remove this then.

env.example
Reza-Rajabi: I thought we might let the developer change the Redis without touching Dockerfiles, but you are right, it doesn't make sense to have it here. I will move this to its own file.

humphd: If someone wants a custom Redis version, they can install it separate from Docker.  If they use Docker, we'll give them whatever version we specify in our image.

.dockerignore
Reza-Rajabi: That would make my life easier. Currently, I can't do `git add .` 😀

humphd: Always feel free to add things to `.gitignore` that would help you with your particular environment.  We have lots of stuff in there that's only specific to some people in the project, but not all.  We want everyone to have their life be easier :)

Dockerfile
Reza-Rajabi: Sure, I will do.

docker-compose.yml
Reza-Rajabi: @manekenpix solved it today actually, while I posted this last night as a draft, mostly because @Grommers00 asked me so.

Dockerfile
Reza-Rajabi: That's true. It was obviously a lot of learning for me rather than me being helpful for the project.

env.example
Reza-Rajabi: @humphd should we have it `redis:latest` same as `node:latest` or I need to keep the version `5.0.7`
There are some comments around `latest` saying that it's not the best choice, but `latest` makes it easier to maintain. So what would you suggest for both?

humphd: For node, we want `lts` (long term stable vs. latest).  It's the "stable" version vs. the latest.

For Redis, let's use `latest`, which is really the [current stable version](https://github.com/docker-library/redis/blob/d42494ab2d96070c8d83f37a7542fbbffd999988/5.0/Dockerfile#L42).  Then we can get rid of this, and let Docker/Redis manage this for us.

Reza-Rajabi: I didn't know that acronym. Thanks for the clarification.

Reza-Rajabi:
@humphd, sorry for keeping you busy with this issue. I am applying requested changes and I am not sure about some parts:
- You mentioned we better not have `nodemon` for production. These files are intended for use in development and I will be having another one for production. Do you want me to remove `nodemon` from these (development version)?
- In Docker Hub there is not any node image with version tag `lts`, do you want me to use the last stable version which I think it's `node:12` or the latest version `node:latest` is better?

Thank you Dave.

humphd:
1. Yes, remove `nodemon` completely from the project
1. The site has an `lts` tag (second last in image below), but maybe we can't use that?  I haven't tired, so maybe I'm wrong

Screen Shot 2019-11-27 at 5 58 23 PM

Reza-Rajabi:
I will try it. I didn't notice that before. Thank you.
(my last efforts for keeping nodemon failed 😄)

Reza-Rajabi:
> > I did some more digging and to address the problem you mention [here](https://github.com/Seneca-CDOT/telescope/issues/4#issuecomment-558002660), I tried this (also removed line 20 `network_mode: host` ):
> > `docker-compose.yml`
> > ```yaml
> > version: "3"
> > 
> > ...
> > 
> > telescope:
> >    environment:
> >       - REDIS_URL=redis://redis_telescope
> > 
> > ...
> > 
> > redis:
> >   container_name: redis_telescope
> > ```
> > 
> > 
> > This way, all the containers are properly exposed:
> > Screenshot from 2019-11-27 10-19-49
> > Also, I think we should change the wild card access for all the containers (`0.0.0.0`):
> > Screenshot from 2019-11-27 10-28-00
> > Maybe only allowing access to `telescope` and/or `ssoipd`?
> > Please test it and let me know what you think
> 
> @manekenpix This works for mac. Awesome! 👏

@manekenpix May I change the name of container from `redis_telescope` (although I like it better) to `redis_db` because of better output:

Screen Shot 2019-11-27 at 6 58 18 PM

Screen Shot 2019-11-27 at 7 03 35 PM


 or any other smaller name if you prefer?

manekenpix:
@Reza-Rajabi Absolutely, I just used that name for debugging purposes. I'm ok with any name you choose.

sukhbeersingh:
I tried this on Ubuntu18.04 and works well. Switched to windows and Docker complains about not being able to share drive. I have been looking at some online help. This is from [docker website](https://docs.docker.com/docker-for-windows/#firewall-rules-for-shared-drives), [another one, ](https://stackoverflow.com/questions/42203488/settings-to-windows-firewall-to-allow-docker-for-windows-to-share-drive/43904051#43904051)and yet [another one](https://blog.olandese.nl/2017/05/03/solve-docker-for-windows-error-a-firewall-is-blocking-file-sharing-between-windows-and-the-containers/#conclusion). This is[ what windows has to say](https://support.microsoft.com/en-us/help/4340181/error-0x80071779-when-removing-client-for-microsoft-networks) for what one of the solutions was suggesting.
I am not too comfortable at networking stuff. If anyone knows more please let me know.

humphd:
Naive question: can we simply avoid sharing volumes with the containers and host for this?  We don't really care about keeping them in sync, do we?

The most important use case we have for Dockerizing this is production, not "get it working on random Windows laptops."  So if it works well on Linux, that's fine with me.  We can iterate on getting it running on Windows later.

sukhbeersingh:
Yeah we can sure leave it for later...
But I thought the whole point was to give all developers a better process for helping out in development. Your point is fair too, if this is okay then we can just use it for prod.

humphd:
@sukhbeersingh don't get me wrong: I want it to work on all platforms.  I just don't want @Reza-Rajabi to get discouraged from pushing this into the tree!

I want it all, I'm just patient about when I get it :)

Reza-Rajabi:
@sukhbeersingh 
Thank you very much for helping with that. Would you please apply the changes @manekenpix mentioned and see how it goes:

in docker-compose file delete line 20 (network-mode), then edit the file similar to the image below in the lines 16 (change format),18 (add) and 27(add)

```
// line 13-18
environment:
      # a place-holder for the variable passed by CLI with a default value
      # the passed or default value is a command to run `telescope`
      - script=${script:-start}
      # override the .env variable for `REDIS_URL`
      - REDIS_URL=redis://redis_db

// line 27
container_name: redis_db
```
Screen Shot 2019-11-27 at 10 25 00 PM

Reza-Rajabi:
If you test this code, would you please check the bull board as well? I am having this now:

without Dockerfile
Screen Shot 2019-11-27 at 10 15 43 PM

vs. Dockerfile:

Screen Shot 2019-11-27 at 10 17 26 PM

@manekenpix

Reza-Rajabi:
@sukhbeersingh I think I didn't get your comment. Are you trying to change Ubuntu to Windows on the code or you are trying to test the code in different machines? If you are trying the first one, we don't need to worry about it at all, but if you are trying the code in different machines, it should work and you may try my latter comment about editing the code.

Reza-Rajabi:
@humphd 
> Naive question: can we simply avoid sharing volumes with the containers and host for this? We don't really care about keeping them in sync, do we?

Do you mean line 23 in docker-compose file? if so, I had that part only for making nodemon from the inside of the container aware of the local changes. I deleted that part because without nodemon we don't need it anymore.

env.example
```diff
@@ -16,6 +16,7 @@ FEED_URL=https://wiki.cdot.senecacollege.ca/wiki/Planet_CDOT_Feed_List
 
 # REDIS_URL specifies Redis server info
```
humphd: Let's add another comment here discussing what to do if you're connecting via docker-compose, and need to use `redis://redis:6379`

Dockerfile
```diff
@@ -1,27 +1,51 @@
...
+
+# Use Redis image via compose-file
+
+# Set Shell to exit immediately if a pipeline returns a non-zero status: -e
+# Assume YES for prompts: -y
+# Update OS
+# Upgrade all packages and resolve conflicts
+# Clear out the local repository of retrieved package files
+RUN set -e
+RUN apt-get update -y
```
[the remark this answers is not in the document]
Reza-Rajabi: You are right. I followed the common pattern in other Docker files I passed through, but it doesn't make sense. I will remove them.

env.example
Reza-Rajabi: In this commit, the docker-compose actually overrides this variable, so we don't need to change it.

Dockerfile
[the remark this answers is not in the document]
Reza-Rajabi: True. I will remove them.

PR #358, opened 2019-11-27
move backend code from src to backend

miggs125:
I organized the backend file structure and moved files from src/ into the backend directory. I also created a utils directory under src where a lot of the files previously in src/ ended up. I also updated the paths in all the files that were affected by the restructure as well as the scripts in package-json.


humphd:
@miggs125 do any of the files in `/docs` need to be updated due to this reorganization?

miggs125:
@humphd nice catch, I missed that. Fixed.

Issue #362, opened 2019-11-28
Running "npm start" causes error, start script doesn't seem to be pointing to the right path


lucacataldo:
Not sure if this is an issue with my current system / setup, but it looks like the npm start script is pointing to /src which is not where the index.js entry point is located. The start script in package.json currently reads:
`node src`

It looks like it should be 
`node src/backend`

Again, I could be wrong but the above fix seems to work for me.

Steps to reproduce:

1.  Clone repo
2.  Run npm install
3.  Run npm start

Output:
![image](https://user-images.githubusercontent.com/35984346/69778088-03be4800-1171-11ea-96ee-389a2b87c332.png)

humphd:
Thanks for filing, @lucacataldo.  It looks like this got lost in a rebase during https://github.com/Seneca-CDOT/telescope/commit/3cd0686db1b7e99c9250939d359b1e9aa2c44db7.  cc @lozinska.

We really need to add a test to make sure that `npm start` works.

humphd:
I filed a new issue on adding a test for this, #363

cagomezr:
working on it

humphd:
Fixed by #364.

PR #364, opened 2019-11-28
Issue#362

cagomezr:
Issue  #362  broke project by removing/ changing the lines on package.json 

 issue fixed by  modifying  json to  original line  calling to src to its proper call,  server and hints.

the new setup is like this: 
26   "jest-watch":"cross-env MOCK_REDIS=1 jest --watch",
27   "start": "node src/backend",
28  "server": "node src/backend/web/server",
29  "hint": "hint http://localhost:3000",
30  "webhint": "cross-env PORT=3000 npm-run-all -r -p server hint"
31 "debug-server": "nodemon server.js"

instead of

26 "jest-watch":"cross-env MOCK_REDIS=1 jest --watch",
27 "start": "node src",
28 "debug-server": "nodemon server.js"

My only concern is that  debug server didn't move during npm  install

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,10 @@
     "test": "npm run jest",
     "jest": "cross-env MOCK_REDIS=1 jest",
     "jest-watch":"cross-env MOCK_REDIS=1 jest --watch",
-    "start": "node src",
+    "start": "node src/backend",
+    "server": "node src/backend/web/server",
+    "hint": "hint http://localhost:3000",
+    "webhint": "cross-env PORT=3000 npm-run-all -r -p server hint"
     "debug-server": "nodemon server.js"
   },
   "husky": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -23,8 +23,11 @@
     "pretest": "npm run lint",
     "test": "npm run jest",
     "jest": "cross-env MOCK_REDIS=1 jest",
-    "jest-watch":"cross-env MOCK_REDIS=1 jest --watch",
-    "start": "node src",
+    "jest-watch": "cross-env MOCK_REDIS=1 jest --watch",
+    "start": "node src/backend",
+    "server": "node src/backend/web/server",
```
sukhbeersingh: Do we need server because of webhint? What exactly would adding webhint do in this PR?

cagomezr: based on the running I just did  without it I would argue it  helps set the setup, as it seems to go a bit off the rails without, also since we are  reverting  items to previous version I rather be safe

PR #369, opened 2019-11-28
Fix issue 260 implement addPost and getPost

KarlaChen:
Added addPost() function using guid as unique key for each post and store posts as sorted set by published date.
Added getPost() by guid and getPosts() by range of dates.
Added test for testing above mentioned functions.


humphd:
This caused #374.

PR #306, opened 2019-11-21
Fix #305: Fix test failure with logger.debug()

manekenpix:
The problem was the way `src/lib/queue.js` and `src/lib/redis.js` were importing `logger`.
Adding destructuring fixed it. 

Please test it before approving.

PR #370, opened 2019-11-28
Fix #309: Added tests for logger methods

### LINKED ISSUE #309: Create a test for logger
There was a recent PR #306, we don't have a test for it.

jpjjulie:
I have added tests to make sure that logger functions exist. This pull request is a result of failure of logger.debug in Fix #306

### DIFF
--- a/test/logger.test.js
+++ b/test/logger.test.js
@@ -8,4 +8,7 @@ test('logger.methods to be functions', () => {
   expect(typeof logger.info).toBe('function');
   expect(typeof logger.child).toBe('function');
   expect(typeof logger.error).toBe('function');
+  expect(typeof logger.trace).toBe('function');
+  expect(typeof logger.warn).toBe('function');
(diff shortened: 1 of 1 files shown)


test/logger.test.js
```diff
@@ -0,0 +1,11 @@
+const { logger } = require('../src/backend/utils/logger');
+
+// This test helps determine whether the type of
+// logger methods are functions
+
+test('logger.methods to be functions', () => {
+  expect(typeof logger.debug).toBe('function');
+  expect(typeof logger.info).toBe('function');
+  expect(typeof logger.child).toBe('function');
+  expect(typeof logger.error).toBe('function');
```
humphd: Perfect.  Let's also add cases for these:

- logger.trace()
- logger.warn()
- logger.fatal()

jpjjulie: Sure. Will update.

PR #372, opened 2019-11-28
Fix #289: deal with log levels that crash the server

### LINKED ISSUE #289: Server crashes if LOG_LEVEL is upper case
```
$ LOG_LEVEL=INFO npm start
/Users/humphd/repos/telescope/node_modules/pino/lib/levels.js:153
    throw Error(`default level:${defaultLevel} must be included in custom levels`)
    ^

Error: default level:INFO must be included in custom levels
    at assertDefaultLevelFound (/Users/humphd/repos/telescope/node_modules/pino/lib/levels.js:153:11)
    at pino (/Users/humphd/repos/telescope/node_modules/pino/pino.js:96:3)
    at Object.<anonymous> (/Users/humphd/repos/telescope/src/lib/logger.js:24:12)
    at Module._compile (module.js:653:30)
    at Object.Module._extensions..js (module.js:664:10)
    at Module.load (module.js:566:32)
    at tryModuleLoad (module.js:506:12)
    at Function.Module._load (module.js:498:3)
    at Module.require (module.js:597:17)
    at require (internal/module.js:11:18)
```

Sbaah:
Fixes #231  Server crashes if LOG_LEVEL is upper case (FIXED)

- Created a new local variable to to hold the value form ```process.env.LOG_LEVEL.toLowerCase()``` . <br>

- Have an statement to check if the variable coming in is either in lowercase or one of the documented levels on  ``` https://getpino.io/#/docs/api?id=level-string``` .

- If it's not then the new local variable will be set to  level parameter ``` info ```.

### DIFF
--- a/src/backend/utils/logger.js
+++ b/src/backend/utils/logger.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-lonely-if */
 const pino = require('pino');
 const expressPino = require('express-pino-logger');
 const os = require('os');
@@ -5,13 +6,28 @@ const os = require('os');
 require('../lib/config');
 
 let logger;
+let logLevel;
 
 /* If in Development mode:
  * Output logs to console (which is by default)
  * Enable prettyPrint option and translate time from epoch time to local time.
  * Set log level to LOG_LEVEL environment variable with 'debug' as default level.
  */
-if (process.env.NODE_ENV === 'development') {
+
+/* Created a local variable to to hold the value form process.env.LOG_LEVEL.toLowerCase()
+ * The if statement will check the variable if the variable is anything other than levels
+ * stated within file it will automatically set to info.
+ * https://getpino.io/#/docs/api?id=level-string Look for the title ``` level (String) ```
+ */
+
+logLevel = (process.env.LOG_LEVEL || 'info').toLowerCase();
+if (!pino.levels.values[logLevel]) {
+  logLevel = 'info';
+}
+
+if (logLevel !== 'info') {
+  logLevel = 'info';
+} else if (process.env.NODE_ENV === 'development') {
   logger = pino({
     prettyPrint: {
       translateTime: 'SYS: yyyy-mm-dd HH:MM:ss.l ',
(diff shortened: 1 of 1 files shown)


src/backend/utils/logger.js
```diff
@@ -5,13 +5,24 @@ const os = require('os');
 require('../lib/config');
 
 let logger;
+let lLog;
```
humphd: Let's change this to `let logLevel` instead of `let lLog` so it's more clear.

src/backend/utils/logger.js
```diff
@@ -5,13 +5,24 @@ const os = require('os');
...
  * Enable prettyPrint option and translate time from epoch time to local time.
  * Set log level to LOG_LEVEL environment variable with 'debug' as default level.
  */
-if (process.env.NODE_ENV === 'development') {
+
+/* Created a local variable to to hold the value form process.env.LOG_LEVEL.toLowerCase()
+ * The if statement will check the verible if the verible is anything other than levels
+ * stated within file it will automaticaly set to info.
+ * https://getpino.io/#/docs/api?id=level-string Look for the title ``` level (String) ```
+ */
```
humphd: We need to be careful here.  You've got a good start at dealing with issues but there are more cases we have to worry about.  For example, we care about `process.env.LOG_LEVEL` being undefined, non-lowercase (both of which you have), and also whether it is one of the default log levels (trace, info, debug, etc).

To check the latter, you can so something like this:

```js
logLevel = (process.env.LOG_LEVEL || 'info').toLowerCase();
// Make sure logLevel is one of the default pino log levels (debug, trace, error, info, etc)
if(!pino.levels.values[logLevel]) {
  logLevel = 'info';
}
```

After this, `logLevel` should be safe to use, and should be `info` or one of the other levels.

src/backend/utils/logger.js
```diff
@@ -5,13 +5,24 @@ const os = require('os');
...
+lLog = process.env.LOG_LEVEL;
+lLog = lLog.toLowerCase();
+if (lLog !== 'info') {
+  lLog = 'info';
+} else if (process.env.NODE_ENV === 'development') {
```
humphd: This shouldn't be an `else if`, it needs to be separate from the previous `if`.

src/backend/utils/storage.js
```diff
@@ -4,6 +4,8 @@ const { redis } = require('../lib/redis');
 const FEED_ID = 'feed_id';
```
humphd: All these other changes shouldn't be in this PR.  It looks like you've merged with master, and pulled in unrelated commits.

Sbaah: How do i resolve this?

humphd: We need to rebase your changes on top of `master`, but it looks like you might have merged (maybe via a `pull`)?  I can help you on Monday if you like, after class?

humphd:
Fixed and merged into `master`!  Thanks @Sbaah.

PR #375, opened 2019-11-28
Fixes #315: Added password to the Bull queue constructor

### LINKED ISSUE #315: Specify Redis username and password in .env and pass it to the Bull queue constructor
According to suggestion introduced in #112, we should include Redis username and password into .env file to be passed to the Bull queue constructor.

birtony:
Fixes #315 
- Added REDIS_PASSWORD to the env.example
- Added password to the Bull queue constructor

In order to pass password to the Redis constructor, I had to change the parameter list to the following, according to the [documentation](https://github.com/luin/ioredis/blob/master/API.md#new-redisport-host-options):
```js
new Redis(port, host, redisPassword);
```

### DIFF
--- a/env.example
+++ b/env.example
@@ -15,9 +15,10 @@ LOG_FILE=
 # FEED_URL url used to access feed list
 FEED_URL=https://wiki.cdot.senecacollege.ca/wiki/Planet_CDOT_Feed_List
 
-# REDIS_URL specifies Redis server info
+# Redis Server info, password may be optional (e.g., leave empty if you don't set one)
 REDIS_URL=redis://127.0.0.1:6379
-REDIS_PORT=6379
+REDIS_PORT=6739
+REDIS_PASSWORD=
 
 # MOCK_REDIS=1 will use an in-memory, mock Redis instance. Useful for testing.
 MOCK_REDIS=
--- a/src/backend/lib/queue.js
+++ b/src/backend/lib/queue.js
@@ -1,13 +1,14 @@
 const Bull = require('bull');
-const { Redis, redisUrl } = require('./redis');
+const { createRedisClient } = require('./redis');
 const { logger } = require('../utils/logger');
 
 /**
  * Shared redis connections for pub/sub, see:
  * https://github.com/OptimalBits/bull/blob/28a2b9aa444d028fc5192c9bbdc9bb5811e77b08/PATTERNS.md#reusing-redis-connections
  */
-const client = new Redis(redisUrl);
-const subscriber = new Redis(redisUrl);
+
+const client = createRedisClient();
+const subscriber = createRedisClient();
 
 /**
  * Create a Queue with the given `name` (String).
@@ -25,7 +26,7 @@ function createQueue(name) {
         case 'subscriber':
           return subscriber;
         default:
-          return new Redis(redisUrl);
+          return createRedisClient();
       }
     },
   })
--- a/src/backend/lib/redis.js
+++ b/src/backend/lib/redis.js
@@ -1,7 +1,7 @@
 require('./config');
 const Redis = require('ioredis');
 const MockRedis = require('ioredis-mock');
-
+const url = require('url');
 const { logger } = require('../utils/logger');
 
 // If you need to set the Redis URL, do it in REDIS_URL
@@ -13,8 +13,16 @@ const useMockRedis = process.env.MOCK_REDIS;
 // RedisConstructor is one of Redis or MockRedis
 const RedisConstructor = useMockRedis ? MockRedis : Redis;
 
-// Export either a real Redis instance, or Mocked.
-const redisInstance = new RedisConstructor(redisUrl);
+function createRedisClient() {
+  try {
+    const { port, host } = url.parse(redisUrl, true);
+    return new RedisConstructor(port, host, { password: process.env.REDIS_PASSWORD });
+  } catch (err) {
+    const message = `Unable to parse port and host from "${redisUrl}"`;
+    logger.error({ err }, message);
+    throw new Error(message);
+  }
+}
 
 // If using MockRedis, shim info() until https://github.com/stipsan/ioredis-mock/issues/841 ships
 if (useMockRedis && typeof MockRedis.prototype.info !== 'function') {
@@ -23,10 +31,8 @@ if (useMockRedis && typeof MockRedis.prototype.info !== 'function') {
 }
 
 module.exports = {
-  // Redis URL to use if using the constructor
-  redisUrl,
   // If callers need to create a new redis instance, they'll use the ctor
-  Redis: RedisConstructor,
+  createRedisClient,
   // Otherwise they can use this shared instance (most should use this)
-  redis: redisInstance,
+  redis: createRedisClient(),
 };


src/backend/lib/queue.js
```diff
@@ -1,13 +1,16 @@
...
 
 /**
  * Shared redis connections for pub/sub, see:
  * https://github.com/OptimalBits/bull/blob/28a2b9aa444d028fc5192c9bbdc9bb5811e77b08/PATTERNS.md#reusing-redis-connections
  */
-const client = new Redis(redisUrl);
-const subscriber = new Redis(redisUrl);
+const { port, host } = url.parse(redisUrl, true);
+
+const client = new Redis(port, host, redisPassword);
```
Reza-Rajabi: Based on the documentation you have addressed for Redis, the third argument passed to the constructor (aka [options]) is an object. So we probably need to pass it like: `{password: redisPassword}`  and actually they have used this example in their document:
```var authedRedis = new Redis(6380, '192.168.100.1', { password: 'password' });```

File: src/backend/lib/queue.js [hunk 2]
```diff
@@ -25,7 +28,7 @@ function createQueue(name) {
         case 'subscriber':
```
Reza-Rajabi: line 24 doesn't make sense for me. I know it is not your code. I am just trying to understand it. In the `const queue` I can't see who is going to pass the type to Bull constructor. What do you think?

src/backend/lib/queue.js
```diff
@@ -1,13 +1,16 @@
...
+const { Redis, redisUrl, redisPassword } = require('./redis');
 const { logger } = require('../utils/logger');
 
 /**
  * Shared redis connections for pub/sub, see:
  * https://github.com/OptimalBits/bull/blob/28a2b9aa444d028fc5192c9bbdc9bb5811e77b08/PATTERNS.md#reusing-redis-connections
  */
-const client = new Redis(redisUrl);
-const subscriber = new Redis(redisUrl);
+const { port, host } = url.parse(redisUrl, true);
```
humphd: Move all this into the `redis.js` file.  We should process all the `env` variables in there, and either expose `redisUrl` like we do now, or a `redisOptions` or something, that people can use to pass to the `Redis()` constructor.

src/backend/lib/queue.js
birtony: I am actually not sure how is it going to be passed either. What we would probably need to do is to pass it as a parameter to the _createQueue_ function, right?
```js
function createQueue(name, type) {
  const log = logger.child({ module: `queue:${name}` });
  const queue = new Bull(name, {
    createClient: type => {
      switch (type) {
        case 'client':
          return client;
        case 'subscriber':
          return subscriber;
        default:
          return new Redis(redisPort, redisHost, { password: redisPassword });
      }
    },
  })
// ...
return queue;
}
```

src/backend/lib/queue.js
birtony: Moved to `redis.js` and added `redisOptions` containing `port`, `host`, and `password` to `_module.exports_`.

src/backend/lib/queue.js
Reza-Rajabi: essentially the Bull constructor here accepts a name and an object (supposed to be Redis instance?) like this:
```
Bull(name, {createClient: function, ?})
.on(...)
...
.on(...);
```
I looked into their documentation and I didn't find this signature. 
I found [this](https://github.com/OptimalBits/bull/issues/711) code similar to ours though (second comment).
I didn't find anything similar in their documentation.
I don't know. It needs more time. I will get back to this later or if you found something wrong, you may want to make a new issue.

src/backend/lib/redis.js
```diff
@@ -24,7 +31,7 @@ if (useMockRedis && typeof MockRedis.prototype.info !== 'function') {
 
 module.exports = {
   // Redis URL to use if using the constructor
-  redisUrl,
+  redisOptions,
```
humphd: Instead of this, let's only expose a function that can create a Redis client connection using the host, password, port info from within this module, but not leak that out:

```js
function createRedisClient() {
  return new Redis(port, host, { password });
}
```

src/backend/lib/queue.js
humphd: We looked into this in class, and I think what we have is good.  The type doesn't need to be passed in for the flow we use.

env.example
```diff
@@ -18,6 +18,9 @@ FEED_URL=https://wiki.cdot.senecacollege.ca/wiki/Planet_CDOT_Feed_List
 REDIS_URL=redis://127.0.0.1:6379
 REDIS_PORT=6379
 
+# REDIS_PASSWORD specifies Redis server info
```
humphd: Let's group these `REDIS_` variables together, like so:

```
# Redis Server info, password may be optional (e.g., leave empty if you don't set one)
REDIS_PORT=6739
REDIS_PASSWORD=
```

src/backend/lib/redis.js
```diff
@@ -14,7 +16,13 @@ const useMockRedis = process.env.MOCK_REDIS;
 const RedisConstructor = useMockRedis ? MockRedis : Redis;
 
 // Export either a real Redis instance, or Mocked.
-const redisInstance = new RedisConstructor(redisUrl);
+const redisInstance = new RedisConstructor(port, host, {
```
humphd: Instead of doing this here, call your function below...

src/backend/lib/redis.js
```diff
@@ -23,10 +31,8 @@ if (useMockRedis && typeof MockRedis.prototype.info !== 'function') {
 }
 
 module.exports = {
-  // Redis URL to use if using the constructor
-  redisUrl,
   // If callers need to create a new redis instance, they'll use the ctor
-  Redis: RedisConstructor,
+  createRedisClient,
   // Otherwise they can use this shared instance (most should use this)
   redis: redisInstance,
```
humphd: `redis: createRedisClient(),`

src/backend/lib/redis.js
```diff
@@ -14,7 +16,13 @@ const useMockRedis = process.env.MOCK_REDIS;
 const RedisConstructor = useMockRedis ? MockRedis : Redis;
 
 // Export either a real Redis instance, or Mocked.
-const redisInstance = new RedisConstructor(redisUrl);
+const redisInstance = new RedisConstructor(port, host, {
+  password: process.env.REDIS_PASSWORD,
+});
+
+function createRedisClient() {
+  return new RedisConstructor(port, host, { password: process.env.REDIS_PASSWORD });
```
humphd: Question: what happens if the server doesn't set a password?  Will this method of doing it work?  Have you tested with/without a password?

src/backend/lib/redis.js
```diff
@@ -1,11 +1,13 @@
 require('./config');
 const Redis = require('ioredis');
 const MockRedis = require('ioredis-mock');
+const url = require('url');
 
 const { logger } = require('../utils/logger');
 
 // If you need to set the Redis URL, do it in REDIS_URL
 const redisUrl = process.env.REDIS_URL || 'redis://127.0.0.1:6379';
+const { port, host } = url.parse(redisUrl, true);
```
humphd: I think `url.parse()` can throw, so we should probably wrap this in a try/catch, and log an appropriate error before we crash.

humphd:
What's happening with this?  Are you going to fix up so I can merge?

birtony:
@humphd working on it, will push the update soon

src/backend/lib/redis.js
birtony: Testing

birtony: Ran the app with the empty `PASSWORD` and it seems to be working fine. I Have conducted some research on google, and it seems that Redis now supports empty string `PASSWORD` as a `null` value, and build the object only using `port` and `host`. Nevertheless, the constructor is now located inside the try/catch block, which would secure us from getting unexpected errors in any case.

humphd:
Tested locally, works great.  Merged and pushed to `master`.  Thanks!

birtony:
Thank you, @humphd!

PR #376, opened 2019-11-28
Fix #374 removed storage.js under src

### LINKED ISSUE #374: Duplicated file storage.js
`Storage.js` is currently in 2 different places, in `src/` and in `src/backend/utils/`
This probably happened after rebasing.

Screenshot from 2019-11-28 13-33-25

KarlaChen:
Removed ```storage.js``` under ```src``` and update ```storage.js``` under ```backend/utils```.
Fix issue #260, cause we used the ```storage.test.js``` to test the ```storage.js``` under src before. 
After I updated ```backend/utils/storage.js```, it fail the tester.


humphd:
Thanks for fixing this quickly.

PR #378, opened 2019-11-29
Fix issue #340: added functionality to get a number of feeds and posts

### LINKED ISSUE #340: Add API to storage.js for getting number of feeds and posts in Redis
Our storage.js code should make it possible to get the number of feeds and posts in the Redis database:

```
const feedsCount = await storage.getFeedsCount();
...
const postsCount = await storage.getPostsCount();
```

Both of these are implemented as Sets in Redis, and a Redis Set has `SCARD` to determine this, see https://redis.io/commands/scard.

Depends on work happening in #332 right now.

mskuybeda:
Added a functionality to get the number of feeds and posts in the Redis database:

```
const feedsCount = await storage.getFeedsCount();
...
const postsCount = await storage.getPostsCount();
```
No issues encountered. Added missing tests for receiving all feed ids. Added new test for feed and post count, as well as empty storage. Used method scard and zcard to retrieve values from redis arrays. 

Fixes #340.


humphd:
@mskuybeda this description isn't enough, "Please review."  Please update it to include info on what you changed, any issues you ran into, any follow-up we need, how to test it, etc.

test/storage.test.js
```diff
@@ -4,6 +4,11 @@ describe('Tests for storage', () => {
   const feed = { name: 'James Smith', url: 'http://seneca.co/jsmith' };
   const feed2 = { name: 'James Smith 2', url: 'http://seneca.co/jsmith/2' };
 
+  it('check that storage is empty', async () => {
+    const result = await storage.getFeedsCount();
```
humphd: This test is risky, since it depends on the order tests get run in, and could fail if that ever changed.

A better way would be to get the count, add a feed, get the count again, check that it's greater by 1.

test/storage.test.js
```diff
@@ -15,6 +20,19 @@ describe('Tests for storage', () => {
     expect(result).toEqual(feed2);
   });
 
+  it('check feed count', async () => {
+    const result = await storage.getFeedsCount();
```
humphd: Same issue here, and you're proving the problem: you depend on the previous test.  Each test in a test suite should be independent, and not depend on the order of test execution, or on other tests.

test/storage.test.js
```diff
@@ -43,6 +61,11 @@ describe('Tests for storage', () => {
     site: 'wordpress.com',
   };
 
+  it('check there is no posts', async () => {
+    const result = await storage.getPostsCount();
```
humphd: Same issue here.

test/storage.test.js
```diff
@@ -60,4 +83,9 @@ describe('Tests for storage', () => {
     expect(result[0]).toEqual(post2.guid);
     expect(result[1]).toEqual(post.guid);
   });
+
+  it('check post count', async () => {
+    const result = await storage.getPostsCount();
```
humphd: And here.

mskuybeda:
removed redundant tests and fixed existing. Please review

PR #390, opened 2019-12-01
Refactor shutdown logic into separate file

humphd:
I refactored the shutdown logic into its own file in `lib/shutdown.js` to simplify our main entry point.  I also removed a stray `async` call and unnecessary `log()` while I was testing.


humphd:
Windows CI build is not related to this, we can ignore.

humphd:
Rebased on Windows Travis fix that landed on master to pass CI.  This is ready for review.

PR #394, opened 2019-12-01
Fix #393 Add last test for backend/storage.js

### LINKED ISSUE #393: Add last test for backend/storage.js
The are still tests missing for the storage.js
missing

Cadesh:
There was still a missing test on storage.js related to the use of redis.smembers
I added the following test:
```javascript
  it('should return all members of the set value stored at key', async () => {
    await storage.addFeed(feed.name, feed.url);
    await storage.addFeed(feed2.name, feed2.url);
    const feeds = await storage.getFeeds();
    expect(feeds).toEqual(['1', '2', '3', '4']);
  });
```
Now the test suit is complete for this file:
missing2

### DIFF
--- a/test/storage.test.js
+++ b/test/storage.test.js
@@ -30,6 +30,13 @@ describe('Tests for storage', () => {
     expect(result).toContain(feedId2.toString());
   });
 
+  it('should return all members of the set value stored at key', async () => {
+    await storage.addFeed(feed.name, feed.url);
+    await storage.addFeed(feed2.name, feed2.url);
+    const feeds = await storage.getFeeds();
+    expect(feeds).toEqual(['1', '2', '3', '4', '5', '6', '7']);
+  });
+
   const post = {
     guid: 'tag:blogger.com,1999:blog-7100164112302197371.post-522285656016053350',
     author: 'Neil David',


Cadesh:
Conflicts resolved 🥇

test/storage.test.js
```diff
@@ -30,6 +30,13 @@ describe('Tests for storage', () => {
     expect(result).toContain(feedId2.toString());
   });
 
+  it('should return all members of the set value stored at key', async () => {
```
humphd: I have two questions:

1. How is this different from the previous test (above it)?  Do we need both? Do they test different things?
1. I'm concerned that this test has external requirements on test order execution: you assume that the previous tests exist, and have run before this test.  That's a dangerous thing, since anyone changing this code could break your test, without breaking the underlying code.

Cadesh: 1 - Yesterday I forked the repo and noticed that storage.js had this missing test as I showed in issue #393 . In storage.js history there wasn't any modifications. I fixed the missing test and committed, then I got the conflict message. It seems that there were a previous commit to fix this issue that did not showed up when I got my latest code. 
2 -  When I got the conflict I did not wanted to change the original code that showed up above mine. I just changed my code to fix the issue. No, we do not need both codes, mine can be discarded. 
I am sorry for the problem, for some reason I do not understand my code did not show the latest changes.

Cadesh: Now I understand what happened, there was another PR that was committed yesterday, this PR solved the same issue I was trying to solve in the storage.test.js . My PR went to waste.

humphd: Definitely not a waste!  You both had the same good idea.  Sorry that there was confusion, and duplication of effort.  I'll close this, but please still include it in your submissions for the course.

Issue #268, opened 2019-11-17
Improve test coverage


humphd:
Running `npm test` prints test coverage information.  Currently on `master` when I run the tests I see this:

```
--------------------------|----------|----------|----------|----------|-------------------|
File                      |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
--------------------------|----------|----------|----------|----------|-------------------|
All files                 |    74.39 |    62.24 |    67.44 |    73.86 |                   |
 src                      |    68.02 |     61.9 |    62.16 |    67.07 |                   |
  config.js               |      100 |      100 |      100 |      100 |                   |
  count-blog-domain.js    |    95.83 |      100 |      100 |    95.65 |                62 |
  email-sender.js         |    14.29 |        0 |        0 |    14.29 |... 42,50,51,52,54 |
  feed-parser.js          |      100 |      100 |      100 |      100 |                   |
  feed-queue.js           |      100 |      100 |      100 |      100 |                   |
  github-url.js           |    96.55 |    85.71 |      100 |    96.15 |                26 |
  inactive-blog-filter.js |    13.04 |        0 |       10 |    13.04 |... 30,131,132,135 |
  sentiment-analysis.js   |      100 |      100 |      100 |      100 |                   |
  spam-checker.js         |      100 |      100 |      100 |      100 |                   |
  storage.js              |    88.89 |      100 |    66.67 |    88.89 |                23 |
  summarize.js            |      100 |        0 |      100 |      100 |                 3 |
  text-parser.js          |      100 |      100 |      100 |      100 |                   |
  utils.js                |      100 |      100 |      100 |      100 |                   |
  word-counter.js         |      100 |      100 |      100 |      100 |                   |
 src/analysis             |    85.37 |    73.91 |      100 |    85.37 |                   |
  basic_analysis.js       |    85.37 |    73.91 |      100 |    85.37 | 20,22,34,35,40,82 |
 src/backend/web          |      100 |      100 |      100 |      100 |                   |
  app.js                  |      100 |      100 |      100 |      100 |                   |
 src/backend/web/routes   |      100 |      100 |      100 |      100 |                   |
  admin.js                |      100 |      100 |      100 |      100 |                   |
  index.js                |      100 |      100 |      100 |      100 |                   |
 src/lib                  |      100 |       50 |      100 |      100 |                   |
  redis.js                |      100 |       50 |      100 |      100 |                11 |
[...]

PR #395, opened 2019-12-01
Fixes #197, fixes #199: revamp inactive-blog-filter and tests

### LINKED ISSUE #197: add test case for issue #22

### LINKED ISSUE #199: Make inactive blog filter update asynchronous
Right now the `update()` function in the inactive blog filter module is no asynchronous by design. It may be more efficient to change it to an async function that resolves/rejects a promise upon completion.

Silvyre:
Fixes #197: builds upon #211 and #380 to boost testing coverage for `inactive-blog-filter.js` into the green (representing a significant step towards the resolution of #268).

Fixes #199: both the `check()` and `update()` functions of `inactive-blog-filter.js` now use [fs.promises](https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fs_promises_api). Thank you, @MeisterRed, for agreeing to collaborate!

### DIFF
--- a/src/backend/utils/inactive-blog-filter.js
+++ b/src/backend/utils/inactive-blog-filter.js
@@ -98,6 +98,8 @@ async function update(
         // We convert the dateDiff(result) from ms in days
         const timeDiff = Math.ceil(dateDiff(recentPostDate) / (1000 * 3600 * 24));
 
+        // BLOG_INACTIVE_TIME is supplied with a necessary default value, e.g. of 365 days
+        // See: https://github.com/Seneca-CDOT/telescope/issues/396
         if (timeDiff > (process.env.BLOG_INACTIVE_TIME || 365)) {
           log.info(`Blog at: ${feedUrl} is INACTIVE!`);
 
--- a/src/backend/utils/inactive-blog-filter.js
+++ b/src/backend/utils/inactive-blog-filter.js
@@ -34,7 +34,7 @@ function dateDiff(postDate) {
 /**
  * Checks if feed url is redlisted
  * @param {string} feedUrl - url of the feed to check against redlist
- * @param {string} [redlistPath=feeds-redlist.json] - relative path to JSON file storing non-active feeds
+ * @param {string} [redlistPath=feeds-redlist.json] - path to JSON file storing non-active feeds
  */
 async function check(feedUrl, redlistPath = 'feeds-redlist.json') {
   // Read redlist file
@@ -46,7 +46,7 @@ async function check(feedUrl, redlistPath = 'feeds-redlist.json') {
       return Promise.resolve(redList.some(isRedlisted.bind(null, feedUrl)));
     })
     .catch(err => {
-      log.error(`failed to read ${redlistPath}: no feeds`);
+      log.error(`failed to read ${redlistPath}`, err.message);
       return Promise.reject(err);
     });
 }
@@ -59,9 +59,9 @@ async function check(feedUrl, redlistPath = 'feeds-redlist.json') {
  * pass feed data more intuitively
  *
  * Due to amount of operations, this can be run periodically instead of with every feed update
- * @param {string} [feedsPath=feeds.txt] - relative path to .txt file storing active feeds
- * @param {string} [redlistPath=feeds-redlist.json] - relative path to JSON file storing non-active feeds
- * @param {feed} [parser=feedParser] - reference to the function used to parse feeds
+ * @param {string} [feedsPath=feeds.txt] - path to .txt file storing active feeds
+ * @param {string} [redlistPath=feeds-redlist.json] - path to JSON file storing non-active feeds
+ * @param {function} [parser=feedParser] - reference to the function used to parse feeds
  */
 async function update(
   feedsPath = 'feeds.txt',
@@ -69,7 +69,7 @@ async function update(
   parser = feedParser
 ) {
   if (!feedsPath || !redlistPath) {
-    return Promise.reject(new Error('failed to update: bad arguments'));
+    return Promise.reject(new Error('failed to update: bad filepath'));
   }
 
   // Read the feeds list file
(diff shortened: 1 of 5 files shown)


ImmutableBox:
Do you think it's possible to split **updating `inactive-blog-filter` to use promises** and **improving test coverage for `inactive-blog-filter`** to different PRs? It would make reviewing this a lot easier.

Silvyre:
That might be a bit tricky, Paul, as the updated `inactive-blog-filter.js` is incompatible with the current `inactive-blog-filter.test.js`, and vice versa.

ImmutableBox:
> 
> 
> That might be a bit tricky, Paul, as the updated `inactive-blog-filter.js` is incompatible with the current `inactive-blog-filter.test.js`, and vice versa.

Alright, that makes sense. It might take a little bit to review, since it's somewhat of a big change.

Silvyre:
Sure thing; thanks for all the help!

src/backend/utils/inactive-blog-filter.js
```diff
@@ -92,20 +86,19 @@ function update() {
...
-      if (typeof feed[0] !== 'undefined') {
+      if (feed && typeof feed[0] !== 'undefined') {
         recentPostDate = new Date(feed[0].date);
 
         // Check if the blog is inactive
         // We convert the dateDiff(result) from ms in days
         const timeDiff = Math.ceil(dateDiff(recentPostDate) / (1000 * 3600 * 24));
 
-        if (timeDiff > process.env.BLOG_INACTIVE_TIME) {
+        if (timeDiff > (process.env.BLOG_INACTIVE_TIME || 365)) {
```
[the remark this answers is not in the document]
Silvyre: Actually, this is the exact change that was needed to pass my CI tests! Check out #396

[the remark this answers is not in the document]
ImmutableBox: Oh OK. Do you still want to at least make a comment for `365`, since I couldn't tell what the number was at first glance meaning later contributors might also have that problem in the future.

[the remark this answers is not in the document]
Silvyre: Great idea! Adding one right in!

src/backend/utils/inactive-blog-filter.js
```diff
@@ -98,6 +98,8 @@ async function update(
         // We convert the dateDiff(result) from ms in days
         const timeDiff = Math.ceil(dateDiff(recentPostDate) / (1000 * 3600 * 24));
 
+        // BLOG_INACTIVE_TIME is supplied with a necessary default value, e.g. of 365 days
+        // See: https://github.com/Seneca-CDOT/telescope/issues/396
         if (timeDiff > (process.env.BLOG_INACTIVE_TIME || 365)) {
```
ImmutableBox: Good job with the comment 👍 .

But I just realized something:

Lets say `process.env.BLOG_INACTIVE_TIME` value is greater than `365` day (Lets say 500 days). 
If a blog has only been active for 400 days it fails the first check but passes on `365` which is then flagged which I assume is what we don't want.

Maybe something like this would be better? To use only one value instead of checking both?:
```
let inactive_time = process.env.BLOG_INACTIVE_TIME != null? process.env.BLOG_INACTIVE_TIME : 365;
if (timeDiff > inactive_time) {
```

Silvyre: Yup, the code actually does this! `(a || b)` is actually functionally equivalent to `(a ? a : b)` and is called a [short-circuit evaluation.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_Operators#Short-circuit_evaluation)!

ImmutableBox: The expression is this:
timeDiff > (a || b)

Let say the values are:
a = 500 days (process.env.BLOG_INACTIVE_TIME);
b = 365;
timeDiff = 400;

**First step**: 400 > 500 is false but since || is OR it doesn't just stop it moves onto the next expression
**Second step**: 400 > 365 which is true. 

> (a || b) is functionally equivalent to (a ? a : b)

They are the same but in this context, we are setting a default value (We don't check both 500 and 365):
```
let inactive_time = process.env.BLOG_INACTIVE_TIME != null? process.env.BLOG_INACTIVE_TIME : 365;
if (timeDiff > inactive_time) {
```

Silvyre: > 
> 
> The expression is this:
> `timeDiff > (a || b)`
> 

>
> **First step**: 400 > 500 is false but since || is OR it doesn't just stop it moves onto the next expression
> **Second step**: 400 > 365 which is true.

A good thought, but, in this case, `(a || b)` is resolved *before* the binary greater-than comparison with `timeDiff` due to the enclosure of the [short-circuit evaluation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_Operators#Short-circuit_evaluation) by a [grouping operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Grouping) (i.e. parentheses), which provides that expression with [a higher priority](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence#Table).

As such, the logic in question proceeds as follows:
- If `process.env.BLOG_INACTIVE_TIME` is [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) (e.g. undefined, null, etc.), `timeDiff > (process.env.BLOG_INACTIVE_TIME || 365)` evaluates to `timeDiff > 365`. 
- If `process.env.BLOG_INACTIVE_TIME` is [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) (e.g. is defined by a `.env` file), `timeDiff < (process.env.BLOG_INACTIVE_TIME || 365)` evaluates to `timeDiff > process.env.BLOG_INACTIVE_TIME `. 

(On a side note, our professor, @humphd, actually [specifically recommended use of short-circuit evaluation](https://seneca-open-source.slack.com/archives/C8WKYE7E3/p1575216101086600) in these scenarios.)

Thank you for taking the time review my pull request in such detail, I really appreciate it!

ImmutableBox: Oh OK. I think I get it. Thanks for providing the information. I don't have any other problems with this. Need someone else to review this code now 👍 .

PR #397, opened 2019-12-01
Added add-feed script to manually add a feed to the queue

### LINKED ISSUE #273: Write a command line tool for injecting a feed into the queue
In #112, I mentioned a cool idea that I got from [this code](https://github.com/GetStream/Winds/blob/48cda6bef1ae4cd0612368bef86becead4f0e8c5/api/src/commands/winds-og.js) is to add a node.js command-line tool that lets you inject a feed name/URL manually into the running queue.

This would be great for debugging, and for our testing purposes right now, while we wait to get all the feeds loading automatically.

This is somewhat similar to #266, but where that one is an automated process that goes and reads data from the wiki page, this one would be to write a command line tool that works like this:

```
$ node add-feed.js "FirstName LastLastName" https://url.to/user/feed.xml
```

This would cause the user "FirstName LastName", and their feed URL to be inserted into our running feed queue, and make it easier to test.

miggs125:
fix #273. This PR is unfinished but I wanted to create one for feedback before I commit more work into this. I have a working script with some bugs. I've manually tested the script by checking the bull board and confirming the feed job was queued.

To run the script, use ``` npm run add-feed -- --name "bloggers name" --url <feed url> ```. This will make use of the feedQueue.add under ``` src/backend/feed/queue ``` to add a feed to the queue for processing. 

I wanted to make this as an npm script so I changed package.json but if the script is executed using npm, the appropriate error messages seem to be suppressed and it gives a vague npm error message instead. When run as ``` node add-feed.js ..etc ``` the proper error and info messages are displayed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,8 @@
   "dependencies": {
     "bull": "3.12.1",
     "bull-board": "0.5.0",
+    "commander": "4.0.1",
+    "dompurify": "2.0.7",
     "dotenv": "8.2.0",
     "express": "4.17.1",
     "express-healthcheck": "0.1.0",
@@ -56,6 +58,7 @@
     "ioredis": "4.14.1",
     "ioredis-mock": "4.18.4",
     "jsdom": "15.2.1",
+    "minimist": "1.2.0",
     "node-fetch": "2.6.0",
     "node-summarizer": "1.0.7",
     "nodemailer": "6.3.1",
--- a/package.json
+++ b/package.json
@@ -13,6 +13,9 @@
       "./test/lib/setupJest.js"
     ]
   },
+  "bin": {
+    "add-feed": "./src/backend/utils/add-feed.js"
+  },
   "scripts": {
     "eslint": "eslint --ignore-path .gitignore .",
     "eslint-fix": "eslint --fix --ignore-path .gitignore .",
@@ -31,8 +34,7 @@
     "webhint": "cross-env PORT=3000 npm-run-all -r -p server hint",
     "lighthouse-index": "lighthouse http://localhost:3000 --output-path=./lighthouse-report.html --view",
     "lighthouse": "cross-env PORT=3000 npm-run-all -r -p server lighthouse-index",
-    "debug-server": "nodemon server.js",
-    "add-feed": "node src/backend/utils/add-feed.js"
+    "debug-server": "nodemon server.js"
   },
   "husky": {
     "hooks": {
--- a/package.json
+++ b/package.json
@@ -31,7 +31,8 @@
     "webhint": "cross-env PORT=3000 npm-run-all -r -p server hint",
     "lighthouse-index": "lighthouse http://localhost:3000 --output-path=./lighthouse-report.html --view",
     "lighthouse": "cross-env PORT=3000 npm-run-all -r -p server lighthouse-index",
-    "test-ci": "run-s prettier-check test"
+    "debug-server": "nodemon server.js",
+    "add-feed": "node src/backend/utils/add-feed.js"
   },
   "husky": {
     "hooks": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -28,7 +28,8 @@
     "server": "node src/backend/web/server",
     "hint": "hint http://localhost:3000",
     "webhint": "cross-env PORT=3000 npm-run-all -r -p server hint",
-    "debug-server": "nodemon server.js"
+    "debug-server": "nodemon server.js",
+    "add-feed": "node src/backend/utils/add-feed.js"
```
humphd: Let's use `bin` instead, see docs at https://docs.npmjs.com/files/package.json#bin

src/backend/utils/add-feed.js
```diff
@@ -0,0 +1,49 @@
...
+const feedQueue = require('../feed/queue');
+
+const log = logger.child({ module: 'add-feed' });
+
+async function add() {
+  if (
+    !Object.prototype.hasOwnProperty.call(args, 'name') ||
+    !Object.prototype.hasOwnProperty.call(args, 'url')
+  ) {
+    log.error('INVALID NUMBER OF ARGUMENTS');
```
humphd: Nicer to print a help/usage message when the args are wrong, teach the user how to call it.

src/backend/utils/add-feed.js
```diff
@@ -0,0 +1,49 @@
...
+    process.exit(1);
+  }
+  const { name, url } = args;
+  if (!isValidUrl(url)) {
+    log.error('INVALID URL');
+    process.exit(1);
+  }
+  const cleanName = String(name);
+  if (cleanName.match(/[.*+?^${}()|[\]\\`]/)) {
```
humphd: What's the goal here?  Why do we need to validate names?  What if someone's name contains unicode or some such?

src/backend/utils/add-feed.js
```diff
@@ -0,0 +1,49 @@
...
+    log.error('ONLY ALPHANUMERICAL CHARACTERS ALLOWED');
+    process.exit(1);
+  }
+
+  const feed = { name, url };
+
+  log.info(`Enqueuing feed job for ${url}`);
```
humphd: We probably don't need to log when things work, per Unix tradition.

src/backend/utils/add-feed.js
```diff
@@ -0,0 +1,49 @@
...
+        delay: process.env.FEED_QUEUE_DELAY_MS || 60 * 1000,
+      },
+      removeOnComplete: true,
+      removeOnFail: true,
+    })
+    .catch(err => {
+      log.error({ err }, 'Error enqueuing feed');
+      process.exit(1);
+    });
+  log.info('feed added');
```
humphd: Remove this, I think.

src/backend/utils/add-feed.js
miggs125: I was over thinking it. Over engineered, will remove

src/backend/utils/add-feed.js
miggs125: good idea!

miggs125:
@humphd I made changes to reflect the review comments. I haven't added a test yet. Not sure of how much work is involved. I was thinking we could make that a separate issue? I tested it manually and it seems to work. @manekenpix tested and confirmed it works as well. The tool can be used as such 
`add-feed --name "bob dole" --url http://theBestUrlEverDurrr`

src/backend/utils/add-feed.js
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env node
```
humphd: Can this live in `tools/` or something?  It's not really `src/`.

PR #278, opened 2019-11-18
Fix 274 Add pino logging to backend

miggs125:
Closes #274

PR #399, opened 2019-12-02
Updated outdated npm packages

### LINKED ISSUE #296: Update outdated npm packages
We have a few packages that could be updated (`npm outdated` will show you):

```
$ npm outdated
Package                 Current  Wanted  Latest  Location
bull                     3.11.0  3.11.0  3.12.1  telescope
eslint-config-prettier    6.5.0   6.5.0   6.7.0  telescope
nodemon                  1.19.4  1.19.4   2.0.0  telescope
pino-pretty               3.3.0   3.3.0   3.4.0  telescope
stylelint                11.1.1  11.1.1  12.0.0  telescope
```

NOTE: the `pino-pretty` update is happening in #278, so ignore that one.  But the rest might be good to do.  The only two that need a bit of investigation, to make sure they don't break us with API changes, are `nodemon` and `stylelint`.  Go look at these repos and their releases/changelog to see what changed in the new version, and if we have to do anything to switch to it.

NOTE2: If you haven't updated to `master` recently, note that we removed `package-lock.json` and switched to exact versions in `package.json` in #295.  We don't use `^` or `~` semver prefixes on package versions anymore.

shmooey:
**Related issue (if applicable)**: fixes #296 

**Description**:

Updates packages in package.json. Packages that were updated:
- eslint 6.6.0 -> 6.7.2
- eslint-config-prettier 6.5.0 -> 6.70
- bull 3.11.0 -> 3.12.1
- ioredis-mock 4.18.2 -> 4.18.4
- nodemon 1.19.4 -> 2.0.1 
This one mentions a "breaking change"
"Upgrading to chokidar@3 drops support for node@4, so nodemon is doing the same, and now supports node@8+
Chokidar upgrade means: massive CPU & RAM consumption improvements. 17x package & deps size reduction."
I don't believe this affects us

- stylelint 11.1.1 -> 12.0.0

Per #296 I have not updated pino-pretty, but it does not look like it was updated in #278 , should it be updated?


package.json
```diff
@@ -72,21 +72,21 @@
...
+    "eslint-config-prettier": "6.7.0",
     "eslint-plugin-import": "2.18.2",
     "eslint-plugin-prettier": "3.1.1",
     "hint": "5.4.1",
     "husky": "3.1.0",
     "jest": "24.9.0",
     "jest-fetch-mock": "2.1.2",
     "nock": "11.7.0",
-    "nodemon": "1.19.4",
+    "nodemon": "2.0.1",
```
humphd: I think we can rip this out completely, we don't (or shouldn't be) using `nodemon` anymore.  Please also get rid of the line above that uses it, `"debug-server": "nodemon server.js"`

humphd:
We have one issue with the spam-filter not liking our version of node.js, not related to your PR, but somewhat connected to the general outline of what you're doing.

shmooey:
Oh ok, phew!

PR #400, opened 2019-12-02
Fixes 335: Remove duplicity caused by using both async and Promise in a function 

robertbegna:
@humphd 
Sorry for the delay for this fix
As you suggested, I got rid of the additional promises and simplified the try catch block to just throw the error generated, rather than using Promise.reject()... since await function implicitly does that


humphd:
This is merge now, closing.

Issue #259, opened 2019-11-17
Create a Post type to represent the data from a feed article


humphd:
Our [feed-parser](src/feed-parser) is working now, and turns a feed URL into a JavaScript Object representing a feed article.

<details>
<summary>Here is an example of what the Object looks like</summary>
<code><pre>
{ title: 'My First Blog',
  description:
   'I have never done this before, so let\'s make this short. This post is just a test on how my blog post will look like.',
  summary: null,
  date: 2008-09-07T22:12:00.544Z,
  pubdate: 2008-09-07T22:09:00.000Z,
  pubDate: 2008-09-07T22:09:00.000Z,
  link: 'http://nadavid.blogspot.com/2008/09/my-first-blog.html',
  guid:
   'tag:blogger.com,1999:blog-7100164112302197371.post-522285656016053350',
  author: 'Neil David',
  comments:
   'http://www.blogger.com/comment.g?blogID=7100164112302197371&postID=522285656016053350',
  origlink: null,
  image: {},
  source: {},
  categories: [],
  enclosures: [],
  'atom:@': {},
  'atom:id':
   { '@': {},
     '#':
      'tag:blogger.com,1999:blog-7100164112302197371.post-522285656016053350' },
  'atom:published': { '@': {}, '#': '2008-09-07T15:09:00.000-07:00' },
  'atom:updated': { '@': {}, '#': '2008-09-07T15:12:00.544-07:00' },
  'atom:title': { '@': { type: 'text' }, '#': 'My First Blog' },
  'atom:content':
   { '@': { type: 'html' },
     '#':
      'I have never done this before, so let\'s make this short. This post is just a test on how my blog post will look like.' },
  'atom:link':
   [ { '@': [Object] },
     { '@': [Object] },
     { '@': [Object] },
     { '@': [Object] },
     { '@': [Object] } ],
  'atom:author':
   { '@': {},
     name: { '@': {}, '#': 'Neil David' },
     uri:
      { '@': {},
        '#': 'http://www.blogger.com/profile/15120043357385745054' },
     email: { '@': {}, '#': 'noreply@blogger.com' },
     'gd:image': { '@': [Object] } },
  'thr:total': { '@': {}, '#': '0' },
  meta:
   { '#ns':
      [ [Object], [Object], [Object], [Object], [Object], [Object] ],
     '@':
      [ [Object], [Object], [Object], [Object], [Object], [Object] ],
[...]

Issue #352, opened 2019-11-26
Investigate DOMPurify and JSDOM issues in HTML sanitizer


humphd:
In #344 we applied a bandaid fix to a seemingly larger problem: our HTML sanitizer uses DOMPurifier and JSDOM in such a way that certain HTML can cause it to throw.  Specifically, anything that requires a resource to be loaded (e.g., `<img src>`).

We need to get this working in a way we can count on.  Ideas to investigate:

- version issues with JSDOM and DOMPurify, see https://github.com/Seneca-CDOT/telescope/issues/343#issuecomment-558363444
- the way we create our document from snippets, see https://github.com/Seneca-CDOT/telescope/issues/343#issuecomment-557930403

It could also be something else.  DOMPurify is able to pass their unit tests using JSDOM, and doing almost the same thing we were doing when we failed, see https://github.com/cure53/DOMPurify/blob/3bc348bf5882b200e559f65f3fc749733afb812d/test/bootstrap-test-suite.js#L13-L37 and https://github.com/cure53/DOMPurify/blob/dd053931337b457da156530c4db0c1d5c92f66a4/test/jsdom-node.js.

miggs125:
@humphd I've started investigating this after you mentioned it yesterday, I can focus on this one but I may have to give away the issue I am currently assigned #273

humphd:
@miggs125 this isn't critical, I just wanted to get it filed for later.  We can loop back on it.

humphd:
We're going to need a fix for this soon, since we've hooked up the sanitizer.  I think we probably need to change how we/DOMPurify is creating the HTML document that gets parsed, so that it doesn't fail when trying to load resources (e.g., img)

```
[ 2019-12-05 11:15:38.669 ] ERROR (30885 on MacBook-Pro.local): Unable to process feed https://ekatdev.wordpress.com/category/SPO600/feed/ for job 1287
    err: {
      "type": "TypeError",
      "message": "Cannot read property 'fetch' of null",
      "stack":
          TypeError: Cannot read property 'fetch' of null
              at PerDocumentResourceLoader.fetch (/Users/humphd/repos/telescope/node_modules/jsdom/lib/jsdom/browser/resources/per-document-resource-loader.js:17:42)
              at HTMLImageElementImpl._attrModified (/Users/humphd/repos/telescope/node_modules/jsdom/lib/jsdom/living/nodes/HTMLImageElement-impl.js:39:36)
[...]

PR #403, opened 2019-12-02
Fixes Issue #259. Adds a class named Post to store data for each blog post.

jatinAroraGit:
This PR will add a file named post.js which hosts the class Post with some attributes of class. The PR also modifies worker.js in src/backend to process a post and store it as an object of class Post. This PR does not include the capability to sanitize the post description yet but can be added later on once issue #352 is addressed.
Let me know of any changes to be made.


humphd:
@jatinAroraGit I'm continuing your work in #408, I needed to update it to use some other code we just landed.  I'll close this, and continue there.

Issue #367, opened 2019-11-28
Add Logging for the Feed parser


cagomezr:
The Feed parser  doesn't  provide us with   a lot of  feedback,  it be very useful to  have a way to get a log of where failures happened so that we can review later.

giatuongtran9:
I can give it a try.

PavanKKamra:
Hello, I would like to attempt to try this issue!

humphd:
I'm doing this, closing.

PR #404, opened 2019-12-02
Fixed Issue #367: Add Logging to feed parser

PavanKKamra:
I added logging to the feed parser which shows the title, author, and summary.


humphd:
@PavanKKamra I don't think this solves our issue.  We've already got logging for the queue when items enter and leave, what we really need is to understand what failures mean.  I'm going to close this one.  I plan to work on this a bit in class today, so we can discuss further in class.

PR #407, opened 2019-12-02
issue#255 fix, added comment for silent logging options

jayson528:
If we'd like our logger to NOT display any logs on the console, we can use silent as the log level like this (LOG_LEVEL=silent) in the .env file. This will completely disable all logs. So now when we run our tests, it will look a lot less cluttered.

Screen Shot 2019-12-02 at 12 57 20 PM


humphd:
Fixed, rebased, and merged in `master`.

PR #409, opened 2019-12-02
Fixes #314: Server-side changes for server sent events

s-arika:
Fixes #314. 
The issue was broken down into 2 PRs. The client side can be found in PR #410  . 
The goal of this pr is to create a route that sends updates to the browser when there is a job completion.

PR #410, opened 2019-12-02
Fixes #314: Client-side changes for server sent events

### LINKED ISSUE #314: Add Server Sent Events to web server to signal new posts
We need a way to signal to the browser that new posts are available to be downloaded.  When our feedQueue completes processing all the posts, we need to listen for an event, and then trigger a message to the browser.

Let's try using [Server Sent Events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events).

In another PR, we also need to update our frontend index.html to connect to this route, and listen for these updates.

zufishanali:
Fixes issue #314. 

This PR addresses the client-side changes for server sent events.  Changes were made to allow listening to the stream endpoint using EventSource. Receives data whenever a job is completed in the queue.

Server-side PR for this fix can be found in #409

### DIFF
--- a/src/frontend/index.html
+++ b/src/frontend/index.html
@@ -298,5 +298,7 @@
       src="https://translate.google.com/translate_a/element.js?cb=googleTranslateElementInit2"
     ></script>
     <script src="./dialogue.js"></script>
+    <!-- Server Sent Events Script -->
+    <script src="./server-sent-events.js"></script>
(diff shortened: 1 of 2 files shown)


zufishanali:
I have requested a review again as I made a small change to match the update @s-arika made in PR #409.

PR #411, opened 2019-12-02
Fixes #264: Store Posts into Redis

### LINKED ISSUE #264: Feed Worker should store Posts in Redis when done
Our [feed-worker.js](src/feed-worker.js) currently parses Posts, puts them into an array, and console.logs them.

We need to use the work in #260 and #259, we need to update it so that a Post gets stored in the database.  This depends on #260 for the changes to [storage.js](src/storage.js).

c3ho:
This PR is still a WIP and builds on top of the work done for storage.js. 
1. It will process feeds in the workerCallback function using `feedParser`
2. Grab the relevant data from the post as defined by `const processedPost`

On completion of the processing, it should then store each individual post(result) into Redis by using `storage.addPost()`. There's something we may have to look at though as storage.addPost is an async function. Should we make the exports.start function an async function so we can use await?

### DIFF
--- a/src/backend/feed/worker.js
+++ b/src/backend/feed/worker.js
@@ -1,19 +1,48 @@
-const feedparser = require('./parser');
 const feedQueue = require('./queue');
+const feedParser = require('./parser');
+const extractUrls = require('../utils/extract-urls');
+const storage = require('../utils/storage');
+// const logger = require('./../utils/logger');
 
 exports.workerCallback = async function(job) {
   const { url } = job.data;
-  const posts = await feedparser(url);
-  return posts.map(post => ({
-    author: post.author,
-    date: post.date,
-    title: post.title,
-    description: post.description,
-    postURL: post.link,
-  }));
+  try {
+    const posts = await feedParser(url);
+    const processedPosts = [];
+    if (posts.length > 0) {
+      posts.forEach(post => {
+        // We can extract any other information from the post that we need here.
+        const processedPost = {
+          author: post.author,
+          date: post.date,
+          title: post.title,
+          description: post.description,
+          postURL: post.link,
+        };
+        processedPosts.push(processedPost);
+        extractUrls.extract(post.description);
+      });
+      return processedPosts;
+    }
+    return processedPosts;
+    // return processedPosts;
+  } catch (err) {
+    return err;
+  }
 };
 
+// eslint-disable-next-line prettier/prettier
 exports.start = function() {
   // Start processing jobs from the feed queue...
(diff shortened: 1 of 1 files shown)


src/backend/feed/worker.js
```diff
@@ -1,19 +1,48 @@
-const feedparser = require('./parser');
 const feedQueue = require('./queue');
+const feedParser = require('./parser');
+const extractUrls = require('../utils/extract-urls');
+const storage = require('../utils/storage');
+// const logger = require('./../utils/logger');
 
 exports.workerCallback = async function(job) {
   const { url } = job.data;
-  const posts = await feedparser(url);
```
humphd: Can you rebase on `master` please?  We have updated the way this `workerCallback` does its thing.

src/backend/feed/worker.js
```diff
@@ -1,19 +1,48 @@
...
+      return processedPosts;
+    }
+    return processedPosts;
+    // return processedPosts;
+  } catch (err) {
+    return err;
+  }
 };
 
+// eslint-disable-next-line prettier/prettier
```
humphd: Why?

src/backend/feed/worker.js
```diff
@@ -1,19 +1,48 @@
...
 exports.start = function() {
   // Start processing jobs from the feed queue...
-  feedQueue.process(this.workerCallback);
+  feedQueue.process(exports.workerCallback);
+  feedQueue.on('completed', (job, results) => {
+    if (results.length > 0) {
```
humphd: Is this necessary?  If it's 0, won't it just never call the `forEach` callback?  I'd remove it.

src/backend/feed/worker.js
```diff
@@ -1,19 +1,48 @@
...
+      results.forEach(result => {
+        console.log(`Result: ${result.title}`);
+        // We have an issure here, addPost is an async function. Should we make the start
```
humphd: Yes, we also need to update how you do this "loop" when you have promises for each:

```js
await Promise.all(results.map(async result => storate.addPost(result)))
```

And yeah, we should probably make `start` async.

src/backend/feed/worker.js
c3ho: ESLint was complaining about an issue with function () and prettier kept doing changing my code which I had written function() to function ()

src/backend/feed/worker.js
c3ho: Will do!

humphd:
Fixed and merged in class.

PR #413, opened 2019-12-03
Fixed #283 added lighthouse testing

### LINKED ISSUE #283: Add Lighthouse CI
Similar to #272, it would be interesting to experiment with [Lighthouse CI](https://github.com/GoogleChrome/lighthouse-ci).  Lighthouse is Google's tool for finding performance and other frontend web issues in a site.

Like with webhint, I'm not sure we need to run this by default with `npm test`; but it could be an optional thing people can run if they are working on the frontend code.

dbeigi:
This pull request is to address #283. After a discussion with @humphd the decision was to use lighthouse. I have added two npm scripts, lighthouse-index, and lighthouse-login, to allow testing of the front-end. The results are presented in an HTML file and are displayed after executing the commands. Also added the extension to .gitignore to ensure the report results are not pushed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,8 @@ yarn-error.log*
 *.obj
 *.o
 
+lighthouse-report.html
+
 # web-hint report
 hint-report/
 
--- a/package.json
+++ b/package.json
@@ -27,7 +27,9 @@
     "start": "node src/backend",
     "server": "node src/backend/web/server",
     "hint": "hint http://localhost:3000",
-    "webhint": "cross-env PORT=3000 npm-run-all -r -p server hint"
+    "webhint": "cross-env PORT=3000 npm-run-all -r -p server hint",
+    "lighthouse-index": "lighthouse http://localhost:3000 --output-path=./lighthouse-report.html --view",
+    "lighthouse": "cross-env PORT=3000 npm-run-all -r -p server lighthouse-index"
   },
   "husky": {
     "hooks": {
@@ -81,6 +83,7 @@
     "husky": "3.1.0",
     "jest": "24.9.0",
     "jest-fetch-mock": "2.1.2",
+    "lighthouse": "5.6.0",
     "nock": "11.7.0",
     "npm-run-all": "4.1.5",
     "prettier": "1.19.1",


package.json
```diff
@@ -28,7 +28,9 @@
     "server": "node src/backend/web/server",
     "hint": "hint http://localhost:3000",
     "webhint": "cross-env PORT=3000 npm-run-all -r -p server hint",
-    "debug-server": "nodemon server.js"
+    "debug-server": "nodemon server.js",
+    "lighthouse-index": "lighthouse http://localhost:3000 --output-path=./index-report.report.html --view",
```
humphd: We'll need to pre-launch the web server on port 3000 for your script to work.  We did something similar for `webhint` above.  You need to do:

```
"lighthouse-index": "lighthouse http://localhost:3000 --output-path=./lighthouse-report.html --view",
"lighthouse": "cross-env PORT=3000 npm-run-all -r -p server lighthouse-index",
```

Also, I renamed the output path for the report.  We won't have a separate login.html page soon, so there's no point testing it.

package.json
```diff
@@ -28,7 +28,9 @@
...
+    "lighthouse-login": "lighthouse http://localhost:3000/login.html --output-path=./login-report.report.html --view"
```
humphd: You can remove this one, since we won't have a login page once we do SSO fully.

.gitignore
```diff
@@ -28,6 +28,8 @@ yarn-error.log*
 *.obj
 *.o
 
+*.report.html
```
humphd: Update this to match the name change below.

.gitignore
```diff
@@ -28,7 +28,7 @@ yarn-error.log*
 *.obj
 *.o
 
-*.report.html
+*.html
```
humphd: `lighthouse-report.html`

humphd:
This also needs a rebase.

package.json
```diff
@@ -27,7 +27,10 @@
     "start": "node src/backend",
     "server": "node src/backend/web/server",
     "hint": "hint http://localhost:3000",
-    "webhint": "cross-env PORT=3000 npm-run-all -r -p server hint"
+    "webhint": "cross-env PORT=3000 npm-run-all -r -p server hint",
+    "debug-server": "nodemon server.js",
```
humphd: This was removed recently, compare to what's on `master` now: https://github.com/Seneca-CDOT/telescope/blob/master/package.json#L16-L30.  You should make sure you are in sync with `master`.

PR #414, opened 2019-12-03
Closes #262: Add routes for getting posts and a post

### LINKED ISSUE #262: Add routes for getting posts and a post
In #260 we're going to be adding the ability to store Posts in the Redis database.  In #261 we're thinking about how to display the data in the UI.  Once these are in, we'll want to be able to get the posts out of Redis via our web server.

We need some routes to allow for querying the database for posts, and an individual post, as JSON, something like:

- `posts/` should return an array of IDs for all posts in the database.  Later we can update this to support ranges and pagination (only requesting the last 10, or the next 20-30, etc).
- `post/<id>` given an ID of a post, get back JSON for that post

Until the work in #260 is done, this could return dummy data so the frontend team can start working on it.  Reference what is happening in #259 to understand how this data should look.

manekenpix:
This PR only has the endpoints so it can be reviewed and merged easily so front-end contributors can have something to work with. Tests can be added later.

Endpoints added:

`GET /post/:id`
`id` is the identifier for a specific post
Returns:
```js
{
    'title':  '(String) the title of the blog post',
    'author': '(String) the author of the post',
    'content': '(String) the content of the post, what was originally named description run through our HTML sanitizer (#18)',
    'text': '(String) the content of the post after being run through our text parser (no HTML). We could make this a method, and only generate it on demand, or could store it on the Object.'
    'updated': '(Date) the date/time of the most recent update to the post'
    'published': '(Date) the date/time when the post was originally published'
    'url': '(String) the URL of the original post'
    'site': '(String) the URL of the site from which this was published (e.g., blog). This is available in meta.link'
    'guid': '(String) a unique ID for this post (e.g., use the guid property from the feed)'
}
```
The structure of the data returned was decided in #259.

`GET /posts`
Returns:
```js
[
    {
        'id': 'id of the post'
        'url': 'URL to get the post e.g. development: localhost3000/post/{id}',
    },
    {
        'id': 'id of the post'
        'url': 'URL to get the post e.g. production: telescope.cdot.systems:3000/post/{id}',
    }
...
]
```
`NODE_ENV` and `PORT` in `.env` are used for the URLs in the returned array.

### Update
At the moment, `/posts` returns all the stored posts. In the future, filters can be implemented to return sets posts using different types of ranges (date, number of posts...).

### DIFF
--- a/src/backend/utils/storage.js
+++ b/src/backend/utils/storage.js
@@ -57,11 +57,18 @@ module.exports = {
       .exec();
   },
 
-  getPosts: (startDate, endDate) =>
-    // Get all posts between start and end dates in the sorted set
-    redis.zrangebyscore(POSTS, startDate.getTime(), endDate.getTime()),
+  getPosts: () =>
+    // Get all posts
+    redis.zrange(POSTS, 0, -1),
 
   getPostsCount: () => redis.zcard(POSTS),
 
-  getPost: guid => redis.hgetall(guid),
+  getPost: async guid => {
+    const post = await redis.hgetall(guid);
+
+    if (Object.keys(post).length !== 0) post.guid = guid;
+    else return null;
+
+    return post;
+  },
 };
--- a/src/backend/web/routes/post.js
+++ b/src/backend/web/routes/post.js
@@ -1,0 +1,12 @@
+const express = require('express');
+const { getPost } = require('../../utils/storage');
+
+const post = express.Router();
+
+post.get('/:id', async (req, res) => {
+  const redisPost = await getPost(req.params.id);
+
+  res.json(redisPost);
+});
+
+module.exports = post;
--- a/src/backend/web/routes/posts.js
+++ b/src/backend/web/routes/posts.js
@@ -4,12 +4,9 @@ const { getPost, getPosts } = require('../../utils/storage');
 const posts = express.Router();
 
 posts.get('/', async (req, res) => {
-  const arrayOfPosts = [];
-
   const redisGuids = await getPosts();
 
-  redisGuids.forEach(guid => arrayOfPosts.push(getPost(guid)));
-  const processedPosts = await Promise.all(arrayOfPosts);
+  const processedPosts = await Promise.all(redisGuids.map(guid => getPost(guid)));
 
   const listOfPosts = processedPosts
     // Return id and url for a specific post
--- a/src/backend/web/routes/posts.js
+++ b/src/backend/web/routes/posts.js
@@ -1,0 +1,27 @@
+const express = require('express');
+const { getPost, getPosts } = require('../../utils/storage');
+
+const posts = express.Router();
+
+posts.get('/', async (req, res) => {
+  const arrayOfPosts = [];
+
+  const redisGuids = await getPosts();
+
+  redisGuids.forEach(guid => arrayOfPosts.push(getPost(guid)));
+  const processedPosts = await Promise.all(arrayOfPosts);
+
+  const telescopeURL = env === 'production' ? 'telescope.cdot.systems' : 'localhost';
+  const listOfPosts = processedPosts
+    // Return id and url for a specific post
+    .map(processed => {
+      return {
+        id: processed.guid,
+        url: `/post/:${processed.guid}`,
+      };
+    });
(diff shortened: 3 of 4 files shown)


src/backend/web/routes/post.js
```diff
@@ -0,0 +1,18 @@
+const express = require('express');
+const { getPost } = require('../../utils/storage');
+
+const post = express.Router();
+
+post.get('/:id', async (req, res) => {
```
humphd: Are we going to get burned here by URLs that aren't encoded property as the `id`?  A lot of `guid`s for Posts are just URIs or even URLs.  Should we use `/*` and grab everything off the route that way, or are we good if the passed id contains path separators?  I've had issues with this in the past.

src/backend/web/routes/post.js
```diff
@@ -0,0 +1,18 @@
...
+  const redisPost = await getPost(req.params.id);
+
+  /**
+   * If the post retrieved from redis doesn't have the property 'author'
```
humphd: I don't fully understand this comment.  It seems odd to use `author` as a sentinel like this.  Can we change how our storage code works to return null instead?

src/backend/web/routes/post.js
```diff
@@ -0,0 +1,18 @@
...
+   * it means the id passed to the endpoint is invalid, and the endpoint sends
+   * a JSON { 'id': '-1' } so the frontend knows that the passed id was invalid.
+   */
+  redisPost.guid = Reflect.has(redisPost, 'author') ? req.params.id : '-1';
```
humphd: First time I've ever seen `Reflect` used :)

Again, I'm skeptical of `'-1'` here.

src/backend/web/routes/posts.js
```diff
@@ -0,0 +1,53 @@
+require('../../lib/config');
+const express = require('express');
+const { getPost, getPosts } = require('../../utils/storage');
+
+const posts = express.Router();
+
+const env = process.env.NODE_ENV || 'development';
+const port = process.env.PORT || 3000;
```
humphd: Let's avoid needing this, and just return relative URLs for this route.

src/backend/web/routes/posts.js
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * addGuidToPost inserts the redis key into the fetched post from redis.
```
humphd: Let's fix our storage/redis code to just include this instead.  We own all the code, we can change anything to make it fit.

src/backend/web/routes/posts.js
```diff
@@ -0,0 +1,53 @@
...
+  post.guid = Reflect.has(post, 'author') ? guid : '-1';
+
+  return post;
+};
+
+posts.get('/', async (req, res) => {
+  const arrayOfPosts = [];
+
+  /**
+   * Temporary Date() until it is decided how to handle this
```
humphd: Let's alter the `getPosts()` function to accept a different number of arguments, and make some optional.  We don't always need/want to limit the range by a date, but we might.  If we do, we should expose more route options here to pass in ranges.  For now, you could default to return all.

src/backend/utils/storage.js
```diff
@@ -57,11 +57,18 @@ module.exports = {
...
+    // Get all posts
+    redis.zrange(POSTS, 0, -1),
 
   getPostsCount: () => redis.zcard(POSTS),
 
-  getPost: guid => redis.hgetall(guid),
+  getPost: async guid => {
+    const post = await redis.hgetall(guid);
+
+    if (Object.keys(post).length !== 0) post.guid = guid;
```
humphd: Instead of this, why don't we just add `guid` to the `hmset()` call above in `addPost`, so it's always available on every post object in the db.

manekenpix: @humphd `guid` is already in `hmset()` ([line 36](https://github.com/Seneca-CDOT/telescope/blob/2fd026ad4409298ccd69bceb9ffd600db6fad271/src/backend/utils/storage.js#L36)), but `hgetall` returns a set without the guid, that's why I added that.

humphd:
Reviewed and fixed in class, closing.

PR #421, opened 2019-12-03
Fixes #220: Added testcases for wiki-feed-parser

### LINKED ISSUE #220: Wiki feed parser Testing
Wiki Feed parser has the JS file but needs a testing file .

ThomasLuu00:
I added some test cases for wiki-feed-parser for #220 .

The getData seems to escape all the newline characters in the body of the mock fetch response. I am not sure why it does this. 

Also not really sure what's the expected error codes in the cases of failures, such as no pre tag or failed fetch, so I only added placeholders for now.

@cryolis thoughts?

### DIFF
--- a/test/wiki-feed-parser.test.js
+++ b/test/wiki-feed-parser.test.js
@@ -1,0 +1,146 @@
+const parser = require('../src/backend/utils/wiki-feed-parser');
+global.fetch = require('node-fetch');
+
+const mockFeed = `################# Failing Feeds Commented Out [Start] #################
+
+#Feed excluded due to getaddrinfo ENOTFOUND s-aleinikov.blog.ca s-aleinikov.blog.ca:80
+#[http://s-aleinikov.blog.ca/feed/atom/posts/]
+#name=Sergey Aleinikov
+
+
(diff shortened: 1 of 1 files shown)


cryolis:
getData() is passing an array of strings, it doesn't need new line characters there.  As for the 'pre' tag and or failed fetch, it will throw an error as it would be trying to read from null. I believe error was something like this:

TypeError: Cannot read property 'textContent' of null

test/wiki-feed-parser.test.js
```diff
@@ -0,0 +1,146 @@
...
+
+[http://kopay.wordpress.com/category/sbr600-win2011/feed]
+name=Pirathapan Sivalingam
+
+
+[http://jessefulton.wordpress.com/category/SBR600/feed/]
+name=Jesse Fulton
+
+
+[http://eric-spo600.blogspot.com/feeds/posts/default]
```
ThomasLuu00: I pass in a feed list similar with this format.  From what I understand the getData() is supposed to parse this information and as you said, return an array of strings.  However, it doesn't do this and just returns an array with one string where all the newline characters are escaped.

ThomasLuu00:
Alright, I used "TypeError: Cannot read property 'textContent' of null" as the error, but it might be better to generate a more specific error in the case \<pre\> does not exist in the received body.

PR #398, opened 2019-12-01
Fixes #386 - Added specific environment instructions in CONTRIBUTING.md

drwm-base:
Addressing #386 , this pull request updates the CONTRIBUTING.md to include docker and docker-compose instructions. It also includes updated instructions on installing Redis. 

There are variables in the docker-compose file that use the .env file to create the environment, and I am unsure of how to replicate it in the native application instructions. In my mind, we should remove the native set up entirely, and focus on using docker.

Another aspect of this pull request is the .env file. Specifically how to fill out the credentials inside of it. I am working on this aspect but I thought I would put up a pull request and get some feedback before I continue. 
UPDATE: I have talked to some people and understood the .env file better, I will update the environment set up in a new pull request, as it is more complex than I initially thought.

PR #422, opened 2019-12-03
Fix #319: add script to generate ssl certificates

### LINKED ISSUE #319: Create certs for the mock saml2 for docker.
This will allow someone to easily input this information when running docker so they can configure the server for their own specific use.

ragnarokatz:
Fixes issue #319 

This PR does the following:
- Add bash shell script to generate ssl certificates using openSSL
- Add instructions on how to do it in CONTRIBUTING.md.
- Add certs/ folder to .gitignore

Documentation update:
image

Generating a a new certificate:
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,6 @@ hint-report/
 
 #Redis dump file
 dump.rdb
+
+# SAML certificates
+certs/*
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -33,6 +33,9 @@ Some helpful guides:
 - Redis for MacOS - need instructions
   An easier solution would be to use Docker.
 
+**SAML Setup**
+- Run `bash ./generate_ssl_certs.sh` in terminal
+
 **Setup**
 
 1. Navigate to the root directory of telescope.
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -33,3 +33,6 @@ hint-report/
 
 #Redis dump file
 dump.rdb
+
+# SAML certificates
+certs/*
```
manekenpix: Missing nextline

docs/CONTRIBUTING.md
```diff
@@ -33,6 +33,9 @@ Some helpful guides:
 - Redis for MacOS - need instructions
   An easier solution would be to use Docker.
 
+**SAML Setup**
+- Run `bash ./generate_ssl_certs.sh` in terminal
+
```
manekenpix: I don't see the documentation update that you mention in the description, but someone is already working on this so don't worry about it. See #398

generate_ssl_certs.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+mkdir certs
+openssl req -x509 -newkey rsa:4096 -keyout certs/key.pem -out certs/cert.pem -nodes -days 900
```
manekenpix: Suggestion: Instead of using self-signed certificates, we could try something like [Let's Encrypt](https://letsencrypt.org/), or if we're going to use self-signed certificates maybe we don't need a script for that, we could just add these lines to our documentation.

PR #346, opened 2019-11-25
Fixes #263 - Added fetching and rendering of data from the backend to the frontend

lucacataldo:
Added the fetching & rendering of posts in frontend. This includes fetching post & participant data using jQuery (currently dummy data in src/frontend/dummy_data),  and rendering it using a templated <article> element. I picked some posts from the current planet that included elements I though would bring up some edge cases and adjusted CSS accordingly so that when this gets hooked into the actual routes it'll be a bit less garbled and legible. Also, since pagination.js dialogue.js are no longer relevant or being used, I removed them.

PR #426, opened 2019-12-03
Update work for issue 263 to connect to backend

humphd:
We merged a bunch of code today to get the redis and pipeline connected, and added it to the front-end, based on @lucacataldo's work from #346.  This needs more work, but IT WORKS! :fire:

Screen Shot 2019-12-03 at 1 34 02 PM


src/backend/utils/storage.js
```diff
@@ -34,6 +34,8 @@ module.exports = {
       .hmset(
         // using guid as keys as it is unique to posts
         post.guid,
+        'guid',
+        post.guid,
```
cindyledev: #427 just landed and it made these changes. This is no longer needed

humphd: Good catch.  I've rebased and updated this, new commits pushed to this branch.

cindyledev:
Pretty excited to see this land

humphd:
I fixed a few more small things, and now posts have content:

Screen Shot 2019-12-03 at 3 09 26 PM

humphd:
Let's not merge this yet, it's just a WIP to get us moving.

PR #434, opened 2019-12-04
Fix#270 use text-parser.js for wiki-feed-parser.js

haichuan0424:
Use text-parser.js to replace JSDOM for parsing text. When we doing some tests, the res.text() returns a string with length 0.

### DIFF
--- a/src/backend/utils/wiki-feed-parser.js
+++ b/src/backend/utils/wiki-feed-parser.js
@@ -1,8 +1,6 @@
 const fetch = require('node-fetch');
-const jsdom = require('jsdom');
 require('../lib/config');
-
-const { JSDOM } = jsdom;
+const textParser = require('../utils/text-parser');
 
 /*
  * getData() returns Promise { <pending> }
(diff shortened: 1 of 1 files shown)


src/backend/utils/wiki-feed-parser.js
```diff
@@ -1,21 +1,19 @@
...
-module.exports.getData = function() {
+module.exports.getData = async function() {
   return fetch(process.env.FEED_URL)
     .then(res => res.text())
-    .then(data => {
-      const dom = new JSDOM(data);
-      return dom.window.document.querySelector('pre').textContent.split(/\r\n|\r|\n/);
+    .then(async data => {
+      const result = await textParser(data);
+      return result.split(/\n/);
```
humphd: When @giatuongtran9 was working on the text parser, he thought about doing this as well.  We decided against it, because the way it works right now, we don't have the ability to request a specific element (e.g., `<pre>`) to get from within the DOM, we just default to `document.body`.  If we want to do this, we'd have to alter how it works.

To be honest, I'm not sure if it's worth it or not.  I think doing it would involve changing the text parser to take a 2nd optional argument, with a selector string to use when selecting the elements and `innerText`.

haichuan0424: @humphd OK, we thought about the <pre> issue too. However, since it extracts text from all the tags and only lines start with "#[" or "#name" gets processed, we might not need optional argument.  
Even with the original JSDOM version, res.text() seems to return an empty string. We created a test for wiki-feed-parser.js as well but it cannot pass due to this empty string issue when fetch from feed url. Could we submit the test instead as a new issue?

humphd: If there's a bug with empty strings, sure, file another issue.

humphd:
Per our discussion in class, I'm going to close this.

PR #436, opened 2019-12-04
Closes #432, #437: Limit number of posts returned by HTTP /posts route

### LINKED ISSUE #432: Limit number of posts returned by HTTP /posts route
We got the code to run today, and it's already obvious that we can't ever return/use all posts from the database--there are too many.

We should start by updating the `/posts` route to take an optional number: `/posts/:count?`, where `count` should be a number.  We'd use this as the upper limit for the number of posts we request from the database.

By default, I suggest that we should limit this to the most recent 10 posts, unless `count` is provided.  That is:

- `/posts` - should return an Array of 10 elements
- `/posts/15` - should return an Array of 15 elements

manekenpix:
`GET /posts` now accepts a [query parameter](https://guides.emberjs.com/release/routing/query-params/) to request a specific number of posts, with a max of 100 posts.
If no parameter is passed, it'll returned 30 posts.

For testing:
 - Run telescope `npm start` (with redis)
- In a browser, go to:
  - `localhost:3000/posts`, the number of posts shown should be 30
  - `localhost:3000/posts?counter=50`,  the number of posts shown should be 50
  - `localhost:3000/posts?counter=200`,  the number of posts shown should be 100

### DIFF
--- a/src/backend/utils/storage.js
+++ b/src/backend/utils/storage.js
@@ -59,7 +59,7 @@ module.exports = {
       .exec();
   },
 
-  getPosts: (counter = 0) =>
+  getPosts: counter =>
     /**
      * 'counter' needs -1 because 'zrevrange()' includes the element at index 'counter'
      * in the array it returns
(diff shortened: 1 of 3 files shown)


manekenpix:
@humphd sorry, just removed a magic number.

PR #438, opened 2019-12-04
fixes #420 duplicate files using the reading module in both

### LINKED ISSUE #420: Consolidate utils/basic_analysis and utils/reading-level
We have two files doing essentially the same thing:

- src/backend/utils/basic_analysis.js
- src/backend/utils/reading-level.js

Both use the `reading-level` module.  We don't need both, so let's get rid of one, and update the tests to use the proper one.  Also, we aren't importing correctly in both files, see #406.

jrdnlx:
I have removed the reading level module since the basic analysis does what the reading level does and more.


humphd:
This seems sane to me, but let's ask @jordanhui19, who did this work in 26183ceb.  Is there any reason we need this code, separate from the other `reading-level` code we have in our tree?

PR #448, opened 2019-12-05
fix #424 added ability to login and callback

### LINKED ISSUE #424: Add necessary express routes for SAML2
Adding everything that is needed in express to get SAML2 test server working.

Grommers00:
Documentation has to be changed quite a bit and added to create the necessary certifications.

This is a substantial amount of code, and will need some documentation. However, I think its important to keep moving forward, with code, and in the documentation there is a current tutorial that is being used to implement.

With this part completed, we should be able to add in a route that displays the login / logout button for people to access.


humphd:
```
FAIL  test/server.test.js
  ● Test suite failed to run

    ENOENT: no such file or directory, open '/home/circleci/telescope/src/backend/login/certs/certskey.pem'

      20 |     issuer: 'saml-poc',
      21 |     identifierFormat: null,
    > 22 |     decryptionPvk: fs.readFileSync(path.join(__dirname, '/certs/certskey.pem'), 'utf8'),
         |                       ^
      23 |     privateCert: fs.readFileSync(path.join(__dirname, '/certs/certskey.pem'), 'utf8'),
      24 |   },
      25 |   function(profile, done) {

      at Object.<anonymous> (src/backend/login/usingPassport.js:22:23)
      at Object.<anonymous> (src/backend/web/app.js:9:26)
```

Looks like we'll need to add steps to Travis/Circle CI to run your cert script before the tests.

src/backend/login/usingPassport.js
```diff
@@ -1,36 +1,36 @@
-const passport = require('passport');
```
humphd: I don't think this file is in the right place.  I'll let @miggs125 chime in, but for sure it needs to be somewhere under ` src/backend/web/*`

env.example
```diff
@@ -21,21 +21,21 @@ REDIS_PORT=6379
 # MOCK_REDIS=1 will use an in-memory, mock Redis instance. Useful for testing.
 MOCK_REDIS=
 
-#https://developers.onelogin.com/saml - is used for this current portion of code
+#https://medium.com/disney-streaming/setup-a-single-sign-on-saml-test-environment-with-docker-and-nodejs-c53fc1a984c9 - is used for this current portion of code
```
humphd: Spaces after `#` and before the first character in all your comments, please.

env.example
```diff
@@ -21,21 +21,21 @@ REDIS_PORT=6379
...
 
-#OICD_BASE_URI = BASE URL e.g.  https://openid-connect.onelogin.com/oidc
-OICD_BASE_URI=""
+#SAML2 = BASE URL e.g.  http://localhost/login/callback
```
humphd: `SAML2_BASE_URI`.  Please explain what this is, as well as giving an example.

env.example
```diff
@@ -21,21 +21,21 @@ REDIS_PORT=6379
...
+SAML2_BASE_URI=""
 
-#OIDC_CLIENT_ID = CLIENT ID obtained from OpenId Connect Strategy - THIS WILL BE OBTAINED FROM SENECA AT A LATER DATE.
-OIDC_CLIENT_ID=
+#SAML2_CLIENT_ID = CLIENT ID obtained from SAML Strategy default: saml-poc - THIS WILL BE OBTAINED FROM SENECA AT A LATER DATE.
```
humphd: You can get rid of all the `THIS WILL BE OBTAINED FROM SENECA AT A LATER DATE.` notes.

env.example
```diff
@@ -21,21 +21,21 @@ REDIS_PORT=6379
...
+SAML2_CLIENT_ID=
 
-#OIDC_CLIENT_SECRET = CLIENT SECRET obtained from OpenID Connect Strategy - THIS WILL BE OBTAINED FROM SENECA AT A LATER DATE.
-OIDC_CLIENT_SECRET=
+#SAML2_CLIENT_SECRET = CLIENT SECRET obtained from SAML Strategy - THIS WILL BE OBTAINED FROM SENECA AT A LATER DATE.
+SAML2_CLIENT_SECRET=
 
-#OIDC_REDIRECT_URI This is the call back URL, set within the OPENID CONNECT STRATEGY, WE WILL NEED TO GET THIS SET-UP WITH SENECA AT A LATER DATE. http://localhost:3000/oauth/callback is used for tutorial sake.
-OIDC_REDIRECT_URI=
+#SAML2_REDIRECT_URI This is the call back URL, set within the SAML STRATEGY, WE WILL NEED TO GET THIS SET-UP WITH SENECA AT A LATER DATE. http://localhost:4300/login/callback is used for tutorial sake.
```
humphd: `callback` vs. `call back`

env.example
```diff
@@ -21,21 +21,21 @@ REDIS_PORT=6379
...
+SAML2_REDIRECT_URI=
 
-#end portion of code with regards to https://developers.onelogin.com/saml
```
humphd: Get rid of this "footer" we don't need it.

src/backend/login/usingPassport.js
```diff
@@ -1,36 +1,36 @@
...
-OIDC_BASE_URI=https://openid-connect.onelogin.com/oidc
-OIDC_CLIENT_ID=ddf97d-xxx-xxxx-e92920
-OIDC_CLIENT_SECRET=f021f63bxxxx71f2c6a90c9b
-OIDC_REDIRECT_URI=http://localhost:3000/oauth/callback
+/*
+Tutorial is here: https://medium.com/disney-streaming/setup-a-single-sign-on-saml-test-environment-with-docker-and-nodejs-c53fc1a984c9
+SAML2_BASE_URI=https://openid-connect.onelogin.com/oidc
+SAML2_CLIENT_ID=ddf97d-xxx-xxxx-e92920
+SAML2_CLIENT_SECRET=f021f63bxxxx71f2c6a90c9b
+SAML2_REDIRECT_URI=http://localhost:3000/oauth/callback
```
humphd: This should be in the `env.sample`

src/backend/login/usingPassport.js
```diff
@@ -1,36 +1,36 @@
...
 */
 
-// Configure the OpenId Connect Strategy
-// with credentials obtained from OneLogin
-const { OIDC_BASE_URI, OIDC_CLIENT_ID, OIDC_CLIENT_SECRET, OIDC_REDIRECT_URI } = process.env;
+const { SAML2_BASE_URI, SAML2_REDIRECT_URI } = process.env;
```
humphd: Do you need to check for, and log errors if any of these are missing or in the wrong format?  Same for all the other env variables below.  It seems like getting this right could be tricky for someone without full knowledge, can you help a dev get this right via docs, error logs, errors, examples in the `env.sample`, etc?

src/backend/login/usingPassport.js
```diff
@@ -1,36 +1,36 @@
...
-    }
-  )
+const samlStrategy = new saml.Strategy(
+  {
+    callbackUrl: SAML2_REDIRECT_URI,
+    entryPoint: SAML2_BASE_URI,
+    issuer: 'saml-poc',
+    identifierFormat: null,
+    decryptionPvk: fs.readFileSync(path.join(__dirname, '/certs/certskey.pem'), 'utf8'),
+    privateCert: fs.readFileSync(path.join(__dirname, '/certs/certskey.pem'), 'utf8'),
```
humphd: Don't use `readFileSync`, since it blocks.  It would be better to do this async, and use the errors that might happen if files are missing to give good errors to the dev, and shutdown the app.

src/backend/login/usingPassport.js
```diff
@@ -1,36 +1,36 @@
...
+  },
+  function(profile, done) {
+    return done(null, profile);
+  }
 );
+
+// Configure the Saml Strategy
+// with credentials obtained from Saml2 Docker
+// eslint-disable-next-line no-unused-expressions
+// eslint-disable-next-line no-sequences
```
humphd: Do we have to disable eslint?  Why not just:

```js
module.exports = new saml.Strategy(...);
```

src/backend/web/app.js
```diff
@@ -1,16 +1,73 @@
 const express = require('express');
 const healthcheck = require('express-healthcheck');
-
+const session = require('express-session');
+const bodyParser = require('body-parser');
+const cookieParser = require('cookie-parser');
+const passport = require('passport');
 const logger = require('../utils/logger');
 const router = require('./routes');
+const { samlStrategy } = require('../login/usingPassport');
```
humphd: If the only thing you're going to return from this module is your `samlStrategy`, you might as well export it directly vs. as a property of the exports object.

src/backend/web/app.js
```diff
@@ -1,16 +1,73 @@
...
 
 const app = express();
 
+app.use(cookieParser());
+
+app.use(bodyParser.urlencoded({ extended: false }));
+app.use(bodyParser.json());
+app.use(session({ secret: 'secret', resave: false, saveUninitialized: true }));
```
humphd: `secret: 'secret'`, should that be an env variable too?

src/backend/web/app.js
```diff
@@ -1,16 +1,73 @@
...
+app.use(passport.session({}));
+
 app.set('logger', logger);
 
 app.use(logger);
 app.use('/health', healthcheck());
 
 app.use('/', router);
 
+app.get(
```
humphd: This likely needs to follow the pattern all the other routes are using vs. doing it directly in here.  cc @miggs125.

src/backend/web/app.js
```diff
@@ -1,16 +1,73 @@
...
+  '/login',
+  function(req, res, next) {
+    console.log('-----------------------------');
```
humphd: We use `logger.js` in `utils` for all logging.  Also, you probably don't need to log here at all, since our express logger will log this route being hit automatically.

src/backend/web/app.js
```diff
@@ -1,16 +1,73 @@
...
+    console.log('/Start login handler');
+    next();
+  },
+  passport.authenticate('samlStrategy')
+);
+
+app.post(
```
humphd: Same things for this, and all routes below: do them like all our other routes using express Router, and get rid of the console.logs, replacing with logger.info/debug calls if necessary.

src/backend/web/app.js
```diff
@@ -1,16 +1,73 @@
...
+  '/login/callback',
+  function(req, res, next) {
+    console.log('-----------------------------');
+    console.log('/Start login callback ');
+    next();
+  },
+  passport.authenticate('samlStrategy'),
```
humphd: Can you add comments above all these routes too, explaining how and when they are used in the flow of other routes?

src/backend/login/usingPassport.js
Grommers00: It depends on how the project is organized, as this only pertains to logging in, I felt it was better to have a seperate folder to keep problems. Isolated, however I can move it if you think it's better organized a different way?

Grommers00:
I'm going to have to close this issue

PR #449, opened 2019-12-05
Generate SSL Certs in CI builds for Linux, macOS

humphd:
I tweaked the script to generate SSL certs a bit, and tried to get them hooked into CI, which @Grommers00 needs for #448.  Not sure if this will work, I want to test on CI.

NOTE: we don't have a script for Windows, and we'll need one, or our Windows builds will fail for sure.


humphd:
OK, this works for everything but Windows

humphd:
Filed #451 to get the Windows script we need for this.

PR #450, opened 2019-12-05
Closes #431: Add tests for HTTP /posts route

### LINKED ISSUE #431: Add tests for HTTP /posts route
We recently added a `/posts` route to our web server.  It currently works by taking no arguments and returning a JSON Array of objects with `id` and `url` for all posts in the database.  We need tests for:

- 500 case: possibly hard to test, we need to do the right thing if the database goes down
- 200 case: expect this to return array with correct number and type of elements.

To test the latter case, you would need to add something to the database using `storage.js` and `addPost()`, then call this route and make sure it can return it. 

NOTE: there are a few things that might change related to this code in other PRs, so pay attention to related work.

manekenpix:
Initial set of tests for `/posts` route.

It checks for:
-  `/posts` should return 30 items by default
- `/posts?per_page=15` should return 15 items
- `/posts?per_page=150` should return 100 items (100 is our max value)

This can be expanded in the future adding tests for paging and count.

### DIFF
--- a/test/posts.test.js
+++ b/test/posts.test.js
@@ -7,9 +7,9 @@ describe('test /posts endpoint', async () => {
   const requestedItems = 50;
   const maxItems = 100;
 
-  const posts = [...Array(150).keys()].map(index => {
+  const posts = [...Array(150).keys()].map(guid => {
     return {
-      guid: `${index}`,
+      guid: `${guid}`,
       author: 'foo',
       title: 'foo',
       link: 'foo',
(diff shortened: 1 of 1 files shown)


test/posts.test.js
```diff
@@ -0,0 +1,52 @@
+const request = require('supertest');
+const app = require('../src/backend/web/app');
+const { addPost } = require('../src/backend/utils/storage');
+
+describe('test /posts endpoint', async () => {
+  const defaultItems = 30;
+  const requestedItems = 50;
+  const maxItems = 100;
+
+  const posts = [...Array(150).keys()].map(guid => {
```
humphd: This is really smart.  Later on we should extract this out into a testing lib we can use in other tests.

PR #465, opened 2019-12-05
Fix Issue 446

MusaBajwa:
I added a route for '/count'. Please let me know if I have to make changes or if I did something wrong.


src/backend/web/routes/posts.js
```diff
@@ -1,6 +1,7 @@
 const express = require('express');
 const { getPosts } = require('../../utils/storage');
 const { logger } = require('../../utils/logger');
+const { count } = require('../../utils/storage.js');
```
humphd: `const { getPostsCount } = require('../../utils/storage.js');`, there's nothing called `count` to get here.

src/backend/web/routes/posts.js
```diff
@@ -31,4 +32,8 @@ posts.get('/', async (req, res) => {
   );
 });
 
+posts.get('/count', async (req, res) => {
+  res.json(count.getPostsCount());
```
humphd: The call to `getPostsCount` needs to be `await`ed, since it's async, and you'll need to wrap in a `try/catch` and give back a 500 if the redis call fails.

MusaBajwa: Ok thanks for the feedback. I tried to fix this in the following commit.

MusaBajwa:
@manekenpix Could I get some feedback on my recent changes, if possible.

src/backend/web/routes/posts.js
```diff
@@ -33,7 +33,14 @@ posts.get('/', async (req, res) => {
 });
 
 posts.get('/count', async (req, res) => {
-  res.json(count.getPostsCount());
+  try {
+    await res.json(getPostsCount());
```
humphd: Are you testing your code locally?  This won't work.  You need to `await` the portion of this that is asynchronous (i.e., returns a Promise).  The call to `res.json(...)` is synchronous, and will block while it runs.  However, the call to `getPostsCount()` is promise-based, and needs to go and talk to the database without blocking the main thread.  As such, you need to do something like:

```js
try {
  res.json(await getPostsCount());
```

You could also break it up a bit if you want:

```js
try {
  const count = await getPostsCount();
  res.json(count);
```

MusaBajwa: Ok thanks will make the changes right now, yes I am running the code locally (testing using npm test) but it does not show any errors with posts.js.

humphd: Nothing in our tests is calling this route.  You should add a test for this too, which will make it a lot easier to understand if it's working or not.

In our posts test file, you want something like:

```js
  it('returns correct number of posts', async () => {
    const count = await request(app).get('/posts/count');

    expect(res.status).toEqual(200);
    expect(res.get('Content-type')).toContain('application/json');
    expect(res.body).toBe(count);
  });
```

(untested code, fix as makes sense...)

MusaBajwa: Ok I am going to add that now thanks David.

MusaBajwa:
So I tested using the tester that you gave me and I made the changes necessary. It gave me no errors.
image

PR #471, opened 2019-12-06
Fix #464: fix formatting in CONTRIBUTING.md

humphd:
We needed to run prettier over the project, since it looks like the `CONTRIBUTING.md` file got missed.  I did that, and also found some issues with our ignore patterns in the process.

@drwm-base can you take a look at what I did in `CONTRIBUTING.md`?  I might have gotten it wrong.

Also, we should add a step to our CI to test that prettier was run using the `--check` option.

PR #472, opened 2019-12-06
Add test-ci script to also check prettier on CI

humphd:
We've had a few PRs recently that missed running Prettier (probably it wasn't installed via Husky in their environment).  This adds another script to `package.json` to run [`prettier --check`](https://prettier.io/docs/en/cli.html#--check), and use it in CI to help catch PRs with missing prettier formatting.

On development machines, running `npm test` is fine, since we don't want to error if files aren't committed yet (i.e., husky+prettier haven't run yet).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,4 +16,4 @@ jobs:
           command: npm install
       - run:
           name: Execute tests
-          command: npm test
+          command: npm run test-ci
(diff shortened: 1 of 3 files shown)


humphd:
You can see it catching the bug in https://github.com/Seneca-CDOT/telescope/pull/471 now.  These failures are what we'd expect until that lands.

PR #474, opened 2019-12-06
Closes #441: Update /posts route to support paging

### LINKED ISSUE #441: Update /posts route to support paging
We should [follow GitHub's lead](https://developer.github.com/v3/#pagination), they do pagination well.  Specifically:

- `/posts?page=2` should return 30 items, but the second 30 items in our set
- `/posts?page=3&per_page=10` should return 10 items, but the third page of 10 items

We should also add a `Link` header with URLs for `next`, `prev`, `first`, `last`.  These would allow the browser to navigate without having to construct URLs for the other pages.

cc @lucacataldo, @manekenpix

manekenpix:
Paging added to `/posts`.
Using `page` now it's possible to request a set of posts with `per_page` number of posts in it, or `default` number of posts if `per_page` is not present.

Examples:
- `/posts` should return 30 items
- `/posts?page=2` should return 30 items, but the second 30 items in our set
- `/posts?per_page=15` should return 15 items
- `/posts?page=3&per_page=15` should return 15 items, but the third page of 15 items

`Link` headers have been added with URLs for `next`, `prev`, `first`, `last`.
These will allow the browser to navigate without having to construct URLs for the other pages.
Example:
- for `/posts?page3=per_page=15`
  - `</posts?per_page=15&page=4>; rel="next"`
  - `</posts?per_page=15&page=2>; rel="prev"`
  - `</posts?per_page=15&page=1>; rel="first"`
  - `</posts?per_page=15&page=268>; rel="last"`

Also, changed tests in `storage.tests.js` to split feed tests and posts tests in 2 different sections.

### DIFF
--- a/src/backend/utils/storage.js
+++ b/src/backend/utils/storage.js
@@ -59,12 +59,13 @@ module.exports = {
       .exec();
   },
 
-  getPosts: (from, to) =>
-    /**
-     * 'counter' needs -1 because 'zrevrange()' includes the element at index 'counter'
-     * in the array it returns
-     */
-    redis.zrevrange(POSTS, from, to - 1),
+  /**
+   * Gets an array of guids from redis
+   * @param from lower index
+   * @param to higher index, it needs -1 because redis includes the element at this index in the returned array
+   * @return Array of guids
+   */
+  getPosts: (from, to) => redis.zrevrange(POSTS, from, to - 1),
 
   getPostsCount: () => redis.zcard(POSTS),
 
--- a/src/backend/utils/storage.js
+++ b/src/backend/utils/storage.js
@@ -59,12 +59,12 @@ module.exports = {
       .exec();
   },
 
-  getPosts: counter =>
+  getPosts: (from, to) =>
     /**
      * 'counter' needs -1 because 'zrevrange()' includes the element at index 'counter'
      * in the array it returns
      */
-    redis.zrevrange(POSTS, 0, counter - 1),
+    redis.zrevrange(POSTS, from, to - 1),
 
   getPostsCount: () => redis.zcard(POSTS),
 
--- a/src/backend/web/routes/posts.js
+++ b/src/backend/web/routes/posts.js
@@ -6,14 +6,15 @@ const { getPostsCount } = require('../../utils/storage');
 const posts = express.Router();
 
 posts.get('/', async (req, res) => {
+  const defaultNumberOfPosts = 30;
+  const capNumOfPosts = 100;
+  const page = req.query.page || 1;
+
   let redisGuids;
   let perPage;
   let postsInDB;
   let from;
   let to;
-  const defaultNumberOfPosts = 30;
-  const capNumOfPosts = 100;
-  const page = req.query.page || 1;
 
   /**
    * Set 'perPage' to a value under within the limits or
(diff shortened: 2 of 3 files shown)


src/backend/utils/storage.js
```diff
@@ -59,12 +59,12 @@ module.exports = {
       .exec();
   },
 
-  getPosts: counter =>
+  getPosts: (from, to) =>
```
humphd: Can you add a JSDOC style comment above this, documenting the expected type/values for `from` and `to`?  People won't know if they are dates, scalar numbers, etc.

src/backend/web/routes/posts.js
```diff
@@ -1,18 +1,40 @@
...
 
 const posts = express.Router();
 
 posts.get('/', async (req, res) => {
-  let redisGuids;
   const defaultNumberOfPosts = 30;
   const capNumOfPosts = 100;
+  const page = req.query.page || 1;
+
+  let redisGuids;
```
humphd: Let's get rid of the `redis` prefix, we don't need to leak implementation details.

src/backend/web/routes/posts.js
```diff
@@ -1,18 +1,40 @@
...
+  let perPage;
+  let postsInDB;
```
humphd: `postsCount`

src/backend/web/routes/posts.js
```diff
@@ -21,6 +43,21 @@ posts.get('/', async (req, res) => {
     return;
   }
 
+  /**
+   * Add prev, next, first and last in the response's header.
+   * It's been implemented to work circularly.
+   * Once reached the last set of posts, 'next' points at the first set.
+   * Same case with 'prev' and the first set of posts.
+   */
+  const nextPage = to >= postsInDB ? 1 : parseInt(page, 10) + 1;
```
humphd: Let's move this `parseInt()` call earlier to line 10 when you define it, so that it's always a number in the lines above, and here.

test/storage.test.js
```diff
@@ -1,99 +1,107 @@
...
+  getFeeds,
+  getFeedsCount,
+  addPost,
+  getPost,
+  getPosts,
+  getPostsCount,
+} = require('../src/backend/utils/storage');
 
-describe('Tests for storage', () => {
+describe('Tests for feeds', () => {
```
humphd: `'Storage tests for feeds'`

test/storage.test.js
```diff
@@ -1,99 +1,107 @@
...
 
-  const post = {
-    guid: 'tag:blogger.com,1999:blog-7100164112302197371.post-522285656016053350',
-    author: 'Neil David',
-    title: 'My First Blog',
-    link: 'http://nadavid.blogspot.com/2008/09/my-first-blog.html',
-    content:
-      "I have never done this before, so let's make this short. This post is just a test on how my blog post will look like.",
-    text: 'post one text',
+describe('Tests for posts', () => {
```
humphd: `'Storage tests for posts'`

test/storage.test.js
```diff
@@ -1,99 +1,107 @@
...
+  beforeAll(() => {
+    Promise.all([testPost, testPost2, testPost3].map(post => addPost(post)));
+  });
+
   it('should allow retrieving a post by guid after inserting', async () => {
-    await storage.addPost(post);
-    const result = await storage.getPost(post.guid);
-    expect(result.link).toEqual(post.link);
+    const result = await getPost(testPost.guid);
+    expect(result.link).toEqual(testPost.link);
```
humphd: All `link` properties above are `'foo'`, so this probably isn't a useful test.  Let's compare `result.guid`, since they are always different.

humphd:
Filed #483 to add support for this to the frontend.

humphd:
Great stuff.

Issue #313, opened 2019-11-21
Improve test coverage for email-sender.js


neilong31:
Currently the test for email-sender covers 14% of the statements. The current status of the test can be seen at #268

RyanWils:
Hi Neil,

If you want to test nodemailer with an SMTP server you can use the following, this is the one I've been testing it with. 
![image](https://user-images.githubusercontent.com/49354833/69652659-02413280-1040-11ea-88a8-48859d0ac743.png)
```
let transporter = nodemailer.createTransport({
        host: 'smtp.googlemail.com', // Gmail Host
        port: 465, // Port
        secure: true, // this is true as port is 465
        auth: {
            user: 'GMAIL_USERNAME', //Gmail username
            pass: 'GMAIL_PASSWORD' // Gmail password
        }
    });
```

PR #476, opened 2019-12-06
Restructured emailsender and added tests

neilong31:
Helps fix issue #313 

I restructured some of the code in email sender so the logic is broken up into more reusable chunks and easier to test.

### DIFF
--- a/src/backend/utils/email-sender.js
+++ b/src/backend/utils/email-sender.js
@@ -1,3 +1,4 @@
+require('../lib/config');
 const nodemailer = require('nodemailer');
 const { logger } = require('./logger');
 
--- a/src/backend/utils/email-sender.js
+++ b/src/backend/utils/email-sender.js
@@ -15,58 +15,87 @@ const log = logger.child({ module: 'email-sender' });
  message - '<p>An Error has occurred <br/>Error Message:</p>'
 */
 
+// Creates and returns a transporter using the passed in parameters
+exports.createTransporter = function(hosted, portNum, isSecure, username, password) {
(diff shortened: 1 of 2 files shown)


src/backend/utils/email-sender.js
```diff
@@ -15,58 +15,87 @@ const log = logger.child({ module: 'email-sender' });
...
+
+// Sends a message using the passed in parameters
+exports.sendMessage = async function(receipiants, subjectMessage, message) {
+  return new Promise((resolve, reject) => {
+    const transporter = this.createTransporter(
+      process.env.NODEMAILER_SERVER,
+      2222,
+      false,
+      process.env.NODEMAILER_USERNAME,
+      process.env.NODEMAILER_PASSWORD
```
RyanWils: I was going to create an issue for this, but its a one liner. require('../lib/config'); needs to be included at the top to access the environment variables. I think its fine to include it in this PR since your restructuring the entire thing

RyanWils:
It looks good to me, did you produce the coverage report? Does this cover all the functions for the coverage reports for the email-sender.js file?

src/backend/utils/email-sender.js
neilong31: It's been added. Do you mind approving the changes then?

RyanWils: It looks good to me, did you produce the coverage report? Does this cover all the functions for the coverage reports for the email-sender.js file?

neilong31:
Sorry yeah, I forgot to put it up. It's not 100%, but its much much better now.
image

Issue #467, opened 2019-12-06
Content scrolls beyond post title header


humphd:
There's a width/scrolling bug with how the post headers are done, allowing content to scroll to the left/right of it.  The header should extend 100% of the width of the content container, such that the post itself scrolls beneath it.

![scroll-bug](https://user-images.githubusercontent.com/427398/70288648-3f27ca80-17a0-11ea-80a5-cf1df40c02d9.gif)

asyam1:
I would like to work on it

miggs125:
Hi @asyam1, we are currently fixing this in #531 so this issue will be closed. Feel free to contribute there :)

PR #477, opened 2019-12-06
Fixed issue #467, content scrolls under the header

asyam1:
Added 2 CSS properties for ul and ol which makes them shift slightly to the right and under the header.


agarcia-caicedo:
We are doing a complete overhaul of the UI, so this issue will be closed.

PR #480, opened 2019-12-06
Fix #312 updated docs for workflow with git/GitHub

### LINKED ISSUE #312: Update documentation regarding rebasing
We should make it clear how to complete rebasing, including squashing/fixup. Current documentation does not mention anything about squashing or fixup.

varshannagarajan:


docs/CONTRIBUTING.md
```diff
@@ -193,14 +199,44 @@ When working on fixing bugs, please use the following workflow:
    npm test
    ...manually fix any errors yourself, rerunning npm test each time to confirm
    ```
+
+### Squashing Commits
```
neilong31: Maybe we can also show people how to use fixup if they dont want there commit messages?

varshannagarajan: I think the instructions in the `git-rebase-todo` are pretty clear on how to use fix up. I am not sure if its necessary in here as well. Also the example in the step below shows fix up. Thank you for the feedback!

neilong31: That's fair, but some people don't always read everything so it would be good to have it here as well. I see you are using fixup in a squash example, didn't even catch that. Maybe we should change it to a squash since its a squash example?

Issue #354, opened 2019-11-26
Updating the login button to point to the saml2 server


Grommers00:
Once issue #287 is set up, we will need to update the login button to point towards the saml2 server. 

This can be set up to be hard coded at first, but will eventually need to take use of the process.env variables.

agarcia-caicedo:
I'll work on it

Grommers00:
@agarcia-caicedo as soon as issue #319 lands, we can continue with this process. 

https://medium.com/disney-streaming/setup-a-single-sign-on-saml-test-environment-with-docker-and-nodejs-c53fc1a984c9

We will keep going through this.

agarcia-caicedo:
e6224e337d74c7b2357e2af96183849e3092097e

PR #481, opened 2019-12-06
#354 Updated the login button to work for the SAML server

agarcia-caicedo:
Once issue #424 lands, this commit will allow a user wanting to login to be redirected to SAML service provider site


src/frontend/index.html
```diff
@@ -31,8 +31,8 @@
         <!-- NAV LINKS -->
 
         <nav>
-          <a href="login.html">
-            <button class="nav-button">
+          <a href="/login">
```
humphd: Question about this: does it not need to point at the URL for the PHP server running in the docker container and/or whatever URL Seneca will give us in production?

That is to say, should this not somehow be read from the environment?  Or am I wrong?  What lives at `/login`?

PR #486, opened 2019-12-07
Fix#468 add test cases for sanitize-html

haichuan0424:
Added more test cases for sanitize-html.

### DIFF
--- a/test/sanitize-html.test.js
+++ b/test/sanitize-html.test.js
@@ -5,34 +5,26 @@ describe('Sanitize HTML', () => {
     const data = sanitizeHTML('<img src="x" onerror="alert(1)" onload="alert(1)" />');
     expect(data).toBe('<img src="x" />');
   });
-});
 
-describe('Sanitize HTML', () => {
   test('<a><img> should work, but inline js should not', () => {
     const data = sanitizeHTML(
       '<a href="https://www.telescope.com"><img border="0" alt="W3Schools" src="logo.gif" width="100" height="100"></a>'
     );
     expect(data).toBe('<a href="https://www.telescope.com"><img src="logo.gif" /></a>');
   });
-});
 
-describe('Sanitize HTML', () => {
-  test('<p> should work with inline style', () => {
+  test('<p> with inline style, sanitize strips inline style', () => {
     const data = sanitizeHTML('<p style="color:blue;font-size:46px;">Here is color blue</p>');
     expect(data).toBe('<p>Here is color blue</p>');
   });
-});
 
-describe('Sanitize HTML', () => {
-  test('<p> and <strong> should work if there are two tab in the same line', () => {
+  test('<p> with <strong> should work with inline style stripped', () => {
     const data = sanitizeHTML(
       '<p style="color:blue;font-size:46px;">Here is color <strong>blue</strong></p>'
     );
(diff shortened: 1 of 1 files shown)


test/sanitize-html.test.js
```diff
@@ -6,3 +6,57 @@ describe('Sanitize HTML', () => {
...
+describe('Sanitize HTML', () => {
+  test('<a><img> should work, but inline js should not', () => {
+    const data = sanitizeHTML(
+      '<a href="https://www.telescope.com"><img border="0" alt="W3Schools" src="logo.gif" width="100" height="100"></a>'
+    );
+    expect(data).toBe('<a href="https://www.telescope.com"><img src="logo.gif" /></a>');
+  });
+});
+
+describe('Sanitize HTML', () => {
```
humphd: just do 1 `describe('Sanitize HTML', () => {` wrapped around all these tests, not one per test.

test/sanitize-html.test.js
```diff
@@ -6,3 +6,57 @@ describe('Sanitize HTML', () => {
...
+  test('<p> should work with inline style', () => {
```
humphd: You might want to update your comment to indicate that it will strip inline style.

test/sanitize-html.test.js
```diff
@@ -6,3 +6,57 @@ describe('Sanitize HTML', () => {
...
+    const data = sanitizeHTML('<p style="color:blue;font-size:46px;">Here is color blue</p>');
+    expect(data).toBe('<p>Here is color blue</p>');
+  });
+});
+
+describe('Sanitize HTML', () => {
+  test('<p> and <strong> should work if there are two tab in the same line', () => {
```
humphd: What does `two tab in the same line` mean?

test/sanitize-html.test.js
```diff
@@ -6,3 +6,57 @@ describe('Sanitize HTML', () => {
...
+    const data = sanitizeHTML(
+      '<p style="color:blue;font-size:46px;">Here is color <strong>blue</strong></p>'
+    );
+    expect(data).toBe('<p>Here is color <strong>blue</strong></p>');
+  });
+});
+
+describe('Sanitize HTML', () => {
+  test('<figure> with <img> should remove <figure>', () => {
```
humphd: I wonder if we should add `<figure>` to the list of allowed elements?  Seems useful.  Can you do that?

test/sanitize-html.test.js
```diff
@@ -6,3 +6,57 @@ describe('Sanitize HTML', () => {
...
+      '<figure class="wp-block-image size-large"><img src="https://paulopensourceblog.files.wordpress.com/2019/12/image-3.png?w=1024" alt="" class="wp-image-2226"></figure>'
+    );
+    expect(data).toBe(
+      '<img src="https://paulopensourceblog.files.wordpress.com/2019/12/image-3.png?w=1024" />'
+    );
+  });
+});
+
+describe('Sanitize HTML', () => {
+  test('<pre> should work with inline style', () => {
```
humphd: Update comment to indicate that it will strip inline styles.

test/sanitize-html.test.js
KarlaChen: ok, i will file a new issue to add ```<figure>```

PR #488, opened 2019-12-07
Fix #388: reuse the same puppeteer browser and page instance for all calls of textParser.parse()

### LINKED ISSUE #388: Reuse Puppeteer browser and page for text parser
In #383 we are moving to use Puppeteer.  This uses a real Chromium browser to render content in a full DOM.  We can reuse the `browser` and `page` objects, so we don't have more than one instance.  We just have to use a Promise chain pattern, see https://github.com/puppeteer/puppeteer/issues/486#issuecomment-324221441:

```js
const browser = await puppeteer.launch();
const page = await browser.newPage();
let chain = Promise.resolve();
// Request handler 
app.get('/pdf', (req, res) => {
  // Handle request generation in the end of the chain
  chain = chain.then(async () => { 
    await page.goto(req.query.url);
    const pdf = await page.pdf();
    res.send(pdf);
  });
});
```

Depends on #383.

ragnarokatz:
**status update: ready for review**

**Types of change:** enhancement

Reference issue #388 

This PR does the following:
1. reuse the same puppeteer `browser` and `page` instance for all function calls of `textParser.parse()`

2. due to bugs in using `jest` to test `puppeteer`, which results in globally defined variables `browser` and `page` to become `undefined`, resulting in calling `browser.close` later to clean up the resources to fail. Therefore, the package `jest-environment-puppeteer` is used for testing. The setup for this module is here: https://www.npmjs.com/package/jest-environment-puppeteer, which outlines the necessary changes to configurations of jest inside `package.json`.

3. in order to support testing purposes, parse now takes in an optional `page` argument, which normally does not need to be passed in during regular usage.

4. during testing with `jest`, lazy initialization will cause the `browser` and `page` resources to become `undefined` after their creation. This prevents these resources being freed up, therefore jest hangs and returns the error message `jest did not exit one second after completion, check to see if you have asynchronous processes running`. Therefore, `text-parser.js` now needs to be explicitly initialized and shut down. The initialization logic is currently in `index.js`, shutdown logic is in `shutdown.js`.

5. because
```
globals: {
    page: true,
},
``` 
was added to `.eslintrc.js` in order to make `jest-environment-puppeteer` pass linting phase. The instructions for doing this are here: https://github.com/smooth-code/jest-puppeteer, under section **Configure ESLint**

6. because node keeps on throwing the following:
```
UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: woops
DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```
which apparently is due to exceptions thrown or rejected promises in an async function that are not being handled properly with `catch()`, therefore, all `async` functions in `text-parser.js` has been switched to using promise chaining with proper `catch()`
```
func = () => return new Promise((resolve, reject) => {
   resolve();
   reject();
...
});

func.then().catch();
```
instead of
```
async () => {
   await blah();
...
}
```

## Checklist

- [ ] **Documentation**: This PR includes updated/added documentation to user exposed functionalty or configuration variables are added/changed or an explanation of why it does not(if applicable)

~~Switch from `puppeteer` to using `puppeteer-cluster` for `text-parser.js`~~

~~This PR does the following:~~
~~1. add `puppeteer-cluster` node module to list of dependencies in `package.json`~~
~~2. switch `text-parser.js` to using `puppeteer-cluster` instead of `puppeteer`~~
~~3. The same cluster is being used for all calls of `parse()`~~
~~4. implement proper graceful shutdown in `shutdown.js` via `shutdownTextParser()`~~
~~5. update tests and other usages of `text-parser` to call `textParser.parse()`~~
~~6. added cases of failures in tests as well~~

### DIFF
--- a/src/backend/utils/text-parser.js
+++ b/src/backend/utils/text-parser.js
@@ -1,20 +1,65 @@
 const puppeteer = require('puppeteer');
+const { logger } = require('./logger');
 
-module.exports = async function(htmlFragment) {
-  let browser;
-  try {
-    browser = await puppeteer.launch();
-    const page = await browser.newPage();
+let browser;
+let page;
 
-    await page.setContent(htmlFragment, {
+module.exports.initialize = function init() {
+  return new Promise((resolve, reject) => {
+    puppeteer
+      .launch()
+      .then(b => {
+        browser = b;
+        browser
+          .newPage()
+          .then(p => {
+            page = p;
+            resolve();
+          })
+          .catch(err => {
+            logger.error({ err }, 'Failed to create new page');
+            reject(err);
+          });
+      })
+      .catch(err => {
+        logger.error({ err }, 'Failed to launch new browser');
+        reject(err);
+      });
+  });
+};
+
+module.exports.parse = function(htmlFragment, p = page) {
+  return new Promise((resolve, reject) => {
+    p.setContent(htmlFragment, {
       waitUntil: 'domcontentloaded',
-    });
+    })
+      .then(() => {
+        p.evaluate('document.body.innerText')
+          .then(result => {
+            resolve(result);
+          })
+          .catch(err => {
+            logger.error({ err }, 'Failed to evaluate html document');
+            reject(err);
+          });
+      })
+      .catch(err => {
+        logger.error({ err }, 'Failed to set html content on page');
+        reject(err);
+      });
+  });
+};
 
-    const result = await page.evaluate('document.body.innerText');
-    return result;
-  } finally {
-    if (browser) {
-      await browser.close();
-    }
-  }
+module.exports.close = function() {
+  return new Promise((resolve, reject) => {
+    browser
+      .close()
+      .then(() => {
+        resolve();
+      })
+      .catch(err => {
+        logger.error({ err }, 'Failed to gracefully close browser instance');
+        reject(err);
+      });
+  });
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -8,6 +8,9 @@ module.exports = {
     node: true,
     es2017: true,
   },
+  globals: {
+    page: true,
+  },
   rules: {
     'prettier/prettier': ['error'],
 
--- a/package.json
+++ b/package.json
@@ -4,7 +4,9 @@
   "description": "A tool for tracking blogs in orbit around Seneca's open source involvement",
   "jest": {
     "collectCoverage": true,
-    "testEnvironment": "node",
+    "globalSetup": "jest-environment-puppeteer/setup",
+    "globalTeardown": "jest-environment-puppeteer/teardown",
+    "testEnvironment": "jest-environment-puppeteer",
     "coveragePathIgnorePatterns": [
       "/node_modules/"
     ],
@@ -58,6 +60,7 @@
     "get-urls": "9.2.0",
     "ioredis": "4.14.1",
     "ioredis-mock": "4.18.4",
+    "jest-environment-puppeteer": "4.3.0",
     "jsdom": "15.2.1",
     "minimist": "1.2.0",
     "node-fetch": "2.6.0",
@@ -89,6 +92,7 @@
     "husky": "3.1.0",
     "jest": "24.9.0",
     "jest-fetch-mock": "2.1.2",
+    "jest-puppeteer": "4.3.0",
     "lighthouse": "5.6.0",
     "nock": "11.7.0",
     "npm-run-all": "4.1.5",
--- a/src/backend/index.js
+++ b/src/backend/index.js
@@ -1,6 +1,7 @@
 require('./lib/config.js');
 const feedQueue = require('./feed/queue');
 const feedWorker = require('./feed/worker');
+const textParser = require('./utils/text-parser');
 const { logger: log } = require('./utils/logger');
 const wikiFeed = require('./utils/wiki-feed-parser');
 const shutdown = require('./lib/shutdown');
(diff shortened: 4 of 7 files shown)


src/backend/utils/text-parser.js
```diff
@@ -1,20 +1,28 @@
...
-    const page = await browser.newPage();
+let isInitialized = false;
+let cluster;
 
-    await page.setContent(htmlFragment, {
-      waitUntil: 'domcontentloaded',
+module.exports.parse = async function(htmlFragment) {
+  if (!isInitialized) {
+    cluster = await Cluster.launch({
+      concurrency: Cluster.CONCURRENCY_CONTEXT,
```
humphd: Should be safe to use `CONCURRENCY_PAGE`

src/backend/utils/text-parser.js
```diff
@@ -1,20 +1,28 @@
...
-  try {
-    browser = await puppeteer.launch();
-    const page = await browser.newPage();
+let isInitialized = false;
+let cluster;
 
-    await page.setContent(htmlFragment, {
-      waitUntil: 'domcontentloaded',
+module.exports.parse = async function(htmlFragment) {
+  if (!isInitialized) {
```
humphd: Flip this logic around, so you can return early:

```js
if(isInitialized) {
  return cluster.execute(htmlFragment);
}
// rest of code here...
```

humphd: Also, do we not need to `await cluster.execute(...)`?  It returns a Promise.

src/backend/utils/text-parser.js
```diff
@@ -1,20 +1,28 @@
...
+    cluster = await Cluster.launch({
+      concurrency: Cluster.CONCURRENCY_CONTEXT,
+      maxConcurrency: 2,
```
humphd: Probably worth throwing this in the env, so it can be set to something else.  I'd also suggest leaving it at the default for dev, 1.

src/backend/utils/text-parser.js
ragnarokatz: will be fixed shortly. same fix will also be applied to `await page.evaluate`

src/backend/utils/text-parser.js
ragnarokatz: will be fixed soon

src/backend/utils/text-parser.js
ragnarokatz: i'm running into a weird bug here. if I do `maxCurrency: 1`, then the tests will pass. However, if I do `maxConcurrency: process.env.TEXT_PARSER_MAX_CONCURRENCY` (I have the variable setup in `.env`, then running `npm test -- test/text-parser.test.js` will just hang, and after a bit it says:
```
Timeout - Async callback was not invoked within the 5000ms timeout specified by jest.setTimeout.Error: Timeout - Async callback was not invoked within the 5000ms timeout specified by jest.setTimeout.

      at mapper (node_modules/jest-jasmine2/build/queueRunner.js:25:45)
```
followed by
```
Jest did not exit one second after the test run has completed.

This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```

I think this is due to an underlying issue somewhere... the tests didn't pass on travis CI either with the same sort of errors suggesting jest didn't cleanly due to async processes hanging.

humphd: On CI you'll need `process.env.TEXT_PARSER_MAX_CONCURRENCY || 1`, since we don't have an environment there.

ragnarokatz: oh, I see. thank you for pointing it out.

src/backend/utils/text-parser.js
```diff
@@ -1,20 +1,33 @@
...
-module.exports = async function(htmlFragment) {
-  let browser;
-  try {
-    browser = await puppeteer.launch();
-    const page = await browser.newPage();
+let isInitialized = false;
+let cluster;
 
-    await page.setContent(htmlFragment, {
+module.exports.parse = async function(htmlFragment) {
```
humphd: Let's not bother with `.parse` here.  Let's just expose the function directly.

ragnarokatz:
ugh, tests pass locally but not on travis CI
```
Jest did not exit one second after the test run has completed.

This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```

I'm trying to get the tests fail locally but to no avail.

src/backend/utils/text-parser.js
ragnarokatz: will I still be able to call `textParser.close()` if I do this?

humphd: Ah, I see.  OK, leave it like this.

ragnarokatz:
@humphd After some research concerning `jest` and `puppeteer`, it seems that globally defined `browser` and `page` cause trouble in testing with `jest`. these variables become `undefined` some time after parsing elements and before closing.

there seems to be solutions, such as `jest-environment-puppeteer:
https://www.npmjs.com/package/jest-environment-puppeteer

but the usage is arduous. I'll then have to manually modify `text-parser.js` to not initialize the instances of `browser` and `page` when it comes to testing, since `jest-environment-puppeteer` provides them globally thru:
```
{
  "globalSetup": "jest-environment-puppeteer/setup",
  "globalTeardown": "jest-environment-puppeteer/teardown",
  "testEnvironment": "jest-environment-puppeteer"
}
```

but then the test cases are limited too: I have to use the variables `browser` and `page` physically inside the test file itself, like:
```
  it('should display "google" text on page', async () => {
    const text = await page.evaluate(() => document.body.textContent)
    expect(text).toContain('google')
  })
})
```

if I just call the function inside `text-parser.js`, the `browser` and `page` inside there are still referring to the ones inside `text-parser.js`, which are `undefined` if I don't do the initialization.

ragnarokatz:
omg it works. The tests are passing on travis. I don't believe my eyes.

humphd:
I'm going to work on this, so I'll assign myself.  Thanks for starting on this.

humphd:
I'm going to close this in favour of #540. @giatuongtran9, @ragnarokatz, and others who have worked hard on the Puppeteer backend, thank you for all your research and work. It was really cool; but long term, the complexity of it isn't worth it for our use case.  Sometimes you have to "build one to throw away," and unfortunately, this is that.

PR #507, opened 2019-12-12
Issue 430: Added tests for the post/ route

### LINKED ISSUE #430: Add tests for HTTP /post route
We recently added a `/post` route to our web server.  It works by taking a Redis key (guid) for a post in the database, and returning a JSON object for that that post.  We need tests for:

- 404 case: if we give an id that doesn't exist in the database, it should return a 404
- 500 case: possibly hard to test, we need to do the right thing if the database goes down
- 200 case: given a key in the database, expect this to return it.
- Deal with encoded and non-encoded URI guids that contain path separators

To test the latter case, you would need to add something to the database using `storage.js` and `addPost()`, then call this route and make sure it can return it.

izhuravlev:
### Type of Change

Added tests for `post/` route

### Description

In this PR, I have implemented tests for `/post/:guid` route.

Tests cover the following cases:  

- `404` - a post with a passed `guid` not found;
- `200` - a post with a passed `guid` found and returned;
- both _encoded_ and _not encoded_ cases are tested

### Quality Check

While testing on my local machine with the latest version of the `master` branch, `inactive-blogs-filter` was failing - and so it did on my branch.

### DIFF
--- a/test/routes.post.test.js
+++ b/test/routes.post.test.js
@@ -1,29 +1,29 @@
 const request = require('supertest');
 const app = require('../src/backend/web/app');
-const { addPost } = require('../src/backend/utils/storage');
+const Post = require('../src/backend/post');
 
 describe('test /post responses', () => {
   // an array of keys.
-  const keys = [100, 101];
+  const existingGuid = 'existing-guid';
+  const missingGuid = 'missing-guid';
+  const existingNeedingEncoding = 'guid needing/encoding';
+  const missingNeedingEncoding = 'missing needing/encoding';
 
-  // an object to be added to the storage for testing purposes
-  const toBeAdded = {
-    guid: `${keys[0]}`,
-    author: 'foo',
-    title: 'foo',
-    link: 'foo',
-    content: 'foo',
-    text: 'foo',
-    updated: new Date('2009-09-07T22:23:00.544Z'),
-    published: new Date('2009-09-07T22:20:00.000Z'),
-    html: '',
-    url: 'foo',
-    site: 'foo',
-  };
+  // an object to be added for testing purposes
+  const addedPost1 = new Post(
+    'foo',
+    'foo',
+    '',
+    'foo',
+    new Date('2009-09-07T22:20:00.000Z'),
+    new Date('2009-09-07T22:23:00.000Z'),
+    'foo',
+    'foo',
+    existingGuid
+  );
 
   // an object, expected to be returned by a correct querry
-  const returned = {
-    guid: '100',
+  const receivedPost1 = {
     author: 'foo',
     title: 'foo',
     html: '',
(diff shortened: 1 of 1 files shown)


izhuravlev:
I have tested it with `npm run jest` before creating a PR, and working on fixing eslint issues right now.
Thank you @humphd for such a quick reply!

test/routes.post.test.js
```diff
@@ -0,0 +1,74 @@
+const request = require('supertest');
+const app = require('../src/backend/web/app');
+const { addPost } = require('../src/backend/utils/storage');
```
humphd: Instead of using `storage` and `addPost` directly, please use `post.js`, `new Post(...)` and `post.save()`.

test/routes.post.test.js
```diff
@@ -0,0 +1,74 @@
...
+
+describe('test /post responses', () => {
+  // an array of keys.
+  const keys = [100, 101];
+
+  // an object to be added to the storage for testing purposes
+  const toBeAdded = {
```
humphd: `const post1 = new Post({...})`

test/routes.post.test.js
```diff
@@ -0,0 +1,74 @@
...
+    text: 'foo',
+    updated: new Date('2009-09-07T22:23:00.544Z'),
+    published: new Date('2009-09-07T22:20:00.000Z'),
+    html: '',
+    url: 'foo',
+    site: 'foo',
+  };
+
+  // an object, expected to be returned by a correct querry
+  const returned = {
```
humphd: `const post2 = new Post({...})`

test/routes.post.test.js
```diff
@@ -0,0 +1,74 @@
...
+    text: 'foo',
+    published: '2009-09-07T22:20:00.000Z',
+    updated: '2009-09-07T22:23:00.000Z',
+    url: 'foo',
+    site: 'foo',
+  };
+
+  // add the post to the storage
+  beforeAll(() => {
+    addPost(toBeAdded);
```
humphd: The function you are calling is async, and needs to be done like so:

```js
beforeAll(async () => post1.save());
```

test/routes.post.test.js
```diff
@@ -0,0 +1,74 @@
...
+  });
+
+  // tests
+  it("pass a guid that doesn't exist", async () => {
+    const res = await request(app).get(`/post/${keys[1]}`);
+
+    expect(res.status).toEqual(404);
+    expect(res.get('Content-type')).toContain('application/json');
+    expect(res.body.message).toEqual('Post not found for id 101');
```
humphd: Don't do this check, since it might change, and the status code is the more correct thing here.

test/routes.post.test.js
```diff
@@ -0,0 +1,74 @@
...
+  });
+
+  it('pass a guid that exists', async () => {
+    const res = await request(app).get(`/post/${keys[0]}`);
```
humphd: Since you are using these guids to mean specific things, let's name them instead of using `keys[0]` and `keys[1]`.  At the top of this file:

```js
const exisingGuid = "existing-guid";
const missingGuid = "missing-guid;
```

And change all uses of `key[...]` below to one or the other of these.

test/routes.post.test.js
```diff
@@ -0,0 +1,74 @@
...
+
+    expect(res.status).toEqual(200);
+    expect(res.get('Content-type')).toContain('application/json');
+    expect(res.body).toEqual(returned);
+  });
+
+  it("pass an encoded guid that doesn't exist", async () => {
+    const res = await request(app).get(encodeURI(`/post/${keys[1]}`));
```
humphd: In order for this test to make sense, we need a guid that contains something that needs encoding.  An example would be `/` or ` ` (space):

```js
```js
const exisingGuid = "existing-guid";
const missingGuid = "missing-guid;
const guidNeedingEncoding = "guid needing/encoding";
```

humphd: Also note: you need to only encode the key:

```js
const res = await request(app).get(`/post/${encodeURIComponent(guidNeedingEncoding)}`);
```

humphd: We'd need to add that object to the database above in `beforeAll` as well.

test/routes.post.test.js
```diff
@@ -0,0 +1,74 @@
...
+
+    expect(res.status).toEqual(404);
+    expect(res.get('Content-type')).toContain('application/json');
+    expect(res.body.message).toEqual('Post not found for id 101');
+  });
+
+  it('pass an encoded guid that exists', async () => {
+    const res = await request(app).get(encodeURI(`/post/${keys[0]}`));
```
humphd: Similar comments here.

test/routes.post.test.js
```diff
@@ -0,0 +1,100 @@
...
+    '',
+    'foo',
+    new Date('2009-09-07T22:20:00.000Z'),
+    new Date('2009-09-07T22:23:00.000Z'),
+    'foo',
+    'foo',
+    existingGuid
+  );
+
+  // an object, expected to be returned by a correct querry
```
humphd: `query` (sp).

izhuravlev: Thank you very much @humphd! My bad with the spelling 😅 
Should I fix this?

humphd: Don't feel badly, I can't spell either.  I just use tools to hide it https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker :)

Yes, please fix this and I'll merge.

PR #518, opened 2019-12-15
Fix #510: Rename /post/:guid route to /posts/:guid

### LINKED ISSUE #510: Rename /post/:guid route to /posts/:guid
**What happened**:

We added both `/posts` and `/post/:guid` routes.

We should have done `/posts` (i.e., return all posts) and `/posts/:guid` (i.e., return the post with `guid`).

We'll have to update the implementation, tests, and frontend, which calls it.

ragnarokatz:
## Issue This PR Addresses
Fixes #510

## Type of Change

## Description

Rename route for getting a specific post by guid from post/:guid to posts/:guid, update all corresponding tests.

## Checklist


src/backend/web/routes/posts.js
```diff
@@ -80,4 +81,27 @@ posts.get('/count', async (req, res) => {
   }
 });
 
+// Allow for the id to be a URI, possibly containing path separators
+posts.get('/*', async (req, res) => {
```
humphd: Let me ask you a silly question: this won't confuse express?  That is `posts/` and `posts/....` are different enough?  Because we use `*` here vs. a named arg, in theory nothing (i.e., passing nothing after `/`) would match, no?

I feel like it must not be a problem, or your tests would be failing, but can you confirm and educate me?

ragnarokatz: I actually had no idea, so I did some research. according to this source: https://stackoverflow.com/questions/6161567/express-js-wildcard-routing-to-cover-everything-under-and-including-a-path
> If you look at line 331 of the connect router the * in a path is replaced with .+ so will match 1 or more characters

however, wildcard routing seems to be not recommended. Therefore, I've changed it to `posts.get('/:guid')`.
also from the same source:
> For those who are learning node/express (just like me): do not use wildcard routing if possible!
from another source: https://blog.praveen.science/wildcard-routing-is-an-anti-pattern/

src/backend/web/routes/posts.js
```diff
@@ -80,4 +81,27 @@ posts.get('/count', async (req, res) => {
   }
 });
 
+// Allow for the id to be a URI, possibly containing path separators
```
humphd: This comment is now incorrect.  We can't include path separators anymore if using `:guid`.  Please change it to say:

```
// The guid is likely a URI, and must be encoded by the client.
```

ragnarokatz: fixed.

test/posts.test.js
```diff
@@ -58,3 +62,104 @@ describe('test /posts endpoint', () => {
...
+    addedPost2.save();
+  });
+
+  // tests
+  it("pass a guid that doesn't exist", async () => {
+    const res = await request(app).get(`/posts/${missingGuid}`);
+
+    expect(res.status).toEqual(404);
+    expect(res.get('Content-type')).toContain('application/json');
+    expect(res.body instanceof Object).toBe(true);
```
humphd: This won't work, since an Array is an Object:

```
> [] instanceof Object
true
```

Instead of adding this particular test to everything below, just do one test that makes sure that it's *not* an Array.

ragnarokatz: fixed.

Issue #462, opened 2019-12-05
Integrate inactive blog url and include logic to filter inactive feeds


c3ho:
Currently we have work for the inactive blog, but we'll need smarter logic in order to filter feeds, could probably improve on this as we go.

- We will probably need a set for inactive feeds
- Include logic to designate the feeds that are in the feed set and inactive set
- The intersection between feeds and inactive sets should be designated to inactive after the current 8 tries

c3ho:
@humphd for the `check` function in inactive-blog-filter.js can we instead check the feedId and whether it is a member using [SISMEMBER ](https://www.w3resource.com/redis/redis-sismember-key-member.php) of the INACTIVE_FEED set instead of checking using the feedUrl? I would think that both do the same thing.

c3ho:
This changes quite a bit of stuff, will need discussion on how we should approach this. Should we create separate functions or use parameters for existing functions to determine what they should do with the feeds?

humphd:
@c3ho I think we should use a similar solution for this to what I'm suggesting in https://github.com/Seneca-CDOT/telescope/issues/624#issuecomment-579571774.  Specifically, when a feed becomes "invalid" (maybe after we try it a bunch of times, it fails every time, and we give up on it), we set a key in Redis to indicate that.  If the feed in question has an `id` of `ebbb25abaa`, we'd do this:

```
SET t:feed:ebbb25abaa:invalid "Date and Reason why this blog is being marked invalid..."
```

NOTE: I'm using `SET` here, not `SETEX`, since this key won't ever expire.

The value of the key can be some generated message that explains why the feed was made invalid, and when.

Whenever we process a feed, we'd check on these so-called "feed status keys" to see if a feed is invalid:

```
EXISTS t:feed:ebbb25abaa:invalid
```

If this gives back a value (1 for true), we know the invalid status is set for this feed, and can stop processing it.  We can expose this on the `Feed` object like this:

```js
let feed = await Feed.byId('ebbb25abaa');
let isInvalid = await feed.isInvalid();
// isInvalid is now true or false
```

When we remove a feed from our system, we'll also need to remove the `...:invalid` keys (if they exist).

cc @Silvyre, who is working on similar code to this.

PR #520, opened 2019-12-16
Fixes Issue #462 Integrate inactive blog url and include logic to filter inactive feeds

c3ho:
## Issue This PR Addresses
#462 

## Type of Change

-  **New Feature**: Change which adds functionality

## Description

There is no logic for handling inactive feeds in storage.js. I added a few functions that allows for active feeds to be moved to inactive feeds, get all inactive feeds and get count of all inactive feeds. There is one function that should be debated which is getInactiveFeed.

Still a WIP but would like some feedback on current changes before continuing

## Checklist

### DIFF
--- a/src/backend/feed.js
+++ b/src/backend/feed.js
@@ -2,34 +2,31 @@ const storage = require('./utils/storage');
 
 class Feed {
   /**
-   * Saves current feed to appropriate set in Redis
+   * Saves current feed to Redis
    */
   async save() {
     this.feedId = await storage.addFeed(this);
+    await storage.storeFeed(this.feedId);
   }
 
   /**
    * Checks if feed is part active feeds list. If it is, will move feed from active FEEDS set to INACTIVE_FEEDS set
+   * else will move ffeed from INACTIVE_FEEDS to FEEDS
    */
-  async inactive() {
-    const isActive = await this.isActive(this);
+  async move() {
+    const isActive = await storage.isActive(this);
     if (isActive === 1) {
-      await storage.addInactiveFeed();
+      await storage.moveFeed(this.feedId, 'feeds', 'inactive_feeds');
+    } else {
+      await storage.moveFeed(this.feedId, 'inactive_feeds', 'feeds');
     }
-  }
-
-  /**
-   * Marks the feed as active and moves feed to active set
-   */
-  async active() {
-    await this.save();
   }
 
   /**
    * Checks if the feed is in the active feeds(FEEDS) set
    */
   async isActive() {
-    return this.getFeedStatus();
+    return storage.isFeedActive(this.feedId);
   }
 }
 
--- a/src/backend/feed.js
+++ b/src/backend/feed.js
@@ -1,0 +1,33 @@
+const storage = require('./utils/storage');
+
+/**
+ * Feed Object, upon being created all feeds will be active and have a feedId of 0 until saved into Redis
+ * @param url url of feed
+ * @param name author name of feed
+ * @param feedId unique id for feed
+ */
+class Feed {
+  constructor(url, name) {
+    this.url = url;
+    this.name = name;
+    this.inactive = false;
+    this.feedId = 0;
+  }
+
+  /**
+   * Saves current feed to appropriate set in Redis
+   */
+  async save() {
+    this.feedId = await storage.addFeed(this);
+  }
+
+  /**
+   * Marks the feed as inactive
+   */
+  inactive() {
+    this.inactive = true;
+    this.save();
+  }
+}
+
+module.exports = Feed;
--- a/src/backend/feed.js
+++ b/src/backend/feed.js
@@ -1,19 +1,6 @@
 const storage = require('./utils/storage');
 
-/**
- * Feed Object, upon being created all feeds will be active and have a feedId of 0 until saved into Redis
- * @param url url of feed
- * @param name author name of feed
- * @param feedId unique id for feed
- */
 class Feed {
-  constructor(url, name) {
-    this.url = url;
-    this.name = name;
-    this.inactive = false;
-    this.feedId = 0;
-  }
-
   /**
    * Saves current feed to appropriate set in Redis
    */
@@ -22,11 +9,27 @@ class Feed {
   }
 
   /**
-   * Marks the feed as inactive
+   * Checks if feed is part active feeds list. If it is, will move feed from active FEEDS set to INACTIVE_FEEDS set
    */
-  inactive() {
-    this.inactive = true;
-    this.save();
+  async inactive() {
+    const isActive = await this.isActive(this);
+    if (isActive === 1) {
+      await storage.addInactiveFeed();
+    }
+  }
+
+  /**
+   * Marks the feed as active and moves feed to active set
+   */
+  async active() {
+    await this.save();
+  }
+
+  /**
+   * Checks if the feed is in the active feeds(FEEDS) set
+   */
+  async isActive() {
+    return this.getFeedStatus();
   }
 }
 
--- a/src/backend/utils/storage.js
+++ b/src/backend/utils/storage.js
@@ -72,9 +72,9 @@ module.exports = {
   getPost: async guid => redis.hgetall(guid),
 
   /**
-   * Moves a feed from active(FEEDS) to inactive(INACTIVE_FEEDS)
-   * @param feedId lower index
-   * @return feedId of feed being moved
+   * Moves a feed from active(FEEDS) to inactive(INACTIVE_FEEDS), will check if the set exists first before moving
+   * @param feedId key of set being moved
+   * @return feedId of key being moved
    */
   addInactiveFeed: async feedId => {
     if ((await redis.exists(INACTIVE_FEEDS)) === 1) {
@@ -90,11 +90,16 @@ module.exports = {
         .hmset(feedId, 'name', name, 'url', url)
         .sadd(INACTIVE_FEEDS, feedId)
         .exec();
-      return feedId;
     }
     return feedId;
   },
 
+  /**
+   * Gets an array of guids from redis
+   * @param from lower index
+   * @param to higher index, it needs -1 because redis includes the element at this index in the returned array
+   * @return Array of guids
+   */
   getInactiveFeeds: () => redis.smembers(INACTIVE_FEEDS),
 
   getInactiveFeed: feedID => redis.hgetall(feedID),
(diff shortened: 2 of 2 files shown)


c3ho:
I opted to use [SMOVE ](https://redis.io/commands/smove) to move members from one set to another instead of doing an SINT(set intersection).

src/backend/utils/storage.js
```diff
@@ -69,4 +70,39 @@ module.exports = {
...
+    }
+    return feedId;
+  },
+
+  /**
+   * Gets an array of guids from redis
+   * @param from lower index
+   * @param to higher index, it needs -1 because redis includes the element at this index in the returned array
+   * @return Array of guids
+   */
```
manekenpix: Do you need this comment here? You're not passing anything to 'getInactiveFeeds` so I don't know what that comment is for.

c3ho: Hahaha, you're right. I had fixed this in a new commit but didn't push it to origin. I just wanted some thought on it.

c3ho:
I had to double the code unfortunately, because we can't SMOVE if there is no destination.

c3ho:
> This feels like it's duplicating code. Could we not refactor things so that we pass `FEEDS` vs. `INACTIVE_FEEDS` to some private functions that actually do the shuffling of keys between sets?
> 
> This PR makes me want to think harder about our data and how we're structuring it, but I should probably leave that for later. I suspect we need to make some changes soon.

Oh wait I see what you mean. I can probably do that

humphd:
So I wonder if what we really need to do, is create a class for `Feed` like we have now for `Post`, and let it manage some of this state.

```js
const feed = new Feed(url, name, ...);
await feed.save(); // now in Redis, and part of FEEDS set
...
const feed = await Feed.byUrl('https://.....'); // get feed from Redis 
feed.inactive = true; // indicate that we want this to go into INACTIVE_FEEDS when we save
feed.save(); // saved, and the "inactive" flag is used to move it into the right set
```

I'm just thinking out loud, but your code at this level could support that one level above.

c3ho:
@humphd do you want get these changes through first and then file a separate issue to refactor?

src/backend/feed.js
```diff
@@ -0,0 +1,33 @@
...
+ * Feed Object, upon being created all feeds will be active and have a feedId of 0 until saved into Redis
+ * @param url url of feed
+ * @param name author name of feed
+ * @param feedId unique id for feed
+ */
+class Feed {
+  constructor(url, name) {
+    this.url = url;
+    this.name = name;
+    this.inactive = false;
```
humphd: I don't think we should include this in the actual data.  Let's have it be something that gets exposed based on which set it's in.  When we read from the db, we'll figure that out.

src/backend/feed.js
```diff
@@ -0,0 +1,33 @@
...
+    this.feedId = 0;
+  }
+
+  /**
+   * Saves current feed to appropriate set in Redis
+   */
+  async save() {
+    this.feedId = await storage.addFeed(this);
```
humphd: Let's move to using the feed url as the key.

src/backend/feed.js
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  /**
+   * Marks the feed as inactive
+   */
+  inactive() {
+    this.inactive = true;
```
humphd: Instead of storing state for this, put it in the appropriate set.

How we we re-activate a feed?

src/backend/feed.js
c3ho: in storage.js did you want to change this to be the case as well?

humphd: I filed #522 to do this, if you want to leave it til later.

src/backend/feed.js
c3ho: so in this case we just want the this.inactive as the only property?

src/backend/feed.js
c3ho: on it, added an active() function that will do the opposite and adjusted the storage.js addFeed to handle it as well

src/backend/feed.js
humphd: It's going to have to be an async getter/setter, since it requires a database operation vs. storing something on `this`.  We don't need an actual property, just a way to get at the database.

c3ho: Got it, don't we have a problem here then? If we're trying to retrieve from the database, should active/inactive be kept track of in the DB as well?

humphd: My thought here is that the state we're going to store won't be on the Object, but rather, in which of the two sets this hash object lives in Redis.  A feed is either in the feeds or inactive-feeds set, but not both.  So we can use that as a state we read from the database, and expose as a boolean via the JS API.

What about:

```js
const isActive = await feed.isActive(); // checks if this feed is in feeds set
await feed.inactive(); // set this feed inactive (move to inactive feeds)
await feed.active(); // set this feed active (move to feeds)
```

src/backend/feed.js
c3ho: I'll leave this one for later

src/backend/feed.js
c3ho: I guess what I'm confused about is I have code like this now

```
async inactive() {
  const isActive = await this.isActive();
  await this.save();
}
```
I won't be passing isActive to save(). So now my addFeed(feed) function which is called when save() is called can't differentiate between whether it should add it to FEEDS or INACTIVE_FEEDS. The only workaround I can think about this is to separate the addFeed function into a addActiveFeed(feed) and addInactiveFeed(feed).

src/backend/utils/storage.js
```diff
@@ -5,21 +5,52 @@ const FEED_ID = 'feed_id';
...
-      .sadd(FEEDS, feedId)
-      .exec();
-    return feedId;
+  /**
+   * Adds/Moves feed to the active FEEDS set and returns the feedId
+   * @param feed Feed Object
+   * @return feedId
+   */
+  addFeed: async feed => {
+    if (feed.feedId && (await redis.exists(INACTIVE_FEEDS)) === 1) {
```
humphd: These `add*` functions are doing two things, and should be split.  First, you're dealing with creating a hash to store the feed data.  That makes sense, and should stay here.  Then, you're putting the id into a set, and that can get moved out to separate functions.

Also, you probably want to use [sismember](https://redis.io/commands/sismember) vs. exists.

src/backend/utils/storage.js
```diff
@@ -5,21 +5,52 @@ const FEED_ID = 'feed_id';
...
+    return feed.feedId;
+  },
+
+  /**
+   * Moves feed from active FEEDS set to INACTIVE_FEEDS set. A destination must always be initialized
+   * with sadd() before other members can be moved to the destination
+   * @param feed Feed Object
+   */
+  addInactiveFeed: async feed => {
+    if ((await redis.exists(INACTIVE_FEEDS)) === 1) {
```
humphd: This code is identical to what you do above.  The only difference is that one uses `FEEDS` where the other uses `INACTIVE_FEEDS`.  As such, you can refactor this to use 2 arguments passed to a function, and then only do this code once, but call it from two helper functions that deal with the order of these args.

src/backend/utils/storage.js
```diff
@@ -69,4 +100,10 @@ module.exports = {
   getPostsCount: () => redis.zcard(POSTS),
 
   getPost: async guid => redis.hgetall(guid),
+
+  getInactiveFeeds: () => redis.smembers(INACTIVE_FEEDS),
+
+  getInactiveFeedsCount: () => redis.scard(INACTIVE_FEEDS),
+
+  getFeedStatus: feedId => redis.sismember(FEEDS, feedId),
```
humphd: This isn't well named. `isFeedActive` maybe?

src/backend/utils/storage.js
c3ho: I slowly realized this... and have something working in storage.js as with feedSet as a param for the set to send it to

```
addFeed: async (feed, feedSet){

}
```

src/backend/utils/storage.js
c3ho: You're right, I think I was trying to do too much with this function. the exists is checking whether the set exists in the first place to see if we need to do SADD operation first.

Would `addFeed` even be an appropriate name for this function, since we're not technically adding the feed anywhere?

src/backend/utils/storage.js
c3ho: whoops please disregard this as I replied to this before the previous one.

src/backend/feed.js
```diff
@@ -2,34 +2,31 @@ const storage = require('./utils/storage');
...
-   * Saves current feed to appropriate set in Redis
+   * Saves current feed to Redis
    */
   async save() {
     this.feedId = await storage.addFeed(this);
+    await storage.storeFeed(this.feedId);
   }
 
   /**
    * Checks if feed is part active feeds list. If it is, will move feed from active FEEDS set to INACTIVE_FEEDS set
```
humphd: part *of* active feeds list

src/backend/feed.js
```diff
@@ -2,34 +2,31 @@ const storage = require('./utils/storage');
...
+   * else will move ffeed from INACTIVE_FEEDS to FEEDS
```
humphd: s/ffeed/feed/

src/backend/feed.js
```diff
@@ -2,34 +2,31 @@ const storage = require('./utils/storage');
...
    */
-  async inactive() {
-    const isActive = await this.isActive(this);
+  async move() {
```
humphd: So this is really `toggleStatus` vs. `move` right?

humphd: I think you'd be better off having 2 functions for this vs. one.  Let the caller mange deciding which one to use.

src/backend/utils/storage.js
```diff
@@ -9,48 +9,46 @@ const INACTIVE_FEEDS = 'inactive_feeds';
...
 
   /**
-   * Moves feed from active FEEDS set to INACTIVE_FEEDS set. A destination must always be initialized
-   * with sadd() before other members can be moved to the destination
-   * @param feed Feed Object
+   * Stores feed into destinatiion
+   * @param feedId unique ID of feed to be stored
+   * @param destination set feed is being added into
+   */
+  storeFeed: async (feedId, destination) => {
```
humphd: Not sure about this name, but also don't know what to suggest.  Need to think on it.

c3ho:
Thank you for the review, I'll write some tests and make the requested changes.

c3ho:
Will move this PR to 0.6

humphd:
My work in #592 will help with this, I think.  It will also require rebasing.

c3ho:
@humphd thanks! Will review that asap

c3ho:
Code has changed a lot since this PR was submitted, will close and submit a new PR when ready.

PR #527, opened 2019-12-21
Fixes #525: Add eslint-plugin-promise and fix async/Promise issues

### LINKED ISSUE #525: Add eslint-plugin-promise and fix async/Promise issues
We have a bunch of improper uses of `async` (e.g., without `await`) and also a bunch of `Promise` bugs that will affect performance.

A first step on the path to fixing this would be to add some linting for Promises.  We can use https://github.com/xjamundx/eslint-plugin-promise, with the recommended rules defined.  I tested it locally, and it already finds 24 problems we can fix.

We should also eliminate all pure synchronous calls that have been wrapped with Promises unnecessarily.  In essence, if we don't need a Promise, let's get rid of it.

[This recent talk](https://www.youtube.com/watch?v=XV-u_Ow47s0&feature=emb_logo) is a good intro to some of the problems I'm referring to here.

c3ho:
## Type of Change

- **Bugfix**: Change which fixes an issue

## Description

Added eslint rules regarding promises, we have promises that have catches and returns and some don't. These rules ensure consistency

## Checklist

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,6 @@
 module.exports = {
-  extends: ['airbnb-base', 'prettier'],
-  plugins: ['prettier'],
+  extends: ['airbnb-base', 'prettier', 'plugin:promise/recommended'],
+  plugins: ['prettier', 'promise'],
   env: {
     jest: true,
     browser: true,
@@ -23,6 +23,36 @@ module.exports = {
      */
     'no-param-reassign': ['error', { props: false }],
 
+    /**Disallows unnecessary return await
+     * https://eslint.org/docs/rules/no-return-await
+     */
+    'no-return-await': ['error'],
+
+    /**
+     * Disallow using an async function as a Promise executor
+     * https://eslint.org/docs/rules/no-async-promise-executor
+     */
+    'no-async-promise-executor': ['error'],
+
+    /**
+     * Disallow await inside of loops
+     * https://eslint.org/docs/rules/no-await-in-loop
+     */
+    'no-await-in-loop': ['error'],
+
+    /**
+     * Disallow assignments that can lead to race conditions due to
+     * usage of await or yield
+     * https://eslint.org/docs/rules/require-atomic-updates
+     */
+    'require-atomic-updates': ['error'],
+
+    /**
+     * Disallow async functions which have no await expression
+     * https://eslint.org/docs/rules/require-await
+     */
+    'require-await': ['error'],
+
     /**
      * Require or disallow named function expressions
      * https://eslint.org/docs/rules/func-names
@@ -34,5 +64,20 @@ module.exports = {
      * https://eslint.org/docs/rules/func-names
      */
     'linebreak-style': 'off',
+
+    /**
+     * The following are eslint rules from the promise-plugin
+     * https://github.com/xjamundx/eslint-plugin-promise
+     */
+
+    /**
+     * Prefer wait to then() for reading Promise values
+     */
+    'promise/prefer-await-to-then': 'warn',
+
+    /**
+     * Prefer async/await to the callback pattern
+     */
+    'promise/prefer-await-to-callbacks': 'warn',
   },
 };
--- a/src/backend/utils/basic_analysis.js
+++ b/src/backend/utils/basic_analysis.js
@@ -71,18 +71,8 @@ module.exports = function analyzeText(text_) {
 
   // asynchonize function
   // to get a basic analysis information, getAsyAnalysis should be called
-  this.getAsyAnalysis = async function() {
-    return new Promise((res, rej) => {
-      if (isValidString) {
-        res(analysis);
-      } else {
-        res(textInfo);
-      }
-      if (isValidString === undefined) {
-        rej(textInfo);
-      }
-    });
+  this.getAsyAnalysis = function() {
+    return isValidString ? analysis : textInfo;
   };
-
   return this;
 };
--- a/src/backend/utils/email-sender.js
+++ b/src/backend/utils/email-sender.js
@@ -2,8 +2,6 @@ require('../lib/config');
 const nodemailer = require('nodemailer');
 const { logger } = require('./logger');
 
-const log = logger.child({ module: 'email-sender' });
-
 /*
                              HOW TO USE
  Import this file - const sendEmail = require('./email-sender);
@@ -69,39 +67,36 @@ exports.verifyTransporter = function(transporter) {
   transporter.verify(err => {
     // If error then print to console
     if (err) {
-      log.error({ err }, 'Transporter connection failed.');
+      logger.error({ err }, 'Transporter connection failed.');
       return false;
     }
     // else print a ready message
-    log.info('Server is running properly');
+    logger.info('Server is running properly');
     return true;
   });
 };
 
 // Sends a message using the passed in parameters
 exports.sendMessage = async function(receipiants, subjectMessage, message) {
-  return new Promise((resolve, reject) => {
-    const transporter = this.createTransporter(
+  try {
+    const transporter = await this.createTransporter(
       process.env.NODEMAILER_SERVER,
       2222,
       false,
       process.env.NODEMAILER_USERNAME,
       process.env.NODEMAILER_PASSWORD
     );
-    const allGood = this.verifyTransporter(transporter);
-    if (!allGood) {
-      reject(new Error()); // Send promise.reject if an error occurs
+    if (!this.verifyTransporter(transporter)) {
+      throw new Error('Email transport could not be verified');
     }
     // Creates email for to be sent
     const mail = this.createMail(receipiants, subjectMessage, message);
 
     // Send the email with the email content
-    transporter.sendMail(mail, (err, info) => {
-      if (err) {
-        reject(err); // Send promise.reject if an error occurs
-      } else {
-        resolve(info.accepted); // Send promise.resolve if an error occurs
-      }
-    });
-  });
+    const result = await transporter.sendMail(mail);
+    return result.accepted;
+  } catch (error) {
+    logger.error({ error }, 'Unable to send email');
+    throw error;
+  }
 };
--- a/src/backend/utils/github-url.js
+++ b/src/backend/utils/github-url.js
@@ -1,6 +1,7 @@
 require('../lib/config');
 const parseGithubUrl = require('parse-github-url');
 const fetch = require('node-fetch');
+const { logger } = require('./logger');
 
 const githubAPI = 'http://api.github.com';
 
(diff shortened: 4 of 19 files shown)


c3ho:
Having issues with dealing with inactive-blog-filter.js. It returns a lot of promises.

.eslintrc.js
```diff
@@ -23,6 +23,36 @@ module.exports = {
...
+     * usage of await or yield
+     * https://eslint.org/docs/rules/require-atomic-updates
+     */
+    'require-atomic-updates': ['error'],
+
+    /**
+     * Disallow async functions which have no await expression
+     * https://eslint.org/docs/rules/require-await
+     */
+    'require-awaut': ['error'],
```
manekenpix:
```suggestion
    'require-await': ['error'],
```

manekenpix: Also, this one seems to create a conflict with the rule in line 29 `no-return-await` in cases like [line 71 in storage.js](https://github.com/Seneca-CDOT/telescope/blob/cda82ab740464c04d50605bd6576382809381311/src/backend/utils/storage.js#L71).

c3ho: I think we can probably fix line 71 with this

```
getPost: async guid => {
  const posts = await redis.getall(guid);
  return posts;
  },
```

c3ho:
I think all the functions in storage.js are/should be async no?

c3ho:
Running npm test will fail on the wiki-feed-parser.test
image

Otherwise it passes every test. Please go easy on me.

manekenpix:
Tests pass for me.

.eslintrc.js
```diff
@@ -34,5 +64,80 @@ module.exports = {
      * https://eslint.org/docs/rules/func-names
      */
     'linebreak-style': 'off',
+
+    /**
+     * The following are eslint rules from the promise-plugin
+     * https://github.com/xjamundx/eslint-plugin-promise
```
humphd: Question: why not use the recommended set vs. manually enabling all these?

```
{
  "extends": ["plugin:promise/recommended"]
}
```

Are you using ones other than the recommended?

src/backend/index.js
```diff
@@ -42,6 +42,7 @@ async function enqueueWikiFeed() {
 enqueueWikiFeed()
   .then(() => {
```
humphd: Often this means you can just rely on the implicit return value you get with an arrow function.  I think this is better:

```
.then(() => feedWorker.start())
```

src/backend/utils/basic_analysis.js
```diff
@@ -71,17 +71,17 @@ module.exports = function analyzeText(text_) {
...
-        res(analysis);
-      } else {
-        res(textInfo);
-      }
-      if (isValidString === undefined) {
-        rej(textInfo);
-      }
-    });
+  this.getAsyAnalysis = function() {
+    let result;
```
humphd: Drop this `result` arg, and just use `return` on every line where you set `result` below.  [Return early, return often](https://blog.humphd.org/vocamus-1421/).

src/backend/utils/email-sender.js
```diff
@@ -90,18 +90,16 @@ exports.sendMessage = async function(receipiants, subjectMessage, message) {
     );
     const allGood = this.verifyTransporter(transporter);
     if (!allGood) {
-      reject(new Error()); // Send promise.reject if an error occurs
+      logger.error(new Error());
```
humphd: Get rid of the variable, add a message to the error so it's useful, throw.

```js
if(!this.verifyTransporter(transporter)) {
  throw new Error('email transport could not be verified');
}
```

src/backend/utils/github-url.js
```diff
@@ -57,68 +57,66 @@ exports.getGithubUrlData = async incomingUrl => {
       ? `?access_token=${process.env.GITHUB_TOKEN}`
       : '';
 
-  return fetch(`${githubAPI}${subUrl}`)
-    .then(res => res.json())
-    .then(data => {
-      let fetchedData;
+  const fetchResult = await fetch(`${githubAPI}${subUrl}`);
```
humphd: `fetch` can throw, so can `.json()`, so we should wrap these in try/catch.

src/backend/utils/storage.js
```diff
@@ -68,5 +68,8 @@ module.exports = {
 
   getPostsCount: () => redis.zcard(POSTS),
 
-  getPost: async guid => redis.hgetall(guid),
+  getPost: async guid => {
```
humphd: No need for async/await here, this already returns a Promise.

```js
getPost: guid => redis.hgetall(guid)
```

test/email-sender.test.js
```diff
@@ -12,7 +12,7 @@ test.skip('Tests if sendMessage resolves with expected info', async () => {
   const testRecipient = process.env.NODEMAILER_TESTRECIPIENT;
   const testReturnValue = emailsender.sendMessage(testRecipient, testSubjectMessage, testHTML);
   // Tests if the expected resolve is correct
-  expect(testReturnValue).resolves.toBe([testRecipient]);
+  await expect(testReturnValue).resolves.toBe([testRecipient]);
```
humphd: Should this `await` not go with the call to `emailsender.sendMessage` on line 13 above?

test/spam-checker.test.js
```diff
@@ -1,5 +1,6 @@
 const feedparser = require('feedparser-promised');
 const checkForSpam = require('../src/backend/utils/spam-checker');
+// const { logger } = require('../src/backend/utils/logger');
```
humphd: Remove this if not needed.

test/storage.test.js
```diff
@@ -81,7 +82,13 @@ describe('Storage tests for posts', () => {
   };
 
   beforeAll(() => {
-    Promise.all([testPost, testPost2, testPost3].map(post => addPost(post)));
+    Promise.all([testPost, testPost2, testPost3].map(post => addPost(post)))
+      .then(() => {
```
humphd: Why are we mixing async/await and .then?  Do we really need this .then at all?

.eslintrc.js
humphd: What if we do a bit of both here: Extend the recommended set, and for all those ones that aren't in recommended that you want from the [rules set](https://github.com/xjamundx/eslint-plugin-promise#rules), add them manually to the rules below.

src/backend/utils/basic_analysis.js
```diff
@@ -71,18 +71,14 @@ module.exports = function analyzeText(text_) {
...
-        res(analysis);
-      } else {
-        res(textInfo);
-      }
-      if (isValidString === undefined) {
-        rej(textInfo);
-      }
-    });
+  this.getAsyAnalysis = function() {
+    if (isValidString === undefined) {
```
humphd: It looks like this can all be simplified to:

```js
return isValidString ? analysis : textInfo;
```

Make sure I'm right (untested), but just reading this code, there's no reason for all these different code paths that I can see.

src/backend/utils/github-url.js
```diff
@@ -56,69 +56,70 @@ exports.getGithubUrlData = async incomingUrl => {
     typeof process.env.GITHUB_TOKEN !== 'undefined'
       ? `?access_token=${process.env.GITHUB_TOKEN}`
       : '';
+  try {
+    const fetchResult = await fetch(`${githubAPI}${subUrl}`);
+    const data = await fetchResult.json();
```
humphd: Put your `catch` here vs. at the bottom, it's only these 2 lines that can throw.

src/backend/utils/github-url.js
```diff
@@ -56,69 +56,70 @@ exports.getGithubUrlData = async incomingUrl => {
...
+        avatarURL,
+        body,
+        createdAt,
+        branch: ghUrl.branch,
+        repo: ghUrl.name,
+      };
+    }
+    return fetchedData;
+  } catch (error) {
+    throw new Error('Error fetching data');
```
humphd: If all you're going to do in a `.catch` is `throw`, there's no value in doing it, and you can just leave it off.  However, it might be nice to log the error here, before you throw it.  Also, I'd consider just re-throwing the actual error vs. making your own, and losing the original error message and stack.

humphd: What you do in  `src/backend/utils/inactive-blog-filter.js` below is what I mean.

src/backend/utils/storage.js
```diff
@@ -68,5 +68,7 @@ module.exports = {
 
   getPostsCount: () => redis.zcard(POSTS),
 
-  getPost: async guid => redis.hgetall(guid),
+  getPost: guid => {
```
humphd: You don't need the explicit `return` here:

```js
getPost: guid => redis.hgetall(guid);
```

A fat arrow function will automatically return if you give it an expression vs. a `{...}` body block.

src/backend/utils/word-counter.js
```diff
@@ -1,5 +1,8 @@
+/**
+ * https://www.npmjs.com/package/wordcount
+ */
 const wordcount = require('wordcount');
 
-exports.wordCounter = async function wordCounter(blogText) {
-  return Promise.resolve(wordcount(blogText));
+exports.wordCounter = function wordCounter(blogText) {
```
humphd: This file isn't doing anything.  We're pulling in the `wordcount` module, then creating a function called `wordCounter` that calls `wordcount`.  Let's just `git rm` this file, and change the caller(s) to use this module directly.

test/inactive-blog-filter.test.js
```diff
@@ -36,31 +36,20 @@ describe('Redlisted feed checking', () => {
...
-          expect(isRedlisted).toBe(false);
-        })
-        .catch(err => {
-          log.error(err.message);
-          throw err;
-        });
-    });
+    await Promise.all(
+      badURLs.map(async url => {
+        // const inactiveFilterCheck = await inactiveFilter.check(url);
```
humphd: Why is this line here?  Remove or uncomment, please.

test/inactive-blog-filter.test.js
```diff
@@ -129,54 +118,57 @@ describe('Testing of update()', () => {
...
+      const data = await inactiveFilter.update(
+        pathToInactiveFeeds,
+        pathToMockRedList,
+        mockFeedParser
+      );
+      expect(data.length).toEqual(inactiveRedlist.length);
+      for (let i = 0; i < inactiveRedlist.length; i += 1) {
+        expect(data[i].url).toBe(inactiveRedlist[i].url);
+      }
+      return data;
```
humphd: Who are we returning `data` to here?  Can we get rid of this?

test/inactive-blog-filter.test.js
```diff
@@ -129,54 +118,57 @@ describe('Testing of update()', () => {
...
+    } catch (err) {
+      deleteMockRedlist();
+      log.error(err.message);
```
humphd: I think in unit tests, it's better to just let things throw, and not bother with catching, unless we want to test something in the error handling path.  The test harness will blow up if there's an exception, and print it, so we don't have to.

test/posts.test.js
```diff
@@ -24,8 +25,12 @@ describe('test /posts endpoint', () => {
     };
   });
 
-  beforeAll(() => {
-    Promise.all(posts.map(post => addPost(post)));
+  beforeAll(async () => {
+    try {
```
humphd: Don't bother with try/catch here.  Let it fail loudly in tests.

test/storage.test.js
```diff
@@ -80,8 +81,13 @@ describe('Storage tests for posts', () => {
     site: 'foo',
   };
 
-  beforeAll(() => {
-    Promise.all([testPost, testPost2, testPost3].map(post => addPost(post)));
+  beforeAll(async () => {
+    try {
```
humphd: Drop this try/catch.

src/backend/utils/github-url.js
```diff
@@ -1,7 +1,9 @@
 require('../lib/config');
 const parseGithubUrl = require('parse-github-url');
 const fetch = require('node-fetch');
+const { logger } = require('./logger');
 
+const log = logger.child({ module: 'github-url' });
```
humphd: Let's not do this.  The docs suggest it, but for our needs, it's not really necessary to isolate logs into sub-modules at this stage.  Just use `logger.info()`, etc.

src/backend/utils/inactive-blog-filter.js
```diff
@@ -38,17 +38,14 @@ function dateDiff(postDate) {
...
-      return Promise.resolve(redList.some(isRedlisted.bind(null, feedUrl)));
-    })
-    .catch(error => {
-      log.error({ error }, `failed to read ${redlistPath}`);
-      return Promise.reject(error);
-    });
+  try {
+    const redListRaw = await fsPromises.readFile(redlistPath, 'utf-8');
+    const redList = [].concat(JSON.parse(redListRaw));
+    return redList.some(isRedlisted.bind(null, feedUrl));
```
humphd: Could this not be `return redList.some(redItem => isRedlisted(feedUrl, redItem));`?  It doesn't seem like we need to `.bind()` here.

.eslintrc.js
humphd: This looks undone to me, are you going to modify this further?

humphd:
@c3ho did you ever fix your error above in https://github.com/Seneca-CDOT/telescope/pull/527#issuecomment-573290446?  It means you need to rebuild some binary dependency that was linked against header files for a different version of node/v8.  If you update node, you sometimes have to rebuild.  There's a command to do it, or you can `rm -fr node_modules && npm install`.

c3ho:
@humphd it is flaky for me on Windows. Sometimes it passes with no issues, sometimes I run into the error. I've done what you've usually suggested in OSD600 removing the dependency folders and everything. I've even tried installing nvm on my windows machine. I tested this on Mac and there were no issues.

test/posts.test.js
```diff
@@ -24,8 +24,8 @@ describe('test /posts endpoint', () => {
     };
   });
 
-  beforeAll(() => {
-    Promise.all(posts.map(post => addPost(post)));
+  beforeAll(async () => {
```
humphd: Jest's functions work async if you return a Promise, so you can change this to:

```js
beforeAll(() => Promise.all(posts.map(post => addPost(post))));
```

```

File: test/posts.test.js [hunk 24]
```diff
@@ -120,8 +120,8 @@ describe('test /posts/:guid responses', () => {
 
   // add the post to the storage
   beforeAll(async () => {
```
humphd:
```js
beforeAll(() => Promise.all([addedPost1.save(), addedPost2.save()]));
```

test/storage.test.js
```diff
@@ -80,8 +80,8 @@ describe('Storage tests for posts', () => {
     site: 'foo',
   };
 
-  beforeAll(() => {
-    Promise.all([testPost, testPost2, testPost3].map(post => addPost(post)));
+  beforeAll(async () => {
```
humphd:
```js
beforeAll(() => Promise.all([testPost, testPost2, testPost3].map(post => addPost(post))));
```

test/posts.test.js
humphd: This one got missed.

c3ho:
@humphd if this is good to go, I'm going to rebase again and resubmit, VSC was going weird on me and added your recent code as part of my PR

c3ho:
@humphd its failing on this one 
image

Is the fix just a simple ```await``` in front of ```queue.add()``` (line 33)?

humphd:
I filed https://github.com/Seneca-CDOT/telescope/issues/567 to deal with the problem of knowing whether something returns a `Promise` or not.

humphd:
Yeah, merged!

PR #544, opened 2020-01-18
Closes: #532. Add graphql support to our server

### LINKED ISSUE #532: Add graphql support to our server

Expose the equivalent of our current `/post`, `/posts`, etc. REST endpoints as a [GraphQL](https://graphql.org/).  Much as [GitHub does with its API](https://github.blog/2016-09-14-the-github-graphql-api/), we can support both REST and GraphQL in parallel (maybe later we'll remove the REST API, but for now it's good to have it, since it's working).

In #531 we're adding a [GatsbyJS](https://www.gatsbyjs.org/) frontend.  [GatsbyJS](https://www.gatsbyjs.org/) works best when used in conjunction with a [GitHub does with its API](https://github.blog/2016-09-14-the-github-graphql-api/) API (it can also use REST APIs, but with some loss of functionality).

The fastest path forward here is likely to [use express-graphql](https://graphql.org/graphql-js/running-an-express-graphql-server/).

Assigning this to @manekenpix, who said he wanted to work on it.  For now, I'll add this to the 0.5 project, but we might want to push it out to 0.6 if we run out of time.

manekenpix:
Closes #532 

## Type of Change

## Description

Adds support for [GraphQL](https://graphql.org/) with the endpoint `/graphql`.

Currently, Telescope is not supposed to manage complicated data types, so at the moment it's not going to take advantage of all the features included in the GraphQL query system.

Once Telescope adds support for subscribers/users, and data analysis for posts, it'll be possible to increase the complexity of the queries by adding features like filters, proper GraphQL pagination using [connections](https://blog.apollographql.com/explaining-graphql-connections-c48b7c3d6976), [nested resolvers](https://reactgo.com/nested-resolvers-relationaldata-graphql/), etc.

### Queries

- `getPost(guid: )`: Returns a post for the passed `guid`.

- `getPosts(page: , perPage: )`: Fetches 'perPage' number of `Post objects` from page number 'page'.
**Note**: Telescope's REST implementation of `getPosts` currently returns `guids`, not `Post objects`.

- `getPostsCounts`: Returns the number of Post objects in telescope's database.

I added some queries for Feeds but they can't be tested at the moment.

### Examples:

Start redis and telescope, and navigate to `localhost:{port_number_in_your_env_file}/graphql` and you'll see [GraphQL's playground](https://www.gatsbyjs.org/docs/using-graphql-playground/) to test the queries.

**Important** Every query that returns a Post object **must** include at least one subfield, if not GraphQL request will fail.

No subfields
wrong-request

With subfields
right-request

`getPosts(page: 1, perPage: 20)` example:
request

## Checklist


package.json
```diff
@@ -49,6 +49,7 @@
   },
   "homepage": "https://github.com/Seneca-CDOT/telescope#readme",
   "dependencies": {
+    "apollo-server-express": "2.9.16",
```
humphd: Good choice.  I notice the [GraphQL docs recommend this as one of the preferred options](https://graphql.org/learn/serving-over-http/#node).

src/backend/web/app.js
```diff
@@ -19,6 +24,15 @@ app.use(logger);
...
 app.use(cors());
 
+const server = new ApolloServer({
+  typeDefs,
+  resolvers,
+  context: { storage },
+});
+
+// Add the Apollo server to app and define GraphQL's endpoint
+server.applyMiddleware({ app, path: '/gql' });
```
humphd: Can we change this to `/graphql` to match the more common/expected name APIs seem to use?

From [the docs]:

> a GraphQL server operates on a single URL/endpoint, usually /graphql

For example, [GitHub uses https://api.github.com/graphql](https://developer.github.com/v4/guides/forming-calls/#the-graphql-endpoint).

src/backend/web/app.js
```diff
@@ -3,6 +3,11 @@ const express = require('express');
 const expressHandlebars = require('express-handlebars');
 const healthcheck = require('express-healthcheck');
 const cors = require('cors');
+const { ApolloServer } = require('apollo-server-express');
+
+// Required for GraphQl
+const storage = require('../utils/storage');
+const { typeDefs, resolvers } = require('./graphql/queries');
```
humphd: If the only thing in the `./graphql` dir is this one file, let's rename it to `./graphql/index.js` so you can do `require('./graphql');` here.

src/backend/web/app.js
```diff
@@ -3,6 +3,11 @@ const express = require('express');
 const expressHandlebars = require('express-handlebars');
 const healthcheck = require('express-healthcheck');
 const cors = require('cors');
+const { ApolloServer } = require('apollo-server-express');
+
+// Required for GraphQl
```
humphd: capital `L`: `GraphQL`

src/backend/web/graphql/queries.js
```diff
@@ -0,0 +1,99 @@
+const { gql } = require('apollo-server-express');
+
+const maxPostsPerPage = 30;
```
humphd: Should this be configurable via `env`?  It seems like we now have 2 places where we do it (here and in the `/posts` pagination code).

humphd: Should this be configurable via `env`?  It seems like we now have 2 places where we do it (here and in the `/posts` pagination code).

src/backend/web/graphql/queries.js
```diff
@@ -0,0 +1,99 @@
...
+    url: String
+  }
+
+  # 'Post' matches our Post type used with redis
+  type Post {
+    author: String
+    title: String
+    html: String
+    text: String
+    published: String
```
humphd: Should be `Date`, same for `updated` on next line.  If we can use the `Post` constructor somehow here vs pulling raw from Redis, we'll get it for free: https://github.com/Seneca-CDOT/telescope/blob/master/src/backend/post.js#L21-L22.

src/backend/web/app.js
```diff
@@ -19,6 +24,15 @@ app.use(logger);
 
 app.use(cors());
 
+const server = new ApolloServer({
+  typeDefs,
+  resolvers,
+  context: { storage },
```
humphd: Question: why not use node's module resolution for this instead of passing this down via `context`?  We can just `requrie()` storage where you need it when defining the queries and use it directly.

src/backend/web/graphql/queries.js
```diff
@@ -0,0 +1,99 @@
...
+    updated: String
+    url: String
+    site: String
+    guid: String
+  }
+
+  # Queries to fetch data from redis
+  type Query {
+    getFeed(guid: ID!): Feed
+    getFeeds: [ID]
```
humphd: Another question: what do GraphQL clients expect when requesting data: getting back IDs (like a REST API); or automatic rehydration with full Object data?  I don't have a lot of experience designing these APIs, so I'm unclear how we should model the data.  Should I be able to request a `feed` or an Array of `feeds` using essentially the same query?

src/backend/web/graphql/queries.js
```diff
@@ -0,0 +1,99 @@
...
+      return storage.getFeedsCount();
+    },
+
+    /**
+     * @description Takes a guid and returns a Post object
+     * @param guid
+     * @return Post object for the passed guid
+     */
+    getPost: (parent, { guid }, { storage }) => {
+      return storage.getPost(guid);
```
humphd: We can use `Post.byGuid()` here instead to get proper types: https://github.com/Seneca-CDOT/telescope/blob/master/src/backend/post.js#L125

src/backend/web/graphql/queries.js
```diff
@@ -0,0 +1,99 @@
...
+    },
+
+    /**
+     * @description Fetches 'perPage' number of Post objects from page number 'page'.
+     * @param page Number of page from which Posts objects are fetched
+     * @param perPage Number of Post objects in every page
+     * @return Array of 'perPage' number of Post objects
+     */
+    getPosts: async (parent, { page, perPage }, { storage }) => {
+      const prPage = perPage > maxPostsPerPage ? maxPostsPerPage : perPage;
```
humphd: Can we share any/all of this code with the REST API?  Seems pretty similar.

src/backend/web/graphql/queries.js
```diff
@@ -0,0 +1,99 @@
...
+
+      const numOFPosts = await storage.getPostsCount();
+      const first = page * prPage;
+
+      if (first < numOFPosts) {
+        const last = first + prPage < numOFPosts ? first + prPage : numOFPosts;
+        const posts = await storage.getPosts(first, last);
+
+        return Promise.all(posts.map(post => storage.getPost(post)));
```
humphd: We can use `Post.byGuid()` here.

c3ho:
I'll review this later tonight if no one else is taking a look at this.

src/backend/web/app.js
manekenpix: Changed to `index.js` and no we have `require('./graphql')`

src/backend/web/app.js
manekenpix: Fixed

src/backend/web/graphql/queries.js
manekenpix: I added it to our `.env` and we're going to use it both in `graphql/index.js` and `posts.js`.

src/backend/web/app.js
manekenpix: Initially that's what I tried (`require()` in the `.js` file where the queries are), but it seems `Apollo` doesn't allow access to the queries from _outside_ the server, so the only way that it'd let me use storage functions _in_ the queries was by passing them down via `context`.

humphd: Fascinating.  How did it disallow it?  The `storage` via `require()` wasn't defined?

humphd: This is going to (potentially) complicated a refactor I wanted to make, moving data logic into classes like `Post` and `Feed` and slowly getting rid of `storage` completely.  Maybe we would just pass both `Post` and `Feed` down on `context`, and it would be OK.

src/backend/web/graphql/queries.js
manekenpix: >  Should I be able to request a feed or an Array of feeds using essentially the same query?

Yes, that's essentially the idea.
Our REST API has `GET /post` and `GET /post/:guid`. The first endpoint returns an array of `guids` that can be used with the second one.

With GraphQL, `/grapql` returns any subfield(s) included in the query, so:

`GET /post` becomes
```
getPosts()
{
    author
    title
    html
    text
    published
    updated
    url
    site
    guid
}
```

and `GET /posts/:guid` becomes
```
getPosts()
{
    guid
}
```

All in one query. 
I added this query `getPost(guid: ID!): Post` to return a post for a given ID because `getPosts` needs some sort of filtering functionality added. I did a bit of research for this, but I thought it was better to try get basic functionality in and then tweak this part with future PRs.
Also, the way getPosts works right now is enough to get Telescope to work with Gatsby.

src/backend/web/app.js
manekenpix: Changed to `/graphql`

src/backend/web/app.js
manekenpix: @humphd I tested it again and it turned out it's ok to require storage where the queries are. Maybe I got some sort of syntax error when I was testing it and I interpreted as Apollo not being happy about with what I was doing. Thank you for asking about this, without your question I'd definitely have missed this.

src/backend/web/app.js
```diff
@@ -3,6 +3,10 @@ const express = require('express');
 const expressHandlebars = require('express-handlebars');
 const healthcheck = require('express-healthcheck');
 const cors = require('cors');
+const { ApolloServer } = require('apollo-server-express');
+
+// Required for GraphQL
+const { typeDefs, resolvers } = require('./graphql');
```
humphd: Can we move this down to be closer to line 26, since that's where it's being used?

src/backend/web/graphql/index.js
```diff
@@ -0,0 +1,109 @@
+const { gql } = require('apollo-server-express');
```
humphd: `graphql` would make this more readable

manekenpix: I was trying to follow their own syntax, check [here](https://www.apollographql.com/docs/apollo-server/api/apollo-server/#gql).
I changed it to `graphql`.

src/backend/web/app.js
manekenpix: Fixed

src/backend/web/graphql/index.js
```diff
@@ -0,0 +1,109 @@
+const { gql: graphql } = require('apollo-server-express');
```
humphd: Sorry to make you change this if it's not their standard way.  That said, I think line 15 reads better from a maintenance point of view, with devs who aren't as familiar with graphql, and will get a better sense of what is going on.  Thanks for doing this.

src/backend/web/graphql/index.js
```diff
@@ -0,0 +1,109 @@
...
+      const prPage = perPage > maxPostsPerPage ? maxPostsPerPage : perPage;
+
+      const numOFPosts = await getPostsCount();
+      const first = page * prPage;
+
+      if (first < numOFPosts) {
+        const last = first + prPage < numOFPosts ? first + prPage : numOFPosts;
+        const posts = await getPosts(first, last);
+
+        return Promise.all(posts.map(post => Post.byGuid(post)));
```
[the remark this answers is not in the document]
humphd: We don't have to include one, no, since it will just propagate the error to the caller.  It does mean that all callers of `getPosts` need a `try/catch`.

PR #545, opened 2020-01-18
Refactor src/backend/feed/parser.js to use package directly

humphd:
This simplifies the feed parser code, using the feedparser package directly instead of adding an unnecessary level of indirection.  I've removed the tests for this, since that module has its own tests (i.e., ours are incomplete and duplicating what upstream has).


humphd:
@c3ho if you stay in this game long enough, you [get to see all your best work deleted](https://blog.humphd.org/vocamus-1626/).

PR #547, opened 2020-01-18
Parallelize feed parser, make it run forever

### LINKED ISSUE #503: Telescope needs to continually update

In order to be useful, our data has to be current.  And in order to be current, we need to continually re-download and parse all the blog feeds into Redis.

Luckily, this is a feature easily supported by Bull.  We can use a [cron job style string](https://github.com/OptimalBits/bull#repeated-jobs) to define repeated jobs.  Later we'll have to add code to purge deleted/errorred feeds from our list, but that can be done after this.

We need our feeds to stay current.  Currently they only update once, and then sit in that state forever.

humphd:
## Issue This PR Addresses

Fixes #503

## Type of Change

## Description

This PR improves our feed parser.  Specifically, it makes it run forever, continually downloading the feeds and parsing them into Redis.  It also allows for parallelization, so we can run multiple instances of our feed worker in different processes.

In order to make this work, I've also done a few other things:

- Moved the feed parsing to its own file, `src/backend/feed/processor.js`, so that we can run it in a separate process or processes
- Altered our retry attempts and delay timing when feeds fail.  What I have still isn't quite right, but what we had before was way to extreme, and the queue would back-up with failed jobs
- Added a new `env` variable: `FEED_QUEUE_PARALLEL_WORKERS`.  This is the number of parallel worker instances to run in separate processes, and can be 1, 2, 3... up to `*`, which means "one per CPU."  I cap this at the CPU count so it doesn't bring the server down.
- I've started passing HTTP options to `request` via the feedparser `parse` function.  I wanted to shorten the timeout length, so dead blogs don't block a worker for so long.  I've also enabled gzip, which makes things faster.
- I've used the `drained` event on the feed queue to trigger another round of updates.  Whenever we finish processing one set of feeds, and `drained` occurs, we'll start over.  This is our infinite loop.
- I've changed the `job.id` to be the feed URL instead of an auto-incrementing integer.  This allows some extra logic so that multiple jobs for the same feed don't get added to the queue (Bull will reject attempts to add the same job/URL more than once).

This is only the beginning of this work.  We can do a bunch more to improve things, especially with blacklisting blogs that fail over and over, adding cache headers so we don't re-process feeds that haven't changed since the last time we updated, etc.  But for now, this will get us a server that does what we expect: auto update in an infinite loop.

## Checklist

### DIFF
--- a/env.example
+++ b/env.example
@@ -57,7 +57,12 @@ BLOG_INACTIVE_TIME=360
 GITHUB_TOKEN=
 
 # Feed job queue attempts
-FEED_QUEUE_ATTEMPTS=8
+FEED_QUEUE_ATTEMPTS=2
 
 # Feed job queue delay(ms)
-FEED_QUEUE_DELAY_MS=60000
+FEED_QUEUE_DELAY_MS=30000
+
+# Number of concurrent worker processors to run. Use * if you want to run
+# one per CPU. Use a number if you want to set it manually, up to a max
+# of the CPU count.  If not set, we'll assume 1.
+FEED_QUEUE_PARALLEL_WORKERS=1
--- a/tools/add-feed.js
+++ b/tools/add-feed.js
@@ -27,10 +27,10 @@ async function add() {
 
   await feedQueue
     .add(feed, {
-      attempts: process.env.FEED_QUEUE_ATTEMPTS || 8,
+      attempts: process.env.FEED_QUEUE_ATTEMPTS || 2,
       backoff: {
         type: 'exponential',
-        delay: process.env.FEED_QUEUE_DELAY_MS || 60 * 1000,
+        delay: process.env.FEED_QUEUE_DELAY_MS || 30 * 1000,
       },
       removeOnComplete: true,
       removeOnFail: true,
--- a/tools/add-feed.js
+++ b/tools/add-feed.js
@@ -23,22 +23,13 @@ async function add() {
     process.exit(1);
   }
 
-  const feed = { name, url };
+  const feedInfo = { name, url };
 
-  await feedQueue
-    .add(feed, {
-      attempts: process.env.FEED_QUEUE_ATTEMPTS || 2,
-      backoff: {
-        type: 'exponential',
-        delay: process.env.FEED_QUEUE_DELAY_MS || 30 * 1000,
-      },
-      removeOnComplete: true,
-      removeOnFail: true,
-    })
-    .catch(err => {
-      log.error({ err }, 'Error enqueuing feed');
-      process.exit(1);
-    });
-  process.exit(0);
+  try {
+    await feedQueue.addFeed(feedInfo);
+    process.exit(0);
+  } catch (err) {
+    process.exit(1);
+  }
 }
 add();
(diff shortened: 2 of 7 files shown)


env.example
```diff
@@ -57,7 +57,12 @@ BLOG_INACTIVE_TIME=360
 GITHUB_TOKEN=
 
 # Feed job queue attempts
-FEED_QUEUE_ATTEMPTS=8
+FEED_QUEUE_ATTEMPTS=2
```
Grommers00: Although I agree 8 is a bit high, should this have a default variable of anything higher then 1?

If the individual isn't using the feed queue at the time, shouldn't this just be 1 to maximize efficiency for them? 

I get it from a more post-production standpoint we would play with this number more, but for development purposes wouldn't it be best for the developer to choose depending on what they need to test.

tools/add-feed.js
```diff
@@ -27,10 +27,10 @@ async function add() {
 
   await feedQueue
     .add(feed, {
-      attempts: process.env.FEED_QUEUE_ATTEMPTS || 8,
+      attempts: process.env.FEED_QUEUE_ATTEMPTS || 2,
       backoff: {
         type: 'exponential',
-        delay: process.env.FEED_QUEUE_DELAY_MS || 60 * 1000,
+        delay: process.env.FEED_QUEUE_DELAY_MS || 30 * 1000,
```
Grommers00: am I right in thinking this says there's a delay of 30000, or a delay of (30 * 1000) if thats the case, why do we multiply opposed to just write 30000?

manekenpix: @Grommers00  I think this is just to provide some context:
`1sec = 1000msecs`
`30 * 1000secs = 30000 msecs = 30 secs`

Sometimes it's good to break down some values to explain where they're coming from. If I do:
`const val = 24 * 60 * 60 * 1000`
it's easy to guess that I'm using `milliseconds in a day`, as opposed to:
`const val = 86400000`
which seems like a random number coming out of nowhere.

Grommers00: Makes sense.

env.example
humphd: We can play with these numbers and try to find a balance.  As you point out, any dev, and any server deployment, can override them with whatever value they like.  So a dev doing testing, who doesn't care about retrying feeds, can set it to 1 (or 0).  Whereas, in production, we can be more liberal with what we retry.

The way that these numbers work in practice is that when a job fails (e.g., can't download a feed, feed can't be parsed, etc.) it will get put into a "delayed" state, and Bull will automatically retry it later.

The number of times it will be retried is `FEED_QUEUE_ATTEMPTS` and `FEED_QUEUE_DELAY_MS` is how long to delay each retry.  NOTE: we are using an `exponential` backoff strategy, which means it will wait 30s (30,000ms) before retrying the first time, 60s the second, and so on (attempts * delay) until it finally works or gives up.

From experience running with the changes in this PR for a while, if I have too many attempts, and wait too long between attempts, the delayed jobs start to back up.

There are a few long-term fixes we can do for this, including better ways to deal with errors (see #548), blacklisting blogs that fail and won't ever work, and also playing with these numbers.

Does this make sense?

humphd:
While I was working on this PR, I noticed we had duplicate code to add feeds.  I've included another commit (7642ea152d4bd8332f6003d6ac6cf015bcff7112) to refactor this out into its own function.  This way when we play with these retry/backoff numbers, we won't have to remember to change things in multiple places.

env.example
Grommers00: Yes, thanks!

PR #550, opened 2020-01-19
Fixes #396: add default values to environment variables

### LINKED ISSUE #396: Ensure that all environment variable calls are properly handled
Since our CI testing does not define any of the environment variables that we set locally via our `.env` file, calls to any such variables (e.g. `process.env.BLOG_INACTIVE_TIME`) will return undefined during CI testing. As such, _environment variables should always be accompanied by a [default value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_Operators#Short-circuit_evaluation)_.

For example, replacing `process.env.BLOG_INACTIVE_TIME` with `(process.env.BLOG_INACTIVE_TIME || 365)` [effectively provides a default value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_Operators#Short-circuit_evaluation) of 365 to that environment variable.

Every instance of environment variables usage within our codebase should be reviewed to ensure that we are able to account for cases where those variables may be undefined.

In select cases, when an environment variable is undefined, it may be prudent to take further action (in addition to falling back on default value): logging a warning, throwing an error, etc.

Silvyre:
## Issue This PR Addresses

Fixes #396

## Type of Change

## Description

Ensures that every `process.env` call within the code base is provided with a default value (as specified in #396) to ensure that they have some value on CI, and thus do not cause unwanted behaviour during CI testing.

The default value provided to each `process.env` is either one specified within `env.example` (if one exists), or `null`.

## Checklist

### DIFF
--- a/src/backend/lib/config.js
+++ b/src/backend/lib/config.js
@@ -1,3 +1,12 @@
 const dotenv = require('dotenv');
+const { logger } = require('../utils/logger');
 
-dotenv.config();
+const result = dotenv.config();
+
+if (result.error && logger) {
+  logger.error(
+    '\n\n\t💡  It appears that you have not yet configured a .env file.',
+    '\n\t   Please refer to our documentation regarding environment configuration:',
+    '\n\t   https://github.com/Seneca-CDOT/telescope/blob/master/docs/CONTRIBUTING.md\n'
+  );
+}
--- a/src/backend/utils/wiki-feed-parser.js
+++ b/src/backend/utils/wiki-feed-parser.js
@@ -14,7 +14,7 @@ const { JSDOM } = jsdom;
  */
 async function getWikiText(url) {
   try {
-    const response = await fetch(process.env.FEED_URL);
+    const response = await fetch(url);
     const data = await response.text();
 
     const dom = new JSDOM(data);
@@ -36,8 +36,13 @@ async function getWikiText(url) {
  * }
  */
 module.exports = async function() {
-  // NOTE: we expect this URL to the CDOT wiki feed list to exist in .env
-  const url = process.env.FEED_URL;
+  let url = process.env.FEED_URL;
+
+  if (!url) {
+    url = 'https://wiki.cdot.senecacollege.ca/wiki/Planet_CDOT_Feed_List';
+    logger.debug(`No value found for FEED_URL in env, using default ${url} instead`);
+  }
+
   const nameCheck = /^\s*name/i;
   const commentCheck = /^\s*#/;
 


src/backend/lib/redis.js
```diff
@@ -8,15 +8,15 @@ const { logger } = require('../utils/logger');
 const redisUrl = process.env.REDIS_URL || 'redis://127.0.0.1:6379';
 
 // Set MOCK_REDIS=1 to mock, MOCK_REDIS= to use real redis
-const useMockRedis = process.env.MOCK_REDIS;
+const useMockRedis = process.env.MOCK_REDIS || null;
```
manekenpix: Do we need all the `nulls` when env vars are used as logical values (true, false)? `undefined` is a [falsy value](https://developer.mozilla.org/en-US/docs/Glossary/Falsy).

tools/add-feed.js
```diff
@@ -30,7 +30,7 @@ async function add() {
       attempts: process.env.FEED_QUEUE_ATTEMPTS || 8,
       backoff: {
         type: 'exponential',
-        delay: process.env.FEED_QUEUE_DELAY_MS || 60 * 1000,
+        delay: (process.env.FEED_QUEUE_DELAY_MS || 60) * 1000,
```
manekenpix: I think the value in `FEED_QUEUE_DELAY_MS` is already in milliseconds (notice the MS at the end), so I think this should be:
`delay: process.env.FEED_QUEUE_DELAY_MS || (60 * 1000)`, which might not even need
(would have to check operator priority).

src/backend/index.js
```diff
@@ -30,7 +30,7 @@ async function enqueueWikiFeed() {
         attempts: process.env.FEED_QUEUE_ATTEMPTS || 8,
         backoff: {
           type: 'exponential',
-          delay: process.env.FEED_QUEUE_DELAY_MS || 60 * 1000,
+          delay: (process.env.FEED_QUEUE_DELAY_MS || 60) * 1000,
```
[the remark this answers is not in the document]
Silvyre: Good call; reverted these changes: https://github.com/Seneca-CDOT/telescope/pull/550/commits/68a4ad098973259828fbccbecef44a4aaebbda86

tools/add-feed.js
Silvyre: Good call; reverted this changes: https://github.com/Seneca-CDOT/telescope/pull/550/commits/68a4ad098973259828fbccbecef44a4aaebbda86

src/backend/lib/redis.js
Silvyre: I was thinking that the `nulls` might ideally and eventually be replaced with sensible default values. At least, I think that having the `nulls` there might serve as a reminder that calls to `process.env` need to be handled with care.

humphd: Agree, let's not use `null`.  This will already use `undefined` if not set, which is what we want.

Silvyre: Sure; `null` default values removed: https://github.com/Seneca-CDOT/telescope/pull/550/commits/273499f61abe90b37a46842b98f996b4f4f16a62

Silvyre:
> Also, can you add some code to `src/backend/lib/config.js` where we load the `env` using `dotenv`: we can check to see if it worked or not (i.e., if there is an `.env` file): https://www.npmjs.com/package/dotenv#config
> 
> If we get back `result.error`, we should probably either warn (like you did with redis) or perhaps throw and let the app crash now vs. fail later with random missing data. I think we should warn, so we don't have to mess with CI too much at this point.

How's this?
https://github.com/Seneca-CDOT/telescope/blob/1074030835be3adc9007d1ba4e327fba8a5d2eab/src/backend/lib/config.js#L1-L12

Because `src/backend/utils/logger.js` and `src/backend/lib/config.js` currently require each other, I added the `&& logger` condition to ensure that `logger.error()` doesn't get called before `logger` is defined.

Also, it is worth mentioning that `src/backend/lib/config.js` is imported by [a dozen other modules](https://github.com/Seneca-CDOT/telescope/search?p=2&q=require+config&unscoped_q=require+config), which will result in a dozen warning messages being logged if `.env` is missing. I am not sure how we might ensure that only one warning message is logged in this case... Any suggestions?

src/backend/lib/config.js
```diff
@@ -1,3 +1,12 @@
...
 
-dotenv.config();
+const result = dotenv.config();
+
+if (result.error && logger) {
+  logger.error(
+    '\n\n\t💡  It appears that you have not yet configured a .env file.',
+    '\n\t   Please refer to our documentation regarding environment configuration:',
+    '\n\t   https://github.com/Seneca-CDOT/telescope/blob/master/docs/CONTRIBUTING.md\n'
+  );
```
manekenpix: In the future, I think it'd be a good idea to add something like [process.exit(1)](https://stackoverflow.com/a/5266239/8628446) here.

src/backend/utils/logger.js
```diff
@@ -9,7 +9,7 @@ require('../lib/config');
 let logLevel = (process.env.LOG_LEVEL || 'info').toLowerCase();
 if (!pino.levels.values[logLevel]) {
   // Use `debug` by default in development mode, `info` otherwise.
-  logLevel = process.env.NODE_ENV === 'development' ? 'debug' : 'info';
+  logLevel = !process.env.NODE_ENV || process.env.NODE_ENV === 'development' ? 'debug' : 'info';
```
humphd: I don't understand this change.  We need to set one of `debug` or `info`.  What is the first part of this doing?  If `process.env.NODE_ENV` is `undefined`, the original code will return `'debug'` already.  I think you can revert this unless I'm missing something.

src/backend/utils/wiki-feed-parser.js
```diff
@@ -11,7 +11,9 @@ const { JSDOM } = jsdom;
  * That data is then returned as a Promise
  */
 module.exports.getData = function() {
-  return fetch(process.env.FEED_URL)
+  return fetch(
+    process.env.FEED_URL || 'https://wiki.cdot.senecacollege.ca/wiki/Planet_CDOT_Feed_List'
```
humphd: This is probably wise, since this URL doesn't change very often (if ever).  But could you add a log that says we're doing this? Something like:

```js
let ur = process.env.FEED_URL;
if(!url) {
  url = 'https://wiki.cdot.senecacollege.ca/wiki/Planet_CDOT_Feed_List'
  logger.debug('No value found for FEED_URL in env, using default ${url} instead');
}
```

This will help us debug things later, if we can't figure out why it's always loading a URL we don't intend.

src/backend/web/routes/login.js
```diff
@@ -13,7 +13,7 @@ router.use(cookieParser());
 router.use(bodyParser.urlencoded({ extended: false }));
 router.use(bodyParser.json());
 
-const { SAML2_CLIENT_SECRET } = process.env;
+const SAML2_CLIENT_SECRET = process.env.SAML2_CLIENT_SECRET || 'secret';
```
humphd: This seems like a security issue.  Let's not provide any default fallback if we're missing a secret.  We should fail loudly if we are.

src/backend/utils/logger.js
Silvyre: > 
> If `process.env.NODE_ENV` is `undefined`, the original code will return `'debug'` already.

Sorry, Dave, but my understanding is that, since `undefined === 'development'` evaluates to false, the enclosing ternary operation would return 'info'. However, I did also assume that CI would want to utilize a more verbose log level, and may be wrong about this. Let me know.

src/backend/login/usingPassport.js
```diff
@@ -12,7 +12,8 @@ SAML2_REDIRECT_URI=http://localhost:3000/oauth/callback
 */
 let cert = null;
 
-const { SAML2_BASE_URI, SAML2_REDIRECT_URI } = process.env;
+const SAML2_BASE_URI = process.env.SAML2_BASE_URI || '';
```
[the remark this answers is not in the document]
Silvyre: Sure, done: https://github.com/Seneca-CDOT/telescope/pull/550/commits/e4c5a12a32b21a76efdeaf8feede880ddb3f31d6

Currently, `env.example` assigns an empty string to be the value of `SAML2_BASE_URI`:
https://github.com/Seneca-CDOT/telescope/blob/08c55d4d694cf7b8a2c37b7945606c792976191d/env.example#L28-L29

Let me know if you'd like me to remove this value, as well.

src/backend/utils/wiki-feed-parser.js
Silvyre: Great idea! Done: https://github.com/Seneca-CDOT/telescope/pull/550/commits/e4c5a12a32b21a76efdeaf8feede880ddb3f31d6

src/backend/web/routes/login.js
Silvyre: For sure... done: https://github.com/Seneca-CDOT/telescope/pull/550/commits/e4c5a12a32b21a76efdeaf8feede880ddb3f31d6

Currently, `env.example` mentions `secret` to be the default value of `SAML2_CLIENT_SECRET`:
https://github.com/Seneca-CDOT/telescope/blob/08c55d4d694cf7b8a2c37b7945606c792976191d/env.example#L34

Let me know if you'd like me to remove this piece of information, as well.

I'll also ask if you'd like me to do the same for `SAML2_CLIENT_ID`—its default value is likewise mentioned:
https://github.com/Seneca-CDOT/telescope/blob/08c55d4d694cf7b8a2c37b7945606c792976191d/env.example#L31

Silvyre:
Rebase completed; thank you everyone for you help!

PR #565, opened 2020-01-21
Closes #522: Standardize on base64 encoded, SHA-256 hashed, normalized URIs as Redis keys

### LINKED ISSUE #522: Standardize on base64 encoded, SHA-256 hashed, normalized URIs as Redis keys

Our `Feed` (see #520) and `Post` objects need to live in Redis.  I've been doing a bit of research on the best way to do this, and here is what I suggest:

1. A `Feed` is a Redis Hash stored using the feed's URL as its key.  Let's use [normalize-url](https://www.npmjs.com/package/normalize-url) to clean it up a bit (so we don't have slight variations leading to duplication) and then hash it using MD5 and base64 encode it (see below).  We should also be namespacing our keys.  I suggest `t:feed:<base64 encoded MD5 hash>`, where `t=Telescope`.
2. A `Post` is a Redis Hash stored using the article's `guid`, which is (likely) a URI or URL.  Let's hash using MD5 and then base64 encode it.  I suggest `t:post:<base64 encoded MD5 hash>`.

To hash and encode these URIs in node, do this:

```
const crypto = require('crypto')

function toBase64(url) {
  return crypto.createHash('md5').update('some_string').digest('base64');
}
```

We'd also need a way to compare a URL with a key, so we can check if a URI is already in the database.

I noticed [this post](https://groups.google.com/forum/m/#!topic/redis-db/dPW3m9t6vk4) by Redis' creator suggesting this approach for storing URLs.

manekenpix:
## Issue This PR Addresses
Closes #522 

## Type of Change

## Description

In order to standardize feeds and posts:
Feeds are normalized using [normalize-url](https://www.npmjs.com/package/normalize-url), and both feeds and posts are hashed using SHA256,  base64 encoded and namespaced (`t:feed:` or `t:post:`) before being stored.

## Checklist


package.json
```diff
@@ -54,6 +54,7 @@
     "bull-board": "0.5.0",
     "cookie-parser": "1.4.4",
     "cors": "2.8.5",
+    "crypto": "1.0.1",
```
humphd: You can remove this, it's part of node itself now, see https://www.npmjs.com/package/crypto.

src/backend/utils/standardize.js
```diff
@@ -0,0 +1,25 @@
+const crypto = require('crypto');
```
humphd: I don't think that we should let the key naming code leak out of `storage.js`.  It's a detail only for communication from Redis, and the rest of our code should be isolated from it.

If someone wants to get/set a feed, they think in terms of a `url`.  If someone wants to get/set a post, they think in terms of a `guid`.  We should do all the key namespacing quietly at the point of interaction with Redis, and then strip it out when we communicate things back to the user.

humphd: The actual logic of what you're doing in here is great, I just want to hide it all from the rest of our code vs. making it a module you can require elsewhere.

src/backend/utils/storage.js
```diff
@@ -29,11 +27,13 @@ module.exports = {
   getFeedsCount: () => redis.scard(FEEDS),
 
   addPost: async post => {
+    const encrGuid = standardize(post.guid, 'post');
```
humphd: This isn't encrypted, so maybe call it `hashedGuid` or just `key` to better communicate its use.

test/posts.test.js
```diff
@@ -126,15 +127,15 @@ describe('test /posts/:guid responses', () => {
 
   // tests
   it("pass a guid that doesn't exist", async () => {
-    const res = await request(app).get(`/posts/${missingGuid}`);
+    const res = await request(app).get(`/posts/${standardize(missingGuid, 'post')}`);
```
humphd: Question about this.  Given the following data:

- a URL of `https://some-string.com/feed`
- hashed/encoded URL is `VM/xbiXC58KgqEqAdrzDZKR4d4UmXwuubY3/dNDlLhY=`

Our API can expose one of three things:

1. `/posts/VM/xbiXC58KgqEqAdrzDZKR4d4UmXwuubY3/dNDlLhY=`
1. `/posts/t:post:VM/xbiXC58KgqEqAdrzDZKR4d4UmXwuubY3/dNDlLhY=`
1. `/posts/https://some-string.com/feed`

I think we should automatically eliminate 2., because it once again leaks an internal implementation detail with Redis.  That leaves the raw URL string, or the hashed/encoded version.  Of the two I think that 3. is the better choice, and the more common.

As such, my proposal would be this:

- When we `addPost()` into Redis, we take the `guid` given in the post and immediately hash/encode it, then throw away the original.  We use this hash/encoded version instead for the rest of the time we deal with this post.
- We continue to use this new `guid` the same way we used the old one.
- When we get the array of posts from Redis, it will contain keys.  We need to `posts.map(key => key.replace(/^t:post:/, '')` and only return the guid vs. the namespaced-guid-key.
- When we call `Post.byGuid()` we will pass it this guid, and it will work with the storage/Redis code to namespace it before trying to look it up.
- For testing, we'll probably want to move the code to hash/encode the URI to the `Post` class, so that unit tests can call it.
- The process of hash/encoding and namespacing can thus be separated.

Does this make sense?  I'm thinking out loud and what I've written might have issues.  Let's discuss.

src/backend/utils/standardize.js
manekenpix: I moved all the logic to `storage.js`, and removed `standardize.js`.

src/backend/utils/storage.js
manekenpix: Renamed `encrGuid` to `key`.

package.json
manekenpix: Removed `crypto`.

test/posts.test.js
manekenpix: I moved all the logic to `storage.js` so now tests and other components are completely blind to how redis stores keys. Now `/posts/` returns an array of unhashed, unencrypted guids.

src/backend/utils/storage.js
```diff
@@ -1,16 +1,38 @@
+const crypto = require('crypto');
+const normalizeUrl = require('normalize-url');
+const { logger } = require('./logger');
 const { redis } = require('../lib/redis');
 
+const standardize = (url, type) => {
```
humphd: Can we rename this `createKey(uri, prefix)`?  We will either pass a URL or a URI, so URI makes sense here, and the "type" is really just a string prefix in our key's namespace.

src/backend/utils/storage.js
```diff
@@ -1,16 +1,38 @@
...
+  let namespace = 't:post:';
```
humphd:
```js
const namespace = `t:${type}:`;
```

src/backend/utils/storage.js
```diff
@@ -1,16 +1,38 @@
...
+  let processed = url;
+
+  if (type === 'feed') {
```
humphd: This feels brittle to me. Can we separate out the `normalizeUrl(url)` call, so that it happens before we pass in the URL to this function?  That way you don't need to worry about it here, and the caller can decide if it's needed.

src/backend/utils/storage.js
```diff
@@ -1,16 +1,38 @@
...
+    );
+  } catch (error) {
+    logger.error(`There was an error processing ${url}`);
+    throw error;
+  }
+};
+
 // Redis Keys
-const FEED_ID = 'feed_id';
 const FEEDS = 'feeds';
```
humphd: Based on what we're doing with namespaces, this should probably become `t:feeds` and `t:posts` below.

src/backend/utils/storage.js
```diff
@@ -1,16 +1,38 @@
...
 
 const POSTS = 'posts';
 
 module.exports = {
   addFeed: async (name, url) => {
-    // "If the key does not exist, it is set to 0 before performing the operation"
-    // https://redis.io/commands/INCR
-    const feedId = await redis.incr(FEED_ID);
+    const feedId = standardize(url, 'feed');
```
humphd: Let's call this what it is:

```js
const key = ...;
```

src/backend/utils/storage.js
humphd: One way to keep this a single function call would be to do:

```js
function createPostKey(uri) {
  const normalizedUri = normalizeUrl(url);
  return createKey(normalizedUri, 'post');
}

function createFeedKey(uri) {
  return createKey(uri, 'feed');
}
```

And let the common bits happen in `createKey()`.

src/backend/utils/storage.js
```diff
@@ -54,19 +78,34 @@ module.exports = {
...
    * @param from lower index
    * @param to higher index, it needs -1 because redis includes the element at this index in the returned array
    * @return Array of guids
    */
-  getPosts: (from, to) => redis.zrevrange(POSTS, from, to - 1),
+  getPosts: async (from, to) => {
+    const keys = await redis.zrevrange(POSTS, from, to - 1);
+
+    /**
+     * 'zrevrange returns an array of encrypted hashed guids.
```
humphd: NOTE: they aren't encrypted, only hashed (sha256) and encoded (base64).

src/backend/utils/storage.js
```diff
@@ -54,19 +78,34 @@ module.exports = {
...
+     * This array is used to return the 'guid' property in
+     * Post objects, which contains the unencrypted, unhashed version
+     * of the guid
+     */
+    return Promise.all(
+      keys.map(async key => {
+        const { guid } = await redis.hgetall(key);
+        return guid.replace('/t:post:/', '');
```
humphd:
```js
/^t:post:/
```

src/backend/utils/storage.js
```diff
@@ -1,25 +1,49 @@
...
 const FEEDS = 'feeds';
 
 const POSTS = 'posts';
 
 module.exports = {
   addFeed: async (name, url) => {
-    // "If the key does not exist, it is set to 0 before performing the operation"
-    // https://redis.io/commands/INCR
-    const feedId = await redis.incr(FEED_ID);
+    const key = createFeedKey(url, 'feed');
```
humphd: You can drop second arg here.

src/backend/utils/storage.js
```diff
@@ -29,11 +53,13 @@ module.exports = {
   getFeedsCount: () => redis.scard(FEEDS),
 
   addPost: async post => {
+    const key = createPostKey(post.guid, 'post');
```
humphd: You can drop second arg here too

src/backend/utils/storage.js
```diff
@@ -54,19 +80,34 @@ module.exports = {
...
+        const { guid } = await redis.hgetall(key);
+        return guid.replace('/^t:post:/', '');
+      })
+    );
+  },
 
   getPostsCount: () => redis.zcard(POSTS),
 
-  getPost: async guid => redis.hgetall(guid),
+  getPost: async guid => redis.hgetall(createPostKey(guid, 'post')),
```
humphd: And here, too, drop `'post'` second arg.

src/backend/utils/storage.js
humphd: The `FEEDS` and `POSTS` keys still need namespacing `t:...`

src/backend/utils/storage.js
manekenpix: Renamed to `createKey(uri, prefix)`.

src/backend/utils/storage.js
manekenpix: Added your suggestion, definitely clearer and easier to understand. Thank you!

src/backend/utils/storage.js
manekenpix: Renamed to `key`.

src/backend/utils/storage.js
manekenpix: Fixed comment

src/backend/utils/storage.js
manekenpix: Added `^`.

src/backend/utils/storage.js
```diff
@@ -1,25 +1,49 @@
...
       // Using hmset() until hset() fully supports multiple fields:
       // https://github.com/stipsan/ioredis-mock/issues/345
       // https://github.com/luin/ioredis/issues/551
-      .hmset(feedId, 'name', name, 'url', url)
-      .sadd(FEEDS, feedId)
+      .hmset(key, 'name', name, 'url', url)
+      .sadd(FEEDS, key)
       .exec();
-    return feedId;
+    return key;
```
humphd: Question: should we return the key, or the key with the namespace removed?

humphd: I wonder if we should return nothing here.  The only place I see us using this is in one unit test that checks the return type.  We could just remove that test.  I don't think we ever need to use the key that we get back, since we know that it's based on the URL.

Yeah, let's remove this line and the test.

src/backend/utils/storage.js
```diff
@@ -1,39 +1,73 @@
+const crypto = require('crypto');
```
humphd: Remove this, it's not needed.

src/backend/utils/storage.js
```diff
@@ -1,39 +1,73 @@
...
+};
+
+const createFeedKey = uri => {
+  return createKey(uri, feedNamespace);
+};
+
 // Redis Keys
-const FEED_ID = 'feed_id';
-const FEEDS = 'feeds';
+const FEEDS = 't:feeds';
```
humphd: Let's move these two key names up to the top where you define your other namespaces, and make the names more consistent.  You have `feedNamespace` and `postNamespace`, which is good.  Let's rename `POSTS` to `postsKey` and `FEEDS` to `feedsKey`, so it's clear what all of these things are.

src/backend/utils/storage.js
```diff
@@ -1,39 +1,73 @@
...
+const createKey = (url, prefix) => {
+  try {
+    return prefix.concat(
+      crypto
+        .createHash('sha256')
+        .update(url)
+        .digest('base64')
+    );
+  } catch (error) {
+    logger.error(`There was an error processing ${url}`);
```
humphd:
```js
logger.error({ error }, `There was an error hashing the url ${url}`);
```

src/backend/utils/storage.js
```diff
@@ -1,39 +1,73 @@
...
 
-  getFeeds: () => redis.smembers(FEEDS),
+  getFeeds: async () => {
+    const keys = await redis.smembers(FEEDS);
+
+    /**
+     * It's necessary to remove the namespace from all the ids
+     * before returning them
+     */
+    return keys.map(key => key.replace(/^t:feed:/, ''));
```
humphd:
```js
const feedNamespaceRe = new RegExp(`^${feedNamespace}`);
return keys.map(key => key.replace(feedNamespaceRe, ''));
```

src/backend/utils/storage.js
```diff
@@ -54,19 +88,27 @@ module.exports = {
...
    */
-  getPosts: (from, to) => redis.zrevrange(POSTS, from, to - 1),
+  getPosts: async (from, to) => {
+    const keys = await redis.zrevrange(POSTS, from, to - 1);
+
+    /**
+     * It's necessary to remove the namespace from all the ids
+     * before returning them
+     */
+    return keys.map(key => key.replace(/^t:post:/, ''));
```
humphd:
```js
const postNamespaceRe = new RegExp(`^${postNamespace}`);
return keys.map(key => key.replace(postNamespaceRe, ''));
```

test/posts.test.js
```diff
@@ -1,8 +1,17 @@
+const crypto = require('crypto');
+const normalizeUrl = require('normalize-url');
 const request = require('supertest');
 const app = require('../src/backend/web/app');
 const Post = require('../src/backend/post');
 const { addPost } = require('../src/backend/utils/storage');
 
+const encodeKey = url => {
```
humphd: In a follow-up, I'm going to move this into the logic of the `Post` and `Feed` classes, so it's not something we have to do externally (i.e., it should happen automatically for us).  For now, this is fine.

src/backend/utils/storage.js
humphd: Not enough ☕️ yet this morning, ignore this.  I'm wrong.

PR #389, opened 2019-12-01
Fix #190: Create an template for email-sender

RyanWils:
I created a template for the email-sender, this will make it easier for the users to identify the errors that are coming in from the Telescope project. 

Currently when you send an email, it would send plain text for the message that is passed. 
image

The functionality for the email sender did not change, but is displayed in the following format
image

I created the email template using HTML Table elements and made a simplistic design which will make it easier for the code to be maintained/changed if necessary.

PR #571, opened 2020-01-21
Reduce eslint Promise warnings, refactor wiki parser

humphd:
## Type of Change

## Description

This is follow-up to the work done to add eslint for Promises and async/await.  I've reduced the warnings we get down to 1.  Do do it, I made the following changes:

- switched to async/await in the shutdown code
- rewrote the wiki-feed-parser to use async/await
- refactored the wiki-feed-parser so that it doesn't expose the `getData()` function anymore.  There is no need to expose this.  I've also removed the tests for this.  The other tests we have already run that code.

There is still one warning:

```
/Users/humphd/repos/telescope/src/backend/utils/email-sender.js
  67:22  warning  Avoid callbacks. Prefer Async/Await  promise/prefer-await-to-callbacks
```

My guess is we'll rip this code out soon, so I didn't bother fighting with it.

### DIFF
--- a/src/backend/index.js
+++ b/src/backend/index.js
@@ -2,7 +2,7 @@ require('./lib/config.js');
 const feedQueue = require('./feed/queue');
 const feedWorker = require('./feed/worker');
 const { logger } = require('./utils/logger');
-const wikiFeed = require('./utils/wiki-feed-parser');
+const getWikiFeeds = require('./utils/wiki-feed-parser');
 const shutdown = require('./lib/shutdown');
 
 // Start the web server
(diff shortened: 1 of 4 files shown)


cindyledev:
I can comment on the `email-sender.js` since I reviewed #389 Create a template for email-sender. I agree that it should be ripped out because it uses nodemailer with gmail and this will only work in development, not in production. I tried building a website using this method and gmail actually blocked my website from sending emails because my server was located in America and it thought someone was trying to hack into my account. I ended up scrapping the whole thing and used a different method. 

According to [nodemailer](https://nodemailer.com/usage/using-gmail/), the preferred method is to use OAuth2 authentication but I think we're already trying to get SSO working so this `email-sender.js` is useless.

PR #572, opened 2020-01-21
Use Let's Encrypt for certs on staging 

### LINKED ISSUE #519: Use Let's Encrypt for certs on staging

When we ship to staging (and eventually production), we'll need proper certs that aren't self-signed.  We can [use Let's Encrypt](https://www.humankode.com/ssl/how-to-set-up-free-ssl-certificates-from-lets-encrypt-using-docker-and-nginx) for free.  We'll need to do this automatically when we push to staging.

We'll need to run our app over HTTPS when not in a dev setup.

miggs125:
closes #519 

## Type of Change

## Checklist

Ok so this should add ssl to telescope on the specified port. I've beenr eading a few guides and most of them keep port 80 open and the redirect any connection to the ssl server. I have not included this as we have not had a discussion as to what we would like to do with http (shut it down or simply redirect), but apparently the usualy thing to do is redirect any http connections to https and keep both servers running.

In order to get the certificate the first time, letsencrypt asks for the domain name of the machine running the server and attempts to fetch a static file with a random generated name and content. To serve this file, adding the `app.use(express.static(`${__dirname}/static`, { dotfiles: 'allow' }));
` line in server.js is needed. I am not sure if we need to do this for the auto-renewal so I've left the line in just in case. We can get rid of it in the future if needed. 

letsencrypt certs expire every 90 days. We can automate this using a crone job as such:
`0 */12 * * * <user/root> path/to/certbot/bin renew >/dev/null 2>&1` 
The above says ‘run it every 12 hours, every day: at 00:00 and at 12:00’.

These are the guides I used for reference:

https://advancedweb.hu/how-to-use-lets-encrypt-with-node-js-and-express/#scenario-3-http-redirect-with-a-public-folder

https://flaviocopes.com/express-letsencrypt-ssl/#setup-the-renewal

https://medium.com/@yash.kulshrestha/using-lets-encrypt-with-express-e069c7abe625

image

### DIFF
--- a/env.example
+++ b/env.example
@@ -3,6 +3,7 @@ NODE_ENV=development
 
 # PORT is the port used by the web server
 PORT=3000
+HTTPS=4443
 
 # LOG_LEVEL is used to set the level of debugging for the logs.
 # info, error and debug are commonly used levels. See http://getpino.io/#/docs/api?id=level for more info on levels.
--- a/src/backend/web/app.js
+++ b/src/backend/web/app.js
@@ -4,6 +4,7 @@ const expressHandlebars = require('express-handlebars');
 const healthcheck = require('express-healthcheck');
 const cors = require('cors');
 const { ApolloServer } = require('apollo-server-express');
+const helmet = require('helmet');
 
 const logger = require('../utils/logger');
 const router = require('./routes');
@@ -29,7 +30,11 @@ app.set('view engine', 'handlebars');
 app.set('logger', logger);
 app.use(logger);
 
+app.use(express.static(`${__dirname}/static`, { dotfiles: 'allow' }));
+
 app.use(cors());
+
+app.use(helmet());
 
 app.use('/health', healthcheck());
 
--- a/src/backend/web/server.js
+++ b/src/backend/web/server.js
@@ -1,11 +1,30 @@
 require('../lib/config');
+const https = require('https');
+const fs = require('fs');
+
 const app = require('./app.js');
 
 const { logger } = app.get('logger');
 const HTTP_PORT = process.env.PORT || 3000;
+const HTTPS_PORT = process.env.HTTPS_PORT || 4443;
+
+const { key, cert } = (() => {
+  const certdir = fs
+    .readdirSync('/etc/letsencrypt/live', { withFileTypes: true })
+    .filter(file => file.isDirectory())[0].name;
+
+  return {
+    key: fs.readFileSync(`/etc/letsencrypt/live/${certdir}/privkey.pem`),
+    cert: fs.readFileSync(`/etc/letsencrypt/live/${certdir}/fullchain.pem`),
+  };
+})();
 
 const server = app.listen(HTTP_PORT, () => {
   logger.info(`Telescope listening on port ${HTTP_PORT}`);
 });
+
+https
+  .createServer({ key, cert }, app)
+  .listen(HTTPS_PORT, () => console.log(`HTTPS server listening on ${HTTPS_PORT}`));
 
 module.exports = server;
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
     "express-session": "1.17.0",
     "feedparser-promised": "2.0.1",
     "get-urls": "9.2.0",
+    "helmet": "3.21.2",
     "ioredis": "4.14.1",
     "ioredis-mock": "4.18.4",
     "jsdom": "15.2.1",


env.example
```diff
@@ -3,6 +3,7 @@ NODE_ENV=development
 
 # PORT is the port used by the web server
 PORT=3000
+HTTPS=4443
```
humphd: Let's prefix these both, and update the comment above to mention the difference.

```
HTTP_PORT=3000
HTTPS_PORT=4443
```

src/backend/web/app.js
```diff
@@ -29,8 +30,12 @@ app.set('view engine', 'handlebars');
 app.set('logger', logger);
 app.use(logger);
 
+app.use(express.static(`${__dirname}/static`, { dotfiles: 'allow' }));
```
humphd: Add a comment above this.  You can use your explanation from the PR about why we need this and what it's used for.

src/backend/web/server.js
```diff
@@ -1,11 +1,30 @@
...
 
 const server = app.listen(HTTP_PORT, () => {
   logger.info(`Telescope listening on port ${HTTP_PORT}`);
 });
 
+https
+  .createServer({ key, cert }, app)
+  .listen(HTTPS_PORT, () => console.log(`HTTPS server listening on ${HTTPS_PORT}`));
+
 module.exports = server;
```
humphd: We'll need to be able to close both of these servers during shutdown, so we should probably expose both on exports.

src/backend/web/server.js
```diff
@@ -1,11 +1,30 @@
...
+
 const app = require('./app.js');
 
 const { logger } = app.get('logger');
 const HTTP_PORT = process.env.PORT || 3000;
+const HTTPS_PORT = process.env.HTTPS_PORT || 4443;
+
+const { key, cert } = (() => {
+  const certdir = fs
+    .readdirSync('/etc/letsencrypt/live', { withFileTypes: true })
```
humphd: These paths should probably go in the `.env`.  We won't change them (often), but if we do deploy and need them updated, it's going to be a pain to find these.

src/backend/web/server.js
```diff
@@ -1,11 +1,30 @@
...
+const https = require('https');
+const fs = require('fs');
+
 const app = require('./app.js');
 
 const { logger } = app.get('logger');
 const HTTP_PORT = process.env.PORT || 3000;
+const HTTPS_PORT = process.env.HTTPS_PORT || 4443;
+
+const { key, cert } = (() => {
```
humphd: If the HTTPS port isn't set, can we skip loading these?  If we leave it like this, it will break for every developer locally.

src/backend/web/server.js
```diff
@@ -1,11 +1,30 @@
...
+  const certdir = fs
+    .readdirSync('/etc/letsencrypt/live', { withFileTypes: true })
+    .filter(file => file.isDirectory())[0].name;
+
+  return {
+    key: fs.readFileSync(`/etc/letsencrypt/live/${certdir}/privkey.pem`),
```
humphd: Same for these, move to `.env`

humphd:
Reading more about this, I think doing our HTTPS config in node is not the right way to do this long term.  We could go with this PR to get us moving forward on deployment, but the better approach would be to have a reverse proxy (probably nginx) deal with all traffic on 80/443 and route to our node app, which can just deal with "web traffic" in general vs. secure vs. insecure web traffic.

Whenever I've done what we're doing in production with Mozilla, we've always used a setup with nginx (or a load balancer) sitting between the internet and the node app.  Probably that's where we want to head in the near term.

One advantage to going with this more established method is that the whole Let's Encrypt workflow is more straightforward, and devs working locally can ignore all the SSL bits that this PR is struggling to do without breaking changes for everyone.  One of many examples describing it:

https://www.digitalocean.com/community/tutorials/how-to-secure-a-containerized-node-js-application-with-nginx-let-s-encrypt-and-docker-compose

How should we handle this?  Should we go with this solution now, and later rip it out in favour of a reverse proxy/ingress solution, or should we do that first?

@miggs125, @manekenpix, @c3ho: you three seem the most invested in the deployment work at the moment, what are you thoughts?

cindyledev:
Not sure why the Travis check is showing up as 'pending' for so long here (it's been like this for a day); it's passing when you click on Details.
Annotation 2020-01-22 174851

humphd:
I'll restart it on Travis, and see if it updates with GitHub.  The githook must not have fired.

humphd:
I think we should hold-off on this, and see if we can do it via nginx as a reverse proxy on staging, and not do HTTPS in our node app.  We can morph this into scripts/docs that we land to tell people how to deploy our staging box.

humphd:
https://medium.com/intrinsic/why-should-i-use-a-reverse-proxy-if-node-js-is-production-ready-5a079408b2ca is a useful read on the subject of preferring a reverse proxy like nginx for this.

Issue #591, opened 2020-01-23
Add filters to GraphQL to remove getPost


manekenpix:
**What happened**:

Recently #565 was merged and the way we store posts and feeds changed. That affected GraphQL, specifically our [getPost](https://github.com/Seneca-CDOT/telescope/blob/08c55d4d694cf7b8a2c37b7945606c792976191d/src/backend/web/graphql/index.js#L40) query. `getPost` was kept for legacy reasons, so this issue wont' prevent Telescope from serving all available posts.
Instead of patching `getPost`, it'd be better if we implement filters that would work with our getPosts query, that way we could filter by any subfield included in our Post type.

**How to reproduce it (as precise as possible)**:

Use the getPost query passing a `guid` as argument.

c3ho:
#592 may or may not fix some of the concerns

c3ho:
Got an issue with filters, Prisma doesn't support Redis as of yet... So all the documents on Apollo Server doesn't apply. I just spent 3/4 hours figuring this out.

humphd:
I don't follow, we can connect to redis ourselves, we don't need Prisma afaict.

c3ho:
https://www.howtographql.com/graphql-js/8-filtering-pagination-and-sorting/. I think to use any of this it relies on things generated by Prisma.

humphd:
This stuff is more for when you have a really big database/query set, and need to automate a lot of the code.  For our needs, we're going to have a handful of things we need to return and query by.  Let's find a way to accomplish what we need without worrying about Prisma.

c3ho:
Ok, so you're fine with the way I kind of had it in the blog to test out nested queries? I was trying to find some "elegant" way to do it like a traditional DB, instead of retrieving all the posts then filtering it.

humphd:
I appreciate you trying to do this in an elegant, idiomatic way; but I want it to ship on time even more.  Let's go with the stuff you were doing in your blog, and refine that as we learn more about our needs.  Shipping Code > Perfect Code (that never ships).

PR #592, opened 2020-01-23
Refactor storage to use Post and Feed data classes

### LINKED ISSUE #498: Increase test coverage of Post type in posts.js

In #497 I updated the `Post` class, and added a bunch of new code.  We don't have full test coverage for all of it yet.

```
$ npm test
...
--------------------------|----------|----------|----------|----------|-------------------|
File                      |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
--------------------------|----------|----------|----------|----------|-------------------|
  post.js                 |    42.86 |        0 |        0 |    42.86 |... 15,16,20,23,24
```

humphd:
## Issue This PR Addresses

Fixes #498 

## Description

This reduces the amount code that needs to interact directly with `storage.js`, and instead provides data APIs for a `Post` and `Feed`.  I've also added a lot more tests.

The `Post` and `Feed` data objects have been augmented with an `.id` property, which is the first 10 characters of a base64 encoded sha256 hash.  I've reduced the length of our hashes, since we don't have enough data to require all that precision (this is like what git/GitHub do when they shorten commit SHAs--you only need the first 8-10 characters in practice).

I've also changed how we store items in our `t:feeds` and `t:posts` Redis Sets.  Instead of storing keys, I now store just the hashed id.  This reduces complexity when going from external web APIs to internal data storage: the Redis code adds the key namespace when necessary, but clients can ignore all that.  I think this makes it easier.

I've added a way to query for Feeds by both id and URL, though I'm pretty sure my `typeDefs` for GraphQL are wrong.  I could use some help with that.

I'd still like to figure out a sane API for getting all Posts and Feeds, as well as counts, using the `Post` and `Feed` classes.  Doing so would let us always go through our data classes to get at Redis, and not have data type problems when we expect our data to look a certain way.


src/backend/data/post.js
```diff
@@ -66,6 +68,17 @@ class Post {
...
+      logger.debug('article missing pubdate, substituting current date');
+      article.pubdate = today;
+    }
+    if (!article.date) {
+      logger.debug('article missing date, substituting current date');
+      article.date = today;
+    }
+
     let sanitizedHTML;
     let plainText;
```
c3ho: I don't see us using these vars or am I wrong?

c3ho:
I think what I'm trying to wrap my head around is, aren't all the APIs for ioredis async? Don't we need await/async for the methods such as ```getFeed: feedID => redis.hgetall(feedNamespace.concat(feedID))```

manekenpix:
@c3ho I think the reason is that if you have this
`async foo () => await bar()`
you'd be (a)waiting for the promise returned by bar, and then once it's done, foo itself returns a promise.
So instead of waiting for a promise, and then returning another one, you can just let the original promise (the one returned by bar) "go through". 
Consider this:

```js
const bar = () => {
return 5
};

const foo = () => {
const rt = bar();
return rt;
}
```

Wouldn't you do?
```js
const foo() => bar();
```

Imagine the same situation, but instead of `5`, with a promise

src/backend/data/post.js
humphd: Lines 87-89 below.  It's not shown with the default level of context, but it looks like this:

```js
let sanitizedHTML;
let plainText;

try {
  // The article.description is frequently the full HTML article content.
  // Sanitize it of any scripts or other dangerous attributes/elements
  sanitizedHTML = sanitizeHTML(article.description);
  // Also generate plain text from the sanitized HTML
  plainText = textParser(sanitizedHTML);
} catch (err) {
  logger.error({ err }, 'Unable to sanitize and parse HTML for feed');
  throw err;
}
```

src/backend/web/graphql/index.js
```diff
@@ -34,49 +31,52 @@ module.exports.typeDefs = graphql`
...
+    getFeedById(id: ID!): Feed
+    getFeedByUrl(url: String!): Feed
+    getFeeds: [Feed]
+    getFeedsCount: Int
+    getPost(id: ID!): Post
     getPosts(page: Int, perPage: Int): [Post]
-    getPostsCount: Float
+    getPostsCount: Int
   }
 `;
```
manekenpix: Just a comment.
The proposal in #591 aims to have fewer queries, but very flexible ones. Instead of having a query for one object and another one for a list, we can add filters to the one that returns a list. Something like:
```
{
  getPosts(page:0, perPage: 20, { filter: { author: Josue } } ) {
    title
    published
    text
    html
  }
}
```
That combined with restructuring our schema would lead to something like:
```
type Feed {
  id: String
  name: String
  url: String
  posts: [Post]    <- Array of Posts
}
```
And our queries could become more flexible and effective:
```
{
  getFeeds( { filter: { author: Josue } } ) {    <- that could be id, url...
    url
    posts {
      title
      html
      published
      ...
    }
  }
}
```

src/backend/web/graphql/index.js
```diff
@@ -1,26 +1,23 @@
...
+const Post = require('../../data/post');
+const Feed = require('../../data/feed');
 
 const maxPostsPerPage = process.env.MAX_POSTS_PER_PAGE || 30;
 
 module.exports.typeDefs = graphql`
   # 'Feed' matches our Feed type used with redis
   type Feed {
+    id: String
     name: String
```
manekenpix: We have author in our `Feed` class, we could use that here for consistency.

humphd:
@c3ho the explanation by @manekenpix  is right.  Here's another example to consider.

Imagine that I have a function `getUserFromDatabase()`, and it is asynchronous.  We know that this mean that "it returns a Promise," since the main thread can't block while we go talk to an external resource (like the database, filesystem, network, etc).

So we could do this:

```js
async function f() {
  const user = getUserFromDatabase();
  return user;
}
```

What that cod really says is:

```js
function f() {
  return new Promise(resolve => {
    getUserFromDatabase().then(user => resolve(user));
  });
}
```

Now, what if instead of assigning `user` to a variable, I just return it:

```js
async function f() {
  return await getUserFromDatabase();
}
```

I'm really saying:

```js
function f() {
  return Promise.resolve(getUserFromDatabase());
}
```

This will work.  But it will cause a performance penalty, since I now have two off-thread tasks that have to get done (a Promise happens off the main thread).  What if I just returned the Promise from `getUserFromDatabase()` directly, let it "go through" as @manekenpix says, and didn't bother wrapping it in another Promise.resolve()?

```js
function f() {
  return getUserFromDatabase();
}

// or more commonly...
const f = () => getUserFromDatabase();
```

So here's one way to think about it:

- Any function with `async` on it will get its `return` wrapped in a `Promise.resolve()` and `throw` in `Promise.reject()` automatically.  So when you see `async function() {...}` or `async () => {...}` you can tell yourself, "This function returns a Promise."
- A function *without* `async` can also return a Promise, either because you create one and return it, or because you return the result of another function which returns one.  The `async` isn't required for a function that returns a Promise.
- You *can* add `async` to a function where you don't `await`, but it isn't necessary if you already return a Promise *manually*.  Our linting will recommend you don't every do this.
- When you `await`, you must use `async`, as you well know.

Does this help?  It's hard, and takes a lot of trial and error to understand.  Keep fighting with it, you're close.

src/backend/data/post.js
```diff
@@ -1,41 +1,45 @@
...
     this.title = title;
     this.html = html;
     this.text = text;
-    this.published = toDate(datePublished);
-    this.updated = toDate(dateUpdated);
+    this.published = datePublished ? toDate(datePublished) : new Date();
+    this.updated = dateUpdated ? toDate(dateUpdated) : new Date();
     this.url = postUrl;
     this.site = siteUrl;
     this.guid = guid;
```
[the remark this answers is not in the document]
humphd: Interesting idea.  I wonder if we'd be wise to keep it around, since it comes [from the RSS/Atom article standard](https://cyber.harvard.edu/rss/rss.html#ltguidgtSubelementOfLtitemgt) (that's where the name came from here), and is the field we use to calculate the hash.  If we throw it away, we can't ever confirm/debug our hashes.  I'm inclined to leave it for now, and we can ditch it later if we find that we really don't ever use it, after our code and data backend settle a bit more.

src/backend/web/graphql/index.js
humphd: Awesome.  What, if anything, should I be doing in this PR in anticipation of that?

Also, can you look at this line that I added:

```
getFeedByUrl(url: String!): Feed
```

I'm pretty sure that's wrong, but I don't know what it should be.  All of my changes to the graphql code feel uncertain to me, so let me know what I need to fix.

src/backend/web/graphql/index.js
humphd: Good idea, I'll do that.

src/backend/web/graphql/index.js
manekenpix: I think we can make changes to our queries once we add filters and we start combining our types (and maybe add some new ones).
Right now, what you have here is perfect.
`getFeedByUrl` looks good to me, same logic as `getFeedById` and `getPost`.

Just tested it. It's empty but GraphQL didn't complain about the query itself
Screenshot_20200123_210209

humphd:
I've made a few more changes based on testing:

1. I switched from base64 to hex encoding for the hash, since the former requires extra effort to use on a URL, whereas we can use hex string values raw on a URL.
1. I've added code to use the `Feed` and save them to Redis to make other parts of the code easier to test.  I think @Silvyre's work on OMPL, and @c3ho's work on the inactive feed PR will both benefit from this.

Here is a graphql query that uses the Feeds:

Screen Shot 2020-01-23 at 9 30 56 PM

I could use some more review/testing on this.

@manekenpix should our `Post` type include the `id` of the `Feed` so we can do "joins"?

humphd:
It also now skips invalid URLs instead of adding them to the list of feeds we process and store in redis:

```
[ 2020-01-23 21:43:05.730 ] INFO  (55088 on brightness.local): Skipping invalid wiki feed url blog.cataldo.ca/index.php/category/open-source/feed?v=1 for author Luca  Cataldo
[ 2020-01-23 21:43:05.731 ] INFO  (55088 on brightness.local): Skipping invalid wiki feed url https:https://robertbegnatechblog.wordpress.com/category/open-source/feed for author Robert Begna
[ 2020-01-23 21:43:05.732 ] INFO  (55088 on brightness.local): Skipping invalid wiki feed url https://danwithopensource.blogspot.com/feeds/posts/default/-/open source for author Name = Daniel Beigi
```

manekenpix:
@humphd I think that'd be great! I'm still getting familiar with "joins" so I don't know if Post should include Feed's subfields or the other way around.

[Here is an example](https://www.apollographql.com/docs/apollo-server/data/data/#resolver-map) from the Apollo people.

humphd:
I don't fully understand the graphql resolvers, so let's wait on the "join" part.  It would be pretty easy to do one of:

1. add the `Feed.id` to a `Post` when we create it
1. do some kind of Redis Set for each feed, which could contain all the post ids for this blog.  We could add something like `t:feed:{hash-of-feed}:posts`, which would be a Set (or Ordered Set), and make it trivial to look-up all posts for a given feed.

But the graphql stuff might guide our choices once we understand it better.

By the way, all the work @manekenpix did on the hashes has already paid off.  My Redis database isn't growing forever anymore: adding duplicate data (e.g., same feed, same post) has no effect now, and we only have 1 item as you'd expect.  So good!

PR #612, opened 2020-01-27
fix #443: support Accept header in /posts route

### LINKED ISSUE #443: Support Accept header in /posts route
Our `/post` route returns JSON for a post, given a key.  Let's add support for `Accept: text/plain`.  Using [`req.accepts()`](https://expressjs.com/en/api.html#req.accepts), we can check for text or json, and give back either JSON, or return the pure text content of the post.  We might also want to support `text/html` and just send the HTML content?

yatsenko-julia:
## Type of Change

## Description

Fixes #443 but the test for this issue is not done completely. 
 
## Checklist


Silvyre:
We figured it what `prettier` was complaining about - an extra newline was added in the last commit:
https://github.com/Seneca-CDOT/telescope/blob/0767ec6ccf4050cf33f976de718e1915fd1bff59/src/backend/web/routes/posts.js#L74-L77

For some reason, `prettier` was not flagging this when @yatsenko-julia was running `npm test`. However, when I checked out this branch and ran `npm test`, this issue was flagged as expected.

src/backend/web/routes/posts.js
```diff
@@ -84,7 +85,22 @@ posts.get('/:id', async (req, res) => {
         message: `Post not found for id ${id}`,
       });
     } else {
-      res.json(post);
+      const accept = accepts(req);
+
+      switch (accept.type(['json', 'text', 'html'])) {
```
Silvyre: I'm attempting to test this now (via [Postman](https://www.getpostman.com/)). From my understanding, if this code ([which I am currently checking out locally](https://imgur.com/Pvygcp8)) is working correctly, I should be able to send a GET request to `http://localhost:3000/posts` using an `Accept: text` header and receive a repsonse with `Content-Type: text/plan`.

I have attempted several such requests, but have only been receiving `application/json` content in response, e.g.

image

I've tried multiple `Accept` directives (e.g. `text`, `text/plain`, `html`), but always receive the same JSON-formatted response.

@yatsenko-julia Were you able to receive a different result during your testing?
(@humphd I would greatly appreciate some guidance if I'm obviously going about testing this the wrong way, here!)

manekenpix: @Silvyre I think this is for `posts/:id`, not `/posts`. Try postman again but requesting a specific post, and then you should get the text back. I'll do some testing myself too.

Silvyre: @manekenpix Ah, `/posts/:id` appears to function as I expected:
image

(`Accept: text/plain` also worked fine for GET http://localhost:3000/posts/e69f9ca5df)

* * *

I may have misinterpreted what @humphd was requesting in https://github.com/Seneca-CDOT/telescope/issues/443#issuecomment-574774894; my previous understanding was that we wished for the `/posts` route to support the `Accept` header. Now that we appears to have the `Accept` header working `/posts/:id`, would it be useful if we implement the same functionality for `/posts`?

manekenpix: @Silvyre `/posts` only returns an array of ids that can be used with `posts/:id`, I don't know if it'd be useful for that case.
For `/posts/:id`, this is what I got:
**HTML**
html

**Text**
text
:+1:

Silvyre: Alright, cool.

humphd:
I filed #615 for the tests.

humphd:
Filed #616 to fix a small follow-up for this.

yatsenko-julia:
Thanks, @humphd! I'll look into it today.

PR #618, opened 2020-01-27
Use feed.id as jobId to de-dupe job queue

### LINKED ISSUE #608: On startup, server should complete existing jobs before adding new ones

When we start the server, one of the first things we do is to download the wiki feeds and create feed jobs in our job queue.  We do this regardless of whether or not there are already jobs in the queue for this feed's URL.

Let's allow the queue to `drain` existing jobs before we add our first batch of feed jobs.  This will let the server pick-up where it left off.

The way things work now, it's easy to get thousands, or tens-of-thousands of jobs in the queue without realizing it.

To see what I mean, follow these steps:

* Start the server `npm start`
* Open your browser to `http://localhost:3000/admin/queues` so you can see a live view of what's in the feed queue
* Notice the number of jobs in the `waiting` area of the queue.  It should be something like 500-700 or less, for sure less than 1,000.  Here's what mine looks like when I do this:

Screen Shot 2020-01-26 at 11 32 05 AM

* Now kill the server with `ctrl+c` and then restart it.  Look at the jobs again:

Screen Shot 2020-01-26 at 11 32 17 AM

We can use [`const jobCounts = await queue.getJobCounts()`](https://github.com/OptimalBits/bull/blob/develop/REFERENCE.md#queuegetjobcounts) or [`const waitingJobsCount = await queue.getWaitingCount()`](https://github.com/OptimalBits/bull/blob/develop/REFERENCE.md#queuegetwaitingcount) to figure out how many jobs are waiting in the queue right now, and if it's > 0, we should wait wait on the queue to be fully `drained` (i.e., empty) before we add our new jobs.

### LINKED ISSUE #611: Variable declared twice
**What happened**:

Our `env.example` has the variable `FEED_QUEUE_DELAY_MS` declared twice:

```env
# Feed job queue delay(ms)
FEED_QUEUE_DELAY_MS=30000

# Number of concurrent worker processors to run. Use * if you want to run
# one per CPU. Use a number if you want to set it manually, up to a max
# of the CPU count.  If not set, we'll assume 1.
FEED_QUEUE_PARALLEL_WORKERS=1
FEED_QUEUE_DELAY_MS=60000
```

According to the [repo's history](https://github.com/Seneca-CDOT/telescope/commit/72a6c5798546ead49d94f0750f7a256f66d5691d#diff-005950e1aff25afa320a11650c8979d6), the second declaration was added in a commit related to a different part of the code that doesn't use this variable.

This second declaration should be deleted.

humphd:
## Issue This PR Addresses

1. Fixes #611 
1. Fixes #608

## Type of Change

## Description

I was trying to understand why our staging box had 68K+ delayed jobs in the queue, and discovered that we are using the wrong id name for our jobs.

When adding a job to a Bull queue, you can either use the default job number, an auto-incrementing integer, or [you can specify your own](https://github.com/OptimalBits/bull/blob/develop/REFERENCE.md#queueadd).  The problem with the latter is that it means that you can easily add multiple jobs for the same task, since each one is keyed on a number that has nothing to do with the data in the job.  If you instead provide a key that is unique to the job itself, you can have Bull ignore duplicates.

We actually tried to do this before, and instead of using `jobId` we used `id`, so Bull is ignoring it.  This change switches to use the correct version.

Because of this change, our backoff strategy now makes sense.  If a job doesn't work, we need to try it over and over again, at longer and longer intervals.  When the job finally fails for real, we can put it in the blacklist (cc @c3ho).

When we land this, we should update the `.env` on the staging box to reflect this changes here to our retry strategy (cc @manekenpix).

### DIFF
--- a/env.example
+++ b/env.example
@@ -45,16 +45,15 @@ BLOG_INACTIVE_TIME=360
 GITHUB_TOKEN=
 
 # Feed job queue attempts
-FEED_QUEUE_ATTEMPTS=2
+FEED_QUEUE_ATTEMPTS=5
 
 # Feed job queue delay(ms)
-FEED_QUEUE_DELAY_MS=30000
+FEED_QUEUE_DELAY_MS=60000
 
 # Number of concurrent worker processors to run. Use * if you want to run
 # one per CPU. Use a number if you want to set it manually, up to a max
 # of the CPU count.  If not set, we'll assume 1.
 FEED_QUEUE_PARALLEL_WORKERS=1
-FEED_QUEUE_DELAY_MS=60000
 
 # Max number of posts per page
 MAX_POSTS_PER_PAGE=30
--- a/src/backend/feed/queue.js
+++ b/src/backend/feed/queue.js
@@ -15,24 +15,24 @@ setQueues(queue);
  * The `feedInfo` is an Object with `name` (i.e., name of author) and `url`
  * (i.e., url of feed) properties, matching what we parse from the wiki.
  */
-queue.addFeed = async function(feedInfo) {
+queue.addFeed = async function(feed) {
   const options = {
-    // Use the feed URL as the job key, so we don't double add it.
-    // Bull will not add a job there already exists a job with the same id.
-    id: feedInfo.url,
-    attempts: process.env.FEED_QUEUE_ATTEMPTS || 2,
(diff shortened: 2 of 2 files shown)


humphd:
@manekenpix I'll merge this, and next time you update staging, if you could flush redis and update the `.env` with values from this PR, it will help.

PR #619, opened 2020-01-28
Integrate GatsbyJS's package.json with root version

### LINKED ISSUE #579: Deal with gatsby front end directories in .gitignore
certain directories in the frontend need to be prefixed in the project's root .gitignore file. Some of these include `public/` and `.cached` which are created when running `npm i`

prefix directories we want to ignore in the frontend inside .gitignore

humphd:
## Issue This PR Addresses

Fixes #579

## Type of Change

## Description

This change integrates the `src/frontend/package.json` scripts with our main `package.json`.  I've added commands in the root `package.json` to call all the relevant GatsbyJS scripts.  Doing all of the following in the root dir will work as expected:

```
npm run build
npm run develop
npm run serve
npm run clean
```

This works using the `--prefix path/name` feature of `npm` to receive a different root path to use for running the command.  I've also updated how we do our `postinstall` to use the same, which is more compatible with Windows, dropping the `&&`.

I've also updated the Prettier integration.  I added `.jsx` to our file types, and dropped the Prettier requirement and script from the Gatsby app itself--we'll do it from the root.

I've done the same with `test`, and we can add tests for the frontend to be called from our main app's `test` script when we write them.

It turns out that no changes are needed to our `.gitignore`.


humphd:
Two more things:

1. I noticed you can pass args along with the `npm scripts`, so I added pass through for arguments from the root
1. I added some docs so people can find more info about what the commands mean

humphd:
Another commit, this one dealing with prettier, stylelint, and Jest, which all need to be taught to ignore the `.cache/` and `public/` folders GatsbyJS makes.

humphd:
I can take no credit for `clean`, it was in the default scripts Gatsby installed, but thank you!

Issue #610, opened 2020-01-26
Blogs not showing up in chronological order


cindyledev:
**What happened**:
Our http://dev.telescope.cdot.systems/planet groups blogs by the same person differently than in http://zenit.senecac.on.ca/~chris.tyler/planet/.
Chronologically, the "How I Learned to Stop Worrying..." blog shouldn't be grouped with Ana's other two blogs because it was posted at a different time.
![Annotation 2020-01-26 140235](https://user-images.githubusercontent.com/36822198/73140331-c4fc1f80-4045-11ea-94ff-aaecc3f82e0b.png)

**What should have happened**:
Blog posts should be listed chronologically, newest on top.

**How to reproduce it (as precise as possible)**:
Visit http://dev.telescope.cdot.systems/planet and http://zenit.senecac.on.ca/~chris.tyler/planet/.

**Anything else we need to know?**:
It was probably grouped together because all three blogs were posted by the same person on the same day but if we look at the time stamps, there should be other blogs between.

**Environment**:

- OS: Windows
- Browser: Google Chrome

humphd:
This isn't a high priority issue, but if someone wants to work on this, here are some details:

- the code that does the template for this lives in https://github.com/Seneca-CDOT/telescope/blob/master/src/backend/web/planet/views/planet.handlebars
- the data for it comes from https://github.com/Seneca-CDOT/telescope/blob/master/src/backend/web/routes/planet.js
- the code in https://github.com/Seneca-CDOT/telescope/blob/master/src/backend/web/routes/planet.js#L16-L51 is supposed to group the posts by Day, Author, Post, but it looks like it's not doing the "by Author" part correctly and/or the template isn't reading the data out correctly.

humphd:
I was reading the new planet just now, and I think what it's actually doing is showing all posts by the author, ignoring chronology.  The multiple posts by @agarcia-caicedo are a good test case right now.

Compare:

a. http://zenit.senecac.on.ca/~chris.tyler/planet/
b. http://dev.telescope.cdot.systems/planet

PR #621, opened 2020-01-28
Match original planet, sort feed list

humphd:
This sorts the planet's feed list the same as the original planet did.


humphd:
I think #610 is different, no?  I thought that one was about the grouping vs. order.  This just sorts the list of names on the right.  I know we talked about it, but I couldn't find a bug for it.

cindyledev:
Oooooh this is for the participates list on the right XD

PR #629, opened 2020-01-30
Add /planet/compare route

humphd:
## Description

This adds a new route for debugging `/planet/compare` which lets you load our old and new planets side by side in the same window to easily compare.  I got the idea from @cindyledev, who has been doing this in some issues, and it gives us a quick way to see differences.

NOTE: the two planets aren't currently pixel perfect with one another (#621 for example).

Here is what it looks like:

Screen Shot 2020-01-30 at 10 14 45 AM

### DIFF
--- a/src/backend/web/planet/views/compare-planets.handlebars
+++ b/src/backend/web/planet/views/compare-planets.handlebars
@@ -1,0 +1,24 @@
+<style>
+  html, body {
+    height: 100%;
+    width: 100%;
+    margin: 0;
+    padding: 0;
+  }
+  main {
+    display: flex;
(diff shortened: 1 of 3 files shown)


cindyledev:
I really like this PR!
However, I fetched your branch but didn't get the same result as your screenshot
VirtualBox_Fedora_30_01_2020_14_07_25

humphd:
That just means that your local redis db isn't caught up with what's actually on the web.  If you give it time, it will re-process @c3ho's feed and show it.

What's nice about this PR is that you can see that something is "wrong."  It's great for testing, so I'd say it's working for you.

PR #630, opened 2020-01-30
Fixes #591, Add filters to GraphQL

### LINKED ISSUE #591: Add filters to GraphQL to remove getPost

c3ho:
## Issue This PR Addresses
Fixes #591 

## Type of Change

## Description

Rough work for now. You can test this by doing the following: 

image

filters are optional, to change another one simply change the last line,
to use two such as fromDate and toDate

```
{
  getPosts(page: 10, perPage: 10, filter: {fromDate: "Enter fromDate here", toDate: "Enter ToDate here"}){
    author
  }
}
```

Pretty sure the above one doesn't work yet as I'm messing up over the date/time things

## Checklist


src/backend/web/graphql/index.js
```diff
@@ -36,9 +36,31 @@ module.exports.typeDefs = graphql`
...
+    ne: String
+    in: String
+    nin: String
+    regex: String
+  }
+
+  input IntFilter {
+    eq: Int
+    ne: Int
+  }
```
c3ho: probably don't need these

src/backend/web/graphql/index.js
```diff
@@ -94,7 +116,24 @@ module.exports.resolvers = {
...
+          // check if author name is equal to what we're searching for.
+          if (filter.author) {
+            return result.filter(posts => posts.author === filter.author);
+          }
+          // check if published date is between
+          if (filter.fromDate && filter.toDate) {
+            const fromDate = new Date(filter.fromDate);
+            const toDate = new Date(filter.toDate);
+            return result.filter(
+              posts => parseInt(posts.published, 10) >= fromDate && posts.published <= toDate
```
humphd: Filter usually defines a plural form, and the callback a singuluar:

```js
posts.filter(post => ...)
```

Also, a `post` uses `Date` for `published` and `updated` (it's only a number of MS in Redis, but we convert it when you call `Post.byId`).  Also, in my PR for #631, I'm adding a way to query Redis for posts only within a date range.  It might make this easier, so you don't have to request them all, and filter.  You'd only have to filter for other fields within that bounded set of posts that meet the date range specified.

c3ho:
can iterate on this, with the other libraries and stuff. Just want to get the filters kind of working so I can give a demo

src/backend/web/graphql/index.js
```diff
@@ -36,9 +36,38 @@ module.exports.typeDefs = graphql`
...
-    getPosts(page: Int, perPage: Int): [Post]
+    getPosts(filter: PostFilter, page: Int, perPage: Int): [Post]
     getPostsCount: Int
   }
+
+  # Post filters
+  input PostFilter {
+    author: String
+    fromDate: Int
+    toDate: Int
```
humphd: Why use `Int` for these dates?  Won't most callers of this expect to be able to use Date objects?

src/backend/web/graphql/index.js
humphd: Should you remove them then?

src/backend/web/graphql/index.js
```diff
@@ -94,7 +123,26 @@ module.exports.resolvers = {
...
+        if (filter) {
+          // check if author name is equal to what we're searching for
+          if (filter.author) {
+            return result.filter(post => post.author === filter.author);
+          }
+          // check if published date is between two provided dates
+          if (filter.fromDate || filter.toDate) {
+            const fromDate = filter.fromDate ? new Date(filter.fromDate * 1000) : new Date();
+            const toDate = filter.toDate ? new Date(filter.toDate * 1000) : new Date();
+            return result.filter(post => post.published >= fromDate && post.published <= toDate);
```
humphd: So you're not using the stuff I landed for stats yet.  Are you going to move toward that?

src/backend/web/graphql/index.js
```diff
@@ -94,7 +122,27 @@ module.exports.resolvers = {
...
+            const datePostIds = await getPostsByDate(fromDate, toDate);
+            return result.filter(post => datePostIds.includes(post.id));
+          }
+          // check if url is equal to what we're searching for
+          if (filter.url) {
+            return result.filter(post => post.url === filter.url);
+          }
+        }
+
+        return result;
```
c3ho: I don't know why the results are being the way they are. 

It might also be I'm not entirely sure of how the flow should be at the moment.
How I'm thinking it should work

1. We get the page results, so say we're getting the first 30 results of page 1.
2. We filter those 30 results by the date filter now

or should we ...
1. Get the results with the filter
2. Apply pagination on the filtered results

Playing around with the filter, I have no idea why the results are so different:
for example with 
```
{
  getPosts(page: 1, perPage: 10, filter: {fromDate: "2020-01-01"}){
    author
    published
    title
  }
}
```
image

but if I change perPage to say 20
```
{
  getPosts(page: 1, perPage: 20, filter: {fromDate: "2020-01-01"}){
    author
    published
    title
  }
}
```
image

The results seem like they're going backwards..?

c3ho:
I think something else to think about, do we want to do a == match or something like string contains `string.match(regExp)`  based on user input?

src/backend/web/graphql/index.js
```diff
@@ -94,7 +122,27 @@ module.exports.resolvers = {
...
+        if (filter) {
+          // check if author name is equal to what we're searching for
+          if (filter.author) {
+            return result.filter(post => post.author === filter.author);
+          }
+          // check if post's date is between two provided dates
+          if (filter.fromDate || filter.toDate) {
+            const fromDate = filter.fromDate ? filter.fromDate : new Date();
+            const toDate = filter.toDate ? filter.toDate : new Date();
+            const datePostIds = await getPostsByDate(fromDate, toDate);
```
humphd: I think you need to check if you're doing a date-filter earlier, since there is no point getting all posts on line 123 if you are going to do it again here by date.  Move your filter checks up, and decide which way to get posts based on that.

src/backend/web/graphql/index.js
```diff
@@ -94,7 +122,27 @@ module.exports.resolvers = {
...
+            return result.filter(post => datePostIds.includes(post.id));
+          }
+          // check if url is equal to what we're searching for
+          if (filter.url) {
+            return result.filter(post => post.url === filter.url);
```
humphd: We should call the code in `normalize-url` here, so the URLs match more closely.  See https://github.com/Seneca-CDOT/telescope/blob/11700c087602f83838a9a9a67ffe403f8db2cce1/src/backend/data/feed.js#L1

src/backend/web/graphql/index.js
humphd: I might be that you get counts based on not filtering, but then sometimes filter (by date).  The posts and count should only get processed once you know about the filters you are using.

src/backend/web/graphql/index.js
c3ho: I think this was the question I wanted to ask, are we going to do the pagination first. As in

1. I want to see 30 posts on page 0
2. Out of those 30 posts I want to see the ones that are from "February 2, 2020" to "February 07, 2020". I'll then compare which of those posts on page 0 are within the results from `getPostsByDate()`

humphd: Pagination only makes sense in the context of your data, so calculate the posts you have first, then figure out pages after that

src/backend/web/graphql/index.js
```diff
@@ -113,37 +113,56 @@ module.exports.resolvers = {
      * @return Array of 'perPage' number of Post objects
      */
     getPosts: async (parent, { filter, page, perPage }) => {
+      let postIds;
+      const numOfPosts = await getPostsCount();
```
humphd: Do this later: depending on how we filter, it will be different (all posts vs. date range)

src/backend/web/graphql/index.js
```diff
@@ -113,37 +113,56 @@ module.exports.resolvers = {
...
 
-      const numOFPosts = await getPostsCount();
-      const first = page * prPage;
-
-      if (first < numOFPosts) {
-        const last = first + prPage < numOFPosts ? first + prPage : numOFPosts;
-        const postIds = await getPosts(first, last);
-
+      // For Date Filter
+      if (filter) {
```
humphd: Flip this logic: `if (!filter) {...}` since the code for the other case is smaller, and then you can `return` and not indent all this.

Even better, split this out into a separate function.

src/backend/web/graphql/index.js
```diff
@@ -113,37 +113,56 @@ module.exports.resolvers = {
...
+        if (filter.fromDate || filter.toDate) {
+          const fromDate = filter.fromDate ? filter.fromDate : new Date();
+          const toDate = filter.toDate ? filter.toDate : new Date();
+          postIds = await getPostsByDate(fromDate, toDate); // Get all results within the date
+          // check if # of filtered results are less than max results allowed per page.
+          if (postIds.length < prPage) {
+            return Promise.all(postIds.map(id => Post.byId(id))); // return all the posts' info
```
humphd:
```js
return Promise.all(postIds.map(Post.byId));
```

src/backend/web/graphql/index.js
```diff
@@ -113,37 +113,56 @@ module.exports.resolvers = {
...
+          }
+          // Otherwise do some pagination here
+          const pageResult = postIds.slice(page - 1 * perPage, page * prPage);
+          return Promise.all(pageResult.map(id => Post.byId(id)));
```
humphd: Same thing here, shorten to `(Post.byId)`

src/backend/web/graphql/index.js
```diff
@@ -113,37 +113,56 @@ module.exports.resolvers = {
...
+      }
 
-        return result;
+      const first = page * prPage;
+
+      if (first < numOfPosts) {
+        const last = first + prPage < numOfPosts ? first + prPage : numOfPosts;
+        postIds = await getPosts(first, last);
+
+        return Promise.all(postIds.map(id => Post.byId(id)));
```
humphd: `map(Post.byId)`

src/backend/web/graphql/index.js
```diff
@@ -84,18 +112,57 @@ module.exports.resolvers = {
...
+          const authorResults = result.filter(post => post.author === filter.author);
+          // check if # of filtered results are less than max results allowed per page.
+          if (authorResults.length < prPage) {
+            return authorResults;
+          }
+          return authorResults.slice(page - 1 * perPage, page * prPage);
+        }
+
+        if (filter.author) {
+          const urlResults = result.filter(post => post.url === normalize(filter.url));
```
humphd: You should normalize both of these URLs before you compare:

```js
const urlResults = result.filter(post => normalize(post.url) === normalize(filter.url));
```

src/backend/web/graphql/index.js
```diff
@@ -84,18 +112,57 @@ module.exports.resolvers = {
...
+          if (postIds.length < prPage) {
+            return Promise.all(postIds.map(id => Post.byId(id))); // return all the posts' info
+          }
+          // Otherwise do some pagination here
+          const pageResult = postIds.slice(page - 1 * perPage, page * prPage);
+          return Promise.all(pageResult.map(id => Post.byId(id)));
+        }
+        // Other filters use the getPosts() instead of getPostsByDate()
+        postIds = await getPosts(0, numOfPosts);
+        const result = await Promise.all(postIds.map(id => Post.byId(id)));
```
humphd: `posts` vs. `result`?

src/backend/web/graphql/index.js
```diff
@@ -84,18 +112,57 @@ module.exports.resolvers = {
...
+
+        if (filter.url) {
```
manekenpix: I think you have `if(filter.url)` here (but this deals with authors, and `if(filter.author)` below this where you deal with urls.

src/backend/web/graphql/index.js
c3ho: You can do that..?

humphd:
```js
// v1
map(function(id) { return Post.byId(id) })

//v2
map(id => Post.byId(id));

// v3
const fn = id => Post.byId(id);
map(fn)

// v4
map(Post.byId)
```

Because `map` takes a function that accepts an `id` argument and transforms it into a Post, and since `Post.byId` is a function that accepts an `id` argument and transforms it into a Post, you can use it directly without calling it via an outer function.

c3ho: TIL

src/backend/web/graphql/index.js
c3ho: Opting to do this in another PR.

c3ho:
Kind of weird, 2020-01-01, 2020-01-02, 2020-01-03 don't work but 2020-01-04 does.
image

c3ho:
Fixed slicing logic. I did a console.log of the length of array returned, the other dates worked because the results were less than 30, no issue as there was no pagination involved. 

results > 30 triggered my faulty slicing logic.

```
const pageResult = postIds.slice(
  page * perPage - 1 < 0 ? 0 : page * perPage,
  (page + 1) * prPage
);
```

if user sets page as 0 and perPage as 30, logic will now become
slice(0, 30)

if user sets page as 1 and perPage as 30, logic will now become
slice(30, 60)

c3ho:
Merging as this was approved by both.

Issue #595, opened 2020-01-24
Write Jest tests for src/backend/web/routes/opml.js


Silvyre:
As requested in https://github.com/Seneca-CDOT/telescope/pull/594#pullrequestreview-348101295, Jest tests should be written to test the logic used within the new `/opml` API endpoint.

Issue #517, opened 2019-12-15
Automatic PR deployments for frontend


humphd:
**What would you like to be added**:

I'd love to see if we can get our frontend to be auto-deployed to [Zeit Now](https://zeit.co/docs/v2/more/now-for-github) (or a similar free service) so that we can easily test frontend changes whenever someone makes a change and wants to have it get merged.

**Why would you like this to be added**:

A lot of projects use this for static portions of their app (e.g., frontend) so that they can do more responsible reviews.  No one is going to pull every PR and build/test it locally.  We need a way to automate this so that it's possible to really see the impact of a PR.

Doing this will be dependent on us separating our frontend from our express server.  Once we have a React or similar setup, this should be much easier.

ragnarokatz:
@humphd 

So I've been playing around with this. I created an account on zeit, and I was able to do some manual deployment so far: https://frontend-zeta-beige.now.sh/

steps in order to deploy manually:
```
npm i -g now
cd src/frontend
now login
now
```

I have linked up my own fork of telescope on github with my zeit app as well. However, the automatic deployment after every PR creation does not seem to work. It is unsupported at the moment according to this source:
https://spectrum.chat/zeit/now/git-deployment-from-subdirectory~eb838339-47c8-43a9-a24b-211e197fb1f3

humphd:
@cindyledev I notice you [can do this with Netlify as well](https://www.netlify.com/blog/2016/07/20/introducing-deploy-previews-in-netlify/).

One thing I'm not clear on how we should do is to specify the backend URLs to use for accessing our REST/GraphQL endpoints when we do this.  We'd need to tell the version deployed to zeit/netflify to use our staging box vs. the same box serving the content.  This is somewhat related to what @agarcia-caicedo is doing in #569.  We just need some kind of flag/env variable to be able to configure gatsby with a different base URL to use when accessing backend data routes.

humphd:
https://www.gatsbyjs.org/docs/environment-variables/ has some good ideas on this, and https://www.gatsbyjs.org/docs/environment-variables/#example gives an example very close to what we need to do.

Issue #539, opened 2020-01-14
Add "image" to our post object and extract during processing


humphd:
**What would you like to be added**:

If a post has an image associated with it (e.g., social media image, first image in the post, etc), we should extract it out and include it as part of our Post object's JSON.

**Why would you like this to be added**:

Our frontend design may need to show a cover image for a post, and/or we want to allow better sharing with social media, which usually expects an image.  This should make it easier to provide that.

Take a look at this example UI: https://gatsby-starter.haezl.at/

In it, you have "cover" images (also sometimes called "poster" images) for post summaries.  Our feed parser [returns an image](https://www.npmjs.com/package/feedparser#list-of-article-properties) as part of the data extracted from a feed article.  If that's missing, we could also use JSDOM to query (e.g., `document.querySelector('img')`) the HTML for a post and obtain it that way.

PR #633, opened 2020-02-01
Expose /feed for RSS, Atom, JSON, OPML, and wiki feeds

humphd:
## Type of Change

## Description

This PR consolidates our exported "feeds" into a `/feed` route.  Currently we have:

1. `/opml` which returns an OPML (XML) list of our feeds

After this PR, we have the following:

1. `/feed/opml`which returns an OPML (XML) list of our feeds (I've moved it onto `/feed/*`
1. `/feed/wiki` which returns a Plain Text version of our feed list, suitable for replacing the wiki feed list.  We can use this later to update the wiki feed list with only active feeds.
1. `/feed/rss` which returns an RSS (XML) feed of the most recently indexed posts
1. `/feed/atom` which returns an Atom (XML) feed of the most recently indexed posts
1. `/feed/json` which returns a JSON feed of the most recently indexed posts

NOTE: The `/feed` route is where we exported feeds of various kinds, and `/feeds` is our route for working with our internal Feed objects.  I've left a comment in the code to clarify this, and I wish I had better names so it was less confusing, but I can't come up with something better.

I've also added a new `env` variable, `BASE_URL`, which is the base URL for our backend.  This will be useful for a number of things, where we need a way to configure our URL (e.g., #517, cc @cindyledev, @agarcia-caicedo).

## Checklist

I have not done tests here, since I didn't want to mess up the work that @Silvyre is doing in #595.  I can rebase on that when it comes in.

I also have a few TODOs that I need to clean up in here.  I either need to file follow-ups, or fix things here.  I wanted to get a first review before I did that.


humphd:
Filed #637, #638 for follow-up, and this already links to #539.

PR #639, opened 2020-02-02
Fixes #462, Integrate inactive blog url and include logic to filter inactive feeds

### LINKED ISSUE #462: Integrate inactive blog url and include logic to filter inactive feeds

c3ho:
#462 

## Type of Change

## Description

As per discussion, removed the inactive-blog-filter.js files (sorry Rafi). Added new functions in ```storage.js``` + ```feed.js``` to handle invalid feeds. Modified ```processor.js``` to use the new functions for the appropriate errors.

## Checklist


src/backend/utils/storage.js
```diff
@@ -7,11 +7,15 @@ const postsKey = 't:posts';
 // Namespaces
 const feedNamespace = 't:feed:';
 const postNamespace = 't:post:';
+// Suffix
+const invalidNamespace = ':invalid';
```
humphd: `invalidSuffix`

src/backend/utils/storage.js
```diff
@@ -47,6 +51,13 @@ module.exports = {
 
   getFeedsCount: () => redis.scard(feedsKey),
 
+  addInvalidFeed: (feed, reason) => {
+    const key = createInvalidFeedKey(feed.id);
+    redis.set(key, reason);
+  },
+
+  isInvalid: feed => redis.exists(feed.id),
```
humphd:
```js
redis.exists(createInvalidFeedKey(feed.id))
```

src/backend/data/feed.js
```diff
@@ -30,6 +30,17 @@ class Feed {
     addFeed(this);
   }
 
+  /**
+   * Saves the current Feed to the database as a key and value as a reason
+   */
+  saveInvalid(reason) {
```
humphd: `setInvalid(reason)`

src/backend/utils/storage.js
```diff
@@ -47,6 +51,13 @@ module.exports = {
 
   getFeedsCount: () => redis.scard(feedsKey),
 
+  addInvalidFeed: (feed, reason) => {
```
humphd: The rest of our Redis feed code accepts an `id` vs. a `feed`.  Let's do that here too.

src/backend/data/feed.js
```diff
@@ -30,6 +30,17 @@ class Feed {
...
+    addInvalidFeed(this, reason);
+  }
+
+  isInvalid() {
+    isInvalid(this);
```
humphd: `isInvalid(this.id)` (see note below about using `id` in Redis).

src/backend/feed/processor.js
```diff
@@ -96,7 +96,7 @@ module.exports = async function processor(job) {
   const feed = Feed.parse(job.data);
   let articles;
   let info;
-
+  // if (!feed.isInvalid()) {
```
humphd: Why is this commented out?

src/backend/feed/processor.js
```diff
@@ -108,9 +108,11 @@ module.exports = async function processor(job) {
           break;
         case 404:
           logger.warn(`${info.status} Feed not found: ${feed.url}`);
+          feed.saveInvalid(`${info.status} Feed not found: ${feed.url}`);
```
humphd: I don't think we want to set feeds invalid in here.  Instead, you should do it when the feed queue `.on('fail')` event occurs, since this will give the feed queue a chance to retry the job (e.g., could be a temporary 404).

src/backend/feed/processor.js
c3ho: Commented out for now as it was failing current tests, when it expects those switch res codes, it is instead always receiving 200.

src/backend/feed/processor.js
c3ho: Any issues with line 159 and 163 and the use of ```feed.setInvalid()``` there?

humphd: I think we only want to invalidate a feed if it comes via the fail event in the queue.

c3ho:
Still need to work on the logic and the failedReason

c3ho:
To test, the stuff for now:

1. set the .env var ```FEED_QUEUE_ATTEMPTS``` to 1 AND set removeOnFail: to ```removeOnFail: true```
2. npm start
3. check for failed jobs in localhost:(yourPort#Here)/admin/queues
4. open up a terminal to interact with redis and use the following command
```redis exists t:feed:jobId:invalid``` jobId being whatever id you see in the failed queue, if it returns 1, its there
to check the reason why it was invalid use:
```redis get t:feed:jobId:invalid``` 

for example I have a job id #937dda54ec
```redis exists t:feed:937dda54ec:invalid``` should return 1
```redis get t:feed:937dda54ec:invalid``` should also give me a reason.

**Comment by c3ho** [comment]
Further testing is making it weird, its also saving the posts it can't process as part of the ```t:feed:**:invalid``` set to Redis. Not entirely sure how the err is supposed to behave as there's a discrepancy between the bull board failureReason and the reason in redis

File: src/backend/index.js [hunk 8]
```diff
@@ -54,6 +54,10 @@ function processFeeds(feeds) {
       const currentFeed = await updateFeed(feed);
       // Add a job to the feed queue to process all of this feed's posts.
       await feedQueue.addFeed(currentFeed);
+      // If failed, set the feed to invalid and save to Redis.
+      feedQueue.on('failed', (job, err) => {
+        currentFeed.setInvalid(err.message);
```
humphd: When you have a callback in an event handler that is synchronous, and you need to invoke an `async` function, you need to add a `.catch()` so a failed Promise doesn't blow things up.  Also, you shouldn't hold a reference to the `feed` in a closure like this.  Use the `job.data` to recreate a new `Feed` object:

```js
async function invalidateFeed(feedData) {
  const feed = Feed.parse(feedData);
  await feed.setInvalid(feedData.reason || 'unknown reason');
}

feedQueue.on('failed', (job, err) => invalidateFeed(job.data).catch(error => logger.error({ error }, 'Unable to invalidate feed in database.')));
```

NOTE: we'll need to attach the `reason` to the `job.data` before we fail it.  I'm not 100% sure where/how to do this without doing it myself.

test/storage.test.js
```diff
@@ -17,8 +19,22 @@ describe('Storage tests for feeds', () => {
...
+    'last-modified'
+  );
+
+  beforeAll(() =>
+    Promise.all([
+      addFeed(feed1),
+      addFeed(feed2),
+      addFeed(feed3),
+      addFeed(feed4),
+      setInvalidFeed(feed5.id, 'This just fails'),
```
humphd: You haven't added `feed5` to Redis, so this can't work.  You can only invalidate a feed that exists in the db.

c3ho: We have mock-redis so I'm guessing I can save this in mock-redis and it should be fine?

humphd: Correct, you can do whatever you want in the test code, it's all being done in-memory with the mock redis, so nothing will get broken.

humphd: Probably `setInvalidFeed` should check that the feed exists, and throw if it doesn't.  We should deal with that problem higher up in the call stack.

I'm unclear how a feed could not be in our feed set in Redis.  Can you trace this back, and make sure that we're adding the feed before we start to rely on it?

c3ho: In the `processor.js` file
the first if statement with 
```
if (invalid) {
//code
}
```
should always be skipped at the start, because there shouldn't be any invalid feeds in the db to retrieve with the `GET` command. Otherwise later on in the code you have the line `await feed.save()` which actually will save the feed to Redis

I think the way I implemented your solution though might be wrong. We don't actually need a feed to be saved to the database to get a key,

In `storage.js` this is the following code to set an invalid feed
```
  setInvalidFeed: (id, reason) => {
    const key = createInvalidFeedKey(id);
    redis.set(key, reason);
  }
```
as long as an ID and a reason is provided to the function, Redis will create a key value pair using the `SET` command.

humphd: Right, so we'll always process a feed once, which will `feed.save()` it into Redis before we can possibly try and invalidate it.  Good.

I want to avoid orphaned feed keys, where we create some invalid key for a feed that doesn't exist.  It's not a problem per se, but it just wastes space in the db with something we'll never find again.

src/backend/feed/queue.js
```diff
@@ -26,7 +26,7 @@ queue.addFeed = async function(feed) {
       delay: process.env.FEED_QUEUE_DELAY_MS || 60 * 1000,
     },
     removeOnComplete: true,
-    removeOnFail: true,
+    removeOnFail: false,
```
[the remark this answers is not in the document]
c3ho: Sorry, enabled that for now just so I can refer to which jobs failed or not and can manually find them in redis

src/backend/index.js
c3ho: I tried your above code, it does work... just like you said all of them have a value of 'unknown reason'

Another side effect of it being failed, I'm not sure if its a bug. Bull can't differentiate between feeds failing and posts failing, it adds both to Redis

humphd:
I think you need to rebase to get the code where I fix the posts not failing.

c3ho:
Nice, that did the trick.
image

src/backend/index.js
```diff
@@ -43,6 +43,15 @@ async function updateFeed(feed) {
   return currentFeed;
 }
 
+/**
+ * Invalidates a feed
+ * @param {Feed} feed
+ */
+async function invalidateFeed(feedData) {
+  const feed = Feed.parse(feedData);
+  await feed.setInvalid(feedData.reason || 'unknown reason');
```
humphd: Can you file a follow-up issue so we can get the `reason` working properly?

Also, let's add a 

```js
logger.info(`Invalidating feed ${feedData.url} for the following reason: ${feedData.reason || unknown reason}`)
```

src/backend/index.js
```diff
@@ -43,6 +43,15 @@ async function updateFeed(feed) {
   return currentFeed;
 }
 
+/**
+ * Invalidates a feed
+ * @param {Feed} feed
```
humphd: This is wrong.  `feedData` is a regular JS Object with properties of a feed, but not a Feed object.  Because we process feeds in a separate process, they get passed across process boundaries as JSON.

src/backend/utils/storage.js
```diff
@@ -47,6 +51,13 @@ module.exports = {
 
   getFeedsCount: () => redis.scard(feedsKey),
 
+  setInvalidFeed: (id, reason) => {
+    const key = createInvalidFeedKey(id);
```
humphd: File a follow-up issue to have us check to make sure the `id` exists in our database already, and throw if not.

c3ho: #654

src/backend/index.js
c3ho: #653

src/backend/index.js
```diff
@@ -54,6 +67,12 @@ function processFeeds(feeds) {
       const currentFeed = await updateFeed(feed);
       // Add a job to the feed queue to process all of this feed's posts.
       await feedQueue.addFeed(currentFeed);
+      // If failed, set the feed to invalid and save to Redis.
+      feedQueue.on('failed', job =>
```
humphd: Move this event handler down around 107, not inside `processFeeds`.

c3ho:
Testing this on windows dev environment, I didn't get all those errors popping up unlike the demo this morning when shutting down the service with ctrl + c

c3ho:
@Silvyre when you're testing this use the following command in Redis to see if it is correctly invalidating blogs `redis-cli get t:feed:e35052b2c8:invalid` should return a result with `unknown reason`. Otherwise the terminal should start displaying messages after the initial processing of skipping resouce because it has been previously marked invalid or something along those lines

Silvyre:
> 
> 
> @Silvyre when you're testing this use the following command in Redis to see if it is correctly invalidating blogs `redis-cli get t:feed:e35052b2c8:invalid` should return a result with `unknown reason`. Otherwise the terminal should start displaying messages after the initial processing of skipping resouce because it has been previously marked invalid or something along those lines

Thanks, going to attempt this!

humphd:
@Silvyre hop on Slack and join the Review Party so you can ask him questions as you go.

c3ho:
Going to just merge this once it passes CI, since I'm just rebasing stuff.

PR #640, opened 2020-02-02
Refactor SAML auth, make /admin/queues require auth

### LINKED ISSUE #202: Restrict access to admin route
The admin routes (`/admin/*`) should only be accessible to developers, we should add some kind of auth service to prevent this.

humphd:
Fixes #202

## Description

This is a WIP experimental PR to try and get our authentication code working smoothly.  I've focused on the node/passport bits, and gotten it to the point that it works in basic HTML.  The next step is connecting this into our GatsbyJS app.

Here's some of what I've done:

- Refactored the code to centralize it, and reduce how much we need
- Added default values to our `env.example` for all `SAML_*` values, so things "Just Work" (@manekenpix is suggesting we try and do this more often, and I think it's a good idea).
- moved `/login` to `/auth` and we now have these routes: `/auth/login`, `/auth/logout`, and `/auth/callback` which enable our SAML SSO flow
- I've gutted the current homepage and added links for logging in, logging out, and also a link to the Admin Queues area.  I've made this so you have to be authenticated to get to it, which lets us test things easily.
 
I have a bunch of `TODO`s in the code, which need work.  The main thing I'm stuck on is that you should be able to do this:

```js
router.get('/some/route/you/want/authenticated', passport.authenticate('saml'), (req, res) => {
  // This only gets called if you are now authenticated...
  const user = req.user;
  // You can use user now.
});
```

I have a stupid redirect issue, where it goes back to the homepage instead of falling through to the proper handler.

To test this, pull my branch and do this:

```
docker-compose up --build
```

Now go to `http://localhost:3000`.  You can try going to any of the buttons at the top:

Screen Shot 2020-02-01 at 7 44 18 PM

When you're logging in/out, you can look at your cookies:

Screen Shot 2020-02-01 at 7 44 53 PM


src/frontend/index.html
```diff
@@ -5,106 +5,43 @@
...
+      <a href="/auth/logout">
+        <button class="nav-button" class="login">
+          Logout
+        </button>
+      </a>
+    </nav>
+
+    <h1>Telescope</h1>
+
+    <p>NOTE: this page is going away soon, just for testing login/logout.</p>
```
Grommers00: Not sure what this is referring too?

src/backend/web/routes/auth.js
```diff
@@ -0,0 +1,69 @@
...
+ * they were trying to go, or to the home page
+ */
+router.post('/login/callback', passport.authenticate('saml'), (req, res) => {
+  let returnTo;
+  if (req.session) {
+    returnTo = req.session.returnTo;
+    delete req.session.returnTo;
+  }
+
+  res.redirect(returnTo || telescopeHomeUrl);
```
Grommers00: interesting - i was just saying that seneca works needs this ability.

.eslintrc.js
```diff
@@ -93,5 +93,10 @@ module.exports = {
     'react/jsx-uses-vars': 'error',
 
     'react/jsx-filename-extension': [1, { extensions: ['.js', '.jsx'] }],
+
+    /**
+     * Some passport properties are only exposed via underscore names
+     */
+    'no-underscore-dangle': 'off',
```
Grommers00: What's the advantage/disadvantage of this?

src/frontend/index.html
humphd: This entire web page will be replaced by our GatsbyJS app.  I just have this here now for testing that the authenticated routes work.  I added this note so people didn't wonder why I had this page.

src/backend/web/routes/auth.js
humphd: Yeah, it's really critical, and you have to hack it in place yourself.

.eslintrc.js
humphd: I just needed it to get eslint to pass, since I (currently) have to access the logout function I need via `passport._strategy` and our rules don't allow names with leading `_`.  I might be able to fix this later so I don't need to do this.

humphd: In general, anything named `_name` is a "private" variable, and you aren't usually supposed to access it.

humphd:
This is a great, short article on the kind of login flow that I'm using here to protect these api routes: https://auth0.com/docs/login/spa/authenticate-with-cookies

PR #645, opened 2020-02-03
Fixes #517 Automatic PR deployment for frontend using Zeit Now

### LINKED ISSUE #517: Automatic PR deployments for frontend

cindyledev:
## Type of Change

## Description

This adds the configuration files for Zeit Now, after this gets merged, whenever commits are pushed into a branch, automatic deployment will trigger.

## Checklist

### DIFF
--- a/now.json
+++ b/now.json
@@ -1,0 +1,9 @@
+{
+  "builds": [
+    {
+      "src": "package.json",
+      "use": "@now/static-build",
+      "config": { "distDir": "src/frontend/public" }
+    }
+  ]
+}
--- a/package.json
+++ b/package.json
@@ -47,7 +47,8 @@
     "webhint": "cross-env PORT=3000 npm-run-all -r -p server hint",
     "lighthouse-index": "lighthouse http://localhost:3000 --output-path=./lighthouse-report.html --view",
     "lighthouse": "cross-env PORT=3000 npm-run-all -r -p server lighthouse-index",
-    "test-ci": "run-s prettier-check test"
+    "test-ci": "run-s prettier-check test",
+    "now-build": "npm run build"
   },
   "husky": {
     "hooks": {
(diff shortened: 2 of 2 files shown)


now.json
```diff
@@ -0,0 +1,9 @@
+{
+  "builds": [
```
humphd: I notice that the Zeit config docs mention that [the`builds` property has been deprecated](https://zeit.co/docs/configuration#project/builds), and that we should instead rely on the [`build` step with zero config](https://zeit.co/guides/migrate-to-zeit-now).  I don't know if it will work for us or not, since we don't put the build output in the expected location (i.e., `./src/frontend/public` vs. `./public/`).

It looks like [we can't change the directory](https://github.com/gatsbyjs/gatsby/issues/14703) that Gatsby builds to, so we can consider trying one of the following:

- leave things as you have them now, since it's working.  The risk is that it will break in the future.
- try removing the `now.json` file and see what Zeit does automatically.  I think it will fail to find our `public/` dir, but maybe I'm wrong and it will just work?

Probably we should just leave this how you have it now.  I've [asked on Twitter](https://twitter.com/humphd/status/1224488848828305408).

humphd: Here's the [answer from Zeit](https://twitter.com/notquiteleo/status/1224489188269137921):

> Currently, you could run `mv path/to/public public` after your build command to make it work. We're releasing something later this week that will make this work without requiring any code changes, however!

Let's take this as is, and migrate to their new solution when it ships.

PR #649, opened 2020-02-04
Closes #519: Use Let's Encrypt and NGINX for certs on staging

### LINKED ISSUE #519: Use Let's Encrypt for certs on staging

manekenpix:
## Issue This PR Addresses

Closes #519 

## Type of Change

## Description

This PR adds 2 new containers to our docker compose file: [nginx](https://www.nginx.com/) and [certbot](https://certbot.eff.org/).

-  Nginx redirects requests to `dev.telescope ` or `login.telescope` and makes sure that all the requests use SSL.
-  Certbot is responsible for checking and renewing our certificates if needed.

To avoid conflict between development settings and production/staging, a second `docker-compose` file was added. The original `docker-compose.yml`will be the one used for development. This way, no changes are needed in our current documentation.
`docker-compose-staging.yml` will be used in staging and production, and is the one that reflects the changes in this PR.

## Checklist


docker-compose-prod.yml
```diff
@@ -0,0 +1,63 @@
...
+      # a place-holder for the variable passed by CLI with a default value
+      # the passed or default value is a command to run `telescope`
+      - script=${script:-start}
+      # override the .env variable for `REDIS_URL`
+      - REDIS_URL=redis://redis:${REDIS_PORT}
+    depends_on:
+      - redis
+    ports:
+      - '$PORT:$PORT'
+    restart: always
```
humphd: I think this has been replaced with `restart_policy`, see https://docs.docker.com/compose/compose-file/#restart_policy

docker-compose-prod.yml
```diff
@@ -0,0 +1,63 @@
...
+  # Second service
+  redis:
+    image: redis:latest
+    ports:
+      - '$REDIS_PORT:$REDIS_PORT'
+
+  # Third service
+  # SSO Identity Provider test service, https://simplesamlphp.org
+  # Access to the authentication page via http://localhost:8080/simplesaml or https://localhost:8443/simplesaml
+  ssoipd:
```
humphd: I wonder if we should rename this to `login`, which is more recongnizable to developers who don't fully understand how SAML works (SSO IPD is right, but not an obvious name).

docker-compose-prod.yml
```diff
@@ -0,0 +1,63 @@
...
+    environment:
+      - SIMPLESAMLPHP_SP_ENTITY_ID=${SAML2_CLIENT_ID:-saml-poc}
+      - SIMPLESAMLPHP_SP_ASSERTION_CONSUMER_SERVICE=${SAML2_REDIRECT_URI}
```
humphd: Note: these are changing in #640, so one of us will need to rebase and adjust (probably me, if you beat me into master, so remind me).

docker-compose-prod.yml
```diff
@@ -0,0 +1,63 @@
...
+    ports:
+      # These ports + port 80 are pre-defined in the latter image
+      # http port
+      - '8080:8080'
+      # https port
+      - '8443:8443'
+    restart: always
+  nginx:
+    image: nginx:latest
+    container_name: production_nginx
```
humphd: Curious why you named this one with a custom name and not the others?

docker-compose-prod.yml
```diff
@@ -0,0 +1,63 @@
...
+    # image owner's blog post https://medium.com/disney-streaming/setup-a-single-sign-on-saml-test-environment-with-docker-and-nodejs-c53fc1a984c9
+    image: kristophjunge/test-saml-idp
+    ports:
+      # These ports + port 80 are pre-defined in the latter image
+      # http port
+      - '8080:8080'
+      # https port
+      - '8443:8443'
+    restart: always
+  nginx:
```
humphd: Above this file has `# First Service`, `# Second Service`, so I'd either continue this pattern, or remove those previous comments, since we now have more services.

docker-compose-prod.yml
```diff
@@ -0,0 +1,63 @@
...
+    container_name: production_nginx
+    volumes:
+      - ./nginx.conf:/etc/nginx/nginx.conf
+      - ../certbot/conf:/etc/letsencrypt
+      - ../certbot/www:/var/www/certbot
+    ports:
+      - 80:80
+      - 443:443
+    restart: always
+    command: '/bin/sh -c ''while :; do sleep 6h & wait $${!}; nginx -s reload; done & nginx -g "daemon off;"'''
```
humphd: Can you add a comment above this line explaining why you're doing this 6h dance?

docker-compose-prod.yml
```diff
@@ -0,0 +1,63 @@
...
+  certbot:
+    image: certbot/certbot
+    volumes:
+      - ../certbot/conf:/etc/letsencrypt
+      - ../certbot/www:/var/www/certbot
+    entrypoint: "/bin/sh -c 'trap exit TERM; while :; do certbot renew; sleep 12h & wait $${!}; done;'"
```
humphd: Same deal here, add a comment above this with some notes so we can maintain this.

nginx.conf
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  server {
+    listen 443 ssl;
+    server_name dev.telescope.cdot.systems;
+
+    include /etc/letsencrypt/options-ssl-nginx.conf;
+    ssl_dhparam /etc/letsencrypt/ssl-dhparams.pem;
+
+    ssl_certificate /etc/letsencrypt/live/dev.telescope.cdot.cdot.systems/fullchain.pem;
```
humphd: Typo here, too many `.cdot` in `dev.telescope.cdot.cdot.systems`

docker-compose-prod.yml
humphd: We might also end up with 3 of these files: dev, staging, and prod.  In prod we won't want the login stuff, since we'll use Seneca's SSO server.

docker-compose-prod.yml
manekenpix: I tried to add `restart_policy`, and it needs a `deploy` key, which is not supported by compose. I might need to revert to `restart:always` (if I don't use `deploy`, compose complains).
deploy

docker-compose-prod.yml
manekenpix: Since we're going to have 3 these, maybe I should rename the new one added in this PR to `docker-compose-stage.yml`?

docker-compose-prod.yml
manekenpix: I did some testing that involved naming the containers and I forgot to remove that one. Thank you for reminding me.

docker-compose-prod.yml
humphd: Noted.  Thanks for testing.

Issue #487, opened 2019-12-07
Add <figure> to allowedTags in sanitize-html.js 


KarlaChen:
**add ```<figure>``` allowedTag to ```sanitize-html.js```**:

**```<figure>``` is a useful tag for blog post**:

humphd:
We have code now to examine all feeds and their posts, and do an analysis of the elements that are contained in the HTML.  I ran it just now, and here is what people use:

```
1. <br> (68,321, 35.367%)
2. <p> (35,193, 18.218%)
3. <a> (19,244, 9.962%)
4. <div> (15,139, 7.837%)
5. <li> (8,681, 4.494%)
6. <td> (7,270, 3.763%)
7. <strong> (6,440, 3.334%)
8. <img> (6,217, 3.218%)
9. <pre> (3,914, 2.026%)
10. <code> (3,798, 1.966%)
11. <b> (3,170, 1.641%)
12. <em> (2,613, 1.353%)
13. <ul> (2,273, 1.177%)
14. <tr> (1,836, 0.950%)
15. <h3> (1,704, 0.882%)
16. <figure> (1,460, 0.756%)
17. <i> (1,417, 0.734%)
18. <h4> (1,114, 0.577%)
19. <blockquote> (992, 0.514%)
20. <ol> (554, 0.287%)
21. <hr> (457, 0.237%)
22. <table> (376, 0.195%)
23. <tbody> (375, 0.194%)
24. <th> (344, 0.178%)
25. <h5> (133, 0.069%)
26. <iframe> (61, 0.032%)
27. <strike> (38, 0.020%)
28. <h6> (26, 0.013%)
29. <thead> (11, 0.006%)
30. <caption> (5, 0.003%)
```

Our code to strip/allow HTML elements is here:

https://github.com/Seneca-CDOT/telescope/blob/master/src/backend/utils/sanitize-html.js#L8-L10

And the docs on what is allowed by default is here:

https://github.com/apostrophecms/sanitize-html#what-are-the-default-options

So let's morph this bug into being about making sure that all of the element types we use in the list above are reflected in the setup that we have.

yatsenko-julia:
It appeared that `<figure>` tag is already added to allowedTags in sanitize-html.js. 

@humphd  I went through all tags that you mentioned and compared them to those that are allowed by default. It appeared all the tags we are using in the project are default except `<img>` and `<figure>` that are already in sanitize-html.js. I think we can close this issue.

[...]

PR #652, opened 2020-02-05
fix-556: updated sinitize-html.js and test for iframe tag added

yatsenko-julia:
## Type of Change

## Description

Updated sanitize-html.js, so all needed tags are allowed, the rest are disallowed. Needed tags are taken from the list provided by @humphd in [issue-487](https://github.com/Seneca-CDOT/telescope/issues/487). Also, test for <iframe> tag added. 

## Checklist

### DIFF
--- a/src/backend/utils/sanitize-html.js
+++ b/src/backend/utils/sanitize-html.js
@@ -7,12 +7,14 @@ module.exports = function(dirty) {
   return sanitizeHtml(dirty, {
     // Add <img> to the list of allowed tags, see:
     // https://github.com/apostrophecms/sanitize-html#what-are-the-default-options
-    allowedTags: sanitizeHtml.defaults.allowedTags.concat([
+    allowedTags: [
       'img',
       'figure',
       'br',
       'p',
       'a',
+      'ol',
+      'nl',
       'div',
       'li',
       'td',
@@ -24,8 +26,10 @@ module.exports = function(dirty) {
       'ul',
       'tr',
       'h3',
+      'td',
       'i',
       'h4',
+      'pre',
       'blockquote',
       'hr',
       'table',
@@ -36,7 +40,6 @@ module.exports = function(dirty) {
       'h6',
       'thread',
       'caption',
-    ]),
-    disallowedTagsMode: 'discard',
+    ],
   });
 };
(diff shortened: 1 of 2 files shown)


src/backend/utils/sanitize-html.js
```diff
@@ -7,6 +7,36 @@ module.exports = function(dirty) {
   return sanitizeHtml(dirty, {
     // Add <img> to the list of allowed tags, see:
     // https://github.com/apostrophecms/sanitize-html#what-are-the-default-options
-    allowedTags: sanitizeHtml.defaults.allowedTags.concat(['img', 'figure']),
+    allowedTags: sanitizeHtml.defaults.allowedTags.concat([
```
humphd: This won't remove `<iframe>`, only add the tags in your list:

```js
> ['iframe', 'a', 'img'].concat(['img', 'figure', 'p'])
[ 'iframe', 'a', 'img', 'img', 'figure', 'p' ]
```

You would have to remove items from the original array (i.e., sanitizeHtml.defaults.allowedTags).  However, another approach, since you're listing them all manually below, would be to just define your own list:

```js
allowedTags: [
  // your list of tags here...
], 
```

If you're going to do it this way, please consolidate the list you have below with what's in https://github.com/apostrophecms/sanitize-html#what-are-the-default-options, since you will miss some otherwise.

test/sanitize-html.test.js
```diff
@@ -34,6 +34,13 @@ describe('Sanitize HTML', () => {
     );
   });
 
+  test('<iframe> gets removed after run through sanitizeHTML', () => {
+    const data = sanitizeHTML(
+      '<iframe src="https://www.telescope.com" style="border:none;">Telescope</iframe>'
+    );
+    expect(data).toBe('<iframe>Telescope</iframe>');
```
humphd: So it hasn't stripped the `<iframe>` but rather, removed the `src` attribute.  Is that expected?  I would have thought it would remove the entire tag.  See my comment above.

src/backend/utils/sanitize-html.js
```diff
@@ -36,7 +40,6 @@ module.exports = function(dirty) {
       'h6',
       'thread',
       'caption',
-    ]),
-    disallowedTagsMode: 'discard',
```
humphd: You removed this, but I think it's probably a good idea.  Do we not need it to remove tags that aren't listed here, or is this the default?

PR #686, opened 2020-02-08
Refactor Login frontend component to be functional

humphd:
@Grommers00, @agarcia-caicedo and I were discussing functional vs. class components in React.  I refactored the recent Login component we were working on at the time to be functional so we would have an example to discuss.

The main changes here:

- split out the `LoggedIn` and `LoggedOut` components to their own files
- switch from using an ES6 class to a function for Login
- use `const [email, setEmail] = useState(null);` to deal with state.  The `useState()` React function returns an array with 2 things: the current value of the state, a function to update it.
- switch from `componentDidMount()` to `useEffect()`.  Notice the use of the `[]` as the last arg, which will mean it only fires once on startup.
- moved away from Promise-based `fetch` code to use async/await

### DIFF
--- a/src/frontend/src/components/Login/LoggedIn.jsx
+++ b/src/frontend/src/components/Login/LoggedIn.jsx
@@ -1,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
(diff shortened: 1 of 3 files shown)


humphd:
Rebased, and will merge once CI is green.

PR #687, opened 2020-02-08
adds: docker multistage builds for reduce container sizes

raygervais:
## Type of Change

## Description

image

Generates docker containers using a the native multistage build process, specifically for the staging environment. This reduces the current container size from ~1.91GB (640MB with @humphd's experiment) to  ~251MB!

```
telescope_telescope           latest              95b308226ca8        4 minutes ago       251MB
<none>                        <none>              5957bf2d5b55        4 minutes ago       940MB
<none>                        <none>              986285b4fdda        8 minutes ago       773MB
node                          12.15.0-alpine3.9   72805eb6a2b6        46 hours ago        84.9MB
redis                         latest              44d36d2c2374        7 days ago          98.2MB
kristophjunge/test-saml-idp   latest              a85d01fab7df        2 years ago         457MB
```

**IMPORTANT**
To test locally, use the staging.Dockerfile  instead of developer.Dockerfile  in the docker-compose.yml file.

## Checklist

- [ ] **Quality**: This PR builds and passes our npm test and works locally _Tested locally on W10, unable to verify npm test due to the reduced node packages on final container_

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
-# This Docker file tends to be used only
-# for `development` and `test` stage
+# This Docker file is used  for
+# `development`, `test`, and `staging` enviornments
 #
 # CLI command to run this file: script="some-name" docker-compose up --build
 # `some-name` is one of the names provided under `scripts` tag in `package.json` file
@@ -8,37 +8,45 @@
 
 
 # Dockerfile
+#
+# -------------------------------------
+# Context: Build Context
+FROM node:lts-alpine as build
 
-# Use `node` long term stable image as the parent to build this image
-FROM node:lts
+# Tini Entrypoint for Alpine
+RUN apk add --no-cache tini
+ENTRYPOINT [ "/sbin/tini", "--"]
 
-# Use Redis image via compose-file
-
-# Change working directory on the image
+# Set Working Directory Context
 WORKDIR "/telescope"
 
-# Copy package.json and .npmrc from local to the image
-# Copy before bundle app source to make sure package-lock.json never gets involved
-COPY package.json ./
-COPY .npmrc ./
-# Bundle app source
+# Copy shared context
+COPY package.json .
+
+# -------------------------------------
+# Context: Dependencies
+FROM build AS dependencies
+
+# Get Entire Codebase
 COPY . .
 
-# Install all Node.js modules on the image
-RUN npm install --no-package-lock
-
-# Build Frontend before deployment
+# Install Production Modules, Build!
+RUN npm install --only=production --no-package-lock
 RUN npm run build
 
-# ADD . src/frontend/public
 
-# User is root by default (only for development)
+# -------------------------------------
+# Context: Release
+FROM build AS release
 
-# Expose the server port from compose file
+# GET deployment code from previous containers
+COPY --from=dependencies /telescope/node_modules /telescope/node_modules
+COPY --from=dependencies /telescope/src/frontend/public /telescope/src/frontend/public
+COPY ./src/backend ./src/backend
 
 # Environment variable with default value
 ENV script=start
 
 # Running telescope when the image gets built using a script
-# `script` is one of the scripts from `packege.json`, passed to the image
+# `script` is one of the scripts from `package.json`, passed to the image
 CMD ["sh", "-c", "npm run ${script}"]
--- a/developer.Dockerfile
+++ b/developer.Dockerfile
@@ -1,0 +1,46 @@
+# This Docker file tends to be used only
+# for `development` and `test` stage
+#
+# CLI command to run this file: script="some-name" docker-compose up --build
+# `some-name` is one of the names provided under `scripts` tag in `package.json` file
+# example: $ script="test" docker-compose up --build    --> will run telescope with `test` script
+# default: $ docker-compose up --build                  --> will run telescope with `start` script
+
+
+# Dockerfile
+
+# Context: Build Container
+# Use `node` long term stable image as the parent to build this image
+FROM node:lts-alpine as build
+
+# Use Redis image via compose-file
+
+# Change working directory on the image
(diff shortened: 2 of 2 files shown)


docker-compose-staging.yml
```diff
@@ -7,7 +7,7 @@ services:
     container_name: telescope_staging
     build:
       context: .
-      dockerfile: Dockerfile
+      dockerfile: staging.Dockerfile
```
humphd: Can we leave this as `Dockerfile` for the default/dev case?  In staging/prod we can do more work to pick the right name.

staging.Dockerfile
```diff
@@ -0,0 +1,74 @@
+# This Docker file tends to be used only
+# for `development` and `test` stage
```
humphd: This comment seems wrong now, in the context of this file.

staging.Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+# `some-name` is one of the names provided under `scripts` tag in `package.json` file
+# example: $ script="test" docker-compose up --build    --> will run telescope with `test` script
+# default: $ docker-compose up --build                  --> will run telescope with `start` script
+
+
+# Dockerfile
+#
+# ------------------
+# Context: Base Container
+FROM node:12.15.0-alpine3.9 AS base
```
humphd: Question: why switch from `FROM node:lts` as you do above, and not use `FROM node:lts-alpine`?

staging.Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+FROM build AS dependencies
+
+# Install Production Modules
+RUN npm install --only=production --no-package-lock
+
+# Make Copy of Production Modules
+RUN cp -R node_modules prod_node_modules
+
+# Install Dev Modules (skip if we're not testing)
+RUN npm install --no-package-lock
```
humphd: Why bother with this?  If we're going to do this, should it be gated on a flag we check from the env?

staging.Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+
+#
+# Context: Testing
+# FROM dependencies AS test
+# COPY . .
+# RUN npm run pretest && npm run test
```
humphd: Should we move this kind of thing to `staging-test.Dockerfle` or something?

staging.Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+
+#
+# Context: Frontend
+FROM build as frontend
+COPY . .
+RUN cd /telescope/src/frontend
```
humphd: How will this interact with our `postinstall` script in the root package.json?  Are we now running this twice?

staging.Dockerfile
raygervais: We can, commented at the time due to the node version failing. I'll revise and ensure it's working before we merge this.

staging.Dockerfile
raygervais: Ah I didn't notice this before, I can redo the container flow logic to account for this, or we can remove the call for `postInstall` since it's handled in the container context now (would have to update developer Dockerfile too).

staging.Dockerfile
raygervais: Agreed, will clean up file and explain some choices / logics in comments.

staging.Dockerfile
raygervais: Node:LTS is based off of Debian Stretch https://github.com/nodejs/docker-node/blob/46f5203674c748b0701135c8eeea4b250b3ecb6d/12/stretch/Dockerfile which is bigger, great for ensuring we are able to install / compile dependencies before moving the final artifacts to the smaller Alpine container.

staging.Dockerfile
raygervais: Opting to test the former, since it's cleaner for the development workflow and is the least intrusive.

staging.Dockerfile
raygervais: I misunderstood what you meant before, looking now I see a better implementation which addresses your comment. Testing now.

staging.Dockerfile
humphd: Yeah, let's not break our dev's workflow.

Dockerfile
```diff
@@ -8,37 +8,45 @@
 
 
 # Dockerfile
+#
+# -------------------------------------
+# Context: Build Context
+FROM node:12.15.0-alpine3.9 as build
```
humphd: The only downside of this, is that we'll have to keep updating this manually as the LTS version progresses.  Maybe that's a good thing, since a change could break us, but it's something we should remember.

developer.Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+# `some-name` is one of the names provided under `scripts` tag in `package.json` file
+# example: $ script="test" docker-compose up --build    --> will run telescope with `test` script
+# default: $ docker-compose up --build                  --> will run telescope with `start` script
+
+
+# Dockerfile
+
+# Context: Build Container
+# Use `node` long term stable image as the parent to build this image
+FROM node:lts
```
humphd: This is not mirroring the version you have above.  Let's keep them the same.

Dockerfile
raygervais: We could look into a better way to manage, I'd argue being explicit is better for deployment behavior compared to using `latest` for example.

humphd: `lts` is ["Long Term Stable"](https://nodejs.org/en/about/releases/) for node vs. "Latest."  If we always use whatever is the current LTS, we're a sweet spot for maintenance long term.  If we pin to some release, it will drift, and go out of date without intervention.

The LTS line doesn't break things very often.  If we can find a way to say on that, I'd like to.

raygervais: Good point, corrected in latest and tested.

Issue #691, opened 2020-02-09
Allow authentication tokens to make it through nginx proxy


humphd:
Our login on staging doesn't work correctly anymore:

<img width="973" alt="Screen Shot 2020-02-09 at 11 23 42 AM" src="https://user-images.githubusercontent.com/427398/74105761-a65e5400-4b2e-11ea-8121-0f7e6c489a3c.png">

I think the issue here is that our HTTPS ngnix reverse proxy to the node app is not allowing authentication headers to pass through to the client.

I'm not 100% how to fix this.  Some initial reading points at http://nginx.org/en/docs/http/ngx_http_auth_request_module.html and also these extra nginx config settings:

```
       proxy_pass_header  Authorization;
       proxy_set_header Authorization $http_authorization;
       proxy_pass_request_body off;
       proxy_set_header Content-Length "";
       proxy_set_header X-Original-URI $request_uri;
```

We might also need to do something in our node app to have it trust tokens that have come across from a proxy.

c3ho:
I’ll take this issue on, want to learn about this stuff.

humphd:
I ran into another issue with login today on HTTPS for staging.  When I try to login, our express server is using a GET request to `/auth/login/callback` but it should be a POST.

You can try to Login on https://telescope-eq0rhvdfk.now.sh/ and after you do, you'll see that it gives a 404 while trying to GET `/auth/login/callback`.  We have that, but it's POST.

humphd:
@c3ho @manekenpix, have we covered all this?  I think it's working well.  Can we close this?

humphd:
I'm going to close this until we hit a bug that relates to needing it.

PR #696, opened 2020-02-11
Fix: Change response from 301 to 307 and redirect request to the right route

manekenpix:
## Issue This PR Addresses

Progress towards #691 

## Type of Change

## Description

This is initial work to be able to use `SSO` with `HTTPS`.
This PR makes sure that `POST` requests sent from a client don't turn into `GET` requests due to a [301](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301) response in our `nginx.conf`. For that, [307](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307) will be used instead.

> The only difference between 307 and 302 is that 307 guarantees that the method and the body will not be changed when the redirected request is made. With 302, some old clients were incorrectly changing the method to GET: the behavior with non-GET methods and 302 is then unpredictable on the Web, whereas the behavior with 307 is predictable.

Also, `login.telescope.cdot.systems` now redirects to the correct route.

## Checklist

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -4,7 +4,7 @@ http {
   server {
     listen 80 default_server;
     server_name _;
-    return 301 https://$host$request_uri;
+    return 307 https://$host$request_uri;
   }
 
   server {
@@ -30,7 +30,7 @@ http {
     listen 80;
     server_name login.telescope.cdot.systems;
     location / {
-      proxy_pass http://telescope_staging/login;
+      proxy_pass http://telescope_staging:3000/auth/login;
     }
   }
 }


nginx.conf
```diff
@@ -30,7 +30,7 @@ http {
     listen 80;
     server_name login.telescope.cdot.systems;
     location / {
-      proxy_pass http://telescope_staging/login;
+      proxy_pass http://telescope_staging:3000/auth/login;
```
humphd: We also have the same flow for `/auth/logout`, should we add that too?

manekenpix: I'm not sure.
I think `/auth/logout` will be handled in:
```
server {
    listen 80 default_server;
    server_name _;
    return 307 https://$host$request_uri;
  }
```
`/auth/login` needs to be mention so we can tell nginx where to redirect all the requests sent to `login.telescope`, but (I might be wrong about this) I think `/auth/logout` will always be part of a `dev.telescope` request and it will go through the `307`, like when we requests `posts`, `feeds`, etc.

humphd: Make sense.  Let's go with what you have and fix it later if it's wrong.  I bet you're right, though.

c3ho:
I asked how I can test this. I was told I couldn't. I'm going to blindly approve this.

Issue #665, opened 2020-02-06
Nested Queries in GraphQL


c3ho:
**What would you like to be added**:
GraphQL filters should be going in soon (crosses fingers), we should get nested queries going in too.

**Why would you like this to be added**:
Gives our front-end people more flexibility when using GraphQL APIs

c3ho:
@manekenpix you've previously done work/research on nested queries so I'm going to volunteer you for this.

manekenpix:
Nice, I'll need your help since you worked on the filters so we can come up with something good combining filters and nested queries.

c3ho:
Closed via #718.

manekenpix:
I think I can do some more work with this one.

PR #718, opened 2020-02-13
Remove author, site from Post, add feed id

### LINKED ISSUE #716: Update Post class to have a feedId property
To support the addition of new feeds, the `post` class in `post.js` will have to be updated to support.
We need to add a new property called `feedId` and perhaps remove `url` and `site` as that information can be retrieved from the feed class (not too sure). Some of this might block on #706 as it will need functions to interact with Redis.

We need these changes to support the addition of new feeds

humphd:
## Issue This PR Addresses

Fixes #716, #665

## Type of Change

## Description

This is a WIP to remove duplicated feed info such as `author` and `site` from a Post, and replace it with the id of a `feed`.  For example:

```
# Post
curl -H 'Accepts: application/json' http://localhost:3000/posts/00107ba358
{
  "id": "00107ba358",
  "title": "Part 3: ZFS Setup",
  "html": "<p>...</p>",
  "published": "2019-12-18T04:43:05.000Z",
  "updated": "2019-12-18T04:43:05.000Z",
  "url": "https://ebraublog.wordpress.com/2019/12/18/part-3-zfs-setup/",
  "guid": "http://ebraublog.wordpress.com/?p=58",
  "feed": "4f53e7478d"
}

# Feed
curl -H 'Accepts: application/json' http://localhost:3000/feeds/4f53e7478d
{
  "id": "4f53e7478d",
  "author": "Eric Brauer",
  "url": "https://ebraublog.wordpress.com/category/CDOT/feed/",
  "etag": "W/\"1e7899a6c0d9861dcc15f2434092f4d3\"",
  "lastModified": "Thu, 13 Feb 2020 03:41:58 GMT"
}
```

Since we know the `feed` its `id` when we process each `post`, it's easy to add the `feed`'s `id` right then.

Some questions I have about things I have to solve before we can merge this:

1. What should `Post.byId("00107ba358")` return for the feed info?  Options include: a) raw id "feed": "4f53e7478d"; b) expanded feed data `feed: { id: ..., name, ..., url, ...}`.  The latter would make this easier to use in the frontend, so you don't have to do more queries to get the feed info.
1. How do I fix the GraphQL to use this properly?  If i do b) above, it would mean you already have the data you need; otherwise I have to somehow go and get it in the query.

Based on these, I'll need to fix the frontend to use this data.  I also need to move the site URL info into the feed itself.

### DIFF
--- a/src/backend/index.js
+++ b/src/backend/index.js
@@ -22,22 +22,22 @@ process.on('uncaughtException', shutdown('UNCAUGHT EXCEPTION'));
 /**
  * Adds the feed to the database if necessary, or gets a more complete
  * version of the feed if we have better data already.
- * @param {Feed} feed - a feed Object parsed from the wiki feed list.
- * Returns the most appropriate Feed Object to use.
+ * @param {Object} feedData - feed data parsed from the wiki feed list.
+ * Returns Promise<Feed>, with the most appropriate Feed Object to use.
  */
-async function updateFeed(feed) {
+async function updateFeed(feedData) {
   let currentFeed;
 
-  // If we have an existing feed in the database for this id, prefer that,
-  // since it might have updated cache info (e.g., etag)
-  const existingFeed = await Feed.byId(feed.id);
+  // If we have an existing feed in the database for this URL, prefer that,
+  // since it might have updated cache info (e.g., etag).
+  const existingFeed = await Feed.byUrl(feedData.url);
   if (existingFeed) {
     // We have a version of this feed in the database already, prefer that
     currentFeed = existingFeed;
   } else {
     // First time we're seeing this feed, add it to the database
-    currentFeed = feed;
-    await currentFeed.save();
+    const id = await Feed.create(feedData);
+    currentFeed = await Feed.byId(id);
   }
 
   return currentFeed;
@@ -47,13 +47,11 @@ async function updateFeed(feed) {
  * Invalidates a feed
  * @param feedData - Object containing feed data
  */
-async function invalidateFeed(feedData) {
-  const feed = Feed.parse(feedData);
-  await feed.setInvalid(feedData.reason || 'unknown reason');
-  logger.info(
-    `Invalidating feed ${feedData.url} for the following reason: ${feedData.reason ||
-      'unknown reason'}`
-  );
+async function invalidateFeed(id) {
+  const feed = await Feed.byId(id);
+  // TODO: we need to bubble up the reason for the failure
+  await feed.setInvalid('unknown reason');
+  logger.info(`Invalidating feed ${feed.url} for the following reason: ${'unknown reason'}`);
 }
 
 /**
@@ -66,7 +64,7 @@ function processFeeds(feeds) {
       // Save this feed into the database if necessary.
       const currentFeed = await updateFeed(feed);
       // Add a job to the feed queue to process all of this feed's posts.
-      await feedQueue.addFeed(currentFeed);
+      await feedQueue.addFeed({ id: currentFeed.id });
     })
   );
 }
@@ -104,7 +102,7 @@ feedQueue.on('drained', loadFeedsIntoQueue);
  * and save to Redis
  */
 feedQueue.on('failed', job =>
-  invalidateFeed(job.data).catch(error => logger.error({ error }, 'Unable to invalidate feed'))
+  invalidateFeed(job.data.id).catch(error => logger.error({ error }, 'Unable to invalidate feed'))
 );
 
 /**
--- a/src/backend/web/graphql/index.js
+++ b/src/backend/web/graphql/index.js
@@ -26,15 +26,14 @@ module.exports.typeDefs = graphql`
   # 'Post' matches our Post type used with redis
   type Post {
     id: String
-    author: String
     title: String
     html: String
     text: String
     published: String
     updated: String
     url: String
-    site: String
     guid: String
+    feed: String
   }
 
   # Queries to fetch data from redis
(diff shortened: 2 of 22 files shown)


humphd:
For reference, it's interesting to see what GitHub does in this case with their API:

https://developer.github.com/v3/pulls/#response

If I want to get data on a pull request, I also get things like user info, labels, milestone, etc. all expanded in place vs. just an id pointing to those things.  My gut says I should expand `feed` info.  I'll try that.

c3ho:
So everytime we're checking a post, we're also expecting a feedId so we can combine all the info together and store it into Redis?

src/backend/web/graphql/index.js
```diff
@@ -26,15 +26,14 @@ module.exports.typeDefs = graphql`
...
     title: String
     html: String
     text: String
     published: String
     updated: String
     url: String
-    site: String
     guid: String
+    feed: String
   }
```
c3ho: We can make feed return a type [Feed] if we go with the latter option (I think). The problem is I don't know if you can do nested queries both ways. When I played with nested queries, I changed the Feed schema so that it also has posts like below: 
image

So when you get Feed, you can see posts by that author.

What I'm not sure of is if you can do that for Feed AND change the schema for Post so that it does the following at the same time.
```
 type Post {
    id: String
    title: String
    html: String
    text: String
    published: String
    updated: String
    url: String
    guid: String
    feed: [Feed]
  }
```

humphd:
This is getting close.  I have a few more spots where we rely on the old layout of `Post` with `author`, `site` properties instead of a `feed`.  I can't get the Planet to use the feed author properly yet.  But the existing Gatsby components and graphql are working:

Screen Shot 2020-02-13 at 10 34 42 PM

Give me another few days to test this, but it's pretty close.  I'll probably document it a bit too.

c3ho:
This review is gonna take a while 20 files...

humphd:
@c3ho it will probably be closer to 25 by the time I'm done :(  Turns out we have a lot of code that depends on the layout of our Feed and Post.  Changing data structures is painful, which is why I took this issue.

humphd:
OK, I think this is ready.  I'm 98% sure I've got all the places where we assume an `author` lives on a `Post`.

All tests pass, and all the code I can think to test works.  I've updated the authors filter in graphql to work for partial matches that don't care about case.  But I'm not 100% sure how to test this.

If people want to review this, I can help do it on Slack or in person.

This will help with the work happening on Adding a Feed from Gatsby.  You will be able to do this:

```js
const Feed = require('../src/backend/data/feed');

const id = await Feed.create({
  author: 'Feed Author',
  url: 'http://feed-url.com/',
});
```

This will create the feed in Redis, add it to the Feeds set, and return the `id`.  The `POST` route from Gatsby can basically just do that, and it will work (without dealing with ownership, which is fine for now).

c3ho:
Can help review

src/backend/web/routes/feed.js
```diff
@@ -67,8 +67,8 @@ const feedRoute = type => async (req, res) => {
       content: post.html,
       author: [
         {
-          name: post.author,
-          link: post.site,
+          name: post.feed.author,
+          // TODO: add site info for this feed https://github.com/Seneca-CDOT/telescope/issues/724
```
Grommers00: this will be for gatsby correct?

src/backend/web/graphql/index.js
```diff
@@ -26,15 +26,14 @@ module.exports.typeDefs = graphql`
...
-    author: String
     title: String
     html: String
     text: String
     published: String
     updated: String
     url: String
-    site: String
     guid: String
+    feed: Feed!
```
Grommers00: Why is it Feed!  and not Feed

(Curiosity)

src/backend/index.js
```diff
@@ -104,7 +102,7 @@ feedQueue.on('drained', loadFeedsIntoQueue);
  * and save to Redis
  */
 feedQueue.on('failed', job =>
-  invalidateFeed(job.data).catch(error => logger.error({ error }, 'Unable to invalidate feed'))
+  invalidateFeed(job.data.id).catch(error => logger.error({ error }, 'Unable to invalidate feed'))
```
Grommers00: Wouldn't this mean it's unable to validate the feed? 

am I wrong thinking this is a double negative?

src/backend/web/routes/feed.js
humphd: No, this part, the `/feed/rss` type routes are for generating an RSS/ATOM/etc feed for our post content.  That is, if someone wanted to import all the blogs in our system into their feed reader, they would use this.

src/backend/web/graphql/index.js
humphd: @c3ho was telling me it means "this field is required"

src/backend/index.js
humphd: This is like saying:

```js
try {
  // Try and add an 'invalid' key for this feed in Redis
  await invalidateFeed(job.data.id);
} catch(error) {
  // Something went wrong trying to set the invalid key in Redis, report it and move on
 logger.error({ error }, 'Unable to invalidate feed'); 
}
```

However, when you have a callback, you can't mix callbacks and async functions, so we're using a `.catch()` to deal with the error on the Promise we get from Redis.

c3ho:
I'm not familiar enough to review the `Login.jsx` file, but otherwise the code looks good.

humphd:
The `Login.jsx` change is for an eslint issue.

humphd:
Thanks for the reviews.  This has the potential to break something that I missed, since it's so wide-ranging in its effect.

@cindyledev you're the master of finding fallout from my changes, if you notice anything break (esp. around the author for feeds) let me know.

Also, everyone will need to `redis-cli flushall` their database to clear the existing data and get the new layout.  @manekenpix will have to redo staging before the backend data will work with local Gatsby apps, too.

PR #719, opened 2020-02-13
fix #707 add routes - work in progress

### LINKED ISSUE #707: Add a POST route to allow a user to add a feed
Add a POST route to allow a user to add feeds

Once #706 goes in we need to allow users to enter their feeds so we can add them to Redis

Grommers00:
## Issue This PR Addresses
    USAGE: Fixes #707 

## Type of Change

- [x ] **New Feature**: Change which adds functionality

## Description

This is just to get the ball rolling, and start building the express routes. 
Going to need @c3ho to help with the redis part, so we can understand how redis works from a coding standpoint and to move forward. 

## Checklist

 *This is a work in progress at the moment*

### DIFF
--- a/src/backend/web/routes/feeds.js
+++ b/src/backend/web/routes/feeds.js
@@ -1,5 +1,6 @@
 const express = require('express');
 const Feed = require('../../data/feed');
+const { protect } = require('../authentication');
 const { getFeeds } = require('../../utils/storage');
 const { logger } = require('../../utils/logger');
 
@@ -47,6 +48,21 @@ feeds.get('/:id', async (req, res) => {
     logger.error({ err }, 'Unable to get feeds from Redis');
     res.status(503).json({
       message: 'Unable to connect to database',
+    });
+  }
+});
+
+feeds.post('/', protect, async (req, res) => {
+  const { newFeed, email } = req.body;
+  try {
+    /* check if email has a key in redis
+  if not create key
+  if user has a key in redis, check if feed already exists, (not sure how fast this is? or if its necessary), if not store feed in feeds.
+  */
+  } catch (error) {
+    logger.error({ error }, 'Unable to add feed to redis');
+    res.status(503).json({
+      message: 'Unable to add feed to database.',
     });
   }
 });
--- a/src/backend/web/routes/users.js
+++ b/src/backend/web/routes/users.js
@@ -1,0 +1,17 @@
+const express = require('express');
+
+const { logger } = require('../../utils/logger');
+const feeds = require('../../data/feeds');
+const { getFeeds } = require('../../utils/storage');
+
+const router = express.Router();
+
+router.get('/:id', async (req, res) => {
+  // this route will return one user, and there list of feeds, this can be utilized in two places. One our search function, and
+  // two for our add feed component to show a list of feeds that have been attached to the user.
+  /* this should return
+    { id: "some-ud-id, url: "/users/someuserId/ (multiple URLs?)}
+  */
+});
+
+module.exports = router;


Grommers00:
Also note - I want to build it bit by bit, so there's progression happening.

src/backend/web/routes/feeds.js
```diff
@@ -51,4 +52,19 @@ feeds.get('/:id', async (req, res) => {
   }
 });
 
+feeds.post('/', protect, async (req, res) => {
+  const { newFeed, email } = req.body;
+  try {
+    /* check if email has a key in redis
+  if not create key
+  if user has a key in redis, check if feed already exists, (not sure how fast this is? or if its necessary), if not store feed in feeds.
```
Silvyre: >not sure how fast this is?

From what I understand, Redis is super fast (plus can fetch that info in constant time), so we probably don't have to worry about speed in this use case.

>or if its necessary

Also from what I understand, it is impossible to add a duplicate item to Redis set. I'm not sure yet what the end result of such an attempt is (I'm guessing an error being thrown, which you've set up handling for, here...). So, I infer that it's not necessary to implement a uniqueness check, but going to defer to @c3ho on this one.

src/backend/web/routes/users.js
```diff
@@ -0,0 +1,22 @@
...
+
+const router = express.Router();
+
+router.get('/', async (req, res) => {
+  // this route will return all users.
+  /* [{id: "some-user-id", url: "/users/some-user-id", ...] */
+});
+
+router.get('/:id', async (req, res) => {
+  // this route will return one user, and there list of feeds, this can be utilized in two places. One our search function, and
```
Silvyre: >and there list of feeds

(For future reference/reminder: as per discussion in today's class, implementing this will require that we first ensure that feed information stored in Redis includes the user's ID.)

src/backend/web/routes/feeds.js
c3ho: Yeah, Redis won't error, but will just overwrite the value of the key-value 
`redis-cli set hi bye` will create a key called `hi` with the value `bye`
`redis-cli get hi` will return `bye`
`redis-cli set hi hi` will overwrite the current value of `bye` with `hi`
`redis-cli get hi` will now return `hi`

src/backend/web/routes/feeds.js
```diff
@@ -51,4 +52,19 @@ feeds.get('/:id', async (req, res) => {
...
+  */
+  } catch (err) {
+    logger.error({ err }, 'Unable to add feed to redis');
```
c3ho: We're using `(error)` instead of  `(err)`

src/backend/web/routes/feeds.js
Grommers00: so if i'm not mistaken the key will be linked to the user? so, what if I post a feed, and say Rafi does to a blog, would that duplicate the blog feed? (I only ask because I do believe some school staff have a seneca email and a seneca work email)

src/backend/web/routes/users.js
Grommers00: Yes - so that's the part I think I'll need help with is, figuring out how the hashing function works. 

I believe right now we hash the URL for the feed, but going forward we would be hashing the username if I'm not mistaken?

src/backend/web/routes/feeds.js
Grommers00: okay cool, ill fix accordingly.

src/backend/web/routes/feeds.js
c3ho: t:users will contain
`t:user:grommers00@myseneca.ca:feeds`
`t:user:silvyre@myseneca.ca:feeds`

`t:user:grommers00@myseenca.ca:feeds` can contain
`t:feed:a3e61 (a feed for OSD600)` <-- not too sure about the exact wording or way, but it will be implemented something like this
`t:feed:a3eb1 (a feed for OSD700)`

The key should be linked to a specific user


>so, what if I post a feed, and say Rafi does to a blog, would that duplicate the blog feed? (I only ask because I do believe some school staff have a seneca email and a seneca work email)

Not sure what you mean by this, but we could have a duplication if a user signs up with their seneca email and also a seneca work email as we'll have two users now
`t:user:grommers00@myseneca.ca:feeds`
`t:user:grommers00@work.myseneca.ca:feeds`

src/backend/web/routes/users.js
c3ho: @humphd should we hash usernames?

src/backend/web/routes/feeds.js
humphd: `t:users` will contain `grommers00@myseneca.ca` and to get the user info, you'd create a full key from that "id": `t:user:grommers00@myseneca.ca`.  We probably don't need the `:feeds` suffix, since that's the only thing we'll store for a user (i.e., their list of feeds).

Then, `t:user:grommers00@myseneca.ca` will be a Set that contains just the id for each of the feeds that this user owns.  To get those feeds, you do `t:feed:{id}`

src/backend/web/routes/users.js
humphd: Personally, I'm not sure we need to expose Users to the frontend at all.  We want it to exist internally in Redis, so we can match logged in users with their feeds; but we don't need to list all our users, so we can probably get rid of `/users` as a route.

Grommers00: I have removed the /users route, so it would be /users/:id only.

src/backend/web/routes/feeds.js
Silvyre: @humphd Makes sense!

PR #729, opened 2020-02-16
fix #707 adding post route for /feeds

### LINKED ISSUE #707: Add a POST route to allow a user to add a feed

Grommers00:
## Issue This PR Addresses

## Description

This will allow us to connect our front end and back end to post the data to add the feeds to the master list. 

## Checklist

### DIFF
--- a/src/backend/web/routes/feeds.js
+++ b/src/backend/web/routes/feeds.js
@@ -2,6 +2,7 @@ const express = require('express');
 const Feed = require('../../data/feed');
 const { getFeeds } = require('../../utils/storage');
 const { logger } = require('../../utils/logger');
+// const { protect } = require('../authentication'); https://github.com/Seneca-CDOT/telescope/issues/734
 
 const feeds = express.Router();
 
@@ -10,8 +11,8 @@ feeds.get('/', async (req, res) => {
 
   try {
     ids = await getFeeds();
-  } catch (err) {
-    logger.error({ err }, 'Unable to get feeds from Redis');
+  } catch (error) {
+    logger.error({ error }, 'Unable to get feeds from Redis');
     res.status(503).json({
       message: 'Unable to connect to database',
     });
@@ -33,9 +34,8 @@ feeds.get('/:id', async (req, res) => {
   const { id } = req.params;
 
   try {
-    const feed = await Feed.byId(id);
-
     // If the object we get back is empty, use 404
+    const feed = await Feed.byId(id);
     if (!feed) {
       res.status(404).json({
         message: `Feed not found for id ${id}`,
@@ -43,10 +43,33 @@ feeds.get('/:id', async (req, res) => {
     } else {
       res.json(feed);
     }
-  } catch (err) {
-    logger.error({ err }, 'Unable to get feeds from Redis');
+  } catch (error) {
+    logger.error({ error }, 'Unable to get feeds from Redis');
     res.status(503).json({
       message: 'Unable to connect to database',
+    });
+  }
+});
+
+// add protect here for development. https://github.com/Seneca-CDOT/telescope/issues/734
+
+feeds.post('/', async (req, res) => {
+  const feedData = req.body;
+  try {
+    if (!(feedData.url && feedData.author)) {
+      return res.status(400).json({ message: `URL and Author must be submitted` });
+    }
+    if (await Feed.byUrl(feedData.url)) {
+      return res.status(409).json({ message: `Feed for url ${feedData.url} already exists.` });
+    }
+    const feedId = await Feed.create(feedData);
+    return res
+      .status(201)
+      .json({ message: `Feed was successfully added.`, id: feedId, url: `/feeds/${feedId}` });
+  } catch (error) {
+    logger.error({ error }, 'Unable to add feed to Redis');
+    return res.status(503).json({
+      message: 'Unable to add to Feed',
     });
   }
 });
--- a/test/feeds.test.js
+++ b/test/feeds.test.js
@@ -4,7 +4,7 @@ const app = require('../src/backend/web/app');
 const Feed = require('../src/backend/data/feed');
 const hash = require('../src/backend/data/hash');
 
-describe('test /feeds endpoint', () => {
+describe('test GET /feeds endpoint', () => {
   const createdItems = 150;
 
   // Array of feeds
@@ -60,5 +60,61 @@ describe('test /feeds/:id responses', () => {
     expect(res.get('Content-type')).toContain('application/json');
     expect(res.body).toEqual(receivedFeed);
     expect(res.body instanceof Array).toBe(false);
+  });
+});
+
+// https://github.com/Seneca-CDOT/telescope/issues/734 Tests below will need to have authorization to pass.
+describe('test POST /feeds endpoint', () => {
+  it('responds with json', async () => {
+    const feedData = {
+      author: 'foo',
+      url: 'http://telescope200.cdot.systems',
+    };
+    const res = await request(app)
+      .post('/feeds')
+      .send(feedData)
+      .set('Accept', 'application/json');
+    expect(res.status).toEqual(201);
+    const { id } = res.body;
+    const feed = await Feed.byId(id);
+    expect(feed.author).toEqual(feedData.author);
+    expect(feed.url).toEqual(feedData.url);
+  });
+
+  it('no author being sent', async () => {
+    const feedData = {
+      author: null,
+      url: 'http://telescope200.cdot.systems',
+    };
+    const res = await request(app)
+      .post('/feeds')
+      .send(feedData)
+      .set('Accept', 'application/json');
+    expect(res.status).toEqual(400);
+  });
+
+  it('no url being sent', async () => {
+    const feedData = {
+      author: 'foo',
+      url: null,
+    };
+    const res = await request(app)
+      .post('/feeds')
+      .send(feedData)
+      .set('Accept', 'application/json');
+    expect(res.status).toEqual(400);
+  });
+
+  it('url already in the feed list', async () => {
+    const feedData = {
+      author: 'foo',
+      url: 'http://telescope0.cdot.systems',
+    };
+    await Feed.create(feedData);
+    const res = await request(app)
+      .post('/feeds')
+      .send(feedData)
+      .set('Accept', 'application/json');
+    expect(res.status).toEqual(409);
   });
 });


src/backend/web/routes/feeds.js
```diff
@@ -51,4 +51,26 @@ feeds.get('/:id', async (req, res) => {
   }
 });
 
+feeds.post('/', async (req, res) => {
```
humphd: This route should get `protect` middleware: you can't do it unless you're logged in.

src/backend/web/routes/feeds.js
```diff
@@ -51,4 +51,26 @@ feeds.get('/:id', async (req, res) => {
...
+  const feedData = req.body;
```
humphd: We need to validate this data, since we require that an `author` and `url` get passed.  If either is missing, you need to send back an HTTP 400, indicating that the request is invalid, and that it's missing data.

src/backend/web/routes/feeds.js
```diff
@@ -51,4 +51,26 @@ feeds.get('/:id', async (req, res) => {
...
+
+  try {
+    const feed = await Feed.create(feedData);
```
humphd: Let's check if this feed already exists in our system before we try to add it again:

```js
if(await Feed.byUrl(feedData.url)) {
  res.status(409).json({message: `Feed for url ${feedData.url} already exists.`});
  return;
}
```

src/backend/web/routes/feeds.js
```diff
@@ -51,4 +51,26 @@ feeds.get('/:id', async (req, res) => {
...
+
+    // If the object we get back is empty, use 404
```
humphd: This "404" block can be removed.  We'll deal with the case of the Redis call failing in your `catch` below.

src/backend/web/routes/feeds.js
```diff
@@ -51,4 +51,26 @@ feeds.get('/:id', async (req, res) => {
...
+    if (!feed) {
+      res.status(404).json({
+        message: `Feed not created for url ${feedData.url}`,
+      });
+    } else {
+      res.json(`${feed.url} has successfully been added.`);
```
humphd: When you create a new record, you need to send back a 201, and give them enough info to get the url they need for the record.

```js
res.status(201).json({message: `Feed was successfully added.`, id: feed, url: `/feeds/${feed}` });
```

humphd:
This also needs some tests.

src/backend/web/routes/feeds.js
Grommers00: Just wondering though - if it's a protected route, we will have the author correct? 

So, do I just need to check that there is a URL?

humphd: We'll have a username, but not necessarily a real, fullname.  Consider the comment you left in Julia's design issue:

- "humphd"
- vs. "David Humphrey"

The author isn't a user, but a display name.  Later, we might consolidate this somehow, but we have a ton of feeds that don't have associated users, so we can't currently rely on that mapping.

src/backend/web/routes/feeds.js
```diff
@@ -51,4 +52,26 @@ feeds.get('/:id', async (req, res) => {
   }
 });
 
+feeds.post('/', protect, async (req, res) => {
+  const feedData = req.body;
+  if (!feedData.url && !feedData.author) {
+    res.status(400).json({ message: `URL and Author must be submitted` });
```
humphd: You need to `return` after this so it doesn't keep processing this request.

src/backend/web/routes/feeds.js
```diff
@@ -51,4 +52,26 @@ feeds.get('/:id', async (req, res) => {
...
+  }
+  if (await Feed.byUrl(feedData.url)) {
```
humphd: Throw this `if` block inside your `try/catch`, since `Feed.byUrl()` can throw.

src/backend/web/routes/feeds.js
```diff
@@ -51,4 +52,26 @@ feeds.get('/:id', async (req, res) => {
...
+    res.status(409).json({ message: `Feed for url ${feedData.url} already exists.` });
+    return;
+  }
+  try {
+    const feed = await Feed.create(feedData);
```
humphd: Since what you get back from `Feed.create` is an `id` for your feed, let's call it that instead of `feed`: `const id = ...`

src/backend/web/routes/feeds.js
```diff
@@ -51,4 +52,26 @@ feeds.get('/:id', async (req, res) => {
...
+    res
+      .status(201)
+      .json({ message: `Feed was successfully added.`, id: feed, url: `/feeds/${feed}` });
+  } catch (err) {
```
humphd: Can you use `error` vs. `err` please, it makes it consistent with the rest of our code.

src/backend/web/routes/feeds.js
```diff
@@ -51,4 +52,26 @@ feeds.get('/:id', async (req, res) => {
   }
 });
 
+feeds.post('/', protect, async (req, res) => {
+  const feedData = req.body;
+  if (!feedData.url && !feedData.author) {
```
humphd: This won't cover every case (i.e., this will only get triggered if both are missing, but not one or the other).  You probably want to say, `if(!(feedData.url && feedData.author)) {`

src/backend/web/routes/feeds.js
Grommers00: Also fixing the other get routes in the file.

src/backend/web/routes/feeds.js
Grommers00: Updated.

test/feeds.test.js
```diff
@@ -62,3 +62,65 @@ describe('test /feeds/:id responses', () => {
     expect(res.body instanceof Array).toBe(false);
   });
 });
+
+describe('test POST /feeds endpoint', () => {
+  const createdItems = 5;
+  const feedData = {
+    author: 'foo',
+    url: 'http://telescope200.cdot.systems',
```
Silvyre: I'm curious about this choice of URL; does this test suite only require a random/broken URL?

src/backend/web/routes/feeds.js
```diff
@@ -2,6 +2,7 @@ const express = require('express');
 const Feed = require('../../data/feed');
 const { getFeeds } = require('../../utils/storage');
 const { logger } = require('../../utils/logger');
+// const { protect } = require('../authentication');
```
manekenpix: I think you can remove this

test/feeds.test.js
```diff
@@ -62,3 +62,65 @@ describe('test /feeds/:id responses', () => {
...
+  it('url already in the feed list', async () => {
+    feedData.url = 'http://telescope3.cdot.systems';
+
+    const res = await request(app)
+      .post('/feeds')
+      .send(feedData)
+      .set('Accept', 'application/json');
+    expect(res.status).toEqual(409);
+  });
+});
```
manekenpix: What's the difference between these 2?

Grommers00: Oops.

test/feeds.test.js
```diff
@@ -62,3 +62,65 @@ describe('test /feeds/:id responses', () => {
...
+  it('responds with json', async () => {
+    const res = await request(app)
+      .post('/feeds')
+      .send(feedData)
+      .set('Accept', 'application/json');
+    expect(res.status).toEqual(201);
+  });
+
+  it('no author being sent', async () => {
+    feedData.author = null;
```
Silvyre: Since, to my knowledge, Jest [test methods](https://jestjs.io/docs/en/api#testname-fn-timeout) run in parallel, modifying a shared resource like this might lead to a race condition.

src/backend/web/routes/feeds.js
Grommers00: Once David adds in his code protect will be added back in. @humphd didn't know if I should or just comment in for now as a reminder to add it

test/feeds.test.js
```diff
@@ -62,3 +62,65 @@ describe('test /feeds/:id responses', () => {
...
+    author: 'foo',
+    url: 'http://telescope200.cdot.systems',
+  };
+
+  // Array of feeds
+  const feeds = [...Array(createdItems).keys()].map(item => {
+    return new Feed('foo', `http://telescope${item}.cdot.systems`);
+  });
+
+  beforeAll(() => Promise.all(feeds.map(feed => feed.save())));
```
manekenpix: If you're going to manually add feeds in every test (using POST) you might not need this

test/feeds.test.js
Grommers00: So this is where I was having issues, if I used 6 here. It would throw a 409 error. Which showed as a duplicate. Yet I was only creating 5? So it confused me.

test/feeds.test.js
Grommers00: That makes more sense why I had to use the 200.

test/feeds.test.js
Grommers00: True. But do tests run in parallel or in order? If in order I don't need this. In parallel I would.

Grommers00:
> I think you can include your tests in the `describe` that we already have so you won't have to repeat code ([DRY vs WET](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself#DRY_vs_WET_solutions))

Not sure about this because a Post route is different than a GET route. Unless you were referring to something different?

test/feeds.test.js
Grommers00: And it just requires a URL that hasn't been added previously.

test/feeds.test.js
manekenpix: This part only adds feeds using `feed.save()`, which is not very useful to you since you want to do `POST` and test `json`, `url`, `author`. I think maybe creating 3 Feed objects (one is ok, one without author, one without url) and using them in the specific test could be all you need.

manekenpix:
> Not sure about this because a Post route is different than a GET route. Unless you were referring to something different?

You're right. If you end up not using `beforeAll` maybe it's better to have your tests in a different set.

humphd:
Some really good ideas in here.  I'll just add/answer a few more:

- consider each test as an individual "unit", and don't rely on anything external to that test if possible.  For example, if you need some data to use, create it in that test.  I usually do this by having functions that return data, and call them over and over again in different tests vs. creating variables.  Variables are what make tests fail.
- Tests can and will run in any order.  As such, don't have one test depend on state from another (e.g., don't add something in one test, and try to read it in the next test).
- Tests might not all get run.  A dev might run a single test, so don't rely on anything in other tests.
- If you require the world to be in a certain state for your test, do that in the test itself.  If all your tests need that, do `beforeAll()` or `beforeEach()`, but prefer doing things in the test if possible so you know they happened in the order you need them to, and nothing else changed their state.

In summary: don't share data between tests, don't rely on one test in another, don't rely on test order.

test/feeds.test.js
```diff
@@ -62,3 +62,54 @@ describe('test /feeds/:id responses', () => {
     expect(res.body instanceof Array).toBe(false);
   });
 });
+
+describe('test POST /feeds endpoint', () => {
+  it('responds with json', async () => {
+    const feedData1 = {
```
humphd: Minor nit: all these variables can be named `feedData` without the 1, 2, 3 suffix, since they are local to the function.

src/backend/web/routes/feeds.js
```diff
@@ -34,21 +35,42 @@ feeds.get('/:id', async (req, res) => {
 
   try {
     const feed = await Feed.byId(id);
-
-    // If the object we get back is empty, use 404
```
humphd: Not sure why you removed these lines.  Not critical, but I'd avoid changing things not related to your code.

test/feeds.test.js
```diff
@@ -62,3 +62,55 @@ describe('test /feeds/:id responses', () => {
...
+    const feedData = {
+      author: 'foo',
+      url: 'http://telescope200.cdot.systems',
+    };
+    const res = await request(app)
+      .post('/feeds')
+      .send(feedData)
+      .set('Accept', 'application/json');
+    expect(res.status).toEqual(201);
+  });
```
humphd: Let's modify this test a bit.  I care less about it responding with JSON, and more that it gives back a usable id.  After your code above, do this:

```js
// See if the feed was created as expected.
const id = res.body.id;
const feed = await Feed.byId(id);
expect(feed.author).toEqual(feedData.author);
expect(feed.url).toEqual(feedData.url);
```

This code is untested, but hopefully it gives you enough.

src/backend/web/routes/feeds.js
Grommers00: Misclick :)

PR #735, opened 2020-02-17
Complete Gatsby MaterialUI installation plus 'Add Feeds' page wireframing

### LINKED ISSUE #725: Prototype the Add Feed view in React

Anticipating @yatsenko-julia's delivery of a design/mockup for a new `Add Feed` view, we should take at least the initial steps necessary to implement this design by creating a React component to house this design.

Once @yatsenko-julia delivers her design for the `Add Feed` page, these initial steps task will already be complete, reducing the amount of time it will take to implement it later.

### LINKED ISSUE #730: Pick a React component library as our base UI toolkit

Let's pick a well supported React component library, and use it as a base for building our designs.  The ones I know, and would suggest:

- https://react-bootstrap.github.io/
- https://material-ui.com/

Microsoft's Fabric components https://developer.microsoft.com/en-us/fabric#/controls/web are neat too, but look a bit too "Office" for our needs.

There are a lot of unmaintained ones too, and I'd suggest we avoid.

Initially we've been creating all of our own components, and it's a huge time sink.  So many things have to be done to get these right, from styling to accessibility to UX.  Let's not take on the work of creating everything from scratch when there are so many good frameworks we can use as a base.

NOTE: I'm not suggesting we don't do design work and create our own look-and-feel.  Rather, I'd rather we began from an existing set of components, and customized those vs. vanilla HTML/CSS/JS.

For example, we currently need to create a search component.  Here's how you can do it in Material-UI:

https://material-ui.com/components/text-fields/#customized-inputs

And in React Bootstrap:

https://react-bootstrap.github.io/components/forms/

Both allow you to specify things like colour and font, add buttons and icons, etc.

The cost to using one of these is that we have to learn a bit about how to use their components; but what we gain is that we don't have to implement 90% of what already exists.

Does anyone have a strong preference? I'm most familiar with Material-UI, but am happy to use React Bootstrap, too.

Once we pick one, we can start by porting our existing code over to use the framework's components, then extend.

Silvyre:
## Issue This PR Addresses

Addresses #730 - Pick a React component library as our base UI toolkit 
Resolves #725 - Prototype the Add Feed view in React 

## Type of Change

## Description

- Addresses #730 by completing our frontend's installation of MaterialUI with [`gatsby-theme-material-ui`](https://www.gatsbyjs.org/packages/gatsby-theme-material-ui/#webfonts). The remainder of the MaterialUI installation was previously done by @miggs125.
  - (It is worth noting that, by default, the installation of `gatsby-theme-material-ui` enforces some global styling changes to the frontend, e.g. changing the default font to Roboto. This default styling [can be overwritten](https://www.gatsbyjs.org/packages/gatsby-theme-material-ui/#webfonts), if we'd prefer.)
  - (AFAIK, we're keeping #730 open for the purposes of discussion.)
- Resolves #725 by adding a 'My Feeds' [Gatsby Page](https://www.gatsbyjs.org/docs/creating-and-modifying-pages/) (accessed via [`/myfeeds`](http://localhost:8000/myfeeds)) wireframed with MaterialUI components.
   - This is just a wireframe; none of the implied functionality has been implemented (and I suggest that we finalize the component's design before putting major work into its implementation).

Preview of 'My Feeds' component:

- It is worth noting that this component currently ends up being rendered quite small due to styling enforced by `src/frontend/src/pages/index.css`:
   https://github.com/Seneca-CDOT/telescope/blob/9dc85649f1c29d1a7fceafc9b7ccbda668beb22d/src/frontend/src/pages/index.css#L9-L13

## Checklist

### DIFF
--- a/src/frontend/src/pages/myfeeds.js
+++ b/src/frontend/src/pages/myfeeds.js
@@ -29,7 +29,7 @@ export default function MyFeeds() {
                     <AccountCircle />
                   </Grid>
                   <Grid item>
-                    <TextField disabled label="John Doe" />
+                    <TextField id="author" label="John Doe" disabled />
                   </Grid>
                 </Grid>
               </Grid>
@@ -39,7 +39,7 @@ export default function MyFeeds() {
                     <RssFeed />
                   </Grid>
                   <Grid item>
-                    <TextField label="Blog feed URL" />
+                    <TextField id="url" label="Blog feed URL" />
                   </Grid>
                   <Grid item>
                     <IconButton color="primary" classes={{ root: classes.button }}>
--- a/src/frontend/src/pages/myfeeds.js
+++ b/src/frontend/src/pages/myfeeds.js
@@ -1,0 +1,64 @@
+import React from 'react';
+import { Container, Box, Typography, TextField, Grid, Card, IconButton } from '@material-ui/core';
+import { AccountCircle, RssFeed, HelpOutline, Add } from '@material-ui/icons';
+import { makeStyles } from '@material-ui/core/styles';
+
+const useStyles = makeStyles(theme => ({
+  margin: {
+    margin: theme.spacing(2),
+  },
+  button: {
+    padding: '3px 0 3px 0',
+  },
+}));
+
+export default function MyFeeds() {
+  const classes = useStyles();
+  return (
+    <div className={classes.margin}>
+      <Container maxWidth="xs" bgcolor="aliceblue">
+        <Card>
+          <Box px={2} py={1}>
+            <Typography variant="h3" component="h3" align="center">
+              My Feeds
+            </Typography>
+            <Grid container spacing={5}>
+              <Grid item>
+                <Grid container spacing={1} alignItems="flex-end">
+                  <Grid item>
+                    <AccountCircle />
+                  </Grid>
+                  <Grid item>
+                    <TextField disabled label="John Doe" />
+                  </Grid>
+                </Grid>
+              </Grid>
+              <Grid item>
+                <Grid container spacing={1} alignItems="flex-end">
+                  <Grid item>
+                    <RssFeed />
+                  </Grid>
+                  <Grid item>
+                    <TextField label="Blog feed URL" />
+                  </Grid>
+                  <Grid item>
+                    <IconButton color="primary" classes={{ root: classes.button }}>
+                      <HelpOutline />
+                    </IconButton>
+                  </Grid>
+                </Grid>
+                <Grid container spacing={2}>
+                  <Grid item>
+                    <IconButton classes={{ root: classes.button }}>
+                      <Add />
+                    </IconButton>
+                  </Grid>
+                </Grid>
+              </Grid>
+            </Grid>
+          </Box>
+        </Card>
+      </Container>
+    </div>
+  );
+}
--- a/src/frontend/gatsby-config.js
+++ b/src/frontend/gatsby-config.js
@@ -48,5 +48,6 @@ module.exports = {
         icon: `src/images/logo.svg`, // This path is relative to the root of the site.
       },
     },
+    `gatsby-theme-material-ui`,
   ],
 };
--- a/src/frontend/package.json
+++ b/src/frontend/package.json
@@ -13,6 +13,7 @@
     "gatsby-plugin-react-helmet": "^3.1.16",
     "gatsby-plugin-sharp": "^2.3.5",
     "gatsby-source-filesystem": "^2.1.40",
+    "gatsby-theme-material-ui": "^1.0.8",
     "gatsby-transformer-sharp": "^2.3.7",
     "prop-types": "^15.7.2",
     "react": "^16.12.0",


src/frontend/src/pages/myfeeds.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import { Container, Box, Typography, TextField, Grid, Card, IconButton } from '@material-ui/core';
+import { AccountCircle, RssFeed, HelpOutline, Add } from '@material-ui/icons';
+import { makeStyles } from '@material-ui/core/styles';
+
+const useStyles = makeStyles(theme => ({
```
Grommers00: Do we want to keep styles located in the react component? Is there a way to store this in the css file that should be located with the page to keep with the current format of the front end?

src/frontend/src/pages/myfeeds.js
```diff
@@ -0,0 +1,64 @@
...
+              My Feeds
+            </Typography>
+            <Grid container spacing={5}>
+              <Grid item>
+                <Grid container spacing={1} alignItems="flex-end">
+                  <Grid item>
+                    <AccountCircle />
+                  </Grid>
+                  <Grid item>
+                    <TextField disabled label="John Doe" />
```
Grommers00: Can we add the id = "author"

src/frontend/src/pages/myfeeds.js
```diff
@@ -0,0 +1,64 @@
...
+                  </Grid>
+                </Grid>
+              </Grid>
+              <Grid item>
+                <Grid container spacing={1} alignItems="flex-end">
+                  <Grid item>
+                    <RssFeed />
+                  </Grid>
+                  <Grid item>
+                    <TextField label="Blog feed URL" />
```
Grommers00: And id = "url"

src/frontend/src/pages/myfeeds.js
Silvyre: Yes, it is technically possible to replace this instance of `makeStyles` with pure CSS, i.e.. 

```css
.MuiContainer-root {
  position: relative;
  top: 16px;
}

.MuiIconButton-root {
  padding: 3px 0 3px 0;
}
```

Relying more on [MaterialUI's styling solutions](https://material-ui.com/styles/basics/#hook-api) may become more useful in the future if/when we want to incorporate more MUI functionalities, such as customized [theming](https://material-ui.com/styles/advanced/#theming).

However, since I'm still pretty new to MUI, I'd like to defer to @miggs125 and @humphd regarding whether it would be better, in this case, to embrace or move away from MUI styling solutions.

src/frontend/src/pages/myfeeds.js
Silvyre: Done: https://github.com/Seneca-CDOT/telescope/pull/735/commits/24c6a44423e887991d032f2425cb39a0ab044943

src/frontend/src/pages/myfeeds.js
Silvyre: Done: https://github.com/Seneca-CDOT/telescope/pull/735/commits/24c6a44423e887991d032f2425cb39a0ab044943

src/frontend/src/pages/myfeeds.js
Grommers00: I agree - not sure if having each react component contain our materialUI solutions or if we should grab it from a separate page that would house all the material UI.

I do think we should keep anything CSS related outside of the functionality part. Perhaps we could import it and have a different file such as myFeedsCss.js? (or materials not sure what to call this file)

This would allow the CSS developers to quickly locate where the CSS is that would need to be changed, and not get them stuck in reading everything within a component?

Thoughts @humphd ?

humphd: Depending on the outcome of #730, it looks like we might use MUI as a base.  If we do, then I think moving toward the MUI style approach is nice, because it takes less work to translate what their examples in the docs to to what we'll have.  Doing CSS-in-JS this way is kind of nice.

I think do whatever you want in this PR, and we'll change it to match #730.  I'll add a note about this topic there.

Silvyre:
> 
> 
> Can we also add another option for the main page to navigate here :D. Perhaps in the nav bar?

In my opinion, a higher priority would be better to enable route guarding for this component. However, both of these functionalities would be more appropriately handled within separate issues.

Grommers00:
> > Can we also add another option for the main page to navigate here :D. Perhaps in the nav bar?
> 
> In my opinion, a higher priority would be better to enable route guarding for this component. However, both of these functionalities would be more appropriately handled within separate issues.

Agreed, However nav bar is rendered in a way that we only want this page shown in logged in mode anyways.  

As I don't know if we are adding in protected routes this way..or necessarily how as we don't have a state manager given that it's a simple application having the link only show up when logged in the navbar I think could suffice?

Silvyre:
> 
> 
> > > Can we also add another option for the main page to navigate here :D. Perhaps in the nav bar?
> > 
> > 
> > In my opinion, a higher priority would be better to enable route guarding for this component. However, both of these functionalities would be more appropriately handled within separate issues.
> 
> Agreed, However nav bar is rendered in a way that we only want this page shown in logged in mode anyways.
> 
> As I don't know if we are adding in protected routes this way..or necessarily how as we don't have a state manager given that it's a simple application having the link only show up when logged in the navbar I think could suffice?

That's a good point, especially coupled with the fact that there is talk of designing Telescope as a single page site.

PR #772, opened 2020-03-03
Use MUI for post layout, fix UI to be closer to design

humphd:
I was really happy with the improvements to the frontend in 0.7, and wanted to keep pushing things forward.  I spent a bit of time tonight reworking the post layout so that it was a bit closer to our design (I'm not going to get it all the way in this PR, but I've started us toward it).  I've also reworked things to use MUI.

In particular, this PR does the following:

1. renames the `TextArea` component to `Posts`, since "TextArea" means something different in HTML.  We now have `Posts` and `Post`, with the former being a collection of the latter.
1. tried to use MUI styling, based on discussions in class today.
1. added a space for the avatar and name to go, but it's empty right now.
1. made the posts responsive using `Grid` and `Container` components
1. namespaced all the post content CSS styles to a class, so they won't leak into the page.
1. fixed a CSS issue for `display` with the emojis (we need `inline-block` to set a width)
1. fixed a bunch of font/text/colour/size issues in the layout

I'd love to do more than this, but I don't want this PR to get any larger.


raygervais:
Question, what is the purpose of the sidebar beside each Post? Asking because the darker gray area is jarring unless we do have an intended use coming up for it. _Mentioning because we are looking at the area_, **ducks and hides**.

src/frontend/src/components/Post/Post.jsx
```diff
@@ -1,26 +1,82 @@
...
+        <a href={url} rel="bookmark" className={classes.published}>
+          <time dateTime={date}>{formatPublishedDate(date)}</time>
+        </a>
+      </header>
+
+      <Grid container>
+        <Grid item sm={12} md={9} className={classes.content}>
+          <section className="telescope-post-content" dangerouslySetInnerHTML={{ __html: html }} />
+        </Grid>
+        <Grid item sm={false} md={3} className={classes.sidebar}></Grid>
```
raygervais: Let's remove the sidebar (comment out if we want to preserve) for now since it's an empty grid item. This will allow us to set the `md={12}` and utilize the entire width of the container for Post content. Willing to drop this suggestion if we see it better to preserve the sidebar space -if anything, perhaps color sidebar same color as content background?

agarcia-caicedo:
> 
> 
> Question, what is the purpose of the sidebar beside each Post? Asking because the darker gray area is jarring unless we do have an intended use coming up for it. _Mentioning because we are looking at the area_, **ducks and hides**.

The bar exists because when I made the design in Adobe XD, the width that had been recommended to me made the post component very skinny, and there was a lot of blank space around each post.
So as a way to keep the width, I made this design:

image

It looked ok in XD, but if we feel that it doesn't work with Telescope, we can definitely update the design.

src/frontend/src/components/Post/Post.jsx
humphd: Pushed a new commit to remove the sidebar for now.

Issue #765, opened 2020-03-02
Header covers search bar component


agarcia-caicedo:
**What happened**:
The header covers the search bar component, so the search bar cannot be seen.

![image](https://user-images.githubusercontent.com/49992710/75641868-a15d6380-5c07-11ea-87b9-e2563d651df4.png)

![image](https://user-images.githubusercontent.com/49992710/75641988-fb5e2900-5c07-11ea-9712-dc7a59dccc37.png)

**What should have happened**:

The search bar should be under the header

**How to reproduce it (as precise as possible)**:
Running Telescope as usual and then going to the search page, either through clicking the search button in the header, or by url

**Anything else we need to know?**:

This is a result of the header position being fixed. Any fixes should aim to keep the position fixed, as it allows the header to always be available for the user.

**Environment**:

- OS: Windows 10
- Browser: Firefox

PR #792, opened 2020-03-04
Make search bar return results

### LINKED ISSUE #750: Make search bar return results
Connect our search bar to queries from GraphQL. #747 should be going in soon, we now have access to GraphQL queries in the front end. 

We have the search bar component in place, author result component should be merged soon. All that is left is to connect the search bar to our GraphQL getPosts query so we can render author result component based on results

c3ho:
## Issue This PR Addresses
FIxes #750 

## Type of Change

## Description
WIP, I think this is close? Just need to figure out how to deal with this issue
```
Uncaught Invariant Violation: Could not find "client" in the context or passed in as an option. Wrap the root component in an <ApolloProvider>, or pass an ApolloClient instance in via options.
```

## Checklist


src/frontend/src/components/SearchBar/SearchBar.jsx
```diff
@@ -39,11 +49,12 @@ const useStyles = makeStyles(theme => ({
   },
 }));
 
-export default function CustomizedInputBase() {
+function CustomizedInputBase() {
   const classes = useStyles();
   const [state, setState] = React.useState({
     filter: '',
   });
+  const [results, setResults] = React.useState('[]');
```
humphd: Why is your array in quotes?

src/frontend/src/components/SearchBar/SearchBar.jsx
```diff
@@ -52,6 +63,14 @@ export default function CustomizedInputBase() {
     });
   };
 
+  const [getResults, { loading, error, data }] = useLazyQuery(FEED_SEARCH_QUERY);
+  if (loading) console.log(null);
```
humphd: This will need to change

src/frontend/src/components/SearchBar/SearchBar.jsx
```diff
@@ -75,10 +94,17 @@ export default function CustomizedInputBase() {
           placeholder="Search Telescope"
           inputProps={{ 'aria-label': 'search telescope' }}
         />
-        <IconButton type="submit" className={classes.iconButton} aria-label="search">
+        <IconButton
+          type="submit"
+          className={classes.iconButton}
+          aria-label="search"
+          onClick={() => getResults()}
```
humphd: For debugging, this might be OK, but in general you want to pass your event handlers down from a parent control to a controlled component.  So we'd put the search results logic in the parent of this search component, and then just wire up the button here.

src/frontend/src/components/SearchBar/SearchBar.jsx
```diff
@@ -1,12 +1,22 @@
...
 import Paper from '@material-ui/core/Paper';
 import InputBase from '@material-ui/core/InputBase';
 import IconButton from '@material-ui/core/IconButton';
 import SearchIcon from '@material-ui/icons/Search';
 import FormControl from '@material-ui/core/FormControl';
 import NativeSelect from '@material-ui/core/NativeSelect';
 
+const FEED_SEARCH_QUERY = gql`
+  {
+    getPosts(page: 0, perPage: 30, filter: $filter) {
```
humphd: How does `$filter` get set?

src/frontend/src/components/SearchBar/SearchBar.jsx
```diff
@@ -1,12 +1,22 @@
...
+import { withApollo, useQuery, useLazyQuery } from 'react-apollo';
+import gql from 'graphql-tag';
 import Paper from '@material-ui/core/Paper';
 import InputBase from '@material-ui/core/InputBase';
 import IconButton from '@material-ui/core/IconButton';
 import SearchIcon from '@material-ui/icons/Search';
 import FormControl from '@material-ui/core/FormControl';
 import NativeSelect from '@material-ui/core/NativeSelect';
 
+const FEED_SEARCH_QUERY = gql`
```
humphd: You're querying for posts, so naming this FEED_ seems confusing.

c3ho: Sorry this is real rough as I'm taking whatever from docs and playing around with it!

src/frontend/src/components/SearchBar/SearchBar.jsx
c3ho: you mean like
```
function handleOnClick(){
  useLazyQuery()
}
```

humphd: Here's a very rough outline of what I mean:

```jsx
// In Parent component, which manages the state for the search, and does the gql stuff...
function ParentComponent() {
  const [searchText, setSearchText] = useState('');
  const onSearchClick = () => { /* do your gql search here... */ };

  return (
    <ChildSearchComponent onSearchClick={onSearchClick} currentSearchText={searchText} />
  );
}

// In Child component
function ChildSearchComponent(props) {
  // Get the click handler, and current text to put in the search bar from the parent...
  const { onSearchClick, currentSearchText } = props;
  return (
    <div>
      <TextInput value={currentSearchText} />
      <Button onClick={onSearchClick}>Search</Button>
    </div>
  );
}
```

You manage the search text, and "on click" handler in a parent component, and the search component just displays buttons and your text, and tells you when to do things (run the function).

c3ho: Think I understand. Once I can actually get this working, I'll set it up like this.

c3ho:
@humphd this was the issue I'm receiving. Which is weird as I've wrapped (or to the best of my knowledge) the <Layout /> component with <ApolloProvider > as per instructions. 

image

image

To reproduce this error, click on the search to get the search bar to appear, then click on the search button again. I've read online differing package versions for apollo could cause this issue, so I've removed the apollo-boost. I've tried using '@apollo/react-hooks' package also but that still doesn't get rid of this issue.

humphd:
@c3ho I think the problem you have here is that you're wrapping the components in `pages/index.js` but not `pages/search.js`, which is where the `SearchBar` component is used.

c3ho:
@humphd is there a proper way to set a state here for results? When I'm trying to use the hook like `setResults`, it goes into something like a re-render hell

humphd:
A couple of things I notice:

- you're doing too much in your input base component.  It shouldn't be getting results from your query.  That should happen in a parent component, where the results get consumed and rendered.  Imagine you have a `Search` component, which is the root of your search UI.  It might be where you get/render search results, and also where you manage the state for the current value in the search input box.  You pass that down via props, and also a function to be called when the user clicks on "search".  You have the beginnings of it here, but you need to split it up, and not try to get the results in this component.

- you probably need to read about and add [useEffect()](https://reactjs.org/docs/hooks-effect.html) to manage side effects.  When you start changing state, you're going to re-trigger rendering, which might be what you're seeing.

c3ho:
I tried your comment in class of doing the client stuff for apollo in useEffect, it was giving me the Apolloprovider client issue I had previously, so I just opted to do it outside of the hook unfortunately.

c3ho:
image

cindyledev:
If you're planning to make further changes to this PR, I have a very small nit, when I click on the post title, I don't want it to open in the same tab. I think opening the blog post in a different tab/window would be nicer

c3ho:
@cindyledev done it will now open in a new tab

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -0,0 +1,94 @@
...
+        onClickHandler={onClickHandler}
+        searchText={searchText}
+        onChangeHandler={onChangeHandler}
+      />
+      {// The initial state of results is going to be 0 and we can't map 0,
+      //  this will check if there are any results and return AuthorResult component
+      //  for each with feed guid as key, otherwise default to "No Results"
+      results.length > 0 ? (
+        results.map(result => {
+          console.log(results);
```
raygervais: Do we really want to console.log these results if we're past the development stage? @humphd thoughts?

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -0,0 +1,94 @@
...
+    setSearchText(event.target.value);
+  }
+
+  return (
+    <div>
+      <Header />
+      <div
+        style={{
+          height: '12vh',
+        }}
```
raygervais: For good practice, lets define a min-height as well if we are using `vh` units since this could cause scaling issue.

```
minHeight: '12em`
```

For example will leverage 120px min height if that's what we are going for. Adjust to taste

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -0,0 +1,94 @@
...
+      {// The initial state of results is going to be 0 and we can't map 0,
+      //  this will check if there are any results and return AuthorResult component
+      //  for each with feed guid as key, otherwise default to "No Results"
+      results.length > 0 ? (
+        results.map(result => {
+          console.log(results);
+          return <AuthorResult key={result.id} author={result.author} post={result.post} />;
+        })
+      ) : (
+        <h1>No Results</h1>
```
raygervais: Nit, let's wrap this in a container class with some padding so that it doesn't stick to the left side and is hard to spot when there's no results yet. Likewise I think we should follow this up (or add here) a third state: No text entered... This would be the default state and thus, your three displays: 

1. No search terms entered
2. No results found
3. Results


My recommendation for the styling of the no-text / no-search container: 
```css
margin: auto;
padding: 2rem;
color: coral;
font-weight: 400;
```

That, plus adding some material-grid goodness to centre it similar to the results will make it :100:

c3ho:
I'm inclined at this point to just remove the search button and have search just start whenever the user enters input. I've spent a whole day reading docs and stackoverflow without any success.

@raygervais it does what you want now, kind of. The empty `no search term` will appear up until a user enters input at which point the message will change to `no results found` until the user clicks on the search button. From there it will just keep executing the query based on user input regardless of whether the user clicks on the search button again.

scenarios
0 chars in searchbar:
image

results + chars in searchbar:
image

no results + chars in searchbar:
image

src/frontend/src/components/AuthorResult/AuthorResult.jsx
```diff
@@ -55,12 +51,13 @@ const useStyles = makeStyles(theme => ({
...
+export default function AuthorResult(props) {
   const classes = useStyles();
-  const [open, setOpen] = React.useState(false);
+  const { author } = props;
+  const { postDate, title, postLink } = props.post;
 
-  const handleClick = () => {
-    setOpen(!open);
+  const handleLatestPostClick = () => {
+    window.open(postLink);
```
humphd: This may be problematic.  Gatsby doesn't have a `window` at build time.  Have you tried doing a full build of your code?  Also, why are we opening a new window?

src/frontend/src/components/AuthorResult/AuthorResult.jsx
```diff
@@ -81,30 +78,23 @@ export default function AuthorResult() {
                   <ListItemIcon className={classes.icons}>
                     <PermContactCalendarIcon />
                   </ListItemIcon>
-                  <ListItemText className={classes.font} primary="Username" />
+                  <ListItemText className={classes.font} primary={`Username: ${author}`} />
```
humphd: It's not a `Username`, it's the author's name.  Why not call it that?

src/frontend/src/components/AuthorResult/AuthorResult.jsx
```diff
@@ -81,30 +78,23 @@ export default function AuthorResult() {
...
                 </ListItem>
                 <ListItem button className={classes.infoLine}>
                   <ListItemIcon className={classes.icons}>
                     <EventIcon />
                   </ListItemIcon>
-                  <ListItemText className={classes.font} primary="Date of Last Post: " />
+                  <ListItemText
+                    className={classes.font}
+                    primary={`Date of Last Post: ${postDate}`}
```
humphd: `postDate` should probably get converted to a human readable date.  You can probably use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat

src/frontend/src/components/AuthorResult/AuthorResult.jsx
```diff
@@ -81,30 +78,23 @@ export default function AuthorResult() {
...
+                  />
                 </ListItem>
-                <ListItem button className={classes.infoLine} onClick={handleClick}>
+                <ListItem button className={classes.infoLine} onClick={handleLatestPostClick}>
```
humphd: If the user is supposed to click on it, and it opens content, that's an `<a>`.  Why use a `<button>`?

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -0,0 +1,105 @@
...
+
+const useStyles = makeStyles(() => ({
+  searchReply: {
+    margin: 'auto',
+    padding: '2rem',
+    color: 'coral',
+    fontWeight: 400,
+  },
+  divAfterHeader: {
+    height: '12vh',
```
humphd: This will be huge on larger screens.  I would be careful with `vh` here.  Why not use flexbox or something?

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -0,0 +1,105 @@
...
+    // setting author variable here to use in the query above
+    variables: { author: searchText },
+
+    onCompleted: () => {
+      const returnedIds = [];
+      // We only want to display an author once, this would avoid the issue if
+      // there are multiple posts by the same author returned in our query
+      const finalResults = data.getPosts
+        .map(result => {
+          if (!returnedIds.includes(result.feed.id)) {
```
humphd: This isn't right, `map()` needs to always return a value in its callback.  Sounds like you want to `filter()` first.

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -0,0 +1,105 @@
...
+              post: {
+                title: result.title,
+                postLink: result.url,
+                postDate: result.published,
+              },
+            };
+          }
+        })
+        // Filters undefined items that might have been pushed/mapped when we're checking for duplicate author results
+        .filter(item => item);
```
humphd: Flip the order to `filter().map()`

humphd: Also, for future, the way you generally do this is: `.filter(Boolean)`.  This means you end up returning `Boolean(item)` which will give you true or false, depending on whether the object exists.  But here we can skip this step by filtering first.

src/frontend/package.json
```diff
@@ -2,18 +2,24 @@
...
   "private": true,
   "dependencies": {
+    "@apollo/react-hooks": "3.1.3",
     "@material-ui/core": "4.9.5",
     "@material-ui/icons": "4.9.1",
     "@material-ui/styles": "4.9.0",
+    "apollo-boost": "0.4.7",
+    "apollo-link-http": "1.5.16",
     "dotenv": "8.2.0",
+    "es6-promise": "4.2.8",
```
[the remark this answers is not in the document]
c3ho: removed, I think I forgot to remove it when I was trying to get apolloclient to work during `npm run build`

src/frontend/src/components/SearchPage/SearchPage.jsx
c3ho: So just use `<Box display="flex"> </Box>` I tried this and my SearchBar disappeared

src/frontend/src/components/AuthorResult/AuthorResult.jsx
cindyledev: I think opening a new window when the user clicks on a post makes sense. I'm gonna need to ask you more about `window` for the PWA thing on a later date

src/frontend/src/components/SearchPage/SearchPage.jsx
cindyledev: The `vh` is just a bandaid solution for #765 . We don't have a solution to fix it yet. The SearchBar is actually underneath the Header bar

humphd:
Ping me when this is rebased, and I'll re-review.

c3ho:
@humphd rebased and fixed conflict. I'll squash this after I get a pass on review since this is a ridiculous amount of commits.

src/frontend/package.json
```diff
@@ -13,7 +16,9 @@
     "gatsby-plugin-sharp": "2.4.5",
     "gatsby-source-filesystem": "2.1.48",
     "gatsby-transformer-sharp": "2.3.16",
+    "graphql-tag": "2.10.3",
     "highlight.js": "9.18.1",
+    "node-fetch": "2.6.0",
```
humphd: Why do we need this here?

c3ho: required for node implementation of apolloclient. Otherwise it won't build properly with `npm run build`

c3ho:
Merging this as it was approved prior to rebase.

PR #818, opened 2020-03-08
Improve colours for links, text in heading on post

humphd:
This plays with the colours in the heading and text of links in posts, which I want to connect more closely to our colour theme:

Screen Shot 2020-03-07 at 6 59 35 PM

### DIFF
--- a/src/frontend/src/components/Post/Post.jsx
+++ b/src/frontend/src/components/Post/Post.jsx
@@ -16,8 +16,9 @@ const useStyles = makeStyles({
   },
   header: {
     backgroundColor: '#335A7E',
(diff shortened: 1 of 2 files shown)


humphd:
I'm not sure if the flipped colours in the header are better or not.  I think having the title be more prominent vs. the published info, is good.  See what you think.

raygervais:
Not to block this PR, but I wonder if we could introduce an accent colour (following the Material guidelines) that would work better for the date/link line as opposed to the darker blue. Aside from this, I do like the change.

humphd:
@raygervais can you file that for follow-up? I think it would be good to do.

PR #822, opened 2020-03-09
Closes #537: Add full-text indexing and search to our backend

### LINKED ISSUE #537: Add full-text indexing and search to our backend

When we process feeds in our backend, and add them to Redis, let's also do full-text search indexing.

One common approach to this would be to use [Elasticsearch](https://www.elastic.co/products/elasticsearch):

- https://blog.logrocket.com/full-text-search-with-node-js-and-elasticsearch-on-docker/
- https://blog.patricktriest.com/text-search-docker-elasticsearch/

Since we're already using Redis, we could also try and use a Redis-backed solution:

- https://oss.redislabs.com/redisearch/index.html
- JS API https://github.com/stockholmux/redredisearch (not getting a lot of updates though...)

There are other options as well, so this needs some research.

I'll file a separate issue related to the REST API aspects of this.

In the frontend, we're going to need to be able to search for things in old posts.  This will enable that to work.

manekenpix:
## Issue This PR Addresses

Closes #537 

## Type of Change

## Description

This PR adds an [Elasticsearch](https://www.elastic.co/what-is/elasticsearch) container to telescope as its search engine. With this change, our processor will not only store Post objects in redis, but it will also store text and id ([same id used for redis](https://github.com/Seneca-CDOT/telescope/blob/master/docs/redis.md)) from posts in elasticsearch.
Elasticsearch can be installed and used locally (instructions [here](https://www.elastic.co/guide/en/elastic-stack-get-started/7.6/get-started-elastic-stack.html#install-elasticsearch)), or simply run using `docker-compose up elasticsearch`.

To test elasticsearch, run telescope fully locally, using docker-compose, or a hybrid of both, open your browser and navigate to:

[`http://localhost:9200/_cat/indices?v`](http://localhost:9200/_cat/indices?v) to list information about all the indexes (in this case, we only have one index: posts).

[`http://localhost:9200/posts/_search?q=text:open%20source`](http://localhost:9200/posts/_search?q=text:open%20source) to send a query to elastic search with the text `open source`.

Example of a response (in JSON):
```json
{
  "took":483,
  "timed_out":false,
  "_shards":{
    "total":1,"successful":1,
    "skipped":0,"failed":0},
    "hits":{
      "total":{"value":618,"relation":"eq"},
      "max_score":5.2338853,
      "hits":[
        {"_index":"posts","_type":"post","_id":"ffa154089a","_score":5.2338853,"_source":{"text":"Full blogpost text"}},
      ]
   }
}
```

## Checklist


env.staging
```diff
@@ -25,6 +25,14 @@ REDIS_PASSWORD=
 # MOCK_REDIS=1 will use an in-memory, mock Redis instance. Useful for testing.
 MOCK_REDIS=
 
+# Elasticsearch info
+ELASTIC_URL=http://127.0.0.1
+ELASTIC_PORT=9200
+# Max number of results per query
+ELASTIC_MAX_RESULTS=50
+# Delay to check connectivity with Elasticsearch in ms
+ELASTIC_DELAY=10000
```
humphd: `ELASTIC_DELAY_MS`

src/backend/lib/elastic.js
```diff
@@ -0,0 +1,85 @@
...
+const type = 'post';
+
+/**
+ * Indexes the text and id from a post
+ * @param text from a post
+ * @param postId same id used to store on redis the post object where the text is from
+ */
+const indexPost = (text, postId) => {
+  try {
+    esClient.index({
```
humphd: Does this not need to be `await`'ed?  Docs seem to use it.

src/backend/feed/worker.js
```diff
@@ -25,7 +27,39 @@ function getFeedWorkersCount() {
   return Math.min(count, cpuCount);
 }
 
-exports.start = function() {
+exports.start = async function() {
+  /**
+   * Elasticsearch needs time after deployment for setting up all its components.
```
humphd: I'm curious about this.  The elastic docs don't show this as being necessary.  Obviously you've read something I haven't, but it surprises me that we need to do this synchronous dance around a new connection.  Can you point me at docs to understand why this is necessary?  I wonder if the older JS client needed this and the new Promise-based on doesn't?

env.example
```diff
@@ -33,6 +33,14 @@ REDIS_PASSWORD=
 # MOCK_REDIS=1 will use an in-memory, mock Redis instance. Useful for testing.
 MOCK_REDIS=
 
+# Elasticsearch info
+ELASTIC_URL=http://127.0.0.1
+ELASTIC_PORT=9200
+# Max number of results per query
+ELASTIC_MAX_RESULTS=50
+# Delay to check connectivity with Elasticsearch in ms
+ELASTIC_DELAY=10000
```
humphd: `ELASTIC_DELAY_MS`

src/backend/feed/processor.js
```diff
@@ -102,7 +103,10 @@ function articlesToPosts(articles, feed) {
   return Promise.all(
     articles.map(async article => {
       try {
-        await Post.createFromArticle(article, feed);
+        const id = await Post.createFromArticle(article, feed);
+        // Use the returned id to get the 'text' from the post and index it
+        const { text } = await Post.byId(id);
```
humphd: Can you move this indexing step into `createFromArticle()` here: https://github.com/Seneca-CDOT/telescope/blob/master/src/backend/data/post.js#L134?  We could use `Promise.all()` there and let redis and elastic both work in parallel.

src/backend/feed/worker.js
manekenpix: When I first started testing I noticed how I was trying to index posts and failing because elasticsearch wasn't ready. The idea of waiting for elasticsearch to be ready is from [here](https://blog.logrocket.com/full-text-search-with-node-js-and-elasticsearch-on-docker/) (search `function checkconnection()` in the page).
> Now let’s consider that ElasticSearch is a huge system and can take some seconds to boot up. We can’t connect to ES until it is ready, so we need a function that checks when the ES server is ready:

```js
function checkConnection() {
  return new Promise(async (resolve) => {
    console.log("Checking connection to ElasticSearch...");
    let isConnected = false;
    while (!isConnected) {
      try {
        await esclient.cluster.health({});
        console.log("Successfully connected to ElasticSearch");
        isConnected = true;
      // eslint-disable-next-line no-empty
      } catch (_) {
      }
    }
    resolve(true);
  });
}
```
They use an infinite loop to wait for ES to be ready, but I thought it was better to have a timeout in case ES failed (without a timeout, if ES is down telescope will never launch).

src/backend/data/post.js
```diff
@@ -5,6 +5,7 @@ const textParser = require('../utils/text-parser');
 const Feed = require('./feed');
 const hash = require('./hash');
 const ArticleError = require('./article-error');
+const { indexPost } = require('../lib/elastic');
```
humphd: We've got our redis `storage` code in `utils/` and elastic in `lib/`.  They seem like they should sit beside each other.

src/backend/feed/worker.js
humphd: I read [a bit more into this](https://docs.docker.com/compose/startup-order/) and probably you're right that we make our dependent services something we wait on before we assume they are running and can be used.  I'll file a follow-up to do the same for Redis, since we should be consistent.

I wonder if we should actually crash if elastic isn't there (same for Redis).  Likely search will become more than a nice-to-have once we have it.

humphd: Can you move this whole block of code (down to line 61) into your elastic client code, and call it `waitOnReady()` or `init()` or something, and have it return a `Promise`?  That would simplify this code to:

```js
exports.start = async function() {
  try {
    await Promise.all([elastic.waitOnReady(), redis.waitOnReady()])
    // keep going with starting the workers here...
  } catch(err) {
    // Redis, Elastic, or both failed, we need to crash...
  }
}
```

src/backend/lib/elastic.js
```diff
@@ -0,0 +1,85 @@
...
+    await esClient.index({
+      index,
+      type,
+      id: postId,
+      body: {
+        text,
+      },
+    });
+  } catch (error) {
+    logger.error(`There was an error indexing a post for id ${postId}: ${error}`);
```
humphd: Do errors like this:

```js
logger.error({ error }, 'message...');
```

src/backend/lib/elastic.js
```diff
@@ -0,0 +1,85 @@
...
+  }
+};
+
+/**
+ * Searches text in elasticsearch
+ * @param textToSearch
+ * @return all the results matching the passed text
+ */
+const getIndexResults = async textToSearch => {
```
humphd: s/getIndexResults/search/

humphd:
Filed #828, #829 for follow-up.

src/backend/feed/worker.js
```diff
@@ -25,9 +27,24 @@ function getFeedWorkersCount() {
...
 }
 
-exports.start = function() {
-  const concurrency = getFeedWorkersCount();
-  logger.info(`Starting ${concurrency} instance${concurrency > 1 ? 's' : ''} of feed processor.`);
-  feedQueue.process(concurrency, path.resolve(__dirname, 'processor.js'));
-  return feedQueue;
+exports.start = async function() {
+  try {
+    await elastic.waitOnReady();
```
humphd: Nice.

src/backend/feed/worker.js
```diff
@@ -25,9 +27,24 @@ function getFeedWorkersCount() {
...
+  } catch (error) {
+    /**
+     * If elasticsearch is not initialized, we throw again to terminate Telescope.
+     * According to nodejs.org:
+     * "If it is necessary to terminate the Node.js process due to an error condition,
+     * throwing an uncaught error and allowing the process to terminate accordingly
+     * is safer than calling process.exit()"
+     */
+    logger.error(error);
+    throw new Error();
```
humphd: Let's re-throw the error so we get some context for the crash: `throw error;`

c3ho:
Tested functionality locally with options provided by @manekenpix https://www.elastic.co/guide/en/elasticsearch/reference/current/search-search.html#search-search-api-query-params was able to change size of index returned in API. Its pretty awesome that full text search is so quick and flexible.

src/backend/utils/elastic.js
```diff
@@ -0,0 +1,116 @@
...
+    logger.error({ error }, `There was an error indexing a post for id ${postId}`);
+  }
+};
+
+/**
+ * Searches text in elasticsearch
+ * @param textToSearch
+ * @return all the results matching the passed text
+ */
+const getIndexResults = async textToSearch => {
```
humphd: NIT: I still think this should be called `search` vs. `getIndexResults` but won't block this PR on it.

PR #823, opened 2020-03-09
Fix #757: reject posts without publication date

### LINKED ISSUE #757: Deal with partial/broken post content
Every time I test our app, the first post is always broken.  [This feed](http://compileofcrap.com/SPO600/feed.xml) has a number of problems, all of which we should consider how we want to deal with in our code:

- The feed itself is missing publication date info.  Previously, we rejected posts like this as invalid, but then we [added a fix to stamp them with the current date/time](https://github.com/Seneca-CDOT/telescope/blob/master/src/backend/data/post.js#L98-L107).  I think we should undo this, since it means that these posts rise to the top of everything else.  If a post is missing date info, we need to reject it.

- The [feed content is not valid XML or HTML](https://validator.w3.org/feed/check.cgi?url=http%3A%2F%2Fcompileofcrap.com%2FSPO600%2Ffeed.xml).  Chrome and Firefox both choke on it, and we end up only showing the first sentence of content.  I wonder if we should consider running post HTML through one of our HTML parsers (I'm actually amazed that our sanitizer and feed parser don't already choke on this), and rejecting posts that contain broken markup.  If we allow broken markup in posts, we risk having posts further down in the page break too.

So, the action items for this bug include:

1. Remove the "set missing date to current date" logic in post.js, and fail the post
1. See if we can find a way to validate markup in posts before we add it to our cache. 

cc @PopeSpaceous, who owns this feed.

Silvyre:
## Issue This PR Addresses

Fixes #757 

## Type of Change

## Description

- Posts missing a publication date will no longer be processed/displayed (with tests updated accordingly).
- If a post is missing a date of last update but has a publication date, its date of last update will be set to its publication date.
- (As a result, this PR rids Telescope of an infamous nuisance post...)

## Checklist

### DIFF
--- a/src/backend/data/post.js
+++ b/src/backend/data/post.js
@@ -7,6 +7,28 @@ const hash = require('./hash');
 const ArticleError = require('./article-error');
 
 /**
+ * Makes sure that a given date can be constructed as a Date object
+ * Returns a constructed Date object, if possible
+ * Otherwise throws an Error
+ * @param {Object} date an Object to construct as a Date object
+ * @param {Date} [fallbackDate] an optional second Date to construct in case the first fails to do so
+ */
+function ensureDate(date, fallbackDate) {
+  if (
+    date &&
+    (Object.prototype.toString.call(date) === '[object String]' ||
+      (Object.prototype.toString.call(date) === '[object Date]' && !Number.isNaN(date)))
+  ) {
+    return new Date(date);
+  }
+  if (Object.prototype.toString.call(fallbackDate) === '[object Date]') {
+    return new Date(fallbackDate);
+  }
+
+  throw new Error(`post has an invalid date: ${date}'`);
+}
+
+/**
  * Makes sure that the given feed is a Feed and not just an id.  If the latter
  * it gets the full feed.
  * @param {Feed|String} feed a Feed object or feed id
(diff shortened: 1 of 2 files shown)


src/backend/data/post.js
```diff
@@ -29,8 +22,16 @@ class Post {
...
+    try {
+      this.published = new Date(datePublished);
+    } catch (error) {
+      throw new Error(`post has invalid publication date : ${datePublished}'`);
+    }
+    try {
+      this.updated = new Date(dateUpdated);
+    } catch (error) {
+      throw new Error(`post has invalid date of last update: ${dateUpdated}'`);
+    }
```
manekenpix: Maybe you could have `try and catch` inside of `toDate` and throw there. Also, `toDate` checks if what's passed is already an instance of `Date`, I think it'd be useful to keep it.

manekenpix: Also, what criteria does `Date` follow to throw? I looked around to see how people check for invalid dates and I found some [interesting suggestions](https://stackoverflow.com/a/1353711/8628446).

Silvyre: Made some changes in https://github.com/Seneca-CDOT/telescope/pull/823/commits/24a4fd4fdf0f8c7ce990b8f0ff024912e0f3b96f, let me know what you think!

Silvyre: Apologies for the delay. Made the requested changes in https://github.com/Seneca-CDOT/telescope/pull/823/commits/225ef1ac7607100c12f410c1a4c00f14174927ec, please let me know what you think!

PR #831, opened 2020-03-11
 fixes: ensures docker cache dependencies when package.js is not modified #826 

### LINKED ISSUE #825: Improve local docker build times for code changes.
Current `Dockerfile` takes an average of 10 minutes to complete due to multiple `npm installs` failing to cache, defeating the purpose of the multi-container design for local development.
 
Reduce @Silvyre load time from ~10min to ~1min if possible, establish `Docker Doctor` title till a challenger or learner approaches.

raygervais:
## Issue This PR Addresses
Fixes #825 

## Type of Change

## Description

On my system with changes, the build time for a local docker instance after doing the first successful build (and thus caching dependencies) and modifying a line in the backend code as a test comes out to:
```
real	0m6.243s
user	0m0.375s
sys     0m0.064s
```

**Limitations  for  Docker caching**: any change to package.json and it's related dependencies will force a new container build. Source code changes otherwise should not trigger rebuild of cached containers.

## Checklist

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,20 +20,34 @@ ENTRYPOINT [ "/sbin/tini", "--"]
 # Set Working Directory Context
 WORKDIR "/telescope"
 
-# Copy shared context
+# Copy package.jsons for each service
 COPY package.json .
+COPY ./tools/autodeployment/package.json ./tools/autodeployment/package.json
+COPY ./src/frontend/package.json ./src/frontend/package.json
 
 # -------------------------------------
 # Context: Dependencies
 FROM build AS dependencies
 
-# Get Entire Codebase
+# Install Production Modules! 
+# Disable postinstall hook in this case since we are being explict with installs
+# `postinstall` typically goes on to install front-end and autodeployment modules
+# which though is logical for local development, breaks docker container caching trick.
+RUN npm install --only=production --no-package-lock --ignore-scripts
+
+# Install Frontend Modules!
+RUN cd ./src/frontend && npm install --no-package-lock
+
+# Install Deployment Microservice Modules!
+RUN cd /telescope/tools/autodeployment && npm install --no-package-lock --ignore-scripts
+
+# -------------------------------------
+# Context: Front-end Builder
+FROM dependencies as builder
+
 COPY . .
 
-# Install Production Modules, Build!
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -20,28 +20,41 @@ ENTRYPOINT [ "/sbin/tini", "--"]
...
 # Context: Dependencies
 FROM build AS dependencies
 
-# Get Entire Codebase
+# Install Production Modules! 
+# Disable postinstall hook in this case since we are being explict with installs
+RUN npm install --only=production --no-package-lock --ignore-scripts
+
+# Install Frontend Modules!
+RUN cd ./src/frontend && npm install --no-package-lock
```
humphd: Why won't this get triggered by the `postinstall` script after 34?

Dockerfile
```diff
@@ -20,28 +20,41 @@ ENTRYPOINT [ "/sbin/tini", "--"]
...
+
+# Install Deployment Microservice Modules!
+RUN cd /telescope/tools/autodeployment && npm install --no-package-lock
```
humphd: Same question here...

Dockerfile
raygervais: `--ignore-scripts` causes all hooks to be skipped. This enables me to be very explicit with the installation calls.

raygervais: Doing a clean build now to confirm this, want to go over logs and ensure I am correct on that logic.

Dockerfile
raygervais: Adding the same `--ignore-scripts` to ensure that no other hooks are picked up, one item to note is that this translates to using the default npm scripts and all other ones fail. So, as long as we are doing straight `install`, `build`, and `start` with the front-end not having such since we need the custom build command) it should not cause any trouble.

raygervais:
Putting here for historical archiving: 

`[0m[91mnpm WARN lifecycle @seneca/telescope@0.7.0~postinstall: cannot run in wd @seneca/telescope@0.7.0 cd src/frontend && npm install && cd ../../tools/autodeployment && npm install (wd=/telescope)`

And, to ensure we are not encountering postinstall, doing a search for "packages in" (the standard NPM install output) shows only 3, which are my explicitly called ones. The only postinstall that still runs is Gatsby's built-in.

PR #843, opened 2020-03-18
Include version info on header banner

### LINKED ISSUE #803: Include version info on header banner

Let's find a way to include the current git SHA for the commit we've built and shipped on staging/production.  What if we altered this slightly so it was:

```html
<a href="https://github.com/Seneca-CDOT/telescope/commit/9e50064a8a090fd93975e8f037fe67ef5688fa0c" title="git commit 9e50064a8a090fd93975e8f037fe67ef5688fa0c">v0.7.0</a>
```

This would require the following changes:

- Instead of pulling in the entire `package.json` file [here](https://github.com/Seneca-CDOT/telescope/blob/master/src/frontend/src/components/Banner/Banner.jsx#L4), let's do it in https://github.com/Seneca-CDOT/telescope/blob/master/src/frontend/gatsby-config.js at build-time, and add the version and current git sha so we can build our URL for that link.  We can use something like https://www.npmjs.com/package/git-repo-info to pull this info into the gastby-config.js file.
- Add the version and git sha properties to our sitemetadata hook https://github.com/Seneca-CDOT/telescope/blob/master/src/frontend/src/hooks/use-site-metadata.js and StaticQuery https://github.com/Seneca-CDOT/telescope/blob/master/src/frontend/src/pages/index.js
- Do some CSS to keep the colour/text-decoration like it is now on the banner, but when you hover we can add `text-decoration: underline`.

We currently do this, and on staging it doesn't indicate exactly which commit we are looking at:

Screen Shot 2020-03-05 at 12 27 00 PM

c3ho:
## Issue This PR Addresses
Fixes #803 

## Type of Change

## Description
Version + SHA should now show up on banner.
Installed a package and moved all the info pulling into gatsby.config site metadata part. 

image

ZEIT is failing I think because I installed git-repo-info into our main package.json instead of the front-end one (I tried installing it here and pulling data from here, it didn't work). To use the module from there, I'm pulling it from node_modules.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
     "feed": "4.1.0",
     "feedparser-promised": "2.0.1",
     "get-urls": "9.2.0",
+    "git-repo-info": "2.1.1",
     "graphql-iso-date": "3.6.1",
     "graphql-passport": "0.6.1",
     "helmet": "3.21.2",
--- a/package.json
+++ b/package.json
@@ -84,7 +84,6 @@
     "feed": "4.1.0",
     "feedparser-promised": "2.0.1",
     "get-urls": "9.2.0",
-    "git-repo-info": "2.1.1",
     "graphql-iso-date": "3.6.1",
     "graphql-passport": "0.6.1",
     "helmet": "3.21.2",
--- a/src/frontend/gatsby-config.js
+++ b/src/frontend/gatsby-config.js
@@ -23,7 +23,6 @@ require('dotenv').config({
  * in our now.json file.
  */
 const telescopeUrl = process.env.API_URL || `http://localhost:${process.env.PORT || 3000}`;
-// const telescopeSha = info.sha;
 const { sha } = info;
 const { version } = Version;
 
--- a/src/frontend/gatsby-config.js
+++ b/src/frontend/gatsby-config.js
@@ -1,10 +1,6 @@
 const path = require('path');
-
+const getRepoInfo = require('git-repo-info');
 const Version = require('../../package.json');
-
-const getRepoInfo = require('../../node_modules/git-repo-info');
-
-const info = getRepoInfo();
 
 /**
  * Try to load an .env from the root telescope project
@@ -23,7 +19,7 @@ require('dotenv').config({
  * in our now.json file.
  */
 const telescopeUrl = process.env.API_URL || `http://localhost:${process.env.PORT || 3000}`;
-const { sha } = info;
+const { sha } = getRepoInfo();
 const { version } = Version;
 
 module.exports = {
--- a/src/frontend/gatsby-config.js
+++ b/src/frontend/gatsby-config.js
@@ -1,5 +1,11 @@
 const path = require('path');
 
+const Version = require('../../package.json');
+
+const getRepoInfo = require('../../node_modules/git-repo-info');
+
+const info = getRepoInfo();
+
 /**
  * Try to load an .env from the root telescope project
  */
@@ -17,6 +23,9 @@ require('dotenv').config({
  * in our now.json file.
  */
 const telescopeUrl = process.env.API_URL || `http://localhost:${process.env.PORT || 3000}`;
+// const telescopeSha = info.sha;
+const { sha } = info;
+const { version } = Version;
 
 module.exports = {
   siteMetadata: {
@@ -24,6 +33,8 @@ module.exports = {
     description: `A tool for tracking blogs in orbit around Seneca's open source involvement`,
     author: `SDDS Students and professors`,
     telescopeUrl,
+    sha,
+    version,
   },
   plugins: [
     'gatsby-plugin-top-layout',
--- a/src/frontend/gatsby-config.js
+++ b/src/frontend/gatsby-config.js
@@ -1,5 +1,4 @@
 const path = require('path');
-const { version } = require('../../package.json');
 
 /**
  * Try to load an .env from the root telescope project
@@ -25,7 +24,6 @@ module.exports = {
     description: `A tool for tracking blogs in orbit around Seneca's open source involvement`,
     author: `SDDS Students and professors`,
     telescopeUrl,
-    version,
   },
   plugins: [
     'gatsby-plugin-top-layout',
--- a/src/frontend/gatsby-config.js
+++ b/src/frontend/gatsby-config.js
@@ -1,6 +1,5 @@
 const path = require('path');
-const getRepoInfo = require('git-repo-info');
-const Version = require('../../package.json');
+const { version } = require('../../package.json');
 
 /**
  * Try to load an .env from the root telescope project
@@ -19,8 +18,6 @@ require('dotenv').config({
  * in our now.json file.
  */
 const telescopeUrl = process.env.API_URL || `http://localhost:${process.env.PORT || 3000}`;
-const { sha } = getRepoInfo();
-const { version } = Version;
 
 module.exports = {
   siteMetadata: {
@@ -28,7 +25,6 @@ module.exports = {
     description: `A tool for tracking blogs in orbit around Seneca's open source involvement`,
     author: `SDDS Students and professors`,
     telescopeUrl,
-    sha,
     version,
   },
   plugins: [
--- a/src/frontend/src/components/Banner/Banner.jsx
+++ b/src/frontend/src/components/Banner/Banner.jsx
@@ -234,11 +234,21 @@ export default function Banner() {
           </Typography>
         </ThemeProvider>
 <<<<<<< HEAD
+<<<<<<< HEAD
 
         <div className={classes.version}>v {Version.version}</div>
 =======
         <div className={classes.version}>v {`${version} SHA:${sha}`}</div>
 >>>>>>> 755e450... Include version info on header banner
+=======
+        <a
+          href={`https://github.com/Seneca-CDOT/telescope/commit/${sha}`}
+          title={`git commit ${sha}`}
+          className={classes.version}
+        >
+          v {version}
+        </a>
+>>>>>>> fb053d4... removed sha turned it into a link
 
         <div className={classes.icon}>
           <ScrollDown>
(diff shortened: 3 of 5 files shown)


cindyledev:
For the underline on hover, you can add this to the `version` style in the `Banner.jsx`: 
```
    textDecorationLine: 'none',
    '&:hover': {
      textDecorationLine: 'underline',
    },
```

src/frontend/gatsby-config.js
```diff
@@ -1,5 +1,11 @@
 const path = require('path');
 
+const Version = require('../../package.json');
+
+const getRepoInfo = require('../../node_modules/git-repo-info');
```
humphd: You should be able to `require('git-repo-info')` and the way that node resolves is that it will go up to parent directories to find it.  That said, why not just install this in the frontend package.json vs. the root?

src/frontend/gatsby-config.js
```diff
@@ -1,5 +1,11 @@
...
+
+const info = getRepoInfo();
```
humphd:
```js
const { sha } = getRepoInfo();
```

c3ho:
Alright, I goofed. I found the issue, was trying to do `const getRepoInfo = require('./package.json')` when I was supposed to do `const getRepoInfo = require('git-repo-info')`

src/frontend/gatsby-config.js
```diff
@@ -1,4 +1,6 @@
 const path = require('path');
+const getRepoInfo = require('git-repo-info');
```
humphd:
```js
const { sha } = require('git-repo-info')();
```

src/frontend/gatsby-config.js
```diff
@@ -1,4 +1,6 @@
...
+const Version = require('../../package.json');
```
humphd:
```js
const { version } = require('../../package.json');`
```

src/frontend/gatsby-config.js
c3ho: ah so much to learn

src/frontend/gatsby-config.js
```diff
@@ -1,4 +1,6 @@
 const path = require('path');
+const { sha } = require('git-repo-info')();
```
humphd: If we can't use this, let's revert to what you did before (even though zeit/gatsby/webpack are wrong!)

c3ho: Zeit just hasn't liked this pr very much since the start, the original way nor the way you suggested. Everything that has worked locally fails to build on zeit. Do we just abandon this pr?

c3ho:
This is going to be stalled until... I have no idea when

cindyledev:
This is so sad... we tried idk how many different methods and Zeit has rejected them all...
Annotation 2020-03-26 083233

Annotation 2020-03-26 083233_

cindyledev:
Wait what does this mean? https://zeit.co/docs/v2/git-integrations/zeit-now-for-github
Included Environment Variables `NOW_GITHUB_COMMIT_SHA`. We had access to this the whole time?

humphd:
@cindyledev we'd be able to use `NOW_GITHUB_COMMIT_SHA` on Zeit, but not with local dev or on staging, where it wouldn't exist.  So we'd have a new set of failures on everything but Zeit.

Good catch though.  I didn't know they had this.

humphd:
See #882.  It would let you use a `fetch()` to get this infro from `/health`

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@seneca/telescope",
   "private": true,
-  "version": "0.8.2",
+  "version": "0.0.1",
```
humphd: Let's not do this.  Can you roll back this commit?

c3ho: Sorry I checked out the wrong package.json should be fixed now

src/frontend/gatsby-config.js
```diff
@@ -24,6 +25,7 @@ module.exports = {
     description: `A tool for tracking blogs in orbit around Seneca's open source involvement`,
     author: `SDDS Students and professors`,
     telescopeUrl,
+    version,
```
humphd: We probably don't need the changes to this file if we're going to get it via /health.

src/frontend/src/components/Banner/Banner.jsx
```diff
@@ -146,7 +149,7 @@ function RetrieveBannerDynamicAssets() {
   useEffect(() => {
     async function getBackgroundImgSrc() {
       // Uses https://unsplash.com/collections/894/earth-%26-planets collection
-      /* Other Options: 
+      /* Other Options:
```
humphd: Careful with unrelated changes like this.  It's fine here, but don't change git blame if you don't have to: it makes it harder to understand the history of this file.

src/frontend/src/components/Banner/Banner.jsx
```diff
@@ -218,6 +221,24 @@ ScrollDown.propTypes = {
...
 export default function Banner() {
   const classes = useStyles();
+  const { version, telescopeUrl } = useSiteMetadata();
+  const [sha, setSha] = useState('');
+
+  async function getSha() {
+    try {
+      const res = await fetch(`${telescopeUrl}/health`);
+      const data = await res.json();
+      return data.info.sha;
```
humphd: Pull version out of here too.

src/frontend/src/components/Banner/Banner.jsx
```diff
@@ -218,6 +221,24 @@ ScrollDown.propTypes = {
 
 export default function Banner() {
   const classes = useStyles();
+  const { version, telescopeUrl } = useSiteMetadata();
+  const [sha, setSha] = useState('');
+
+  async function getSha() {
+    try {
+      const res = await fetch(`${telescopeUrl}/health`);
```
humphd: Error handling please.

src/frontend/src/components/Banner/Banner.jsx
```diff
@@ -218,6 +221,24 @@ ScrollDown.propTypes = {
...
+      return data.info.sha;
+    } catch (error) {
+      console.error(`Error retrieving site's health info`, error);
+      return error;
+    }
+  }
+
+  useEffect(async () => {
+    setSha(await getSha());
+  }, []);
```
humphd:
```js
}, [telescopeUrl]);
```

src/frontend/src/components/Banner/Banner.jsx
```diff
@@ -230,8 +251,13 @@ export default function Banner() {
           </Typography>
         </ThemeProvider>
 
-        <div className={classes.version}>v {Version.version}</div>
-
+        <a
+          href={`https://github.com/Seneca-CDOT/telescope/commit/${sha}`}
```
humphd: I think `/health` has a full GitHubUrl already built that you can use.

src/frontend/src/components/Banner/Banner.jsx
```diff
@@ -221,23 +221,35 @@ ScrollDown.propTypes = {
...
-      const data = await res.json();
-      return data.info.sha;
+
+      // Fetch failure
+      if (!res.ok) {
+        throw new Error(res.statusText);
+      }
+
+      // Checking whether content type is correct
+      if (res.headers.get('content-type').includes('application/json')) {
```
humphd: Don't bother with this.  Ray did it because he was re-using the function, but we know it will be JSON.

c3ho: Removed.

src/frontend/src/components/Banner/Banner.jsx
```diff
@@ -218,6 +221,34 @@ ScrollDown.propTypes = {
 
 export default function Banner() {
   const classes = useStyles();
+  const { telescopeUrl } = useSiteMetadata();
+  const [sha, setSha] = useState('');
+  const [version, setVersion] = useState('');
+  const [githubUrl, setGitHubUrl] = useState('');
```
humphd: Don't bother with this now, but you could use a single state Object vs. splitting this out into three.  It's fine for state to be a nested object in react.

src/frontend/src/components/Banner/Banner.jsx
```diff
@@ -218,6 +221,34 @@ ScrollDown.propTypes = {
...
+        throw new Error(res.statusText);
+      }
+
+      const data = await res.json();
+      setSha(data.info.sha);
+      setGitHubUrl(data.info.gitHubUrl);
+      setVersion(data.info.version);
+    } catch (error) {
+      console.error(`Error retrieving site's health info`, error);
+      throw new Error('Error retrieving site health info', error);
```
humphd: Who are you throwing this to?  Drop this line.

src/frontend/src/components/Banner/Banner.jsx
```diff
@@ -218,6 +221,34 @@ ScrollDown.propTypes = {
...
+    }
+  }
+
+  useEffect(async () => {
+    await getGitData();
```
humphd: You're not allowed to do this in a hook (they have to return `undefined` vs. a `Promise`).  Change it like this:

```js
useEffect(() => {
  async function getGitData() {...}

  getGitData();
});
```

src/frontend/src/components/Banner/Banner.jsx
```diff
@@ -91,6 +89,11 @@ const useStyles = makeStyles(theme => ({
       left: '80%',
       bottom: theme.spacing(18),
     },
+    color: 'white',
+    textDecorationLine: 'none',
+    '&:hover': {
+      textDecorationLine: 'underline',
+    },
```
cindyledev: This is cute, you put this in `icon` instead of `version` XD

cindyledev: This is cute, you put this in `icon` instead of `version` XD

c3ho: The rebase was going weird on me

PR #870, opened 2020-03-24
Initial work to test Seneca SSO setup on staging

humphd:
I need to test this on staging, because it can't be tested locally.  I also need to make sure that what I've done works locally with the old login server.

I could use some help:

1. test this locally, make sure I didn't break anything with the local login flow
1. see if I made any mistakes with this for staging, then let's do a manual test of this branch without merging, and fix what doesn't work.


humphd:
Testing on staging, we're now able to get POST responses back from the IdP, but the passport middleware is crashing:

```
telescope_staging | Error: error:09091064:PEM routines:PEM_read_bio_ex:bad base64 decode
telescope_staging |     at Verify.verify (internal/crypto/sig.js:183:24)
telescope_staging |     at RSASHA256.verifySignature (/telescope/node_modules/xml-crypto/lib/signed-xml.js:137:24)
telescope_staging |     at SignedXml.validateSignatureValue (/telescope/node_modules/xml-crypto/lib/signed-xml.js:405:20)
telescope_staging |     at SignedXml.checkSignature (/telescope/node_modules/xml-crypto/lib/signed-xml.js:353:13)
telescope_staging |     at SAML.validateSignatureForCert (/telescope/node_modules/passport-saml/lib/passport-saml/saml.js:567:14)
telescope_staging |     at /telescope/node_modules/passport-saml/lib/passport-saml/saml.js:537:17
telescope_staging |     at Array.some (<anonymous>)
telescope_staging |     at SAML.validateSignature (/telescope/node_modules/passport-saml/lib/passport-saml/saml.js:536:16)
telescope_staging |     at /telescope/node_modules/passport-saml/lib/passport-saml/saml.js:610:19
telescope_staging |     at _fulfilled (/telescope/node_modules/q/q.js:854:54)
```

humphd:
It looks like we're crashing in https://github.com/bergie/passport-saml/blob/master/lib/passport-saml/saml.js#L541-L568

humphd:
Looking at the [metadata for the IdP](https://login.microsoftonline.com/3ddc4284-8377-4878-94bf-a35840ca61fe/federationmetadata/2007-06/federationmetadata.xml?appid=55d4fc73-6e1b-429c-b796-28b25a669e20), we might be failing because the saml passport provider defaults to sha1: 

```xml
<DigestMethod Algorithm="http://www.w3.org/2001/04/xmlenc#sha256"/>
```

See https://github.com/bergie/passport-saml#security-and-signatures

PR #871, opened 2020-03-25
Closes #634: nginx configuration for staging and production

### LINKED ISSUE #634: nginx configuration for staging and production

A few things we should do in our nginx config for staging and production:

- setup nginx to [serve our static content](https://docs.nginx.com/nginx/admin-guide/web-server/serving-static-content/) (e.g., GatsbyJS frontend) instead of the node app
- setup proper [caching for our endpoints](https://serversforhackers.com/c/nginx-caching) to reduce load on the node app

The nginx server is better a certain tasks than our node app, and we should leverage it where possible.

manekenpix:
## Issue This PR Addresses

Closes #634 

## Type of Change

## Description

This adds using nginx to serve Telescope's static content and caching Telescope's endpoints to reduce load on it.

## Checklist

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -30,8 +30,8 @@ http {
       root /var/www/certbot;
     }
 
-    # Non-cached routes /admin /user /health
-    location ~ ^/(admin|user|health) {
+    # Non-cached routes /admin /user /health /auth
+    location ~ ^/(admin|user|health|auth) {
       proxy_cache_bypass 1;
       proxy_pass http://telescope_staging:3000;
 
--- a/nginx.conf
+++ b/nginx.conf
@@ -1,6 +1,9 @@
 events {}
 
 http {
+
+  proxy_cache_path /tmp/nginx levels=1:2 keys_zone=telescope_cache:10m inactive=60m;
+  proxy_cache_key "$scheme$request_method$host$request_uri";
   server {
     listen 80 default_server;
     server_name _;
@@ -18,7 +21,12 @@ http {
     ssl_certificate_key /etc/letsencrypt/live/spain.cdot.systems/privkey.pem;
 
     root /var/www/data;
+
     location / {
+      proxy_cache telescope_cache;
+      add_header X-Proxy-Cache $upstream_cache_status;
+
+      include proxy_params;
       try_files $uri $uri/ @proxy;
     }
 
--- a/nginx.conf
+++ b/nginx.conf
@@ -2,7 +2,7 @@ events {}
 
 http {
 
-  proxy_cache_path /tmp/nginx levels=1:2 keys_zone=telescope_cache:10m inactive=60m;
+  proxy_cache_path /tmp/nginx levels=1:2 keys_zone=telescope_cache:10m inactive=60m max_size=100M;
   proxy_cache_key "$scheme$request_method$host$request_uri";
   
   server {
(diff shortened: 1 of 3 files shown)


nginx.conf
```diff
@@ -1,6 +1,14 @@
 events {}
 
 http {
+  # levels        - Defines hierarchy levels
+  # keys_zone     - Name for these settings
+  # inactive      - Cached data that are not accessed during the time specified by the inactive parameter get removed from the cache regardless of their freshness
+  # max_size      - When this size is exceeded, it removes the least recently used data
+  # use_temp_path - Use a temp directory before moving all cached files into the cache directory
```
humphd: Thanks for these comments, helpful.

manekenpix:
I'm not sure, I think we'll probably have to do it here since `nginx.conf` has this:
```
# Disables processing of certain response header fields from the proxied server
proxy_ignore_headers "Cache-Control" "Expires";
```

manekenpix:
@c3ho what do you think?

humphd:
From https://docs.nginx.com/nginx/admin-guide/content-cache/content-caching/#select

>To define conditions under which NGINX Plus does not send cached responses to clients, include the proxy_cache_bypass directive. Each parameter defines a condition and consists of a number of variables. If at least one parameter is not empty and does not equal “0” (zero), NGINX Plus does not look up the response in the cache, but instead forwards the request to the backend server immediately.

```
proxy_cache_bypass $cookie_nocache $arg_nocache$arg_comment;
```

>To define conditions under which NGINX Plus does not cache a response at all, include the proxy_no_cache directive, defining parameters in the same way as for the proxy_cache_bypass directive.

```
proxy_no_cache $http_pragma $http_authorization;
```

manekenpix:
@humphd I made changes to avoid caching `/admin`, `/user` and `/health`. Can you review these changes?

nginx.conf
```diff
@@ -17,12 +25,45 @@ http {
...
     ssl_certificate_key /etc/letsencrypt/live/dev.telescope.cdot.systems/privkey.pem;
 
-    location / {
+   # Let's Encrypt route
+    location /.well-known/acme-challenge/ {
+      root /var/www/certbot;
+    }
+
+    # Non-cached routes /admin /user /health
+    location ~ ^/(admin|user|health) {
```
humphd: Let's add `/auth` to this list too

manekenpix: Added `/auth` and comments in `index.js`.

src/backend/web/routes/index.js
```diff
@@ -17,11 +17,19 @@ const query = require('./query');
 
 const router = express.Router();
 
+/**
+ * In staging and production, our revers proxy takes care of serving the content in the public folder.
```
humphd: "reverse"

src/backend/web/routes/index.js
```diff
@@ -17,11 +17,19 @@ const query = require('./query');
...
+ * We're keeping this route for development.
+ */
 router.use(express.static(path.join(__dirname, '../../../frontend/public')));
```
humphd: If this is only for development, should we wrap this in a `if(process.NODE_ENV !== 'production')` or something?

manekenpix: I honestly don't know. In production and staging, this route will never be used, so I'm open to suggestions.

humphd: In `env.staging` we have `NODE_ENV=production`, and `env.example` has `NODE_ENV=development` so let's do this:

```js
if(process.NODE_ENV === 'devleopment') {
  router.use(express.static(path.join(__dirname, '../../../frontend/public')));
}
```

raygervais:
Tested locally, this is awesome!

Issue #874, opened 2020-03-25
Typo in environment-setup.md


cindyledev:
**What happened**:
Typo in our docs `telescope/docs/environment-setup.md` 
Under SAML Setup

![Capture](https://user-images.githubusercontent.com/36822198/77496236-e6e50900-6e20-11ea-894c-d2dc19c39e54.PNG)

`./tools/generate_ssl_certs.sh` should be `./tools/generate-ssl-certs.sh`

**Environment**:

- OS: Windows
- Browser: NA

cindyledev:
#882 closes this along with the other 4 issues.

PR #882, opened 2020-03-26
Clean up SSO code in web app, logout works

### LINKED ISSUE #646: SAML metadata for our SP

We need to provide Seneca with the SAML metadata for our app.  We can use [`generateServiceProviderMetadata()`](https://github.com/bergie/passport-saml#generateserviceprovidermetadata-decryptioncert-signingcert-) in `passport-saml` to do this, and even provide it as an endpoint in the app

I spoke with ITS about connecting our staging server to their staging SSO box, and in order to do it, they need our "Service Provider Metadata" (SP).  To do this we'll have to have our SSL and certs all setup first.

Depends on #640, #634.

### LINKED ISSUE #674: Wrong path to telescope's cert files
**What happened**:

After using our script for SAML setup, a `certs` directory was created inside of the `tools` directory.
At launch, telescope shows this error:
```
[ 2020-02-07 11:20:52.301 ] ERROR (6466 on spain.cdot.systems): Unable to load certs/key.pem
    error: {
      "errno": -2,
      "syscall": "open",
      "code": "ENOENT",
      "path": "/home/josue/Projects/telescope/certs/key.pem"
    }
```
Telescope is trying to load `key.pem` from `telescope/certs`, but the folder `certs` was created in the `tools` directory. The path should be `telescope/tools/certs`.

Execute `generate-ssl-certs.sh` in the `tools` directory and then launch telescope.

**Environment**:

- OS: Fedora 30
- Browser:

### LINKED ISSUE #694: Logout Needs to be fixed on the express server
I would like to add the ability to logout of our site, you can login, and currently you are there forever.

Let's remedy that so people can leave telescope.

### LINKED ISSUE #695: Logout needs to be fixed on the front end.
Another issue. Fix #694  will fix our back end so we can logout correctly.

This also needs to be done on the front end, so that actual user can do it with the ease and click of a button!

This way the front end will let you to logout easy from a click of a button.

humphd:
## Issue This PR Addresses

1. Fixes #674 
2. Fixes #695 
3. Fixes #694 
4. Fixes #646 

## Type of Change

## Description

Now that the SSO code is working with Seneca, I wanted to clean up a bunch of loose ends.  In the process I was also able to get Logout to work (hurray!).

We no longer need to generate dev certs, so I've removed that script from `tools/` and all callers of it.

To test this PR:

- run the login docker container locally (and redis, elastic if you need them via docker)
- build the frontend: `npm run build`
- run our node server: `npm start`
- browse to http://localhost:3000
- Click Login
- login as user1 with password user1pass
- You should go back to the homepage and be logged in
- Click Logout
- You should go back to the homepage and be logged out

I also moved the `/health` route to its own file to clean up our `app.js` and added GitHub commit info and git sha:

Screen Shot 2020-03-26 at 2 27 45 PM

Hopefully this helps with what @c3ho has been fighting with in our build.

NOTE: I have not tested this against Seneca's SSO, because we have to merge to know if it works there.


humphd:
Sorry, I missed a test failure locally, fixed now.

docs/login.md
```diff
@@ -45,41 +45,39 @@ from `src/backend/web/authentication.js`.
...
-
-   # SAML2_CLIENT_ID = CLIENT ID obtained from SAML Strategy default: saml-poc
-   SAML2_CLIENT_ID=saml-poc
-
-   # SAML2_CLIENT_SECRET = CLIENT SECRET obtained from SAML Strategy, default : secret;
-   SAML2_CLIENT_SECRET=secret
-   ```
+# SAML2_CLIENT_SECRET = CLIENT SECRET obtained from SAML Strategy, default : secret;
+SAML2_CLIENT_SECRET=secret
+```
```
Grommers00: What this?

src/backend/web/app.js
```diff
@@ -27,12 +25,16 @@ app.use(helmet());
...
-// TODO: decide if we should do resave=false, https://www.npmjs.com/package/express-session#resave
+// Setup session and passport for authentication
+app.use(bodyParser.json());
+app.use(bodyParser.urlencoded({ extended: false }));
+authentication.init(passport);
 app.use(
   session({
     store: new RedisStore({ client: redis }),
-    secret,
+    secret: process.env.SECRET || `telescope-has-many-secrets-${Date.now()}!`,
```
Grommers00: Shouldn't the variable on the right (telescope-has-many-secrets) have a naming convention that is a bit easier to understand? or is it fine because it's a secret.

src/backend/web/authentication.js
```diff
@@ -105,19 +87,21 @@ function init(passport) {
...
       // TODO: we can probably pick off the user data we actually need here...
       if (!profile) {
-        const error = new Error('SAML Strategy verify callback missing profile');
+        const error = new Error('SAML Strategy verify callback missing user profile');
         logger.error({ error });
         done(error);
       } else {
-        done(null, profile);
+        // We only use the displayname, emailaddress, and nameID (hashed, for use in our db)
+        done(null, {
```
Grommers00: How come this is moved to the strategy opposed to the serialization?

src/backend/web/routes/auth.js
```diff
@@ -29,40 +28,32 @@ router.post('/login/callback', passport.authenticate('saml'), (req, res) => {
...
+
+/**
+ * /auth/logout allows users to clear login tokens from their session
  */
-function logout(req, res) {
+router.get('/logout', (req, res) => {
   try {
-    // TODO: confirm I can use strategy above... was: const strategy = passport._strategy('saml');
-    strategy.logout(req, (error, requestUrl) => {
+    // eslint-disable-next-line no-underscore-dangle
```
Grommers00: interesting way of getting around our ESLint :D. Did not know this was how to do it.

docs/login.md
humphd: Ah, good catch.  I missed this.  I'll fix.

I'm renaming this to `SECRET` because it's not really part of the SSO/SAML stuff.  We just use it for our cookie session.

cindyledev:
Fixes #874 as well

src/backend/web/app.js
humphd: It's just a string we use to make the hashes for the cookies more random.  It doesn't matter what it is.  People often use stupid strings in here, just to make it so it's not guessable.  You want a long string, basically.

src/backend/web/authentication.js
humphd: I did some more reading, and I gave you incorrect advice.  Where I moved it to, when the "profile" data comes back from the SAML response, is where you're supposed to create your internal user object.  The serialize/deserialize functions are where you'd deal with getting data in/out of a database.  For example, you might store the nameID on the session, and then go get it from your database in here.  We don't need to do that, so we can pass the object along untouched.

src/backend/web/routes/auth.js
humphd: Pretend you didn't see this!  I have to do it because Passport doesn't expose the strategy object properly, only as a `._*` param, and our eslint (rightly) says you shouldn't do this, since those are generally private properties not meant to be used externally (only for tests).

humphd:
Thanks @cindyledev, I forgot about that issue.

PR #887, opened 2020-03-27
adds: .git to final build for health check

### LINKED ISSUE #885: Missing git sha in /health route
Screen Shot 2020-03-27 at 9 54 20 AM

This works locally, but I suspect that when we do the build, we lost all git info.  Is there a way around this with our docker pruning?  cc @raygervais.

raygervais:
## Issue This PR Addresses
Fixes #885 

## Type of Change

## Description

Copies .git to the final build, does some compression logic in attempt to reduce size.

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,6 +41,7 @@ RUN cd ./src/frontend && npm install --no-package-lock
 # Install Deployment Microservice Modules!
 RUN cd /telescope/tools/autodeployment && npm install --no-package-lock --ignore-scripts
 
+
 # -------------------------------------
 # Context: Front-end Builder
 FROM dependencies as builder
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -41,6 +41,18 @@ RUN cd ./src/frontend && npm install --no-package-lock
...
+# Context: Git Repo Handling
+FROM alpine/git as git
+
+WORKDIR "/telescope"
+
+COPY .git .git
+
+RUN git reflog expire --expire=now --all
+RUN git repack -a -d --depth=1 --window=10
+RUN git gc --prune=now
```
humphd: I wonder if instead of messing with the `.git` here, we should just do the `--depth=1` here https://github.com/Seneca-CDOT/telescope/blob/master/tools/autodeployment/deploy_stage.sh#L16, so it does the right thing for staging/prod.  What happens locally I don't care about optimizing.

raygervais: Great idea. Updated.

raygervais:
Can I get a review @manekenpix :D

PR #895, opened 2020-03-29
Fix reusing volumes in staging deployment

manekenpix:
## Type of Change

## Description

After landing #871, Telescope kept serving a version of the front end that was 2 commits behind #871. The problem turned out to be the command used in autodeployment for deleting all the docker elements related to Telescope (`docker system prune -af`). This command doesn't get rid of volumes, which is what we're using to share data between containers. As a result, Telescope was serving an old frontend stored in a persistent volume on `dev.telescope.cdot.systems`.

To fix this, adding `--volumes` (as suggested [here](https://linuxize.com/post/how-to-remove-docker-images-containers-volumes-and-networks/)) to the command used in autodeployment will also remove all volumes and new ones will be created for every new deployment.

@raygervais helped me debug this and found out that the problem was persistent volumes.

## Checklist

### DIFF
--- a/docs/environment-setup.md
+++ b/docs/environment-setup.md
@@ -146,6 +146,9 @@ docker-compose up --build redis elasticsearch
 
 _Note: You may need to add your user to the docker group in Linux to use `docker-compose` without `sudo`. To do this, try `groups $USER` in a terminal and check if docker is in the list of groups. If not, add it with `usermod -aG docker $USER` and reboot._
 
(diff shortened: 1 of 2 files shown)


raygervais:
This is a classic `something was cached which shouldn't have been` scenario with Docker, and I'd advise that we should add to the developer documentation to utilize the same command `docker prune ...` when testing to avoid this occurring locally as well. Thoughts on adding that small explanation to said docs @manekenpix ?

Aside from that, this looks good to me!

manekenpix:
I added a small section to [environment-setup.md](https://github.com/Seneca-CDOT/telescope/blob/master/docs/environment-setup.md) to mention how to delete containers. images, volumes... to avoid the same issue that this PR fixes. (Following @raygervais' suggestion)

Ready for review  @humphd @c3ho @cindyledev

Issue #894, opened 2020-03-28
Choose Unsplash images to use with banner


humphd:
**What would you like to be added**:

We've just merged the work @raygervais did to add Unsplash photos to our banner.  As a start he's using the photos in this collection: https://unsplash.com/collections/894/earth-%26-planets.

Let's see if we can create a Collection that can be shared across accounts, or I can make one, and people can tell me which photos to add.

**Why would you like this to be added**:

There's lots of great photos in there, but let's do a bit more work to figure out what we actually want to showcase here.

Let's try and think about the Telescope brand: what are these images trying to say?  It's more than just "looking through a telescope" I think.  It has to do with learning, a diversity of voices, education, etc.

If we choose wisely, we can say a lot without having to do a ton of work here.

humphd:
One thing I'm noticing, as I watch all the new photos go by on staging, is that some of them work better with out colour scheme than others.  We could eliminate those that don't work, and prefer those that do in our collection.

humphd:
From https://github.com/Seneca-CDOT/telescope/pull/897/files#r400181330, we should consider moving the URL to our Unsplash collection to the gatsby-config, so it can be changed at build time.

raygervais:
I'll start the thread I suppose: 
https://unsplash.com/photos/ZMcLVBi9xx4

_From the Earth & Planets Collection_
![From the Earth Collection](https://images.unsplash.com/photo-1441716844725-09cedc13a4e7)

agarcia-caicedo:
Here are some that I think would fit Telescope's vibe.
https://unsplash.com/photos/npxXWgQ33ZQ
![image](https://user-images.githubusercontent.com/49992710/78098393-f4fcd180-73ac-11ea-9a6c-683beac98247.png)

https://unsplash.com/photos/fIq0tET6llw
![image](https://user-images.githubusercontent.com/49992710/78098419-0f36af80-73ad-11ea-88b1-589fa3cd279a.png)

https://unsplash.com/photos/2FPjlAyMQTA
![image](https://user-images.githubusercontent.com/49992710/78098477-39886d00-73ad-11ea-8ac6-5e59758edd80.png)

https://unsplash.com/photos/oMpAz-DN-9I
![image](https://user-images.githubusercontent.com/49992710/78098564-6e94bf80-73ad-11ea-9c78-536cd06145f3.png)

https://unsplash.com/photos/IgUR1iX0mqM
[...]

PR #897, opened 2020-03-29
adds: fade-in transition for stats once loaded

### LINKED ISSUE #892: Fade stats line in only when we have the result from /stats route

We're fading in the image from Unsplash now, and it's cool.  Let's do the same for the line that says how much we've written this year.

It's really jarring how the stats line shows up, but without any numbers, then they come in and it jumps, causing reflow.  I think it would be nicer if you got the Telescope banner, then the image was there, then the stats appeared.

I want to avoid it looking like it's broken every time you load the page.

raygervais:
## Issue This PR Addresses
Fixes #892 

## Type of Change

## Description

This adds the same opacity transition trick used by the Banner background  to the Banner stats component. Makes the loading and API call less jarring to the user since text would simply appear and update the previously blank numeric values. 

## Checklist

### DIFF
--- a/src/frontend/src/components/Banner/Banner.jsx
+++ b/src/frontend/src/components/Banner/Banner.jsx
@@ -46,7 +46,7 @@ const useStyles = makeStyles(theme => ({
     left: theme.spacing(8),
     lineHeight: 'inherit',
     letterSpacing: 'inherit',
-    transition: 'all linear 350ms',
+    transition: 'all linear 1s',
     [theme.breakpoints.between('xs', 'sm')]: {
       textAlign: 'left',
       fontSize: '2rem',
@@ -112,9 +112,13 @@ function ScrollDown(props) {
   );
 }
 
-function RetrieveBackgroundImage() {
+function RetrieveBannerDynamicAssets() {
   const [backgroundImgSrc, setBackgroundImgSrc] = useState('');
   const [transitionBackground, setTransitionBackground] = useState(true);
+  const [stats, setStats] = useState({ stats: { posts: 0, authors: 0, words: 0 } });
+
+  const classes = useStyles();
+  const { telescopeUrl } = useSiteMetadata();
 
   useEffect(() => {
     async function getBackgroundImgSrc() {
@@ -138,7 +142,6 @@ function RetrieveBackgroundImage() {
 
         // Ease in Background
         setBackgroundImgSrc(src);
-        setTransitionBackground(false);
       } catch (error) {
         console.error('Error getting user info', error);
         // Fallback to default image
@@ -146,29 +149,6 @@ function RetrieveBackgroundImage() {
       }
     }
 
-    getBackgroundImgSrc();
-  }, []);
-
-  return (
-    <div
-      className="bannerImg"
-      style={{
-        backgroundImage:
-          backgroundImgSrc === '../../images/hero-banner.png'
-            ? backgroundImgSrc
-            : `url(${backgroundImgSrc})`,
-        opacity: transitionBackground ? 0 : 0.4,
-      }}
-    ></div>
-  );
-}
-
-function RetrieveStats() {
-  const { telescopeUrl } = useSiteMetadata();
-  const [stats, setStats] = useState({ stats: { posts: 0, authors: 0, words: 0 } });
-  const [transitionBackground, setTransitionBackground] = useState(true);
-
-  useEffect(() => {
     async function getStats() {
       try {
         const response = await fetch(`${telescopeUrl}/stats/year`);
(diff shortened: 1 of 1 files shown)


humphd:
I wonder if this needs to use [transitionend](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/transitionend_event) on the image fade in as the moment when we trigger the stats reveal.  Also, I think the image fade in time and the sudden appearance of the stats line aren't the same.

To clarify what I mean: wait for the image to be fully *there* before we start the fade in of the stats line, and have it fade in too.

raygervais:
@humphd, I liked your idea but wanted to approach from a different perspective, my logic being this:

1. background and stats sub-components have about 80% the same code, so why repeat? I'm going about the process of refactoring into a more reusable for those two components, and saw the opportunity to start the creation of a sub-component which would replace the two. Thoughts?

raygervais:
If we don't want to consolidate the common logic, I can revert and follow the order concepts that I have built into the refactored component.

src/frontend/src/components/Banner/Banner.jsx
```diff
@@ -112,87 +112,99 @@ function ScrollDown(props) {
   );
 }
 
-function RetrieveBackgroundImage() {
+async function getDynamicAsset(url, callback, errorCB) {
```
humphd: It's not a `callback` if your first arg isn't an `error` object, per node convention.  So I'd follow the pattern used in other Web APIs and do:

```js
async function getDynamicAsset(url, success, failure) {
   ...
}
```

Each of `success` and `failure` are functions that don't expect to receive an `error` object first.

src/frontend/src/components/Banner/Banner.jsx
```diff
@@ -112,87 +112,99 @@ function ScrollDown(props) {
...
-  useEffect(() => {
     async function getStats() {
-      try {
-        const response = await fetch(`${telescopeUrl}/stats/year`);
-        if (response.status !== 200) {
-          throw new Error(response.statusText);
-        }
-
-        const stat = await response.json();
+      await getDynamicAsset(`${telescopeUrl}/stats/year`, response => {
```
humphd: Tried it on Zeit, and I still dislike the stats line coming in before the image.  What if you initiate this in the success callback of the other call for the background?

src/frontend/src/components/Banner/Banner.jsx
raygervais: Haven't done true node in a while, I appreciate the convention tip. Corrected.

src/frontend/src/components/Banner/Banner.jsx
raygervais: That did it :D

src/frontend/src/components/Banner/Banner.jsx
```diff
@@ -112,87 +112,98 @@ function ScrollDown(props) {
...
+async function getDynamicAsset(url, success, failure) {
+  try {
+    const response = await fetch(url);
+
+    if (response.status !== 200) {
+      throw new Error(response.statusText);
+    }
+
+    if (response.headers.get('content-type').includes('application/json')) {
+      success(await response.json());
```
humphd: You need a `return` here or it will double-call `success()` in the JSON case.

src/frontend/src/components/Banner/Banner.jsx
```diff
@@ -112,87 +112,98 @@ function ScrollDown(props) {
...
-        const dimensions = `${window.innerWidth}x${window.innerHeight}`;
-        const response = await fetch(`https://source.unsplash.com/collection/894/${dimensions}/`);
-
-        if (response.status !== 200) {
-          throw new Error(response.statusText);
+      // Ensure we are using an image which fits correctly to user's viewspace
+      const dimensions = `${window.innerWidth}x${window.innerHeight}`;
+
+      await getDynamicAsset(
+        `https://source.unsplash.com/collection/894/${dimensions}/`,
```
humphd: I wonder if we should move this out to the `gatsby-config.js` like we did with `telescopeUrl`?  Don't do it here, we can tweak that in the other PR I have for #894 if necessary.

src/frontend/src/components/Banner/Banner.jsx
```diff
@@ -112,87 +112,98 @@ function ScrollDown(props) {
...
-      } catch (error) {
-        console.error('Error getting user info', error);
-      }
+        setTransitionBackground(false);
+      });
     }
 
-    getStats();
+    getBackgroundImgSrc();
+    // Ease in Background
```
humphd: Why this comment below the code vs. above?

src/frontend/src/components/Banner/Banner.jsx
```diff
@@ -209,17 +220,14 @@ export default function Banner() {
     <React.Fragment>
       <CssBaseline />
       <div className="heroBanner">
-        <RetrieveBackgroundImage />
+        <RetrieveBannerDynamicAssets />
 
         <ThemeProvider>
           <Typography variant="h1" className={classes.h1}>
             {'Telescope'}
```
humphd: We should file a bug to have this name come from the `gatsby-config.json` via our sitemetadata hook so it is configurable on another install.

humphd:
@agarcia-caicedo do you want to look at this, see if you agree with the transitions?

PR #807, opened 2020-03-05
Add site property to Feeds and Redis

c3ho:
## Issue This PR Addresses
Fixes #724 

## Type of Change

- [ ] **Bugfix**: Change which fixes an issue
- [x] **New Feature**: Change which adds functionality
- [ ] **Documentation Update**: Change which improves documentation
- [ ] **UI**: Change which improves UI

## Description
Added feeds property, also changed some tests.

Test using `redis-cli hgetall t:feed:77a19f8730`
The first time you do it after `npm start`, `link` should be empty
image

Let it run for a bit then use the redis command again and it should be populated
image

To re-test the process, `redis-cli flushall` must be used to clear Redis.

## Checklist

- [x] **Quality**: This PR builds and passes our npm test and works locally
- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [x] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not (if applicable)
- [ ] **Documentation**: This PR includes updated/added documentation to user exposed functionality or configuration variables are added/changed or an explanation of why it does not(if applicable)

PR #898, opened 2020-03-30
#883 user - Ready for Review - Adding User for Redis 

### LINKED ISSUE #883: Redis - add user to own a list of feeds.
Step 1 Add a user field to Redis 
Step 2: Our Feed data type, graphql 
Step 3: Update our tests

So users can remove and add there own feeds, this will let us move in the right direction, and make the next steps easier.

Grommers00:
## Issue This PR Addresses

## Description

This will add the ability for a user to be added to a post and feed, so they'll eventually be able to manage there own feeds. 

## Checklist

### DIFF
--- a/src/backend/data/feed.js
+++ b/src/backend/data/feed.js
@@ -13,7 +13,7 @@ const hash = require('./hash');
 const urlToId = url => hash(normalizeUrl(url));
 
 class Feed {
-  constructor(author, url, etag, lastModified) {
+  constructor(author, url, link, etag, lastModified) {
     if (!url) {
       throw new Error('missing url for feed');
     }
@@ -24,6 +24,7 @@ class Feed {
     this.id = urlToId(url);
     this.author = author;
     this.url = url;
+    this.link = link;
     // We may or may not have these cache values when we create a feed.
     this.etag = etag === '' ? null : etag;
     this.lastModified = lastModified === '' ? null : lastModified;
@@ -76,7 +77,13 @@ class Feed {
    * Returns the newly created Feed's id as a Promise<String>
    */
   static async create(feedData) {
-    const feed = new Feed(feedData.author, feedData.url, feedData.etag, feedData.lastModified);
+    const feed = new Feed(
+      feedData.author,
+      feedData.url,
+      feedData.link,
+      feedData.etag,
+      feedData.lastModified
+    );
     await feed.save();
     return feed.id;
   }
@@ -93,7 +100,7 @@ class Feed {
       return null;
     }
 
-    return new Feed(data.author, data.url, data.etag, data.lastModified);
+    return new Feed(data.author, data.url, data.link, data.etag, data.lastModified);
   }
 
   /**
--- a/src/backend/feed/processor.js
+++ b/src/backend/feed/processor.js
@@ -52,6 +52,7 @@ async function getFeedInfo(feed) {
     status: null,
     etag: null,
     lastModified: null,
+    link: null,
     contentType: null,
     shouldDownload: true,
   };
@@ -62,6 +63,7 @@ async function getFeedInfo(feed) {
     response = await fetch(feed.url, addHeaders({ method: 'HEAD' }, feed));
     info.status = `[HTTP ${response.status} - ${response.statusText}]`;
     info.contentType = response.headers.get('Content-Type');
+    info.link = feed.link;
   } catch (error) {
     logger.error({ error }, `Unable to fetch HEAD info for feed ${feed.url}`);
     throw error;
@@ -186,6 +188,13 @@ module.exports = async function processor(job) {
     // Version info for this feed changed, so update the database
     feed.etag = feed.etag || info.etag;
     feed.lastModified = feed.lastModified || info.lastModified;
+    // If feed.link is empty or there are blog posts
+    if (!feed.link && articles.length) {
+      // Assign link from first post to feed's link
+      const article = articles[0];
+      const { meta } = article;
+      feed.link = meta ? meta.link : null;
+    }
     await feed.save();
   } catch (error) {
     // If the feedparser can't parse this, we get a 'Not a feed' error
--- a/src/backend/utils/storage.js
+++ b/src/backend/utils/storage.js
@@ -39,6 +39,8 @@ module.exports = {
         feed.author,
         'url',
         feed.url,
+        'link',
+        feed.link,
         'etag',
         feed.etag,
         'lastModified',
--- a/src/backend/web/graphql/index.js
+++ b/src/backend/web/graphql/index.js
@@ -21,6 +21,7 @@ module.exports.typeDefs = graphql`
     id: String
     author: String
     url: String
+    link: String
     posts: [Post]
   }
 
--- a/test/posts.test.js
+++ b/test/posts.test.js
@@ -22,7 +22,6 @@ describe('test /posts endpoint', () => {
       guid,
       author: 'author',
       title: 'title',
-      link: 'link',
       html: 'html',
       updated: new Date('2009-09-07T22:23:00.544Z'),
       published: new Date('2009-09-07T22:20:00.000Z'),
@@ -70,7 +69,13 @@ describe('test /posts/:id responses', () => {
   const missingGuid = 'http://missing-guid';
   const randomGuid = 'http://random-guid';
 
-  const feed1 = new Feed('Feed Author', 'http://feed-url.com/', null, null);
+  const feed1 = new Feed(
+    'Feed Author',
+    'http://feed-url.com/',
+    'https://feed-link.com/',
+    null,
+    null
+  );
 
   beforeAll(() => Promise.resolve(addFeed(feed1)));
 
--- a/test/storage.test.js
+++ b/test/storage.test.js
@@ -15,8 +15,14 @@ const hash = require('../src/backend/data/hash');
 describe('Storage tests for feeds', () => {
   const feed1 = new Feed('James Smith', 'http://seneca.co/jsmith');
   const feed2 = new Feed('James Smith 2', 'http://seneca.co/jsmith/2');
-  const feed3 = new Feed('James Smith 2', 'http://seneca.co/jsmith/3', 'etag');
-  const feed4 = new Feed('James Smith 2', 'http://seneca.co/jsmith/4', 'etag', 'last-modified');
+  const feed3 = new Feed('James Smith 2', 'http://seneca.co/jsmith/3', null, 'etag');
+  const feed4 = new Feed(
+    'James Smith 2',
+    'http://seneca.co/jsmith/4',
+    'http://seneca.co/jsmith',
+    'etag',
+    'last-modified'
+  );
 
   beforeAll(() => Promise.all([addFeed(feed1), addFeed(feed2), addFeed(feed3), addFeed(feed4)]));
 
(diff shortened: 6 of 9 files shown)


src/backend/utils/storage.js
```diff
@@ -89,7 +91,9 @@ module.exports = {
         'guid',
         post.guid,
         'feed',
-        post.feed
+        post.feed,
+        'user',
+        post.user
```
manekenpix: I don't think we need to store the user in every post? Posts are already linked to a feed, so by having the user stored in the feed we could assum that the _owner_ of the feed _owns_ all the posts.

src/backend/web/graphql/index.js
```diff
@@ -35,6 +36,7 @@ module.exports.typeDefs = graphql`
     url: String
     guid: String
     feed: Feed!
+    user: String
```
manekenpix: Same here, I don't think we need this here. We have a resolver that gives you a post and the feed that _owns_ the post, so you'd have instant access to the user (since it'd be stored in the feed)

src/backend/utils/storage.js
Grommers00: @humphd I think @manekenpix is right, but I may not be thinking clearly. Any opinions?

src/backend/data/feed.js
```diff
@@ -13,7 +13,7 @@ const hash = require('./hash');
 const urlToId = url => hash(normalizeUrl(url));
 
 class Feed {
-  constructor(author, url, etag, lastModified) {
+  constructor(author, url, user, etag, lastModified) {
```
humphd: I read this PR last night on my phone, and noticed that you had some ordering bugs between callers and this constructor.  It looks like you've maybe fixed them all.

One tip for doing a change like this.  Start with a simple test that calls the constructor and makes sure that all the arguments passed in get set on the correct properties of the resulting object.  That will catch arguments being in the wrong order.

Next, use git to help you find all the callers:

```
$ git grep "new Feed"
src/backend/data/feed.js:   * Creates a new Feed object by extracting data from the given feed-like object.
src/backend/data/feed.js:    const feed = new Feed(feedData.author, feedData.url, feedData.etag, feedData.lastModified);
src/backend/data/feed.js:    return new Feed(data.author, data.url, data.etag, data.lastModified);
src/backend/web/routes/feed.js:  const feed = new FeedMeta({
test/feed.test.js:  const createFeed = () => new Feed(data.author, data.url);
test/feed.test.js:    expect(() => new Feed('author', undefined)).toThrow();
test/feed.test.js:    expect(() => new Feed(undefined, undefined)).toThrow();
test/feed.test.js:      const feed = new Feed('author', 'http://url.com/with/etag');
test/feed.test.js:      const feed = new Feed('author', 'http://url.com/with/lastModified');
test/feeds-wiki.test.js:    await new Feed(testFeedAuthor, testFeedURL).save();
test/feeds.test.js:    return new Feed('foo', `http://telescope${item}.cdot.systems`);
test/feeds.test.js:  const addedFeed = new Feed('foo', existingUrl);
test/posts.test.js:  const feed1 = new Feed('Feed Author', 'http://feed-url.com/', null, null);
test/storage.test.js:  const feed1 = new Feed('James Smith', 'http://seneca.co/jsmith');
test/storage.test.js:  const feed2 = new Feed('James Smith 2', 'http://seneca.co/jsmith/2');
test/storage.test.js:  const feed3 = new Feed('James Smith 2', 'http://seneca.co/jsmith/3', 'etag');
test/storage.test.js:  const feed4 = new Feed('James Smith 2', 'http://seneca.co/jsmith/4', 'etag', 'last-modified');
```

Then you can work through the list and correct them all.

Finally, once you have a long list of things to pass to a constructor like this, it's often wise to switch to an `options` Object pattern: pass 1 argument to the constructor with all the desired data as properties.  This has the advantage of not having to deal with ordering.

src/backend/data/feed.js
```diff
@@ -76,7 +77,13 @@ class Feed {
    * Returns the newly created Feed's id as a Promise<String>
    */
   static async create(feedData) {
-    const feed = new Feed(feedData.author, feedData.url, feedData.etag, feedData.lastModified);
+    const feed = new Feed(
+      feedData.author,
+      feedData.url,
+      feedData.etag,
+      feedData.lastModified,
+      feedData.user
```
humphd: Here you've got an ordering bug, for example.  The `user` needs to be third.

src/backend/feed/processor.js
```diff
@@ -54,6 +54,7 @@ async function getFeedInfo(feed) {
     lastModified: null,
     contentType: null,
     shouldDownload: true,
+    user: null,
```
humphd: We need to think about this one.  Probably having it be `null` is right, but it at least needs a comment above to indicate that we don't have user info at this stage (i.e., wiki feeds will be owned by no user), whereas feeds that get added via the REST API will include it via passport.js user detail on the session.

src/backend/web/graphql/index.js
```diff
@@ -61,6 +62,7 @@ module.exports.typeDefs = graphql`
     id: String
     author: String
     url: String
+    user: String
```
humphd: I don't think we'll expose this to the frontend, so I'm unclear if it's useful to have it here.  cc @c3ho for thoughts.

src/backend/web/routes/feed.js
```diff
@@ -60,6 +60,7 @@ const feedRoute = type => async (req, res) => {
   // Convert these posts to "feed" objects and add to our outgoing feed
   posts.forEach(post =>
     feed.addItem({
+      user: post.user,
```
humphd: A post doesn't have user info.  A user owns a feed which owns a post.

test/posts.test.js
```diff
@@ -70,7 +70,7 @@ describe('test /posts/:id responses', () => {
   const missingGuid = 'http://missing-guid';
   const randomGuid = 'http://random-guid';
 
-  const feed1 = new Feed('Feed Author', 'http://feed-url.com/', null, null);
+  const feed1 = new Feed('Feed Author', 'http://feed-url.com/', '', null, null);
```
humphd: I don't think we want the empty string.  It will either be `null` or a full string.  Let's not introduce this case.

test/posts.test.js
```diff
@@ -113,6 +113,7 @@ describe('test /posts/:id responses', () => {
     // Create a feed
     const feedId = await Feed.create({
       author: 'Feed Author',
+      user: '',
```
humphd: `user: 'user',`

test/storage.test.js
```diff
@@ -56,6 +63,7 @@ describe('Storage tests for feeds', () => {
 
 describe('Storage tests for posts', () => {
   const testPost = {
+    user: 'foo',
```
humphd: Remove `user` from all the `Post` objects in here.

src/backend/web/graphql/index.js
Grommers00: I'm unsure but if a user is logged in and wants to delete a post, won't this need to be exposed on the frontend somehow.

src/backend/data/feed.js
Grommers00: Ah - should I implement this?

src/backend/data/feed.js
Grommers00: fixed.

src/backend/feed/processor.js
Grommers00: So, I was thinking we would be moving the wiki feeds under an admin account or am I wrong?

src/backend/web/routes/feed.js
Grommers00: Removed

test/posts.test.js
Grommers00: Fixed.

test/posts.test.js
Grommers00: fixed.

test/storage.test.js
Grommers00: All removed!

src/backend/web/graphql/index.js
humphd: The server always knows what the current authenticated user's `id` is, since it's stored in the session.  We don't need to round-trip it through the frontend.

src/backend/data/feed.js
humphd: I would leave it as is, but fix your call ordering.  If you refactor it to use an Object, it will touch a lot more code than what you already are doing.

src/backend/feed/processor.js
humphd: `null` for the user would mean "owned by the admins", yes.

Grommers00: Ah - genius.

src/backend/web/graphql/index.js
Grommers00: Ah - I can't envision what you mean, but that can be a road crossed when I'm writing the code for that part :D.

src/backend/feed/processor.js
Grommers00: Fixed

src/backend/data/feed.js
Grommers00: Okay - I think I'm good to resolve this one, for now, if we continue to make the user object larger, we should switch to passing an object through to the constructor opposed too all the different parameters.

src/backend/web/graphql/index.js
c3ho: What @humphd said. I don't imagine us having to use user in the front end as each post displays the author's real name and if we need to uniquely identify two authors we might have the same name + last name we can always use a combination of the feedId + author.

humphd:
```
git checkout master
git pull upstream master
git checkout -B 883-user upstream/master
git cherry-pick 37efb3f5502942caacbf4749edd38e925c9e871e
git push origin 883-user -f
```

src/backend/web/graphql/index.js
Grommers00: Fixed, based on @c3ho and @humphd points..

src/backend/web/graphql/index.js
Grommers00: Fixed.

src/backend/web/graphql/index.js
```diff
@@ -22,6 +22,7 @@ module.exports.typeDefs = graphql`
     author: String
     url: String
     posts: [Post]
+    user: String
```
humphd: I don't think we need this here.  We shouldn't need to expose this to the front end.  If we end up needing it for something in the future, let's add it then.

humphd:
This PR and #807 are *very* similar, so I think you two better review each other's work (cc @c3ho).

Whoever lands first (probably @c3ho) will trigger a rebase for the other.

humphd:
This PR and #807 are both going to require people to manually blow away their redis database.  Remember we need to tell everyone this.

humphd:
@Grommers00 you've got a serious rebase on your hands now that #807 went in:

```
git checkout master
git pull upstream master
git checkout 883-user
git rebase master
...take...your...time...this...won't...be...simple
```

Get @c3ho to review your work after you do that, since your code is literally sitting next to all his, and it will be very easy to break each other.

humphd:
Can we do a tiny postmortem on this rebase?  I'm not wanting to shame or make fun here, but I would really like to understand what happened so I can help people in the future.  What was your workflow, and how could we do better to avoid this in the future?

PR #915, opened 2020-04-01
Fix #734: add mock authentication for testing

### LINKED ISSUE #734: Feeds.js - Routes and Test need to include Protect
**What happened**:
Feeds.js in `src\backend\web\routes\feeds.js`  and `test\feeds.test.js` both need to be updated
once we get protect working on the back end and we can test accordingly. 

Protect should be put on the post route, and this will affect the tests at the same time. 

You can test by not logging in and still using the post route on our express server, this shouldn't be the case. We should not move ahead though with adding protect as it will cause tests to fail.
This will be a future fix.

humphd:
## Issue This PR Addresses

Fixes #734 

## Type of Change

## Description

This adds a Jest mock for our SAML authentication strategy, allowing test code to login, login as an admin, or logout.

I've also added `protect` to our `POST /feeds` route, and updated the tests.

Any code for authenticated routes can use this.

## Checklist

### DIFF
--- a/test/feeds.test.js
+++ b/test/feeds.test.js
@@ -83,10 +83,7 @@ describe('test POST /feeds endpoint', () => {
       user: 'user',
     };
     logout();
-    const res = await request(app)
-      .post('/feeds')
-      .send(feedData)
-      .set('Accept', 'application/json');
+    const res = await request(app).post('/feeds').send(feedData).set('Accept', 'application/json');
     expect(res.status).toEqual(403);
   });
 
--- a/test/feeds.test.js
+++ b/test/feeds.test.js
@@ -6,6 +6,11 @@ const hash = require('../src/backend/data/hash');
 
 jest.mock('../src/backend/utils/elastic');
 
+// Mock the internal authentication strategy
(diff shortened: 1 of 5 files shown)


test/feeds.test.js
```diff
@@ -6,6 +6,11 @@ const hash = require('../src/backend/data/hash');
 
 jest.mock('../src/backend/utils/elastic');
 
+// Mock the internal authentication strategy
+jest.mock('../src/backend/web/authentication');
```
Grommers00, line 10: src/backend/web/__mocks__/authentication.js 

Shouldn't it be pointing here?

humphd:
@Grommers00 the way the mocks work, if you have a file at `src/foo.js` and you want to provide a fake version of it for tests, you create a file at `src/__mocks__/foo.js` and then tell Jest to mock it with `jest.mock('./src/foo')` (original module name).  You don't include the `__mocks__` portion in the name, Jest does this via `jest.mock()`, and any code that needs to require the original module will get the mock one during the test. 

See https://jestjs.io/docs/en/manual-mocks.

PR #916, opened 2020-04-01
Trim deps

### LINKED ISSUE #900: Remove tools/add-feed.js
We have to keep updating and maintaining [tools/add-feed.js](tools/add-feed.js), but we never use it.  Let's delete it.

humphd:
## Issue This PR Addresses

Fixes #900 

## Type of Change

## Description

This removes remaining dead code that we have no plans to integrate for 1.0, as well as any dependencies we aren't using, and tools we aren't using.  I want to get the dep list down to something manageable for maintaining things post 1.0.

@manekenpix sorry to remove your github url code (it's good), but we aren't using it.


cindyledev:
Would removing the Lighthouse dep also stop us from being able to access/generate Lighthouse audit reports?
Annotation 2020-04-01 150451

humphd:
@cindyledev it's built into chrome's dev tools, so we should be fine, see https://developers.google.com/web/tools/lighthouse

cindyledev:
> @cindyledev it's built into chrome's dev tools, so we should be fine, see https://developers.google.com/web/tools/lighthouse

Awesome! Just had to make sure we can still use Lighthouse because I need it

PR #919, opened 2020-04-02
Feed type should support delete, modify (delete + create)

### LINKED ISSUE #908: Authenticated users should be able to add, list, modify, or delete their own feeds
We're going to allow users to add feeds in production.  When we do, we'll need a way for them to manage them.  Basically, we need full CRUD support via our REST API for authenticated users.  This can leverage #717 to figure out what is and isn't allowed for the currently logged in user.

We can break this down into smaller PRs, and do it bit at a time, but I want one bug that tracks it all.  We need:

- support in backend (redis, elastic, node) for adding, modifying, removing feeds and the posts that belong to them
- support in REST API for all necessary operations on feeds
- support in frontend to be able to call the REST API

c3ho:
## Issue This PR Addresses
Part of #908 

## Type of Change

## Description
Calvin's part - Changes to Redis, Feed, ElasticSearch to allow CRUD operations.
Deleting a feed causes the feed's posts in Redis + ElasticSearch to be removed as well.

## Checklist

### DIFF
--- a/src/backend/data/feed.js
+++ b/src/backend/data/feed.js
@@ -4,11 +4,18 @@ const {
   getFeed,
   getFeeds,
   addFeed,
+  removeFeed,
   setInvalidFeed,
   isInvalid,
   setDelayedFeed,
   isDelayed,
+  removePost,
+  getPost,
+  getPosts,
 } = require('../utils/storage');
+
+const { deletePost } = require('../utils/elastic');
+
 const hash = require('./hash');
 
 const urlToId = (url) => hash(normalizeUrl(url));
@@ -38,7 +45,37 @@ class Feed {
    * Returns a Promise.
    */
   save() {
-    addFeed(this);
+    return addFeed(this);
+  }
+
+  /**
+   * Removes current Feed + associated Posts from the databases.
+   * Returns a Promise
+   */
+  async delete() {
+    // Removing feeds and getting all posts, we'll be assigning all the Posts we get back to posts
+    let [, posts] = await Promise.all([removeFeed(this.id), getPosts(0, 0)]);
+
+    // Filter out all posts which do not contain feed id of feed being removed
+    posts = await Promise.all(posts.map((id) => getPost(id)));
+    posts = posts.filter((post) => post.feed === this.id);
+
+    // Remove the post from Redis + ElasticSearch
+    await Promise.all(
+      [].concat(
+        posts.map((post) => removePost(post.id)),
+        posts.map((post) => deletePost(post.id))
+      )
+    );
+  }
+
+  /**
+   * Updates current Feed in the database.
+   * Returns a Promise.
+   */
+  async update() {
+    await this.delete();
+    await this.save();
   }
 
   /**
--- a/src/backend/utils/__mocks__/elastic.js
+++ b/src/backend/utils/__mocks__/elastic.js
@@ -1,22 +1,26 @@
 // Mock storage for our es data
-const db = {};
+const db = {
+  results: 0,
+  values: [],
+};
 
 const indexPost = (text, postId) => {
-  db[postId] = text;
+  db.values.push({
+    id: postId,
+    text,
+    score: 10,
+  });
+  db.results = db.values.length;
   return Promise.resolve();
 };
 
 const deletePost = (postId) => {
-  delete db[postId];
+  db.values = db.values.filter((value) => Object.values(value).includes(postId));
+  db.results = db.values.length;
   return Promise.resolve();
 };
 
-// TODO
-const search = () =>
-  Promise.resolve({
-    results: 0,
-    values: [],
-  });
+const search = () => Promise.resolve(db);
 
 const checkConnection = () => Promise.resolve();
 
--- a/src/backend/utils/storage.js
+++ b/src/backend/utils/storage.js
@@ -63,6 +63,19 @@ module.exports = {
     return redis.set(key, reason);
   },
 
+  /**
+   * Removes a feed entry from redis
+   * @param id id of feed to be removed
+   */
+  removeFeed: async (id) => {
+    const key = createFeedKey(id);
+    await redis
+      .multi()
+      .hdel(key, 'id', 'author', 'url', 'user', 'link', 'etag', 'lastModified')
+      .srem(feedsKey, id)
+      .exec();
+  },
+
   isInvalid: (id) => redis.exists(createInvalidFeedKey(id)),
 
   setDelayedFeed: (id, seconds) => redis.set(createDelayedFeedKey(id), seconds, 1),
@@ -120,4 +133,18 @@ module.exports = {
   getPostsCount: () => redis.zcard(postsKey),
 
   getPost: (id) => redis.hgetall(postNamespace.concat(id)),
+
+  /**
+   * Removes a post entry from redis
+   * NOTE: removing a post from redis should also require the post to be removed from ElasticSearch
+   * @param id id of post to be removed
+   */
+  removePost: async (id) => {
+    const key = createPostKey(id);
+    await redis
+      .multi()
+      .hdel(key, 'id', 'title', 'html', 'published', 'updated', 'url', 'guid', 'feed')
+      .zrem(postsKey, id)
+      .exec();
+  },
 };
--- a/test/feed.test.js
+++ b/test/feed.test.js
@@ -1,9 +1,12 @@
 const normalizeUrl = require('normalize-url');
 
 const Feed = require('../src/backend/data/feed');
+const Post = require('../src/backend/data/post');
+const Elastic = require('../src/backend/utils/elastic');
 const hash = require('../src/backend/data/hash');
 
 const urlToId = (url) => hash(normalizeUrl(url));
+jest.mock('../src/backend/utils/elastic');
 
 describe('Post data class tests', () => {
   const data = {
@@ -17,6 +20,28 @@ describe('Post data class tests', () => {
   };
 
   const createFeed = () => new Feed(data.author, data.url, data.user, data.link, null, null);
+
+  const articleData1 = {
+    guid: 'http://dev.telescope.cdot.systems',
+    id: hash('http://dev.telescope.cdot.systems'),
+    title: 'fooTitle',
+    description: 'fooDescription',
+    link: 'fooLink',
+    pubdate: 'fooDate',
+    length: 'fooLength',
+    date: 'fooDate',
+  };
+
+  const articleData2 = {
+    guid: 'http://example.com',
+    id: hash('http://example.com'),
+    title: 'dooTitle',
+    description: 'dooDescription',
+    link: 'dooLink',
+    pubdate: 'dooDate',
+    length: 'dooLength',
+    date: 'dooDate',
+  };
 
   test('Feed should be a function', () => {
     expect(typeof Feed).toBe('function');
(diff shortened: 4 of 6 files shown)


c3ho:
What are/aren't we allowing users to update? I'm asking this because when we create a feed, we use the url provided by the user to create an ID. Changing a URL means we have to remove the whole thing from Redis.

humphd:
@c3ho modify for the case of a feed URL means "delete the feed and all posts for this feed and make a new one."

src/backend/data/feed.js
```diff
@@ -40,6 +42,22 @@ class Feed {
     addFeed(this);
   }
 
+  /**
+   * Removes current Feed from the database.
+   * Returns a Promise
+   */
+  delete() {
+    removeFeed(this.id);
```
humphd, line 50: Remove all posts that use this `this.id` for their `feed` too.  Also, you need to `return` the Promise here: `return removeFeed()`

src/backend/data/feed.js
```diff
@@ -40,6 +42,22 @@ class Feed {
...
+  }
+
+  /**
+   * Updates current Feed in the database.
+   * Returns a Promise.
+   */
+  update() {
+    updateFeed(this);
```
humphd, line 58: Needs to `return` Promise.

src/backend/utils/storage.js
```diff
@@ -63,6 +63,36 @@ module.exports = {
     return redis.set(key, reason);
   },
 
+  updateFeed: feed => {
```
humphd, line 66: Don't bother with this.  Delete and make a new one.

src/backend/utils/storage.js
```diff
@@ -63,6 +63,15 @@ module.exports = {
     return redis.set(key, reason);
   },
 
+  removeFeed: async id => {
+    const key = createFeedKey(id);
+    await redis
```
humphd, line 68: Probably we should remove the posts here at the redis level when we remove the feed.

test/feed.test.js
```diff
@@ -91,5 +122,47 @@ describe('Post data class tests', () => {
...
+      expect(modifiedFeed.author).toBe(feed.author);
+      expect(modifiedFeed.id).toBe(urlToId(feed.url));
+    });
+
+    test('Removing feeds should also remove posts', async () => {
+      const feed = new Feed(data.author, data.url, data.user, data.link, null, null);
+      await feed.save();
+      postData1.feed = feed.id;
+      postData2.feed = feed.id;
+      await Promise.all([addPost(postData1), addPost(postData2)]);
```
humphd, line 147: Instead of using `addPost`, can we use `new Post()` and `post.save()`?

Same thing below, with `getPost`.  We either need to use the storage methods or the higher order data methods, but shouldn't mix.  I think we should avoid using storage directly as much as possible, since it returns data in different formats that we expect when working with Feeds and Posts.

c3ho: Fixed, I'll be using `Feed.create()` and `Post.create()` instead. Wasn't sure if I should've used the Post class in here since we were testing Feed stuff

test/feed.test.js
```diff
@@ -91,5 +122,47 @@ describe('Post data class tests', () => {
...
+      expect(modifiedFeed.url).not.toBe(createFeed.url);
+      expect(modifiedFeed.author).not.toBe(createFeed.author);
+      expect(modifiedFeed.url).toBe(feed.url);
+      expect(modifiedFeed.author).toBe(feed.author);
+      expect(modifiedFeed.id).toBe(urlToId(feed.url));
+    });
+
+    test('Removing feeds should also remove posts', async () => {
+      const feed = new Feed(data.author, data.url, data.user, data.link, null, null);
+      await feed.save();
```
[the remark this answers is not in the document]
c3ho: Fixed and thank you that fixed all the issues I was having

c3ho:
Should I do the elastic search post removal in this PR as well?

test/feed.test.js
```diff
@@ -35,7 +67,10 @@ describe('Post data class tests', () => {
   });
 
   describe('Get and Set Feed objects from database', () => {
-    beforeAll(() => createFeed().save());
+    beforeAll(async () => {
+      await createFeed().save();
+      jest.mock('../src/backend/utils/elastic');
```
manekenpix, line 72: I think this line might need to be at the top.

c3ho: Thank you, all tests are passing now

c3ho:
@humphd I'm unable to write tests regarding ElasticSearch in `feed.test.js`, Although I can confirm it is working, because when I didn't add the jest.mock(esStuff) I was getting the following error from ElasticSearch trying to delete the test posts

image

I'm unsure if you're satisfied with this or not

humphd:
Should we extend our elastic mock so it holds onto objects in memory, and delete removes them?  We can do more than just return a resolved promise, and then your code would work.

src/backend/data/feed.js
```diff
@@ -37,7 +44,34 @@ class Feed {
...
+  /**
+   * Removes current Feed + associated Posts from the databases.
+   * Returns a Promise
+   */
+  async delete() {
+    await removeFeed(this.id);
+    const posts = await getPosts(0, 0);
+    const filteredPosts = posts.filter(async (post) => (await getPost(post).feed) === this.id);
+    const expandedPosts = await Promise.all(
+      await filteredPosts.map((filteredPost) => getPost(filteredPost))
```
humphd, line 59: I don't think you need to `await` here, you should be returning an Array of Promises.

test/feed.test.js
```diff
@@ -18,6 +21,36 @@ describe('Post data class tests', () => {
...
+    author: 'foo',
+    title: 'foo',
+    link: 'foo',
+    content: 'foo',
+    text: 'foo',
+    updated: new Date('2009-09-07T22:23:00.544Z'),
+    published: new Date('2009-09-07T22:21:00.000Z'),
+    url: 'foo',
+    site: 'foo',
+    feed: '',
```
humphd, line 36: Why the empty string?  At least make this `foo` and `doo` as well.

src/backend/data/feed.js
```diff
@@ -37,7 +44,34 @@ class Feed {
...
   save() {
-    addFeed(this);
+    return addFeed(this);
+  }
+
+  /**
+   * Removes current Feed + associated Posts from the databases.
+   * Returns a Promise
+   */
+  async delete() {
```
humphd, line 54: This function has a lot of Promises, which is not great for performance.  We should see if we can reduce the number in a follow-up PR.

src/backend/utils/storage.js
```diff
@@ -120,4 +129,14 @@ module.exports = {
   getPostsCount: () => redis.zcard(postsKey),
 
   getPost: (id) => redis.hgetall(postNamespace.concat(id)),
+
+  removePost: async (id) => {
+    const key = createPostKey(id);
+    await redis
+      .multi()
+      .hdel(key, 'id', 'title', 'html', 'published', 'updated', 'url', 'guid', 'feed')
+      // sort set by published date as scores
```
humphd, line 138: Remove this comment.

src/backend/utils/storage.js
```diff
@@ -120,4 +129,14 @@ module.exports = {
   getPostsCount: () => redis.zcard(postsKey),
 
   getPost: (id) => redis.hgetall(postNamespace.concat(id)),
+
+  removePost: async (id) => {
```
humphd, line 133: Add a comment above this about having to do the same in elastic (Feed does this).

src/backend/data/feed.js
```diff
@@ -37,7 +44,34 @@ class Feed {
...
+    await removeFeed(this.id);
```
humphd, line 55: Add comments above each line to make it more clear what's happening (eg., `// Remove this feed from redis`)

src/backend/data/feed.js
```diff
@@ -37,7 +44,34 @@ class Feed {
...
+    const posts = await getPosts(0, 0);
```
humphd, line 56: This line and 55 could be `await Promise.all([ ...remove feed, ...get and remove posts])`

src/backend/data/feed.js
```diff
@@ -37,7 +44,34 @@ class Feed {
...
+    const filteredPosts = posts.filter(async (post) => (await getPost(post).feed) === this.id);
```
humphd, line 57: I think this should be:

```js
const feedId = this.id;

// Get posts ids
let posts = await getPosts(0, 0);
// Get full data for post ids
posts = await Promise.all(posts.map(id => getPost(id)));
// Filter out posts that aren't for this feed
posts = posts.filter(post => post.feed === feedId);
// Take this reduced set of feeds, and remove from redis + elasict
await Promise.all(posts.map(post => Promise.all([removePost(post.id), deletePost(post.id)])));
```

untested :)

src/backend/data/feed.js
```diff
@@ -37,7 +44,34 @@ class Feed {
...
+      })
+    );
+  }
+
+  /**
+   * Updates current Feed in the database.
+   * Returns a Promise.
+   */
+  update() {
+    return Promise.all([this.delete(), this.save()]);
```
humphd, line 74: I would:

```js
// Do in order
await this.delete();
await this.save();
```

src/backend/utils/__mocks__/elastic.js
```diff
@@ -11,11 +11,10 @@ const deletePost = (postId) => {
   return Promise.resolve();
 };
 
-// TODO
 const search = () =>
   Promise.resolve({
-    results: 0,
-    values: [],
+    results: Object.keys(db).length,
+    values: db,
```
humphd, line 17: `values` should look like:

```js
[
 { id, text, score },
  ...
]
```

Where `id` is the id of the document (postId), `text` is the text of the post`, and score is a number between 1 and 10.  We could always return 10.

humphd:
```js
  values: Object.entries(db).map((id, text) => { id, text, 10 })
```

Or do it in the `indexPost` above and make it look the way you want, then you can use `Object.values()`

test/feed.test.js
```diff
@@ -91,5 +124,62 @@ describe('Post data class tests', () => {
...
+      expect(modifiedFeed.id).toBe(urlToId(feed.url));
+    });
+
+    test('Removing feeds should also remove posts', async () => {
+      const feed = await Feed.byId(await Feed.create(data));
+      postData1.feed = feed;
+      postData2.feed = feed;
+      await Promise.all([
+        Post.create(postData1),
+        Post.create(postData2),
```
humphd, line 150: These should probably be using `createFromArticle()` and we can probably delete `Post.create()` (in a follow-up PR if you want to file that).

test/feed.test.js
```diff
@@ -91,5 +124,62 @@ describe('Post data class tests', () => {
...
+        Elastic.indexPost(postData1.text, postData1.id),
```
humphd, line 151: Once you use `createFromArticle()` you won't need to do this manually (it does es for you).

test/feed.test.js
```diff
@@ -91,5 +124,62 @@ describe('Post data class tests', () => {
...
+      expect(posts[1].id).toBe(postData2.id);
+      expect(posts[0].feed.id).toBe(postData1.feed.id);
+      expect(posts[1].feed.id).toBe(postData2.feed.id);
+      expect(posts[0].feed.author).toBe(postData1.feed.author);
+      expect(posts[1].feed.author).toBe(postData1.feed.author);
+      // Testing Posts in ElasticSearch here
+      expect(elasticPosts.results).toBe(posts.length);
+      // Post doesn't contain text so we're checking for postData.text here instead
+      expect(Object.values(elasticPosts.values).indexOf(postData1.text)).not.toBe(-1);
+      expect(Object.values(elasticPosts.values).indexOf(postData2.text)).not.toBe(-1);
```
humphd, line 171: If we do a better job in the results from es search() this can be simplified.

test/storage.test.js
```diff
@@ -3,6 +3,7 @@ const {
   getFeed,
   getFeeds,
   getFeedsCount,
+  removeFeed,
```
humphd, line 6: Can you add a test for `removePost` too, since that got added as well?

test/storage.test.js
```diff
@@ -68,6 +69,15 @@ describe('Storage tests for feeds', () => {
     expect(feeds[2].link).toBe('');
     expect(feeds[3].link).toBe('http://seneca.co/jsmith');
   });
+
+  it('feed4 should not exist after being removed', async () => {
+    const feed = await getFeed(feed4.id);
+    await removeFeed(feed.id);
+    const removedFeed = await getFeed(feed.id);
+    const feeds = await getFeeds();
+    expect(removedFeed.id).toBe(undefined);
```
humphd, line 78: This seems odd to me, that `removedFeed` would be an Object vs. `null`.  Can you confirm that this is actually returning something?  If it is, why is it?

humphd: `console.log(removedFeed, typeof removedFeed, ...)`

c3ho: Confirmed on slack with the following image of `console.log`

image

test/storage.test.js
c3ho: Added

test/feed.test.js
c3ho: Simplified!

test/feed.test.js
c3ho: Switched to use `createFromArticle()`

src/backend/data/feed.js
c3ho: Refactored even more

src/backend/data/feed.js
c3ho: Fixed

src/backend/utils/__mocks__/elastic.js
c3ho: Changed the mock to be identical to live ES

src/backend/data/feed.js
c3ho: changed

src/backend/data/feed.js
c3ho: added more comments

src/backend/utils/storage.js
c3ho: removed

src/backend/utils/storage.js
c3ho: added comment, not sure if it is what you wanted

src/backend/data/feed.js
```diff
@@ -37,7 +44,33 @@ class Feed {
...
+  }
+
+  /**
+   * Removes current Feed + associated Posts from the databases.
+   * Returns a Promise
+   */
+  async delete() {
+    // Removing feeds and getting all posts, we'll be assigning all the Posts we get back to posts
+    const [, posts] = await Promise.all([removeFeed(this.id), getPosts(0, 0)]);
+    // Filter out all posts which doesn't contain feed id of feed being removed
```
humphd, line 57: which DON'T vs. DOESN'T

src/backend/data/feed.js
humphd: This is still wrong.  You can't do `posts.filter(async (post) =>...`, filter is synchronous.

src/backend/data/feed.js
c3ho: I can't grammar

src/backend/utils/__mocks__/elastic.js
```diff
@@ -1,22 +1,28 @@
...
+    score: 10,
+  });
+  db.results = db.values.length;
   return Promise.resolve();
 };
 
 const deletePost = (postId) => {
-  delete db[postId];
+  db.values = db.values.filter((value) => {
+    return Object.values(value).indexOf(postId);
```
humphd, line 19:
```js
db.values = db.values.filter((value) => Object.values(value).includes(postId));
```

src/backend/data/feed.js
```diff
@@ -37,7 +44,36 @@ class Feed {
...
+  async delete() {
+    // Removing feeds and getting all posts, we'll be assigning all the Posts we get back to posts
+    let [, posts] = await Promise.all([removeFeed(this.id), getPosts(0, 0)]);
+
+    // Filter out all posts which do not contain feed id of feed being removed
+    posts = await Promise.all(posts.map((id) => getPost(id)));
+    posts = posts.filter((post) => post.feed === this.id);
+
+    // Remove the post from Redis + ElasticSearch
+    return Promise.all(
```
humphd, line 63:
```js
await Promise.all(
  [].concat(posts.map(post => removePost(post.id), posts.map(deletePost => post.id)
)
```

c3ho:
Merging as it received reviews and passed before

Issue #841, opened 2020-03-17
Header Bar text increases size when search button clicked


agarcia-caicedo:
**What happened**:
When a user goes to the search page and clicks the search button, regardless of filter, the text in the header bar grows in size. This is how it looks.

![image](https://user-images.githubusercontent.com/49992710/76881013-e6f86e00-684e-11ea-8d8f-cef895e2e9f4.png)

**What should have happened**:
This is how the header should look:

![image](https://user-images.githubusercontent.com/49992710/76881057-f8417a80-684e-11ea-8787-0e27dc457d04.png)

**How to reproduce it (as precise as possible)**:
Go to the search page and click the search button in the search bar. Typing anything in the bar or changing the filter has no different effect.

- Browser: Firefox

cindyledev:
Does it think the magnify glass button means zooms in or something? 😂

cindyledev:
This is a very weird bug, I can't seem to be able to reproduce this locally on Linux or Windows but it's definitely happening at `https://dev.telescope.cdot.systems/search`

cindyledev:
Let's see if this is still a problem after #792 gets merged...

cindyledev:
I might have found the culprit. I was going through our code looking for .css files to delete for #856 and as a test, I deleted `/src/frontend/src/pages/index.css`. This is the exact look that happens after the search button in the search bar is clicked... I think `index.css` is somehow applying itself here.

Before delete:
![Annotation 2020-03-22 091904](https://user-images.githubusercontent.com/36822198/77250416-40311a80-6c1e-11ea-9a83-dc848f9ffe6b.png)

After delete:
![Annotation 2020-03-22 091904_](https://user-images.githubusercontent.com/36822198/77250445-6b1b6e80-6c1e-11ea-8119-e9d51fffd75f.png)

Unfortunately, everything else gets huge `index.css` gets deleted...
![Annotation 2020-03-22 091904__](https://user-images.githubusercontent.com/36822198/77250505-c3eb0700-6c1e-11ea-9a07-5729ae84f41a.png)
![Annotation 2020-03-22 091904___](https://user-images.githubusercontent.com/36822198/77250507-c5b4ca80-6c1e-11ea-8673-d03bc1d50606.png)

I guess this is another effect of what @Silvyre was talking about last month in #741

cindyledev:
While trying to reproduce #912 , I ran into this hot mess
[...]

Issue #741, opened 2020-02-19
Refactor frontend CSS into CSS-in-JS


Silvyre:
All of our convential frontend CSS styling solutions should be refactored into Material UI's [CSS-in-JS](https://material-ui.com/styles/basics/) styling solution.
<hr>
<details>
<summary><em>(Old issue text)</em></summary>

Currently [`index.css`](https://github.com/Seneca-CDOT/telescope/blob/master/src/frontend/src/pages/index.css) imposes its styling rules upon all Gatsby Pages. While developing [the `Add Feeds` component](https://github.com/Seneca-CDOT/telescope/pull/735), I found some of these rules (namely the scaled-down font size) to be less-than-helpful, [as I noted in my pull request](https://github.com/Seneca-CDOT/telescope/pull/735):

> It is worth noting that this component currently ends up being rendered quite small due to styling enforced by `src/frontend/src/pages/index.css`:
   https://github.com/Seneca-CDOT/telescope/blob/9dc85649f1c29d1a7fceafc9b7ccbda668beb22d/src/frontend/src/pages/index.css#L9-L13

This is not, however, the only styling issue I ran into with `index.css`. I noticed that, due to the stylesheet [enforcing a `body` height of `100vh`](https://github.com/Seneca-CDOT/telescope/blob/9dc85649f1c29d1a7fceafc9b7ccbda668beb22d/src/frontend/src/pages/index.css#L20), the `Add Feed` component's [current use of margins](https://github.com/Seneca-CDOT/telescope/blob/24c6a44423e887991d032f2425cb39a0ab044943/src/frontend/src/pages/myfeeds.js#L8) forces an unnecessary scrollbar to appear:

![margin](https://user-images.githubusercontent.com/13500769/74790231-010d5380-5285-11ea-9c65-f5263d61f791.gif)

<hr>

There is [some discussion](https://github.com/Seneca-CDOT/telescope/issues/730#issuecomment-587138483) about eventually replacing our CSS files, like `index.css`, with CSS-in-JS (JSS). Until we have come to an agreement to that end, I believe that refactoring our CSS files (i.e. to ensure their global styling does not adversely affect new components) prevents a good stop-gap. (Again, at this point, we definitely don't want to outright _remove_ any CSS files...)

<hr>

To summarize what I believe should be done:
[...]

PR #925, opened 2020-04-03
adds: error page front-end concept

### LINKED ISSUE #890: Create custom pages for HTTP 404, 403, 500s

Let's consider adding some custom pages for a few error cases that we have.  I'm not 100% clear on the right way to deal with the difference between what lives in Gatsby and what is just static that we can access via Express.  This will need some research.

What I'd like to deal with are these cases:

- 404s: when we can't find something
- 401s, 403s: when the user isn't allowed to do something
- 500s: we crashed

I don't think there are other cases, but people can add them if I'm missing any.

This is something we could have some fun with, and bring a bit of whimsy or branding to our app.  But it's not a critical component for a 1.0.

In Express we can do custom error handling with a function at the end of the middleware stack.  Here are some links:

- https://stackoverflow.com/questions/6528876/how-to-redirect-404-errors-to-a-page-in-expressjs
- http://expressjs.com/en/starter/faq.html
- http://expressjs.com/en/guide/error-handling.html

raygervais:
## Issue This PR Addresses
Fixes #890 

## Type of Change

## Description

This adds a catch-all route at the end of the routes declaration which redirects to our (WIP) 40X page component. Will add appropriate 50x page as well.

## How To Access

From a fully deployed Docker-compose environment, attempt any url which you'd fat-finger by accident at 3am. This will redirect back to the front-end with the appropriate error response. Example, entering `localhost:3000/AWorldWhereWindowsWins` should redirect straight to `localhost:3000/error?status=404` as seen below.

## Checklist

image


raygervais:
@cindyledev, @humphd, @agarcia-caicedo, looking for your input on how we should style and layout the 40X / 50X error pages. For the above, just to prove the concept I reused Banner layout styling. Was contemplating adding the banner (sans banner text) in the background.

PR is still WIP, looking for input and feedback.

raygervais:
Because of the Zeit setup, you won't be able to follow the instructions first step (attempting to access any non-standard URL) and instead can test with this URL (which is the redirect): `https://telescope-mxr4bagfx.now.sh/404?search=<PUT_WHATEVER_YOU_LIKE>`. Locally, this works end-to-end.

cindyledev:
Tested on Zeit, I like where this is going. Unfortunately, this seems to be suffering from the css bug #841 #741 . We have 3 at least PRs open that are affected by it

cindyledev:
Also tested on Linux
Screenshot from 2020-04-03 22-18-24

src/backend/web/routes/index.js
```diff
@@ -44,4 +44,16 @@ router.use('/stats', stats);
 router.use('/user', user);
 router.use('/query', query);
 
+/**
+ * 404 Handler, Pass to front-end
+ * Leverage .status because adding the `404` status in redirect causes "Not Found. Redirecting to /404?search=" to display.
+ */
+router.use('*', (req, res) => {
```
humphd, line 51: The way this is usually done is as error handling middleware at the end of the routing chain, with an extra error handler at the end:

```js
// Last in the chain
app.use(function (req, res, next) {
  res.status(404).send('...');
});

// Error handling middleware
app.use(function (err, req, res, next) {
  logger.error({ error: err }, 'express web server error');
  res.status(500).send('...')
});
```

src/backend/web/routes/index.js
```diff
@@ -44,4 +44,16 @@ router.use('/stats', stats);
...
+  const originalUrl = req.originalUrl.trim();
+  res
+    .status(404)
+    .redirect(`/404?search=${originalUrl}`)
```
humphd, line 55: You better `encodeURIComponent()` this `originalUrl`

src/backend/web/routes/index.js
```diff
@@ -44,4 +44,16 @@ router.use('/stats', stats);
...
+    .sendFile(path.join(__dirname, '../../../frontend/public/index.html'));
```
humphd, line 56: Why redirect and send this file?  Would you not just redirect?

src/frontend/src/pages/404.jsx
```diff
@@ -0,0 +1,82 @@
...
+    },
+  },
+}));
+
+const ErrorPage = (props) => {
+  const classes = useStyles();
+  const params = new URLSearchParams(props.location.search);
+  let originalUrl = params
+    .get('search')
+    ?.replace('/', '')
```
humphd, line 62: I don't understand what this is doing.  I'm also not sure that we should introduce `?.` to our code base without discussing implications.

src/frontend/src/pages/404.jsx
```diff
@@ -0,0 +1,82 @@
...
+    ?.match(/([A-Z]?[^A-Z]*)/g)
+    ?.slice(0, -1)
+    ?.join(' ');
+
+  return (
+    <div>
```
Silvyre, line 68: I think we should replace this `div` wrapper with [our new `PageBase` wrapper](https://github.com/Seneca-CDOT/telescope/blob/master/src/frontend/src/pages/PageBase.js).

src/frontend/src/pages/404.jsx
```diff
@@ -0,0 +1,82 @@
...
+      <Header />
```
Silvyre, line 69: Once #922 lands, `<Header>` will be included via [the `PageBase` wrapper](https://github.com/Seneca-CDOT/telescope/blob/master/src/frontend/src/pages/PageBase.js). As such, I think this instance of `<Header>` can be safely removed.

src/frontend/src/pages/404.jsx
raygervais: The ?. is the new optional chaining operator, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

It's essentially a replacement of the following semantic: 

``` javascript
const a= params.get('search');
 if (a) {
    const b = a.replace('/', '');

   if (b) {
     const c = b.match(....);

     ... and so on
   }
 }
```

The reason for the actual functions is removing the `/` at the start of the URL search param, which I may be able to remove with the urlEncode that you suggested above. Furthermore, it formats the param client-side to `Capital Spaced` if capital letters exist. Can remove, thought it was a nice little addon.

src/frontend/src/pages/404.jsx
```diff
@@ -0,0 +1,184 @@
...
+  buttonText: {
+    fontSize: '1.5rem',
+    paddingLeft: theme.spacing(1),
+    [theme.breakpoints.between('xs', 'sm')]: {
+      display: 'none',
+    },
+  },
+}));
+
+function RetrieveBackgroundAsset() {
```
humphd, line 105: Isn't this repeated code?  Can we refactor this out into a separate component?

src/frontend/src/pages/404.jsx
```diff
@@ -0,0 +1,184 @@
...
+      }}
+    ></div>
+  );
+}
+
+const ErrorPage = (props) => {
+  const classes = useStyles();
+
+  return (
+    <PageBase title="404">
```
humphd, line 148: Either here or in follow-up, I'd love to generalize this page so that it works for 403s (forbidden) and 500s (errors).

src/backend/web/app.js
```diff
@@ -68,4 +68,12 @@ app.use(logger);
...
 app.use('/', router);
 
+/**
+ * 404 Handler, Pass to front-end
+ * Leverage .status because adding the `404` status in redirect causes "Not Found. Redirecting to /404?search=" to display.
+ */
+app.use((req, res) => {
+  res.status(404).redirect(`/404`);
+});
+
```
humphd, line 78: While you're doing this, can you add a general error handler at the end of this that uses our logger to log the error before sending the 500?

src/frontend/src/pages/404.jsx
raygervais: I can do that here, keep all the scope around 40X and 50X pages starting here.

src/frontend/src/pages/404.jsx
raygervais: I'd like to refactor into a separate component in this PR, is that alright? Or separate follow-up?

src/backend/web/app.js
raygervais: `logger.error`? I had tried this earlier and couldn't access the logger method.

humphd: It's due to how we export that logger.  You probably need `logger.logger`, which is horrible.

src/frontend/src/pages/404.jsx
humphd: Great

src/frontend/src/pages/404.jsx
humphd: Here is fine, nothing is depending on this for timing.

agarcia-caicedo:
> 
> 
> This is looking amazing. I'm not sure about the purple colour you're introducing (cc @agarcia-caicedo). I left some initial comments in the code.

Can't say I'm a fan of the purple. I don't think it works aesthetically, and since it's the only place we've used it, it looks out of place

cindyledev:
I'm digging this so far but I'm not too sure about the purple back button. @agarcia-caicedo , what are your thoughts?
Annotation 2020-04-10 233857

agarcia-caicedo:
If we are to incorporate a purple, we need to make sure we use it in more places than just here AND it needs to be a different purple.

The current hues are too similar in terms of luminosity and contrast, which is an issue for accessibility as blue and purple are colours some color blind users have issues with. While we don't need to steer away from that combination, we do need ensure the colours are distinct enough in terms of contrast.

Aside from accessibility, I just don't think that shade is working. It's relatively desaturated and dark, making the blue and off-white stand out more, so my eyes just get drawn away from it.

raygervais:
Reverted to using the secondary [`light` | `dark`] variants.  Light by default, dark when hovered (second image)
image

image


Thanks for the input, really appreciate you going into your thought process when it comes to color coordination and accessibility. Happily handing you the steering wheel when it comes to UI design on this part. 🎡  🎆

agarcia-caicedo:
Yeah I didn't want to say "I don't like the shade" and leave it at that. It's not really useful, and usually there's a reason why a shade doesn't work. I did take a while to pin point though.

The 404 keeps getting better and better :D

humphd:
Can we pay the CircleCI tax here and keep moving this forward?

```
/home/circleci/telescope/src/backend/web/app.js

  74:25  error    'next' is defined but never used     no-unused-vars



/home/circleci/telescope/src/backend/web/authentication.js

  126:7   error  'err' is never reassigned. Use 'const' instead  prefer-const

  133:20  error  'req' is defined but never used                 no-unused-vars

  133:25  error  'res' is defined but never used                 no-unused-vars
```

src/backend/web/authentication.js
```diff
@@ -122,16 +122,17 @@ function protectWithRedirect(req, res, next) {
...
 function forbidden(req, res) {
-  if (req.accepts('json')) {
-    res.status(403).json({
-      message: 'Forbidden',
-    });
-  } else {
-    // TODO: https://github.com/Seneca-CDOT/telescope/issues/890
-    res.status(403).send('Forbidden');
-  }
+  throwCustomError('', 403);
```
humphd, line 135: Change `''` to `'Forbidden'`

src/backend/web/authentication.js
```diff
@@ -122,16 +122,17 @@ function protectWithRedirect(req, res, next) {
...
+function throwCustomError(message, status) {
+  const err = Error(message);
+  err.status = status;
+
+  throw err;
+}
+
 // If user is not authenticated, return an appropriate 400 error type
+// eslint-disable-line no-unused-vars
 function forbidden(req, res) {
```
humphd, line 134: I think you need to change this to:

```js
function forbidden(next) {
    const err = new Error(message);
    err.status = 403;
    next(err);
}
```

And change the callers to `forbidden()` to pass `next` vs. `req` and `res`

src/backend/web/app.js
```diff
@@ -68,4 +68,25 @@ app.use(logger);
 // Include our router with all endpoints
 app.use('/', router);
 
+/**
+ * Error Handler, Pass to front-end
+ */
+// eslint-disable-line no-unused-vars
+app.use((err, req, res, next) => {
+  logger.logger.error(err);
```
humphd, line 76: `logger.logger.error( { error: err } );`

humphd: For the eslint thing, I think we need `after-used` https://eslint.org/docs/rules/no-unused-vars#args

src/backend/web/authentication.js
raygervais: We wouldn't need to `throw` in this case if using `next`?

raygervais: Like wise we'd have to update many of the function arguments which call `forbidden`, wouldn't we?

src/backend/web/app.js
```diff
@@ -68,4 +68,25 @@ app.use(logger);
...
+    .status(status)
+    .redirect(`/error?status=${status}${err.message ? `&message=${encodeURI(err.message)}` : ``}`);
+});
+
+/**
+ * 404 Handler, Pass to front-end
+ * Leverage .status because adding the `404` status in redirect causes "Not Found. Redirecting to /404?search=" to display.
+ */
+app.use((req, res) => {
+  logger.logger.warn(req.url);
```
humphd, line 88: Add a message here, so we know what the warning is (not found).

src/backend/web/authentication.js
humphd: What if we do this:

- have forbidden take all three `forbidden(req, res, next)` and the third argument.
- update all callers to pass 3 args with next as last
- rework what was here before so you do what you're doing only in the else case, and instead of sending the 403 on `res`, you called `next(err)`.  We shouldn't need to throw at all.

raygervais:
All that is outstanding is standardizing how we display 500s, which could be passed to the front-end via the message param if it's a development environment.

src/backend/web/authentication.js
```diff
@@ -122,21 +122,22 @@ function protectWithRedirect(req, res, next) {
...
+function throwCustomError(message, status) {
+  const err = Error(message);
+  err.status = status;
+
+  throw err;
+}
+
 // If user is not authenticated, return an appropriate 400 error type
-function forbidden(req, res) {
-  if (req.accepts('json')) {
```
humphd, line 127: We still want this case left in, because I don't want to send HTML back to a REST API call.  Only do your custom error path if it isn't expecting JSON.

src/frontend/src/components/DynamicBackgroundContainer.jsx
```diff
@@ -0,0 +1,70 @@
...
+
+      // Ensure we are using an image which fits correctly to user's viewspace
+      const dimensions = `${window.innerWidth}x${window.innerHeight}`;
+      console.log(dimensions);
+      const response = await fetch(`https://source.unsplash.com/collection/894/${dimensions}`);
+
+      if (response.status !== 200) {
+        setBackgroundImgSrc('../../images/hero-banner.png');
+        setTransitionBackground(false);
+        throw new Error(response.statusText);
```
humphd, line 41: This error will fall on the floor, since there is nothing to catch it.  Should you console.error() instead?

src/frontend/src/pages/error.jsx
```diff
@@ -0,0 +1,170 @@
...
+    [theme.breakpoints.up('xl')]: {
+      fontSize: '8rem',
+    },
+  },
+  h3: {
+    fontFamily: 'monospace',
+    fontSize: '2rem',
+    padding: theme.spacing(4),
+    color: theme.palette.grey[300],
+    // backgroundColor: theme.palette.error.light,
```
humphd, line 65: Remove this

src/frontend/src/pages/error.jsx
```diff
@@ -0,0 +1,170 @@
...
+    '401': 'You are not authorized to view this page!',
+    '403': 'Access is not allowed for the requested page!',
+    '404': 'We could not find what you were looking for!',
+    '405': 'Method is not allowed!',
+  };
+
+  if (props.status !== '500' && !props.message) {
+    return (
+      <Typography variant="body1" className={classes.h2}>
+        {errorMessages[props.status]}{' '}
```
humphd, line 114: Should you have some kind of fallback "Something went wrong" error message for any case not covered by the error codes above?  We might add more in the future.

humphd:
Need a rebase for Banner.jsx

raygervais:
This is because we don't have a custom 404 page, instead I'm using `/error` as a wrapper for both. I'd redirect there, or in a follow up PR I create a 404 which is a wrapper around error.

raygervais:
You should be sending to 403, since they are not authorized @Silvyre. `/error?status=403`

Silvyre:
> I'm using `/error` as a wrapper for both. I'd redirect there, or in a follow up PR I create a 404 which is a wrapper around error.

Okay, cool: redirecting to `/error?status=401` seems to work well. We could alternatively have the My Feeds page redirect logged-out users to the login page (which I recall is what @humphd may  have suggested this morning).

<hr>

A few other things...

Navigating to `http://localhost:3000/error` (without a query string) presents:
image

Do you think it would be worth populating the blank card here with the "Something went wrong!" text?

Also, I noticed that navigating to `http://localhost:3000/error?status=40X` renders the numerical status code (e.g. "40X") within the page title, which is really cool: 
image

Do you think it would be worthwhile to add a switch that replaces some numerical status codes with their common names? e.g. being redirected to `http://localhost:3000/error?status=404` would result in a page title of "Not Found | Telescope" instead of "404 | Telescope", etc. Ultimately up to you!

Lastly, I'm not sure whether this next issue affects only my own machine/browser, but when I navigate to a page like `http://localhost:3000/error?status=404`, the card/button component appears to have some margin issues:
image

This does not happen when I decrease the size of my browser window to a size breakpoint smaller:
image

<hr>

Thanks for all of your hard work on this! 👍

src/backend/web/authentication.js
```diff
@@ -132,21 +132,28 @@ function protectWithRedirect(req, res, next) {
...
+
+  throw err;
+}
+
 // If user is not authenticated, return an appropriate 400 error type
-function forbidden(req, res) {
+/* eslint-disable no-unused-vars */
+function forbidden(req, res, next) {
   if (req.accepts('json')) {
     res.status(403).json({
```
humphd, line 146: You need a `return` here so it doesn't throw when it's done.

raygervais:
Thanks for testing so vigorously @Silvyre, that edge (where the user stumbles onto the page) I never thought of. Corrected in upcoming commit. 

image

raygervais:
@Silvyre can I get a rereview?

raygervais:
Thanks for the corrections @cindyledev, can I get a re-review when you have a moment? So Happy to have the eslint warnings gone! 🎆

PR #928, opened 2020-04-03
Fix #741, #841: create a base component for pages

### LINKED ISSUE #741: Refactor frontend CSS into CSS-in-JS

### LINKED ISSUE #841: Header Bar text increases size when search button clicked

humphd:
## Issue This PR Addresses

1. Fixes #741
1. Fixes #841

## Type of Change

## Description

This sets our root element's font size to `10px` using Material UI instead of a stylesheet.  I've also added a CSS Baseline, and created a `PageBase` component, that all other pages can use to get this behaviour for free.

We should figure out how to hook up the theme properly using this, so all pages get it from the top.  But for now, this allows us to dump `index.css` and get the same look on all pages.


humphd:
One thing @Silvyre and I were talking about that @agarcia-caicedo, @cindyledev, @raygervais, and others working on "pages" should consider: this `PageBase` could/should probably get more stuff added to it, so that you don't have to do as much on an individual page.  For example, SEO, the header bar, footer, etc. could all be done in the `PageBase` and removed from each page.

I considered doing it in this PR, but it seemed like scope creep.  Whoever works on other pages in subsequent PRs can do it.

PR #931, opened 2020-04-04
Add a way to receive updates when new posts are available

### LINKED ISSUE #791: Add a way to receive updates when new posts are available
If I leave Telescope running in a tab for a long time, I want it to tell me when there are new posts available.  I think it should work like this:

1. We setup a long-poll (i.e., keep calling `fetch()` or do a graphql query) every few minutes
1. When we get back data that is different from our current state, we know that there are updates
1. Show some kind of "Toast" or "Banner" or some other floating action button that says, "New posts available..." or something, and clicking it loads them in and shows the new data
1. We should also update the favicon in the tab to indicate that there are updates (red/white notification circle).

This is somewhat connected to #790.

c3ho:
## Issue This PR Addresses
Fixes #791 

## Type of Change

## Description
Adding a `setInterval` to the front end that will query telescope to see if the # of posts has changed.

It is currently set at 5 seconds as I'm still working on it. How it should ideally work:
1. The page is loaded: we get posts, total posts count AND start a timer. When we fetched for the total posts count, we assign it to two states `initNumPosts` and `currentNumPosts`.
2. Every 5s the attached timer will fetch for total posts count and assign that value to currentNumPosts.
3. At the return statement portion of `index.js` I have a conditional that will render the component `CustomizedSnackBar`. I set it so that as soon as the `CustomizedSnackBar` component is rendered it will trigger an alert `There are new post(s) available`.

I also updated the package `@material-ui/core` so I can add+use the experimental package `@material-ui/lab` for Alerts

Testing instructions:
You'll need your local backend for this to work.
1. Make sure redis is empty, this includes `redis-flushall` and deleting the redis-data file.
2. Use `npm run develop` make sure the front-end is built
3. Use `npm start`, you can open up the developer tools as I added console.log messages that'll be removed when this goes in.
4. If there's a difference between the initial and the current # of posts an alert should appear at the bottom center of the screen. Keep in mind this alert only appears once and for about 6s.
5. The fun only goes on until all the posts are stored in redis, if this happens start from step1 all over again. To keep track of how many posts are in redis, use the command `redis-cli zcard t:posts`

## Checklist

### DIFF
--- a/src/frontend/package.json
+++ b/src/frontend/package.json
@@ -3,7 +3,7 @@
   "private": true,
   "dependencies": {
     "@apollo/react-hooks": "3.1.3",
-    "@material-ui/core": "4.9.8",
+    "@material-ui/core": "4.9.9",
     "@material-ui/icons": "4.9.1",
     "@material-ui/styles": "4.9.6",
     "apollo-boost": "0.4.7",
--- a/src/frontend/src/components/SnackBar/SnackBar.jsx
+++ b/src/frontend/src/components/SnackBar/SnackBar.jsx
@@ -1,0 +1,38 @@
+import React, { useState } from 'react';
+import Snackbar from '@material-ui/core/Snackbar';
+import IconButton from '@material-ui/core/IconButton';
+import CloseIcon from '@material-ui/icons/Close';
+
+export default function SimpleSnackbar() {
+  const [open, setOpen] = useState(true);
+
+  const handleClose = (event, reason) => {
+    if (reason === 'clickaway') {
+      return;
+    }
+
+    setOpen(false);
+  };
+
+  return (
+    <div>
+      <Snackbar
+        anchorOrigin={{
+          vertical: 'bottom',
+          horizontal: 'center',
+        }}
+        open={open}
+        autoHideDuration={6000}
+        onClose={handleClose}
+        message="There is new content available!"
+        action={
+          <React.Fragment>
+            <IconButton size="small" aria-label="close" color="inherit" onClick={handleClose}>
+              <CloseIcon fontSize="small" />
+            </IconButton>
+          </React.Fragment>
+        }
+      />
+    </div>
+  );
+}
--- a/src/frontend/src/pages/index.js
+++ b/src/frontend/src/pages/index.js
@@ -1,20 +1,23 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useState, useRef } from 'react';
 import PageBase from './PageBase';
 import Banner from '../components/Banner';
 import Posts from '../components/Posts';
 import ScrollToTop from '../components/ScrollToTop';
 import useSiteMetaData from '../hooks/use-site-metadata';
+import CustomizedSnackBar from '../components/SnackBar';
 
 export default function IndexPage() {
   const [numPages, setNumPages] = useState(1);
   const [posts, setPosts] = useState([]);
+  const [initNumPosts, setInitNumPosts] = useState(0);
+  const [currentNumPosts, setCurrentNumPosts] = useState(0);
   const { telescopeUrl } = useSiteMetaData();
+  const savedCallback = useRef();
 
   useEffect(() => {
     async function getPosts(pageNum = 1) {
       try {
         const res = await fetch(`${telescopeUrl}/posts?page=${pageNum}`);
-
         if (!res.ok) {
           throw new Error(res.statusText);
         }
@@ -22,7 +25,6 @@ export default function IndexPage() {
         const postsData = await Promise.all(
           postsUrls.map(async ({ url }) => {
             const tmp = await fetch(`${telescopeUrl}${url}`);
-
             if (!tmp.ok) {
               throw new Error(tmp.statusText);
             }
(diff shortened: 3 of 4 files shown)


src/backend/web/routes/posts.js
```diff
@@ -112,4 +112,8 @@ posts.get('/:id', async (req, res) => {
   }
 });
 
+posts.get('/allPosts', (req, res) => {
+  res.send(getPostsCount());
```
manekenpix, line 116: I think [this line](https://github.com/Seneca-CDOT/telescope/blob/baf12536796cda86f630844fe2e652e3b4f97e1b/src/backend/web/routes/posts.js#L58) will save you some work

c3ho: Thanks that saved me a lot

c3ho:
for some reason when I'm trying to do fetch on line `56` of `src/frontend/src/pages/index.js`. It returns null?

src/frontend/src/pages/index.js
```diff
@@ -42,14 +45,54 @@ export default function IndexPage() {
...
 
+  useEffect(() => {
+    async function getPostsCount() {
+      try {
+        const res = await fetch(`https://dev.telescope.cdot.systems/posts`);
+        // const res = await fetch(`${telescopeUrl}/posts`);
+        if (!res.ok) {
+          throw new Error(res.statusText);
+        }
+        return res.headers.get('x-total-count');
```
manekenpix, line 56: I'm not sure if this could be a problem. The way we do it in the backend is like this:
`res.set('X-Total-Count', postsCount);`
If yours fails it could because `get` is case-sensitive?

src/frontend/src/pages/index.js
```diff
@@ -6,25 +6,28 @@ import Posts from '../components/Posts';
...
+  const [currentNumPosts, setCurrentNumPosts] = useState(0);
   const { telescopeUrl } = useSiteMetaData();
 
   useEffect(() => {
     async function getPosts(pageNum = 1) {
       try {
-        const res = await fetch(`${telescopeUrl}/posts?page=${pageNum}`);
-
+        const res = await fetch(`https://dev.telescope.cdot.systems/posts?page=${pageNum}`);
+        // const res = await fetch(`${telescopeUrl}/posts?page=${pageNum}`);
```
manekenpix, line 22: I know that you're doing this here only for testing changes, but I think you can change `API_URL` in our `.env` so you don't have to commit these changes.

src/frontend/src/pages/index.js
c3ho: We both tested ytd and have verified it is case-insensitive

src/frontend/package.json
```diff
@@ -3,8 +3,9 @@
   "private": true,
   "dependencies": {
     "@apollo/react-hooks": "3.1.3",
-    "@material-ui/core": "4.9.8",
+    "@material-ui/core": "4.9.9",
     "@material-ui/icons": "4.9.1",
+    "@material-ui/lab": "4.0.0-alpha.48",
```
humphd, line 8: This makes me nervous.  Can we not stick with stable components?

src/frontend/src/components/SnackBar/SnackBar.jsx
```diff
@@ -0,0 +1,44 @@
...
+      return;
+    }
+    setOpen(false);
+  };
+
+  return (
+    <div className={classes.root}>
+      <Snackbar open={open} autoHideDuration={6000} onClose={handleClose}>
+        <Alert onClose={handleClose} severity="info">
+          There are {newPosts} new post(s) available!
```
humphd, line 39: I would just say `There are new posts available!` so you don't have to deal with the `is/are` and `post/posts` differences.  Otherwise, you'd have to make this an `if` check on the length of `newPosts` and do one thing for 1, and something else for the other cases.

src/frontend/src/pages/index.js
humphd: Agree, this needs to use the telescopeUrl

src/frontend/src/pages/index.js
```diff
@@ -40,12 +43,62 @@ export default function IndexPage() {
...
 
+  useEffect(() => {
+    async function getPostsCount() {
+      try {
+        const res = await fetch('http://localhost:3000/posts');
+        // const res = await fetch(`${telescopeUrl}/posts`);
+        if (!res.ok) {
+          throw new Error(res.statusText);
+        }
+        return res.headers.get('x-total-count');
```
humphd, line 54: If all you want is this header, you can change your `fetch()` so it does a `HEAD` request:

```js
fetch(url, {method: 'HEAD'})
```

We should confirm that our backend will support this, but it would be a lot lighter weight.

src/frontend/src/pages/index.js
```diff
@@ -40,12 +43,62 @@ export default function IndexPage() {
...
+    // Polls every minute, this timer starts as soon as the page is loaded
+    const interval = setInterval(async () => {
+      await setCurrentNumPosts(await getPostsCount());
+      console.log(currentNumPosts);
+      if (currentNumPosts !== initNumPosts) {
+        console.log('Not the same!');
+      } else {
+        console.log('Same!');
+      }
+    }, 5000);
```
humphd, line 85: This is probably happening too often.  If I'm on mobile, I don't want to send this much traffic.  I'd probably do once per minute, or more.

src/frontend/package.json
c3ho: > I also updated the package @material-ui/core so I can add+use the experimental package @material-ui/lab for Alerts

src/frontend/src/components/SnackBar/SnackBar.jsx
c3ho: Fixed, I was just using it for testing purposes

src/frontend/src/pages/index.js
c3ho: I was thinking every 5 mins or so, but for testing purposes I set it as 5s

src/frontend/src/pages/index.js
c3ho: Thanks for the suggestion! Confirmed it works.

src/frontend/src/pages/index.js
c3ho: Fixed

src/frontend/package.json
humphd: What I'm saying is, I don't think we should be using this:

> This package hosts the incubator components that are not yet ready to move to core

If it's not ready for core, why are we taking it in our tree?  Anything that isn't stable is necessarily unstable, and changing, which means APIs will break.

Why can't we use `import Snackbar from '@material-ui/core/Snackbar';`?

src/frontend/src/components/SnackBar/SnackBar.jsx
```diff
@@ -0,0 +1,43 @@
...
+      return;
+    }
+    setOpen(false);
+  };
+
+  return (
+    <div className={classes.root}>
+      <Snackbar open={open} autoHideDuration={6000} onClose={handleClose}>
+        <Alert onClose={handleClose} severity="info">
+          There are new post(s) available!
```
humphd, line 38: What if we say something more generic:

`There is new content available!`

This equally true of 1 or more posts.

src/frontend/src/pages/index.js
```diff
@@ -40,12 +42,67 @@ export default function IndexPage() {
...
+        console.log({ error });
+      }
+    }
+    setPostsInfo();
+  }, []);
+
+  useEffect(() => {
+    function getCurrentNumPosts() {
+      savedCallback.current();
+      console.log(currentNumPosts);
```
humphd, line 86: Remove this.

src/frontend/src/pages/index.js
```diff
@@ -40,12 +42,67 @@ export default function IndexPage() {
...
+    }
+
+    savedCallback.current = callback;
+    // Polls every 5 minute
+    const interval = setInterval(getCurrentNumPosts, 300000);
```
humphd, line 91: Instead of this, do `5 * 60 * 1000` so that it's more obvious how you arrived at it.

src/frontend/src/pages/index.js
```diff
@@ -40,12 +42,67 @@ export default function IndexPage() {
...
+  }, []);
+
+  useEffect(() => {
+    function getCurrentNumPosts() {
+      savedCallback.current();
+      console.log(currentNumPosts);
+    }
+
+    savedCallback.current = callback;
+    // Polls every 5 minute
```
humphd, line 90: minutes (plural)

src/frontend/src/components/SnackBar/SnackBar.jsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import Snackbar from '@material-ui/core/Snackbar';
+import IconButton from '@material-ui/core/IconButton';
+import CloseIcon from '@material-ui/icons/Close';
+
+export default function SimpleSnackbar() {
+  const [open, setOpen] = React.useState(true);
```
humphd, line 7: Most of our code does `userState(true)` vs. `React.useState(true)`.  We should clean this up in another PR to make them all the same.  This is fine for now.

c3ho:
Will merge as soon as it passes tests as it received two approvals

Issue #839, opened 2020-03-16
Search Bar submit button + onClick()


c3ho:
**What would you like to be added**:
#792 should hopefully go in soon. It doesn't have a search button, but dynamically returns results based on user input. I made it this way currently because I'm unable to prevent the onClick function from triggering non stop after the initial user click

Here's some code snippets that may speed up the process to resolve the issue:

I removed the following code from `SearchBar.jsx`:
line 90
```
        <IconButton
          type="submit"
          className={classes.iconButton}
          aria-label="search"
          onClick={event => handleClick(event)}
        >
          <SearchIcon />
        </IconButton>
```

Code for `SearchPage.jsx`:
```
  function onClickHandler() {
    executeSearch();
  }
```
line 103
```
      <SearchBar
        onClickHandler={onClickHandler}
        searchText={searchText}
        onChangeHandler={onChangeHandler}
      />
```

**Why would you like this to be added**:
Up for debate, should we search upon user input or when the user clicks on button? If this is not a concern, we can close this issue.

humphd:
This really needs to happen in the future, because we waste so much bandwidth searching and pulling in posts while you're typing.  We shouldn't do anything until the user tells us they are done.

raygervais:
Been doing quite a bit of DevOps style items for the past two weeks at work etc, plus having issues with #1008 on the Windows Testing side so I'm going to hash out this one as a weekend-long break. Will return to 1008 after.

PR #937, opened 2020-04-05
Finish Search Feature

### LINKED ISSUE #911: Finish Search feature
We have the beginnings of a search feature, and a backend ready to support it, but right now the search returns no results for me.

I'm also not clear why we have `None` selected by default in the drop-down, and what `Filter2` and `Filter3` do.

We need to connect all the pieces and polish this so that it's usable for 1.0.

c3ho:
## Issue This PR Addresses
Fixes #911 

## Type of Change

## Description
Finishing up missing search functionalities:
- adding a spinner while waiting for results to load (Done)
- adding a way to search posts (Done)
- dropdown working properly and search responds to selected option (Done)

To test head on over to `{telescopeUrl}/search` and type your query in. Both options in the dropdown should work, when searching by authors, you'll get a spinner. I'm not sure if we'll need one for Posts

## Checklist

### DIFF
--- a/src/frontend/src/components/SearchBar/SearchBar.jsx
+++ b/src/frontend/src/components/SearchBar/SearchBar.jsx
@@ -4,6 +4,7 @@ import Paper from '@material-ui/core/Paper';
 import InputBase from '@material-ui/core/InputBase';
 import FormControl from '@material-ui/core/FormControl';
 import NativeSelect from '@material-ui/core/NativeSelect';
+import PropTypes from 'prop-types';
 
 const useStyles = makeStyles((theme) => ({
   root: {
@@ -39,20 +40,14 @@ const useStyles = makeStyles((theme) => ({
 
 function CustomizedInputBase(props) {
   const classes = useStyles();
-  const [state, setState] = React.useState({
-    filter: '',
-  });
-  const { searchText, onChangeHandler } = props;
+  const { searchText, onChangeHandler, onFilterChangeHandler, filter } = props;
 
-  const handleChange = (filter) => (event) => {
-    setState({
-      ...state,
-      [filter]: event.target.value,
-    });
+  const onFilterChange = (event) => {
+    onFilterChangeHandler(event.target.value);
   };
 
   const onTextChange = (event) => {
-    onChangeHandler(event);
+    onChangeHandler(event.target.value);
   };
 
   return (
@@ -61,16 +56,16 @@ function CustomizedInputBase(props) {
         <FormControl className={classes.formControl}>
           <NativeSelect
             disableUnderline
-            value={state.filter}
-            onChange={handleChange('filter')}
+            value={filter}
+            onChange={(event) => onFilterChange(event)}
             name="filter"
             className={classes.selectEmpty}
             inputProps={{ 'aria-label': 'filter' }}
           >
-            <option value="">None</option>
             <option value="author">Author</option>
-            <option value="filter2">Filter2</option>
-            <option value="filter3">Filter3</option>
+            <option selected value="post">
+              Post
+            </option>
           </NativeSelect>
         </FormControl>
         <InputBase
@@ -85,4 +80,10 @@ function CustomizedInputBase(props) {
   );
 }
 
+CustomizedInputBase.propTypes = {
+  searchText: PropTypes.string,
+  onChangeHandler: PropTypes.func,
+  onFilterChangeHandler: PropTypes.func,
+  filter: PropTypes.string,
+};
 export default CustomizedInputBase;
--- a/src/frontend/src/components/SearchPage/SearchPage.jsx
+++ b/src/frontend/src/components/SearchPage/SearchPage.jsx
@@ -5,6 +5,9 @@ import Box from '@material-ui/core/Box';
 import gql from 'graphql-tag';
 import SearchBar from '../SearchBar';
 import AuthorResult from '../AuthorResult';
+import Spinner from '../Spinner';
+import Posts from '../Posts';
+import useSiteMetadata from '../../hooks/use-site-metadata';
 
 const useStyles = makeStyles(() => ({
   searchReply: {
@@ -17,9 +20,14 @@ const useStyles = makeStyles(() => ({
     minHeight: '12em',
     display: 'flex',
   },
+  spinner: {
+    display: 'flex',
+    justifyContent: 'center',
+  },
 }));
 
 const SearchPage = () => {
+  const { telescopeUrl } = useSiteMetadata();
   const classes = useStyles();
   const SEARCH_QUERY = gql`
     query SearchAuthorQuery($author: String!) {
@@ -36,15 +44,16 @@ const SearchPage = () => {
   `;
 
   const [searchText, setSearchText] = useState('');
-  const [results, setResults] = useState([]);
-
-  const [executeSearch, { data }] = useLazyQuery(SEARCH_QUERY, {
-    // setting author variable here to use in the query above
+  const [results, setResults] = useState(undefined);
+  const [filter, setFilter] = useState('post');
+  const [fetchLoading, setFetchLoading] = useState(false);
+  const [searchAuthors, { loading, data }] = useLazyQuery(SEARCH_QUERY, {
+    // Setting author variable here to use in the query above
     variables: { author: searchText },
-
     onCompleted: () => {
-      // We only want to display a feed once, this would avoid the issue if
-      // there are multiple posts by the same author returned in our query
+      // We only want to display a feed once when searching posts for
+      // author, this would avoid the issue if there are duplicate
+      // feed results returned in our query
       const finalResults = data.getPosts
         .filter(
           (element, index, self) =>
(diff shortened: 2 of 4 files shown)


c3ho:
Several concerns, 

Should I break off the searches into their own components or something? The useLazyQuery is a react hook as well and I can't use it conditionally depending on the `filter` chosen. Even more so because the second type of query is an API request to the route `/search`

For some reason while I'm doing `console.log(filterValue)` I keep getting the previous selected one instead of the current one.

Third and last one the spinner. I got it to work. However, if you test it [here](https://telescope-m90xb7vgk.now.sh/), there's one scenario where the spinner will continuously run:
1. Type `jos` into the search input, you should get two results. Wait 2/3s
2. Delete or add an extra character so that the search string becomes `jo` or `josu`
3. You should get a spinner that doesn't stop spinning

src/frontend/src/components/SearchBar/SearchBar.jsx
```diff
@@ -39,16 +39,12 @@ const useStyles = makeStyles((theme) => ({
 
 function CustomizedInputBase(props) {
   const classes = useStyles();
-  const [state, setState] = React.useState({
-    filter: '',
-  });
-  const { searchText, onChangeHandler } = props;
+  const [selectedFilter, setSelectedFilter] = useState(props.filter);
```
humphd, line 42: This is an anit-pattern: you don't mix props and state.  If we're receiving this on props from our parent, we just use that as is, without putting into state here.

src/frontend/src/components/SearchBar/SearchBar.jsx
```diff
@@ -39,16 +39,12 @@ const useStyles = makeStyles((theme) => ({
...
+  const { searchText, onChangeHandler, onFilterChangeHandler } = props;
```
humphd, line 43: Same thing here.

src/frontend/src/components/SearchBar/SearchBar.jsx
```diff
@@ -39,16 +39,12 @@ const useStyles = makeStyles((theme) => ({
...
 
-  const handleChange = (filter) => (event) => {
-    setState({
-      ...state,
-      [filter]: event.target.value,
-    });
+  const onFilterChange = (event) => {
```
humphd, line 45: This needs to call the function props we've received from above, and pass up the value in our input control.

src/frontend/src/components/SearchBar/SearchBar.jsx
```diff
@@ -61,16 +57,14 @@ function CustomizedInputBase(props) {
...
-            value={state.filter}
-            onChange={handleChange('filter')}
+            value={selectedFilter}
+            onChange={(event) => onFilterChange(event)}
             name="filter"
             className={classes.selectEmpty}
             inputProps={{ 'aria-label': 'filter' }}
           >
-            <option value="">None</option>
             <option value="author">Author</option>
```
humphd, line 66: I don't think this should be the default value

src/frontend/src/components/SearchBar/SearchBar.jsx
```diff
@@ -61,16 +57,14 @@ function CustomizedInputBase(props) {
...
-            <option value="filter2">Filter2</option>
-            <option value="filter3">Filter3</option>
+            <option value="post">Post</option>
```
humphd, line 67: Let's make this `selected`.

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -36,15 +40,16 @@ const SearchPage = () => {
...
-  const [executeSearch, { data }] = useLazyQuery(SEARCH_QUERY, {
-    // setting author variable here to use in the query above
+  const [executeSearch, { loading, data }] = useLazyQuery(SEARCH_QUERY, {
+    // Setting author variable here to use in the query above
     variables: { author: searchText },
-
     onCompleted: () => {
       // We only want to display a feed once, this would avoid the issue if
       // there are multiple posts by the same author returned in our query
+      setResults(undefined);
```
humphd, line 52: I wonder if this is what is causing the brief "No results" text to flash before we see the actual results.  Why do you bother with this?

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -36,15 +40,16 @@ const SearchPage = () => {
...
       const finalResults = data.getPosts
         .filter(
```
humphd, line 54: I don't understand this filter.  Can you comment above it?

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -62,38 +67,74 @@ const SearchPage = () => {
...
+      setResults({ type: 'author', searchResults: finalResults });
     },
   });
 
   // Hook that will re-render the page only if the state of searchText changes
   useEffect(() => {
-    executeSearch();
-  }, [searchText]);
+    async function searchPosts() {
+      const res = await fetch(`${telescopeUrl}/query?search=${searchText}`);
```
humphd, line 77: Error handling.

humphd: Also, you need to `encodeURIComponent(searchText)`

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -62,38 +67,74 @@ const SearchPage = () => {
...
+      const searchResults = await res.json();
```
humphd, line 78: Error handling.

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -62,38 +67,74 @@ const SearchPage = () => {
...
+      const postIds = searchResults.values.map((result) => result.id);
+      let posts = await Promise.all(postIds.map((id) => fetch(`${telescopeUrl}/posts/${id}`)));
```
humphd, line 80: combine these two lines (79 and 80), you only need to walk this array once to get the id and look up the result.  Do we not return the URL to use in our JSON results already?

humphd: Also, error handling on the fetch and json calls.

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -62,38 +67,74 @@ const SearchPage = () => {
...
+    if (filter === 'author') executeSearch();
+    else if (filter === 'post') searchPosts();
+  }, [searchText, filter, telescopeUrl, executeSearch]);
 
   // Displays one of three options depending on whether there is a search string, results and no results
   const displayResults = () => {
-    if (searchText.length === 0 && results.length < 1) {
+    if (searchText.length > 0 && loading) {
+      return (
+        <h1 className={classes.searchReply}>
```
humphd, line 93: We need better CSS on this to at least get it in the centre

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -62,38 +67,74 @@ const SearchPage = () => {
...
+          <Spinner />
+        </h1>
+      );
+    }
+
+    if (searchText.length === 0) {
```
humphd, line 99: This is an odd thing to do.  It's not an error to not enter text.  Why are we showing this?

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -62,38 +67,74 @@ const SearchPage = () => {
...
-    //  The initial state of results is going to be 0 and we can't map 0,
-    //  this will check if there are any results and return AuthorResult component
-    //  for each with feed guid as key
-    if (results.length > 0)
-      return results.map((result) => (
-        <AuthorResult key={result.id} author={result.author} post={result.post} />
-      ));
-    return <h1 className={classes.searchReply}>No results found</h1>;
+
+    if (results) {
```
humphd, line 103:
```js
if(!results) {
  return <h1 className={classes.searchReply}>No results found</h1>;
}
```

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -62,38 +67,74 @@ const SearchPage = () => {
...
+      //  If result type is author return AuthorResult component
+      //  for each with feed guid as key
+      if (results.type === 'author' && results.searchResults.length > 0) {
+        return results.searchResults.map((result) => (
+          <AuthorResult key={result.id} author={result.author} post={result.post} />
```
humphd, line 108: Personally, I don't see the value in the `AuthorResult`.  Why don't we just show you all the `<Post>` components for this user?  To me, that would be more logical.

Probably in follow-up we should do an auto-complete for Author, so you start typing and it shows you the people who have in our system and you choose one.  That would make this less necessary.

src/frontend/src/components/SearchPage/SearchPage.jsx
c3ho: I guess the most relevant example would be right now where Krystyna has two posts back to back, if we were to display the results without doing this filter, we'd have two <AuthorResult /> component with Krystyna as opposed to one. We only want unique feed results here.

src/frontend/src/components/SearchPage/SearchPage.jsx
c3ho: I checked with express when creating the route before, it will automatically do this

humphd: This has nothing to do with express.  It will automatically decode it, but you need to create a proper URL.  If I enter "open source" and you put that on the search params, it's not a valid URL any more (you need to encode to "open%20source" or "open+source").

src/frontend/src/components/SearchPage/SearchPage.jsx
c3ho: I think this was asked for in a previous PR by @raygervais

src/frontend/src/components/SearchPage/SearchPage.jsx
c3ho: Removed, I think it was left there when I was trying to find ways to get the spinner to render

src/frontend/src/components/SearchBar/SearchBar.jsx
c3ho: I think this was an old commit as I removed it about 2/3 commits ago

src/frontend/src/components/SearchPage/SearchPage.jsx
c3ho: done

src/frontend/src/components/SearchBar/SearchBar.jsx
c3ho: so don't use useState, and just use props? I removed all instances of assigning and states in `SearchBar.jsx` so now it looks like the below:

image

src/frontend/src/components/SearchBar/SearchBar.jsx
c3ho: Fixed

src/frontend/src/components/SearchPage/SearchPage.jsx
c3ho: Fixed

src/frontend/src/components/SearchPage/SearchPage.jsx
c3ho: so something like 
```
let postsInfo = await Promise.all(postIds.map(async (id) => {
  let post = await fetch(`https://dev.telescope.cdot.systems/posts/${id}`)
  return post.json();
}
```

src/frontend/src/components/SearchBar/SearchBar.jsx
c3ho: Is this to set the default value? If it is, I tried it, but it was giving me issues. I instead set the initialState in `SearchPage.js` using the following line `const [filter, setFilter] = useState('author');`

src/frontend/src/components/SearchBar/SearchBar.jsx
humphd: Red underlines don't tell me anything.  What is it complaining about when you hover?  Are you not using these values anywhere?

src/frontend/src/components/SearchBar/SearchBar.jsx
humphd: Yeah, for the default. Whatever works for making it a default.

src/frontend/src/components/SearchBar/SearchBar.jsx
c3ho: Nono, I didn't mean it like it was erroring. I mean after your requested changes it should look like this without any `useState` right?

src/frontend/src/components/SearchBar/SearchBar.jsx
c3ho: Got it, default to posts

src/frontend/src/components/SearchPage/SearchPage.jsx
c3ho: Ah, also if you're trying to search Posts, the route we're using will only work when searching `strings.length > 3`. This should explain why you're seeing the no results as you're typing.

src/frontend/src/components/SearchPage/SearchPage.jsx
c3ho: It is now in the center

cindyledev:
Tested this locally on Windows using Docker and the spinner is a very nice touch :)

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -62,38 +71,107 @@ const SearchPage = () => {
...
     },
   });
 
   // Hook that will re-render the page only if the state of searchText changes
   useEffect(() => {
-    executeSearch();
-  }, [searchText]);
+    async function searchPosts() {
+      // Setting to check length > 3, as the component will trying to fetch with 0 length causing error
+      if (searchText.length > 3) {
```
humphd, line 82: In another bug I want to add the Search button, since the way it gobbles network bandwidth to do searches while you type is no good on mobile.  But this is fine for now.

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -62,38 +71,107 @@ const SearchPage = () => {
...
+        const encodedSearchText = encodeURIComponent(searchText);
```
humphd, line 83: Thank you!

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -62,38 +71,107 @@ const SearchPage = () => {
...
+              }
+              contentType = res.headers.get('content-type');
+              if (!contentType || !contentType.includes('application/json')) {
+                throw new Error('Response was not json');
+              }
+              return res.json();
+            })
+          );
+          setFetchLoading(false);
+          await setResults({ type: 'post', searchResults: posts });
```
humphd, line 111: Remove `await`

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -62,38 +71,107 @@ const SearchPage = () => {
...
+        } catch (error) {
+          console.log('Something went wrong while fetching data', error);
+        }
+      } else {
+        setResults({ type: 'post', searchResults: [] });
+      }
+    }
+
+    if (filter === 'author') {
+      executeSearch();
```
humphd, line 121: Would `searchAuthors()` be a more logical name for this, given `searchPosts()` below?

src/frontend/src/components/SearchPage/SearchPage.jsx
c3ho: There's a PR up for that, #839

src/frontend/src/components/SearchPage/SearchPage.jsx
c3ho: removed

src/frontend/src/components/SearchPage/SearchPage.jsx
c3ho: renamed

c3ho:
Merging after it passes as it received two approvals

PR #943, opened 2020-04-06
Logged In state for Header 

### LINKED ISSUE #889: Logged In state for Header needs to be updated
**What happened**:
After #882 landed, we are now able to log in AND log out and after #877, we have display names for users who successfully signed in. However, the Header doesn't display them nicely.

Screenshot_2020-03-28 Home

Please fix so that Username and LOGOUT lines up with the rest of the other links in the Header. You can probably use [MUI Divider](https://material-ui.com/components/dividers/#vertical-dividers) instead of `|`. Remove the word "Welcome" as it makes the whole thing unnecessarily long and remove the unavatar.

Using Docker:
- `sudo systemctl start docker` in Linux or run Docker Desktop for Windows or Mac
- `docker-compose up --build redis login elasticsearch`
- `npm start`
- visit `localhost:3000/auth/login` (LOGIN button navigates to `https://dev.telescope.cdot.systems/auth/login`)
- sign in using `Username: user1` `Password: user1pass`

You reproduce this using native redis and elasticsearch too

This issue will likely touch one or more of the following files:
- [LoggedIn](https://github.com/Seneca-CDOT/telescope/blob/master/src/frontend/src/components/Login/LoggedIn.jsx)
- [LoggedOut](https://github.com/Seneca-CDOT/telescope/blob/master/src/frontend/src/components/Login/LoggedOut.jsx)
- [Header](https://github.com/Seneca-CDOT/telescope/blob/master/src/frontend/src/components/Header/Header.jsx)

**Environment**:
- OS: All
- Browser: All

agarcia-caicedo:
## Issue This PR Addresses
fixes #889 

## Type of Change

## Description

This styles the header depending on the state of the app: Logged in vs Logged out

## Checklist


cindyledev:
@humphd @manekenpix  I tried playing around with this PR for fun and I noticed that `localhost:3000` doesn't hot reload so I have keep rebuilding the frontend with `npm clean` and `npm build` every single time I make a small change to the `LoggedIn` component. Is `localhost:3000`'s inability to hot reload normal? If it is, can I get the `LoggedIn` component on `localhost:8000` somehow?

I'm using Docker for Windows btw

humphd:
For the case where you care about authentication, you have to do the full build right now.  It's hard to have the frontend be different from the backend origin with the way authentication works.

When you don't care about authentication, you can use `npm start` in one tab, and `npm run develop` in another, and use localhost:8000 for the frontend.

We might be able to cook up something in a development setting where we proxy :3000/ to :8000/.

humphd:
I wrote a proxy fix in #949

src/frontend/src/components/Header/Header.jsx
```diff
@@ -1,19 +1,61 @@
...
 import useMediaQuery from '@material-ui/core/useMediaQuery';
+import useSiteMetadata from '../../hooks/use-site-metadata';
 
 import MobileHeader from './MobileHeader.jsx';
 import DesktopHeader from './DesktopHeader.jsx';
 
 function Header() {
   const theme = useTheme();
   const matches = useMediaQuery(theme.breakpoints.down('sm'));
+  const LoggedIn = Login();
```
Grommers00, line 12: this line is giving me an error.

src/frontend/src/components/Header/Header.jsx
```diff
@@ -1,19 +1,61 @@
...
+        }
+      } catch (error) {
+        console.error('Error getting user info', error);
+      }
+    }
+
+    getUserInfo();
+  }, [telescopeUrl]);
+
+  return email ? { name } : false;
```
cindyledev, line 58: I think this `return email ? {name} : false; ` might be a problem too

src/frontend/src/components/Header/MobileHeader.jsx
```diff
@@ -137,14 +170,25 @@ const MobileHeader = () => {
...
           </IconButton>
-          <Drawer
-            classes={{ paper: classes.paper }}
-            anchor="right"
-            open={state.right}
-            onClose={toggleDrawer('right', false)}
-          >
-            {sideList('right')}
-          </Drawer>
+          {username === false ? (
```
humphd, line 173: Instead of `username === false ? ` flip the logic and use `username ?`  I don't think the username will be a pure boolean false, will it?  Rely on it being truthy instead.

src/frontend/src/components/Header/DesktopHeader.jsx
```diff
@@ -109,6 +111,32 @@ const Header = () => {
...
+        </ListItem>
+        <Divider className={classes.line} />
+        <div className={classes.footer}>
+          <Footer />
+        </div>
+      </List>
+    </div>
+  );
+
+  console.log('this is a ' + username);
```
humphd, line 138: Remove

src/frontend/src/components/Header/DesktopHeader.jsx
```diff
@@ -126,7 +154,7 @@ const Header = () => {
               Home
             </Link>
           </Button>
-          <Login />
+          {username === false ? <LoggedOut /> : <Link to="/myfeeds">{username}</Link>}
```
humphd, line 157: Same deal here, flip it: `{username ? (<Link to="/myfeeds">{username}</Link>) : <LoggedOut />}`

src/frontend/src/components/Header/Header.jsx
```diff
@@ -4,14 +4,19 @@ import useMediaQuery from '@material-ui/core/useMediaQuery';
...
   const theme = useTheme();
   const matches = useMediaQuery(theme.breakpoints.down('sm'));
 
   return (
     <div>
-      {matches ? <MobileHeader className="mobile-header" /> : <DesktopHeader className="header" />}
+      {matches ? (
+        <MobileHeader className="mobile-header" name={Login()} />
+      ) : (
+        <DesktopHeader className="header" name={Login()} />
```
humphd, line 18: What's happening here with `Login()`?  You are trying to send this component down on props?  Why not have your components require this in directly?

agarcia-caicedo: Made the changes you suggested

src/frontend/src/components/Header/Login.jsx
```diff
@@ -41,7 +34,7 @@ function Login() {
     getUserInfo();
   }, [telescopeUrl]);
 
-  return email ? <LoggedIn name={name} /> : <LoggedOut />;
+  return email ? { name } : false;
```
humphd, line 37: I wonder if we can do something else with this function.  You're calling it like a function, but it's written like a component, with a React lifecycle and hooks.

There's a new API in React for doing what you want to do here, called [Context](https://reactjs.org/docs/context.html).  We need a way to globally manage this `user` info across all of our components, and broadcast changes.  We could pass it down from `PageBase` all the way to every other component via props.  But I think Context might be a better fit.

Let's talk tomorrow on Slack to figure out how to move forward with this.  I don't want to expand the scope so much that we can't land this in time for 1.0.

humphd: Let's discuss this on Slack soon, when everyone is around, and make a plan.  @c3ho is suggesting to we try and do the Context piece in this PR.  If we need to do that, some of us could help make it happen.

Issue #954, opened 2020-04-07
REST API routes for a User to add, delete, and list their feeds


humphd:
We need REST API endpoints to support and use the work in #919 and #717.  @Silvyre and I discussed an API, and he's going to do the tests for it, while @Grommers00 says he'll do the implementation with @manekenpix's help.  I suspect that @c3ho and I will also be involved, at least in reviews.

Here is our proposed API:

1. A user needs to be able to get his/her feeds
1. A user needs to be able to add a new feed
1. A user needs to be able to delete one of their feeds

To support that, we propose:

1. `GET /user/feeds`: allow an authenticated user to get his/her feeds.  Should return an Array identical to what `/feeds` returns, but limited to the feeds owned by this user (i.e., `[{id: 1234, url: '/feeds/1234'}, ...]`.  If the user is NOT logged in, you get back a 403.  If the authenticated user has no feeds (they own none), return the empty array.  NOTE: we are thinking to do this on `/user` vs. `/feeds` so we don't break our current API, and trigger a big rewrite.
1. `DELETE /feeds/:id`: allow an authenticated user to delete his/her feed.  It needs to check if the user who is logged in is the same user as the feed being deleted.  If the user is not logged in, or is not the same user, it should return a 403.  Otherwise, it should do the delete and return a 204.

We've already got code to `POST /feeds` to create a new feed, so that's taken care of.

PR #958, opened 2020-04-08
Fixes #955: add tests for GET /user/feeds, DELETE /feeds/:id

### LINKED ISSUE #955: Tests for REST API for a User to add, delete, list their feeds
This is connected to #954, and relates to writing the tests, which @Silvyre has agreed to do.  We can land these and use `skip()` until the implementation is done. cc @Grommers00, @manekenpix, @c3ho.

These tests need to do (at least) the following

## `GET /user/feeds`

1. `GET /user/feeds` should give a 403 when the user is not logged in
1. `GET /user/feeds` should give a 200 when the user is logged in, and should receive a JSON Array (possibly empty)
1. Create a new Feed Object, set the user to the currently logged in user (I need to fix our test code to support this), save it, login, `GET /user/feeds` and make sure the resulting Array includes this newly created Feed object.

## `DELETE /feeds/:id`

1. `DELETE /feeds:id` when not logged in should return a 403 (i.e., you must be logged in)
1. `DELETE /feeds:id` when logged in as user1 should return a 403 if the feed at `id` is owned by a different user (i.e., users can't delete other user's feeds)
1. `DELETE /feeds:id` when logged in as user1 should return a 204 if the feed at `id` is owned by this user (i.e., users can delete their own feeds)

Silvyre:
## Issue This PR Addresses

Fixes #955

## Type of Change

## Description

Adds a _**skipped**_ test for each of the six scenarios outlined in #955 (concerning testing the `GET /user/feeds` and `DELETE /feeds/:id` API routes currently being developed by @Grommers00 and @manekenpix in https://github.com/Seneca-CDOT/telescope/issues/954).

All of these tests were appended to the existing [`feeds.test.js`](https://github.com/Seneca-CDOT/telescope/blob/master/test/feeds.test.js) file.

Once https://github.com/Seneca-CDOT/telescope/issues/954 lands, these tests should be unskipped.

## Checklist

### DIFF
--- a/test/feeds.test.js
+++ b/test/feeds.test.js
@@ -134,48 +134,38 @@ describe('test POST /feeds endpoint', () => {
 });
 
 describe.skip('test DELETE /feeds/:id endpoint', () => {
-  let feedData;
+  let feedId;
 
-  beforeAll(() => {
+  beforeEach(async () => {
+    logout();
     const user = login('Johannes Kepler', 'user1@example.com');
-    feedData = {
+    const feedData = {
       author: user.name,
       url: 'http://telescope200.cdot.systems',
       user: user.id,
     };
-  });
-
-  beforeEach(() => {
-    logout();
-    login('Johannes Kepler', 'user1@example.com');
+    const res = await request(app).post('/feeds').send(feedData).set('Accept', 'application/json');
+    expect(res.status).toEqual(201);
+    feedId = res.id;
   });
 
   afterAll(() => logout());
 
   it('should respond with a 403 status when not logged in', async () => {
-    let res = await request(app).post('/feeds').send(feedData).set('Accept', 'application/json');
-    const feedId = res.id;
-
     logout();
-    res = await request(app).delete(`/feeds/${feedId}`);
+    const res = await request(app).delete(`/feeds/${feedId}`);
     expect(res.status).toEqual(403);
   });
 
   it('should respond with a 403 status when targeted feed is not owned by user', async () => {
-    let res = await request(app).post('/feeds').send(feedData).set('Accept', 'application/json');
-    const feedId = res.id;
-
     logout();
     login('Galileo Galilei', 'user2@example.com');
-    res = await request(app).delete(`/feeds/${feedId}`);
+    const res = await request(app).delete(`/feeds/${feedId}`);
     expect(res.status).toEqual(403);
   });
 
   it('should respond with a 204 status when targeted feed is owned by user', async () => {
-    let res = await request(app).post('/feeds').send(feedData).set('Accept', 'application/json');
-    const feedId = res.id;
-
-    res = await request(app).delete(`/feeds/${feedId}`);
+    let res = await request(app).delete(`/feeds/${feedId}`);
     expect(res.status).toEqual(204);
 
     res = await request(app).get(`/feeds/${feedId}`);
--- a/test/feeds.test.js
+++ b/test/feeds.test.js
@@ -134,50 +134,15 @@ describe('test POST /feeds endpoint', () => {
 });
 
 describe.skip('test DELETE /feeds/:id endpoint', () => {
-  let feedId;
+  let feedData;
 
-  beforeEach(async () => {
-    logout();
+  beforeAll(() => {
     const user = login('Johannes Kepler', 'user1@example.com');
-
-    const feedData = {
+    feedData = {
       author: user.name,
       url: 'http://telescope200.cdot.systems',
       user: user.id,
     };
-    const res = await request(app).post('/feeds').send(feedData).set('Accept', 'application/json');
-    feedId = res.body.id;
-  });
-
-  afterAll(() => logout());
-
-  it('should respond with a 403 status when not logged in', async () => {
-    logout();
-    const res = await request(app).delete(`/feeds/${feedId}`);
-    expect(res.status).toEqual(403);
-  });
-
-  it('should respond with a 403 status when targeted feed is not owned by user', async () => {
-    logout();
-    login('Galileo Galilei', 'user2@example.com');
-    const res = await request(app).delete(`/feeds/${feedId}`);
-    expect(res.status).toEqual(403);
-  });
-
-  it('should respond with a 204 status when targeted feed is owned by user', async () => {
-    let res = await request(app).delete(`/feeds/${feedId}`);
-    expect(res.status).toEqual(204);
-
-    res = await request(app).get(`/feeds/${feedId}`);
-    expect(res.status).toEqual(404);
-  });
-});
-
-describe.skip('test GET /user/feeds endpoint', () => {
-  let user;
-
-  beforeAll(() => {
-    user = login('Johannes Kepler', 'user1@example.com');
   });
 
   beforeEach(() => {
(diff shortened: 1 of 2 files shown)


test/feeds.test.js
```diff
@@ -132,3 +132,53 @@ describe('test POST /feeds endpoint', () => {
...
+
+  beforeEach(() => {
+    logout();
+    login('Johannes Kepler', 'user1@example.com');
+  });
+
+  afterAll(() => logout());
+
+  it('should respond with a 403 status when not logged in', async () => {
+    let res = await request(app).post('/feeds').send(feedData).set('Accept', 'application/json');
```
humphd, line 156: Everyone of these tests does this same thing at the start, so let's move this to the `beforeEach` above, and also add a line right after this to confirm that your feed got added:

```js
    expect(res.status).toEqual(201);
```

test/user.test.js
```diff
@@ -0,0 +1,57 @@
...
+// Mock the internal authentication strategy
+jest.mock('../src/backend/web/authentication');
+// Use our authentication test helper
+const { login, logout } = require('./lib/authentication');
+
+describe.skip('test GET /user/feeds endpoint', () => {
+  let user;
+
+  beforeAll(() => {
+    user = login('Johannes Kepler', 'user1@example.com');
```
humphd, line 16: Why bother doing this if you're going to always logout on 20?

test/feeds.test.js
Silvyre: >Everyone of these tests does this same thing at the start, so let's move this to the `beforeEach` above

I actually originally had this fetch in `beforeEach()` within my first commit (https://github.com/Seneca-CDOT/telescope/pull/958/commits/210ec80c11c6c76f3974d37799f229fb566499d3):

https://github.com/Seneca-CDOT/telescope/blob/210ec80c11c6c76f3974d37799f229fb566499d3/test/feeds.test.js#L136-L150

However, couldn't this potentially lead to a race condition? As can be seen above, enclosing this call to fetch in the `beforeEach` would necessitate exposing a `feedId` to each test function (L137 in the above code snippet). This could lead to a race condition between the three test functions accessing this thrice-modified variable. Thoughts?

test/user.test.js
Silvyre: I did this to ensure all test functions can access reference user1's information (via the `user` variable on L13).

If I removed this `beforeAll()` call, I still would have to call `login()` within the last test function (even though the user is already logged in) in order to get the user information I need to build [the `feedData` object](https://github.com/Seneca-CDOT/telescope/pull/958/files#diff-a202d1e6aac601eca4c15ad2418135fdR42).

Silvyre: I did this to ensure all test functions can access reference user1's information (via the `user` variable on L13).

If I removed this `beforeAll()` call, I still would have to call `login()` within the last test function (even though the user is already logged in) in order to get the user information I need to build [the `feedData` object](https://github.com/Seneca-CDOT/telescope/pull/958/files#diff-a202d1e6aac601eca4c15ad2418135fdR42).

test/feeds.test.js
humphd: The test cases in a `describe` always run one-by-one, no races.  Test files can be run in parallel in multiple processes.

Silvyre: Wow, I did not know that!

Requested changes made in https://github.com/Seneca-CDOT/telescope/pull/958/commits/5d56f1221e6c3c1d2d0424bec9051ea9b54161fe

test/feeds.test.js
```diff
@@ -132,3 +132,43 @@ describe('test POST /feeds endpoint', () => {
...
+
+  it('should respond with a 403 status when targeted feed is not owned by user', async () => {
+    logout();
+    login('Galileo Galilei', 'user2@example.com');
+    const res = await request(app).delete(`/feeds/${feedId}`);
+    expect(res.status).toEqual(403);
+  });
+
+  it('should respond with a 204 status when targeted feed is owned by user', async () => {
+    let res = await request(app).delete(`/feeds/${feedId}`);
```
humphd, line 168: Let's explicitly do the login again as the correct user, so the test is clear that it's expecting that user to be the current user (your test is right now, but it would be more future-proof to add it here):

```js
const user = login('Johannes Kepler', 'user1@example.com');
```

test/user.test.js
humphd, line 16: Let's remove this `beforeAll()` and capture the user in line 21 below.  This isn't needed.

test/user.test.js
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+  beforeEach(() => {
+    logout();
+    login('Johannes Kepler', 'user1@example.com');
```
humphd, line 21: Change this to be:

```js
user = login('Johannes Kepler', 'user1@example.com');
```

test/user.test.js
```diff
@@ -0,0 +1,57 @@
...
+  afterAll(() => logout());
+
+  it('should respond with a 403 status when not logged in', async () => {
+    logout();
+    const res = await request(app).get(`/user/feeds`);
+    expect(res.status).toEqual(403);
+  });
+
+  it('should respond with a 200 status and JSON array if logged in', async () => {
+    const res = await request(app).get(`/user/feeds`);
```
humphd, line 33: Similar to what you did on line 27, I would be explicit and login here, and you can probably get rid of the global `user` variable (i.e, have each test do the login/logout it needs).  NOTE: these `it()` tests will not run in parallel, so no issue of shared state or races.

```js
const user = login('Johannes Kepler', 'user1@example.com');
```

test/user.test.js
```diff
@@ -0,0 +1,57 @@
...
+    expect(res.status).toEqual(200);
+    expect(res.get('Content-type')).toContain('application/json');
+    expect(res.body instanceof Array).toBe(true);
+  });
+
+  it('should respond with an updated array after a new user feed is added/removed', async () => {
+    const feedCount = (await request(app).get(`/user/feeds`)).body.length;
```
humphd, line 40: Let's login here again before we start.

test/user.test.js
```diff
@@ -0,0 +1,57 @@
...
+
+    const feedData = {
+      author: user.name,
+      url: 'http://telescope200.cdot.systems',
+      user: user.id,
+    };
+    const res = await request(app).post('/feeds').send(feedData).set('Accept', 'application/json');
+
```
humphd, line 48: Add an expect here to make sure you get back a 201 (i.e., that this post worked before you go any further).

test/user.test.js
```diff
@@ -0,0 +1,57 @@
...
+    const incremented = (await request(app).get(`/user/feeds`)).body.length;
+    expect(incremented).toEqual(feedCount + 1);
+
+    await request(app).delete(`/feeds/${res.id}`);
+
```
humphd, line 53: Add an expect here for a 204 to make sure that the delete actually worked.

Silvyre:
Thank you very much, @humphd, @Grommers00 and @manekenpix for the review! I've addressed your most-recent comments in https://github.com/Seneca-CDOT/telescope/pull/958/commits/5cee6b4c37a89f88fa372ea01b50ff1e5cf45b72!

test/user.test.js
Silvyre: Fixed in https://github.com/Seneca-CDOT/telescope/pull/958/commits/5cee6b4c37a89f88fa372ea01b50ff1e5cf45b72

test/user.test.js
Silvyre: Fixed in https://github.com/Seneca-CDOT/telescope/pull/958/commits/5cee6b4c37a89f88fa372ea01b50ff1e5cf45b72

test/user.test.js
Silvyre: Fixed in Fixed in https://github.com/Seneca-CDOT/telescope/pull/958/commits/5cee6b4c37a89f88fa372ea01b50ff1e5cf45b72

test/user.test.js
Silvyre: Fixed in https://github.com/Seneca-CDOT/telescope/pull/958/commits/5cee6b4c37a89f88fa372ea01b50ff1e5cf45b72

test/user.test.js
Silvyre: Fixed in https://github.com/Seneca-CDOT/telescope/pull/958/commits/5cee6b4c37a89f88fa372ea01b50ff1e5cf45b72

test/feeds.test.js
Silvyre: Fixed in https://github.com/Seneca-CDOT/telescope/pull/958/commits/5cee6b4c37a89f88fa372ea01b50ff1e5cf45b72

PR #959, opened 2020-04-08
fix #717 Adding user class

### LINKED ISSUE #717: Create a User class
We currently have a feed class in `feed.js`, a post class in `post.js`, but we need one more, a user class in `user.js` to use the functions we are going to add to `storage.js` in #706 . Good references would be the other existing classes we have currently to give you an idea.

All code that interact directly with Redis will be in `storage.js`, any code that uses the functions we created in `storage.js` will be used by either `feed.js`, `post.js` or the yet to be created `user.js`

We need this to allow users to add feeds

Grommers00:
Co-authored-by: David Humphrey <david.humphrey@senecacollege.ca>
Co-authored-by: Josue <josue.quilon-barrios@senecacollege.ca>

## Issue This PR Addresses

Automatically close the issue when this PR is merged
    USAGE: Fixes #717  

## Type of Change

## Description

This will add the user class for ease of access later on when we get myFeeds working. This is just a draft PR and not ready to be added yet. 

## Checklist

### DIFF
--- a/src/backend/data/feed.js
+++ b/src/backend/data/feed.js
@@ -2,6 +2,7 @@ const normalizeUrl = require('normalize-url');
 
 const {
   getFeed,
+  getFeeds,
   addFeed,
   setInvalidFeed,
   isInvalid,
@@ -114,6 +115,15 @@ class Feed {
     // Use the URL to generate an id
     const id = urlToId(url);
     return this.byId(id);
+  }
+
+  /**
+   * Returns all the feeds
+   * Returns a Promise<Feeds>
+   */
+  static async all() {
+    const ids = await getFeeds();
+    return Promise.all(ids.map(Feed.byId));
   }
 }
 
--- a/src/backend/data/user.js
+++ b/src/backend/data/user.js
@@ -1,0 +1,30 @@
+const Feed = require('./feed.js');
+
+class User {
+  constructor(name, email, id) {
+    this.name = name;
+    this.email = email;
+    this.id = id;
+  }
+
+  toJSON() {
+    return {
+      name: this.name,
+      email: this.email,
+      id: this.id,
+    };
+  }
+
+  owns(feed) {
+    return feed.user === this.id;
+  }
+
+  async feeds() {
+    const feeds = await Feed.all();
+    return feeds.filter((feed) => {
+      return this.owns(feed);
+    });
+  }
+}
+
+module.exports = User;
--- a/src/backend/web/authentication.js
+++ b/src/backend/web/authentication.js
@@ -9,6 +9,7 @@ const SamlStrategy = require('passport-saml').Strategy;
 
 const { logger } = require('../utils/logger');
 const hash = require('../data/hash');
+const User = require('../data/user');
 
 /**
  * Get our SSO/SLO/SAML/Auth env variables, and warn if any are missing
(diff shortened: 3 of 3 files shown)


src/backend/data/feed.js
```diff
@@ -115,6 +116,14 @@ class Feed {
     const id = urlToId(url);
     return this.byId(id);
   }
+
+  /**
+   * Returns all the feeds
+   * Returns a Promise<Feeds>
+   */
+  static all() {
+    return getFeeds();
```
humphd, line 125: We will need to file a follow-up to fix all our other calls to `getFeeds()` and replace with this.

I think that `Feeds.all()` should be returning Feed objects.  So this should be something like:

```js
static async all() {
  const ids = await getFeeds();
}
```

src/backend/data/user.js
```diff
@@ -0,0 +1,34 @@
...
+    return {
+      name: this.name,
+      email: this.email,
+      id: this.id,
+    };
+  }
+
+  async owns(feed) {
+    if (typeof feed.id === 'number') {
+      return this.owns(await Feed.byId(feed));
```
humphd, line 20: With my change above, you'll have a Feed object already here.

src/backend/data/user.js
```diff
@@ -0,0 +1,30 @@
+const Feed = require('./feed.js');
+
+class User {
+  constructor(name, email, id) {
+    this.name = name;
+    this.email = email;
+    this.id = id;
+  }
```
Silvyre, line 8: Our user schema should look like this:
https://github.com/Seneca-CDOT/telescope/blob/b48c614d0e84b4e5bc8036938b7104821861eb94/src/backend/web/__mocks__/authentication.js#L14-L18

- We need an `isAdmin` data member.
- We should not have `id` as a constructor parameter. Instead, we should initialize user IDs by hashing the given email.

src/backend/data/user.js
```diff
@@ -0,0 +1,30 @@
...
+
+  toJSON() {
+    return {
+      name: this.name,
+      email: this.email,
+      id: this.id,
+    };
+  }
```
Silvyre, line 16: Why is this necessary?

src/backend/web/authentication.js
```diff
@@ -70,7 +71,7 @@ function init() {
     done(null, user);
   });
   passport.deserializeUser(function (user, done) {
-    done(null, user);
+    done(null, new User(user.name, user.email, user.id));
```
Silvyre, line 74: Once `user.id` has been refactored as above (to be initialized using a hashed email address), remove `user.id` from this call to construct a new User.

src/backend/data/user.js
humphd: The layer that they are working at is lower down than what you're describing.  The `isAdmin` stuff happens higher up.  Also, they need to accept the `id` as it comes from passport.

src/backend/data/user.js
humphd: This will override how the user object gets stored as JSON (e.g., JSON.stringify() will use it) when express writes this object to the session.

src/backend/web/authentication.js
humphd: They have to keep it in, since we're pulling it off the session for use in later code.

PR #963, opened 2020-04-08
Extend User type and add Admin type

humphd:
Builds on the `User` type, and adds an `Admin`.  I've connected it into our administration middleware, which runs after passport.js, but before the router.  Any user who is an admin will get automatically upgraded.

### DIFF
--- a/src/backend/web/authentication.js
+++ b/src/backend/web/authentication.js
@@ -10,6 +10,7 @@ const SamlStrategy = require('passport-saml').Strategy;
 const { logger } = require('../utils/logger');
 const hash = require('../data/hash');
 const User = require('../data/user');
+const Admin = require('../data/admin');
 
 /**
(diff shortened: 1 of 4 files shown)


src/backend/web/authentication.js
```diff
@@ -189,16 +190,20 @@ function userIsAdmin(id) {
...
+ * whether or not a user is an administrator.
  */
 function administration() {
   return function (req, res, next) {
     if (req.user && req.user.id) {
-      req.user.isAdmin = userIsAdmin(req.user.id);
+      const { user } = req;
+      if (userIsAdmin(user.id)) {
+        // Upgrade User to an Admin
+        req.user = new Admin(user.name, user.email, user.id);
```
Grommers00, line 205: Should this be done in deserialize after our PR lands as well?

humphd: We could refactor to do it all there, yeah.  Either way.

PR #986, opened 2020-04-11
Refactor admin code, add more tests

humphd:
We added a bunch of code for user vs. admin routes, and didn't completely finish it, or fully integrate it.  We also didn't do proper test coverage for the admin side.

This fixes that, specifically:

- removes the `administration` middleware, and puts it in `passport.deserialize()`
- moves the administration account check into `Admin.isAdmin()`
- updates our authentication mock to properly support both user and admin accounts
- fixes a bug with how we serialize a user to the session, adding the `.isAdmin` property
- makes the `/admin/*` routes all require an admin account
- adds test for admins deleting a feed owned by a regular user
- adds tests for `/user/info` for users and admins
- adds tests for `/user/feeds` to make sure admins get all feeds vs. only ones they own

Much of this is changes to our tests, so CI should tell you if it's working.

### DIFF
--- a/src/backend/web/__mocks__/authentication.js
+++ b/src/backend/web/__mocks__/authentication.js
@@ -63,27 +63,31 @@ function checkUser(requireAdmin, redirect, req, res, next) {
   }
 }
 
-// We'll deal with admin via init(), just let this pass
-function administration() {
-  return function (req, res, next) {
-    next();
-  };
+/**
+ * If we're logged in, build and return the appropriate user type (User vs. Admin).
+ * If we're not logged in, return null.
+ */
+function createUser() {
+  if (!loggedInUser) {
+    return null;
+  }
+
+  if (loggedInUser.isAdmin) {
+    return new Admin(loggedInUser.name, loggedInUser.email, loggedInUser.id);
+  }
+  return new User(loggedInUser.name, loggedInUser.email, loggedInUser.id);
 }
 
 function protect(redirect) {
   return function (req, res, next) {
-    if (loggedInUser) {
-      req.user = new User(loggedInUser.name, loggedInUser.email, loggedInUser.id);
-    }
+    req.user = createUser();
     checkUser(false, redirect, req, res, next);
   };
 }
 
 function protectAdmin(redirect) {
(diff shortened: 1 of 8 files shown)


src/backend/web/__mocks__/authentication.js
```diff
@@ -63,33 +63,36 @@ function checkUser(requireAdmin, redirect, req, res, next) {
...
     checkUser(false, redirect, req, res, next);
   };
 }
 
 function protectAdmin(redirect) {
   return function (req, res, next) {
-    if (loggedInUser) {
-      req.user = new Admin(loggedInUser.name, loggedInUser.email, loggedInUser.id);
-    }
+    req.user = createUser();
```
Grommers00, line 90: Can we refactor protect and protectAdmin to be one function which the parameter checkUser could be checked within the function instead of two separate functions that are identical except for the false/true that is passed through?

Grommers00: As well as in `web\authentication.js`

Grommers00: (this is listed as a comment in case there was something else we didn't think of.)

manekenpix: I reviewed this with @Grommers00

humphd: The pattern I'm following with 2 functions managing a boolean is the difference between:

```js
protect();
protectAdmin();

// vs.

protect(true);
protect(false);
```

The former is self-documenting, while the latter requires you to go look up what the boolean is for.

We could have protect take an `options` object, so you could do:

```js
protect({ redirect: true, admin: true })
```

Would that be better?

Grommers00: So, our thought process was could we grab it from the user object, (isAdmin = true) leads and then pass that through, are we wrong in thinking like that way?

humphd: They are different.  `protect` is being defined to work for a user vs. an admin.  The bit about figuring out what you are happens before that.

PR #989, opened 2020-04-12
Teach tools/autodeployment/server.js about a release to master

### LINKED ISSUE #957: Teach tools/autodeployment/server.js about a Commit to master vs. new Release
Our autodeployment server receives push notifications from a GitHub webhook.  Right now it's overly general in what triggers it.

Let's teach it the difference between a commit on `master` (not 100% sure [which event type this is](https://developer.github.com/webhooks/#events)) and a new [Release being created](https://developer.github.com/v3/activity/events/types/#releaseevent).

For staging, we'll want to deploy on every commit to `master`.

For production, we'll want to deploy only when we make a new Release.

c3ho:
## Issue This PR Addresses
Fixes #957 

## Type of Change

## Description
Added a new file `deploy_prod.sh`
Changed `tools/autodeployment/server.js` to fire off the new script when it receives a `release` event

## Checklist


tools/autodeployment/deploy_prod.sh
```diff
@@ -0,0 +1,24 @@
...
+
+rm -rf telescope
+
+git clone https://github.com/Seneca-CDOT/telescope.git --depth=1
+
+
+# Deploy
+cd telescope/
+
+cp env.staging .env
```
humphd, line 22: `env.production`?

Is the only difference between this fine and `deploy_staging.sh` the name of the env file we use?  If so, let's rename this `deploy.sh` and have it take 1 argument: the filename of the env to use.

tools/autodeployment/server.js
```diff
@@ -26,3 +26,11 @@ handler.on('push', (event) => {
     shell.exec('./deploy_stage.sh');
   }
 });
+
+handler.on('release', (event) => {
```
humphd, line 30: Will the `push` handler above not also get triggered all the time?  We'd need a way to ignore that.  We could use an environment variable, and use an `if` to look at `process.env.DEPLOYMENT === 'staging'` vs. `process.env.DEPLOYMENT === 'production'`, and only register the `handler` event for the appropriate situation.

c3ho: I just thought of this properly, you're right. If the if statement above gets executed, it'll wipe the whole telescope folder.

tools/autodeployment/deploy_prod.sh
manekenpix: Also, the passed argument should determine if the script uses `docker-compose-staging.yml` or `docker-compose-production.yml`

tools/autodeployment/deploy.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+if [ $1 = 'production']
+then
+  DOCKER_FILE=docker-compose-production.yml
+else
+  DOCKER_FILE=docker-compose-staging.yml
+fi
```
humphd, line 7: Let's do 2 things here:

1. add a specific check for `$1 = 'staging'` vs using else to assume.
1. if it's neither of these two, the else should print an error message and exit.

tools/autodeployment/deploy.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
humphd, line 1:
```bash
#!/usr/bin/env bash

set -e
set -u
set -x
```

tools/autodeployment/deploy.sh
```diff
@@ -0,0 +1,34 @@
...
+# Shutdown
+cd ../telescope
+
+docker-compose -f $DOCKER_FILE down
+
+docker system prune -af --volumes
+
+
+# Delete and clone only if the first argument is staging
+if [ $1 = 'staging' ]
```
humphd, line 18: Why not for prod too?

tools/autodeployment/deploy.sh
```diff
@@ -0,0 +1,34 @@
...
+then
+  cd ..
+
+  rm -rf telescope
+
+  git clone https://github.com/Seneca-CDOT/telescope.git --depth=1
+fi
```
manekenpix, line 25: I think we should do this for both staging and production.

tools/autodeployment/deploy.sh
```diff
@@ -0,0 +1,34 @@
...
+
+
+# Deploy
+cd telescope/
+
+# Takes the second argument the name of the env file to be used
+cp $2 .env
```
humphd, line 32: We should maybe add a check to make sure that $2 is defined and is an existing file.

```bash
if [[ -f $2 ]]; then
...
```

tools/autodeployment/server.js
```diff
@@ -3,7 +3,7 @@ const http = require('http');
 const createHandler = require('github-webhook-handler');
 const shell = require('shelljs');
 
-const { SECRET, REPO_NAME, DEPLOY_PORT } = process.env;
+const { SECRET, REPO_NAME, DEPLOY_PORT, DEPLOY_TYPE } = process.env;
```
humphd, line 6: Please update env.staging and env.production to add this variable with the correct value.

tools/autodeployment/server.js
```diff
@@ -19,10 +19,18 @@ handler.on('error', (err) => {
   console.error('Error:', err.message);
 });
 
+handler.on('release', (event) => {
```
humphd, line 22:
```js
if(DEPLOY_TYPE === 'production') {
   handler.on('release', ...);
}
```

tools/autodeployment/server.js
```diff
@@ -19,10 +19,18 @@ handler.on('error', (err) => {
...
+  const { name } = event.payload.repository;
+
+  if (name === REPO_NAME) {
+    shell.exec(`./deploy.sh ${DEPLOY_TYPE} env.production`);
+  }
+});
+
 handler.on('push', (event) => {
```
humphd, line 30:
```js
if(DEPLOY_TYPE === 'staging') {
   handler.on('release', ...);
}
```

tools/autodeployment/deploy.sh
manekenpix, line 7: Here you could also include something like:
```sh
then
  DOCKER_FILE=docker-compose-production.yml
  ENV_FILE=env.production
...
```
and then use ENV_FILE later instead of having to pass a second argument to the script.

c3ho: I like this idea.

c3ho: bash newbie here, let me know if my pushed change is what you were expecting

tools/autodeployment/deploy.sh
c3ho: added

tools/autodeployment/deploy.sh
c3ho: added a check, also took @manekenpix suggestion to set the ENV_FILE var here instead of taking a second argument

c3ho: thank you for the suggestion

tools/autodeployment/server.js
c3ho: Added two new files `env.production` + `env.staging`

tools/autodeployment/deploy.sh
c3ho: Going to forego the second argument and set it in the script itself, if that is alright

tools/autodeployment/env.staging
```diff
@@ -0,0 +1,11 @@
+# Port on which the server will be listening
```
humphd, line 1: These files exist already in the root https://github.com/Seneca-CDOT/telescope/blob/master/env.staging

tools/autodeployment/server.js
```diff
@@ -19,10 +19,24 @@ handler.on('error', (err) => {
...
+  handler.on('release', (event) => {
+    const { name } = event.payload.repository;
 
-  if (name === REPO_NAME) {
-    shell.exec('./deploy_stage.sh');
-  }
-});
+    if (name === REPO_NAME) {
+      // Execute the shell script and pass two arguments, our deployment type and the .env file to use
+      shell.exec(`./deploy.sh ${DEPLOY_TYPE}`);
```
humphd, line 28: I would use the string `production` here vs. `${DEPLOY_TYPE}` since you know it's `production` from line 22.

tools/autodeployment/server.js
```diff
@@ -19,10 +19,24 @@ handler.on('error', (err) => {
...
+  });
+}
+
+if (DEPLOY_TYPE === 'staging') {
+  handler.on('push', (event) => {
+    const { name } = event.payload.repository;
+
+    if (name === REPO_NAME) {
+      // Execute the shell script and pass two arguments, our deployment type and the .env file to use
+      shell.exec(`./deploy.sh ${DEPLOY_TYPE}`);
```
humphd, line 39: Same thing here, use `staging`

tools/autodeployment/env.staging
c3ho: I thought this is for the env vars for the autodeployment and not telescope itself?

humphd: Sorry, maybe I'm telling you the wrong thing.  cc @manekenpix

tools/autodeployment/server.js
c3ho: changed to use string

tools/autodeployment/server.js
c3ho: changed to use string

tools/autodeployment/server.js
```diff
@@ -19,10 +19,24 @@ handler.on('error', (err) => {
   console.error('Error:', err.message);
 });
 
-handler.on('push', (event) => {
-  const { name } = event.payload.repository;
+if (DEPLOY_TYPE === 'production') {
```
humphd, line 22:
```js
if(!(DEPLOY_TYPE === 'staging' || DEPLOY_TYPE === 'production')) {
  console.error('DEPLOY_TYPE must be one of 'staging' or 'production');
  process.exit(1);
}
```

Issue #1018, opened 2020-04-16
CPU Pressure and Memory leak with useEffect() calls in main page


Grommers00:
**What happened**:
Youtube videos seem to be slowing down the system and causing massive lag spikes.  It looks it is causing mass memory leaks. 
**What should have happened**:
It shouldn't create a memory leak like this. 
**How to reproduce it (as precise as possible)**:
dev.telescope.cdot.systems and you can see how slow it goes and check console to see the memory application. 

**Anything else we need to know?**:

**Environment**:

- OS:
- Browser:

humphd:
@raygervais has tested with https://github.com/Seneca-CDOT/telescope/commit/702138f2a5a2e839bbee5e303f826f204b8a5813 (1.0.0-alpha) and it doesn't seem to be affected.

humphd:
So the issue is something in https://github.com/Seneca-CDOT/telescope/compare/1.0.0-alpha...master.  The bug seems to be related to use loading and re-loading the posts over and over, and re-rendering the DOM, creating a ton of network traffic and memory thrashing.  The YouTube part isn't related, nor is the highlight.js stuff--they both just get triggered repeatedly.

humphd:
Traced to https://github.com/Seneca-CDOT/telescope/commit/f9a4c872a66b42f06fde99c6aa98c74099741e03, and an infinite render loop due to how we're doing `useEffect()`.  I've reverted for now in https://github.com/Seneca-CDOT/telescope/commit/4cd7ee80c2b4a41b316409e13382acf3193b9177, and we'll test on staging.

humphd:
Verified fixed on staging, closing.

Issue #990, opened 2020-04-12
Look into our eslint warnings


humphd:
Can any of these be fixed?

```
/Users/humphd/repos/telescope/src/backend/utils/elastic.js
  116:10  warning  Prefer await to then()  promise/prefer-await-to-then

/Users/humphd/repos/telescope/src/backend/utils/wiki-feed-parser.js
  59:10  warning  Prefer await to then()  promise/prefer-await-to-then

/Users/humphd/repos/telescope/src/frontend/src/pages/index.js
  43:6  warning  React Hook useEffect has missing dependencies: 'numPages' and 'posts'. Either include them or remove the dependency array. You can also do a functional update 'setNumPages(n => ...)' if you only need 'numPages' in the 'setNumPages' call  react-hooks/exhaustive-deps
  60:8  warning  Prefer await to then()                                                                                                                                                                                                                        promise/prefer-await-to-then
  81:6  warning  React Hook useEffect has a missing dependency: 'getPostsCount'. Either include it or remove the dependency array                                                                                                                              react-hooks/exhaustive-deps
  92:6  warning  React Hook useEffect has a missing dependency: 'callback'. Either include it or remove the dependency array                                                                                                                                   react-hooks/exhaustive-deps

✖ 6 problems (0 errors, 6 warnings)
```

manekenpix:
I can take care of the first 2 (related to #918 and #822)

c3ho:
I'll take a look at the other four

Silvyre:
FYI, the React Hook ones can be 'quick-fixed':
![image](https://user-images.githubusercontent.com/13500769/79276344-d3541d80-7e75-11ea-81d0-a81590d05110.png)

humphd:
I had to revert https://github.com/Seneca-CDOT/telescope/commit/f9a4c872a66b42f06fde99c6aa98c74099741e03, as it cased an infinite loop in the render function for the posts on the main page.  I think we need to fix how we set our dependencies in `useEffect()` (second arg).

c3ho:
I think this can be closed by #1007 + #1019

manekenpix:
Can we close this? @humphd @c3ho

humphd:
`master` has this when I build:

```
warn ESLintError:
/Users/humphd/repos/telescope/src/frontend/src/components/AboutFooter/AboutFooter.jsx
  97:15  warning  img elements must have an alt prop, either with meaningful text, or an empty string for
[...]

PR #1019, opened 2020-04-16
Fix eslint issues in index.js

c3ho:
## Issue This PR Addresses
Part of #990 

## Type of Change

## Description
Refactored our async functions existing outside of `useEffect()` hooks to use `useCallback()`. The only es lint warnings remaining are on line 84 and from my testing, adding the dependencies `post` or `nextPageLink` as recommended will cause the issue in #1018 

I kind of hate the front-end and its dependencies...
From what I understand:
1. `useEffect` line 108 is called every re-render

2. `useEffect` on line 113 is caused by either `getPostsCount`(async function) or change to `currentNumPosts`(state). This sets the current and initial # of posts to whatever the current # of posts in redis is

3. `useEffect` on line 127 should be caused by the callback on line 108? This causes it to start the timer to get the updated # of posts every 5 mins and set currentNumPosts to the new #. Once this is set, it should cause the `useEffect` from 2.

## Checklist

### DIFF
--- a/src/frontend/src/pages/index.js
+++ b/src/frontend/src/pages/index.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useState, useRef } from 'react';
+import React, { useEffect, useState, useRef, useCallback } from 'react';
 import { makeStyles } from '@material-ui/core/styles';
 import { CircularProgress, Button, Grid } from '@material-ui/core';
 import parse from 'parse-link-header';
@@ -88,7 +88,7 @@ export default function IndexPage() {
     setNumPages(numPages + 1);
   }
 
-  async function getPostsCount() {
+  const getPostsCount = useCallback(async () => {
     try {
       const res = await fetch(`${telescopeUrl}/posts`, { method: 'HEAD' });
       if (!res.ok) {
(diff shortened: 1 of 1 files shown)


c3ho:
I kind of hate the front-end and its dependencies...

From what I understand:
1. `useEffect` line 108 is called every re-render

2. `useEffect` on line 113 is triggered by either `getPostsCount`(async function) or change to `currentNumPosts`(state). This sets the current and initial # of posts to whatever the current # of posts in redis is

3. `useEffect` on line 127 should be triggered by the callback on line 108? This causes it to start the timer to get the updated # of posts every 5 mins and set currentNumPosts to the new #. Once this is set, it should trigger the `useEffect` from 2.

src/frontend/src/pages/index.js
```diff
@@ -124,18 +122,17 @@ export default function IndexPage() {
       }
     }
     setPostsInfo();
-  }, []);
+  }, [getPostsCount, currentNumPosts]);
```
humphd, line 125: Does this not need `initNumPosts` too?  Also, why do we call `getPostsCount()` twice on 117 and 118, and not simply re-use the value?

c3ho: We don't care about whatever initNumPosts is going to be in this case. Both currentNumPosts and initNumPosts will be overwritten. I only want this side-effect happening whenever a currentNumPosts value changes

src/frontend/src/pages/index.js
```diff
@@ -88,7 +88,7 @@ export default function IndexPage() {
     setNumPages(numPages + 1);
   }
 
-  async function getPostsCount() {
+  const getPostsCount = useCallback(async () => {
```
humphd, line 91: I'm unclear if you can use an `async` function here.  What I read online and in the docs doesn't help much.

c3ho: https://stackoverflow.com/questions/60835660/proper-use-of-usecallback . Looks like this guy does it in the React hook

c3ho:
Will merge after rebase as it received two approvals

PR #1020, opened 2020-04-16
Add /status to deploy server, fix callback missing bug

humphd:
This updates the deployment server to do a few things:

1. adds a `/status` endpoint, so we can get info on what's happening (idle or building), what the duration of the build is, what the result of the last build was, etc.
1. fixes a bug where we would crash if someone tried to hit an endpoint other than the ones we expect (we needed a callback)
1. added logging for stdout and stderr

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -8,6 +8,14 @@
       "program": "${workspaceFolder}/src/backend/index.js",
       "envFile": "${workspaceRoot}/.env",
       "skipFiles": ["<node_internals>/**"]
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Launch Auto Deployment",
+      "program": "${workspaceFolder}/tools/autodeployment/server.js",
+      "envFile": "${workspaceRoot}/tools/autodeployment/.env",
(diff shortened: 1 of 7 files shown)


humphd:
Would it make sense for me to add a route to the main server that redirects to this?

```
GET /admin/build-status   --->    redirect to :4000/status
```

?

manekenpix:
I think it would. Maybe not redirect but show the info fetched from the deploy server, as a Telescope feature and not something _external_.

humphd:
How should I deal with the fact that the deploy port is specified in an env outside the telescope app?  I either need to be able to read it in Telescope's node server, or we have to hard code it so I can assume it's always the same.

manekenpix:
I'm not sure. I'm thinking that maybe Telescope can get that information from _the other_ `.env`? but I don't know if that'd be a good way to do it.

humphd:
I did some more work on this, trying to add proxying from the main node app to the deployment app for admins, so you can get info and build logs.  It's not fully working yet.

humphd:
@manekenpix should we land this before or after we release 1.0.0?  I'm fine either way.

manekenpix:
I think we can do it before.

PR #1033, opened 2020-04-20
Add UI for Admin to trigger reparsing all posts from feeds

### LINKED ISSUE #1030: Add UI for Admin to trigger reparsing all posts from feeds
We have really effective caching happening for post content, such that we'll skip re-downloading and parsing posts if what we have in Redis has the same `etag` and `last-modified` date.  Because we can't blow away the production database whenever we need to, it's going to be necessary to trigger a "clear cache," such that posts get re-downloaded and re-parsed again on the next cycle.

This could mean adding a `DELETE /feeds/cache` HTTP route, which would delete all `etag` and `lastModified` attributes on every `Feed`, and use `protectAdmin()` middleware.  NOTE: I think this is OK if we put it before `DELETE /feeds/:id`, but I need to confirm that.

We'll also need a way to trigger this via UI in the app, but that could happen in a follow-up.

I want this because I'm working on a switch to our post content to move the syntax highlighting to the server-side from client-side, and I need to have every post get re-parsed to include the proper markup.

c3ho:
## Issue This PR Addresses
Fixes #1030 

## Type of Change

## Description
Allows admin users to "clear cache" for Redis through a put request at the route `feeds/cache`. This sets the `lastModified` + `etag` field to null.

Added a new PUT route - `feeds/cache/`
Added new function in `feed.js` - `clearCache()`
Added new tests to `feed.test.js`
Added new tests to `feeds.test.js`

Testing instructions:
**Requires local backend!**
1. Find an admin account
2. Use `redis-cli smembers t:feeds` to get a list of feedId and pick one of them
3. Use `redis-cli hgetall t:feed:(feedId)` to get info of all the fields for the feed, make sure the `etag` and the `lastModified` field is not null / populated
4. While logged in as admin, hit the route `telescopeUrl/feeds/cache`, this should clear the `etag` + `lastModified` fields in Redis and set it to `null`
5. Use `redis-cli hgetall t:feed:(feedId)` again(with the same Id as before), the `etag` and the `lastModified` field should be `null` this time

Or just run `npm test feed.test.js` + `npm test feeds.test.js` and trust my tests =D

## Checklist

### DIFF
--- a/src/backend/data/feed.js
+++ b/src/backend/data/feed.js
@@ -162,6 +162,21 @@ class Feed {
     const ids = await getFeeds();
     return Promise.all(ids.map(Feed.byId));
   }
+
+  /**
+   * Sets all stored feeds' lastModified + etag field to null. Used for production
+   * Returns a Promise
+   */
+  static async clearCache() {
+    const allFeeds = await this.all();
+    await Promise.all(
+      allFeeds.map((feed) => {
+        feed.etag = null;
+        feed.lastModified = null;
+        return feed.save();
+      })
+    );
+  }
 }
 
 module.exports = Feed;
--- a/src/backend/web/routes/feeds.js
+++ b/src/backend/web/routes/feeds.js
@@ -3,6 +3,7 @@ const Feed = require('../../data/feed');
 const { getFeeds } = require('../../utils/storage');
 const { logger } = require('../../utils/logger');
 const { protect } = require('../authentication');
+const { protectAdmin } = require('../authentication');
 
 const feeds = express.Router();
 
@@ -70,6 +71,18 @@ feeds.post('/', protect(), async (req, res) => {
     logger.error({ error }, 'Unable to add feed to Redis');
     return res.status(503).json({
       message: 'Unable to add to Feed',
+    });
+  }
+});
+
+feeds.delete('/cache', protectAdmin(true), async (req, res) => {
+  try {
+    await Feed.clearCache();
+    return res.status(204).send();
+  } catch (error) {
+    logger.error({ error }, 'Unable to reset Feed data in Redis');
+    return res.status(503).json({
+      message: 'Unable to reset data for Feeds',
     });
   }
 });
--- a/test/feed.test.js
+++ b/test/feed.test.js
@@ -19,6 +19,24 @@ describe('Post data class tests', () => {
     lastModified: null,
   };
 
+  const data2 = {
+    author: 'Post Author2',
+    url: 'https://user2.feed.com/feed.rss',
+    user: 'user2',
+    link: 'https://user2.feed.com/',
+    etag: null,
+    lastModified: null,
+  };
+
+  const data3 = {
+    author: 'Post Author3',
+    url: 'https://user3.feed.com/feed.rss',
+    user: 'user3',
+    link: 'https://user3.feed.com/',
+    etag: null,
+    lastModified: null,
+  };
+
   const createFeed = () => new Feed(data.author, data.url, data.user, data.link, null, null);
 
   const articleData1 = {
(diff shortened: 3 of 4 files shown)


src/backend/data/feed.js
```diff
@@ -162,6 +162,21 @@ class Feed {
...
     return Promise.all(ids.map(Feed.byId));
   }
+
+  /**
+   * Sets all stored feeds' lastModified + etag field to null. Used for production
+   * Returns a Promise
+   */
+  static async clearCache() {
+    const allFeeds = await this.all();
+    return Promise.all(
```
humphd, line 172: `await Promise.all(...` should be good, we don't return anything.

src/backend/data/feed.js
```diff
@@ -162,6 +162,21 @@ class Feed {
...
+      allFeeds.map((feed) => {
+        feed.etag = null;
+        feed.lastModified = null;
+        return this.create(feed);
```
humphd, line 176: `return feed.save()` I think makes more sense.

src/backend/web/routes/feeds.js
```diff
@@ -74,6 +75,18 @@ feeds.post('/', protect(), async (req, res) => {
   }
 });
 
+feeds.put('/cache', protectAdmin(true), async (req, res) => {
+  try {
+    await Feed.clearCache();
+    return res.status(200).json({ message: 'Cache successfully cleared' });
```
humphd, line 81: Probably 204 vs. 200 and return nothing.

test/feed.test.js
```diff
@@ -172,5 +172,19 @@ describe('Post data class tests', () => {
       expect(esSearchDelete.values).toStrictEqual([]);
       expect(esSearchDelete.results).toBe(0);
     });
+
+    test('clearCache should set lastModified and etag to null', async () => {
+      let feed = new Feed(data.author, data.url, data.user, data.link, 'etag', 'lastModified');
+      feed = await Feed.byId(await Feed.create(feed));
+      // Basic check
+      expect(feed.etag).not.toBe(data.etag);
```
humphd, line 180: Why not check for `etag` and `lastModified` directly?

Should we do this for a few feeds vs just 1?

c3ho: Will add more feeds for tests

src/backend/web/routes/feeds.js
c3ho: Thanks! Was debating between the two

src/backend/data/feed.js
```diff
@@ -165,15 +165,14 @@ class Feed {
 
   /**
    * Sets all stored feeds' lastModified + etag field to null. Used for production
-   * Returns a Promise
```
humphd: You can leave this line, `async` functions implicitly return a Promise (good to call it out).

c3ho:
Is it redundant to add the same checks in `feed.test.js` to `feeds.test.js`?

humphd:
Yeah, you should probably add tests for the code you've added to both places.  It will be slightly different (obviously), but also similar.

The tests for doing this with the route will need to use the same `loginAdmin()` stuff I've done in other tests to simulate our authentication stuff.  Let me know if you need help.

I totally forgot about that.

test/feeds.test.js
```diff
@@ -133,6 +134,88 @@ describe('test POST /feeds endpoint', () => {
   });
 });
 
+describe('test PUT /feeds/cache endpoint', () => {
+  it('should respond with a 403 status trying to clear cache without logging in', async () => {
+    const res = await request(app).put('/feeds/cache').send().set('Accept', 'application/json');
```
humphd, line 139: `delete` vs. `put` (same for all tests below)

test/feeds.test.js
```diff
@@ -133,6 +134,88 @@ describe('test POST /feeds endpoint', () => {
...
+    const clearedFeeds = await Promise.all(feedIds.map(Feed.byId));
+    expect(clearedFeeds[0].etag).toBe(null);
+    expect(clearedFeeds[1].etag).toBe(null);
+    expect(clearedFeeds[2].etag).toBe(null);
+    expect(clearedFeeds[0].lastModified).toBe(null);
+    expect(clearedFeeds[1].lastModified).toBe(null);
+    expect(clearedFeeds[2].lastModified).toBe(null);
+
+    // Make sure logging out as admin does not allow us to use the put route
+    logout();
```
humphd, line 213: You've already tested this above, I don't think you need to do it again.

c3ho: removed

test/feeds.test.js
c3ho: Okay, I'm confused about this part since I wrote it as a PUT on line 78 of `feeds.js`

humphd: We could probably argue it both ways.  I think `DELETE` is more accurate, since we're removing something.  With `PUT`, you'd expect some data to accompany the request, as in "here are the new value(s)".

c3ho: Okay, I will update the route to use `DELETE` then. Thanks!

test/feeds.test.js
```diff
@@ -133,6 +134,83 @@ describe('test POST /feeds endpoint', () => {
...
+      link: 'https://user.feed.com/',
+      etag: 'etag',
+      lastModified: 'lastModified',
+    };
+
+    const data2 = {
+      author: 'Post Author2',
+      url: 'https://user2.feed.com/feed.rss',
+      user: 'user2',
+      id: urlToId('https://user2.feed.com/feed.rss'),
```
humphd, line 172: These generated `id` properties (here and below on 182) are never read, so I wouldn't bother generating them, or importing `normalize-url` or hashing them above.  What you're doing is correct, but `Feed.create()` will do this for you, and you're not testing that it works, so don't bother doing its work for it.

c3ho: Totally removed the id property from `data`, `data2`, `data3`

test/feed.test.js
```diff
@@ -19,6 +19,26 @@ describe('Post data class tests', () => {
     lastModified: null,
   };
 
+  const data2 = {
+    author: 'Post Author2',
+    url: 'https://user2.feed.com/feed.rss',
+    user: 'user2',
+    id: urlToId('https://user2.feed.com/feed.rss'),
```
humphd, line 26: This doesn't seem to get used, remove?

test/feed.test.js
```diff
@@ -19,6 +19,26 @@ describe('Post data class tests', () => {
...
+    link: 'https://user2.feed.com/',
+    etag: null,
+    lastModified: null,
+  };
+
+  const data3 = {
+    author: 'Post Author3',
+    url: 'https://user3.feed.com/feed.rss',
+    user: 'user3',
+    id: urlToId('https://user3.feed.com/feed.rss'),
```
humphd, line 36: This doesn't seem to get used, remove?

c3ho:
Merging as it received two approvals prior to rebase.

PR #1035, opened 2020-04-21
Start moving syntax highlighting to the backend

humphd:
I want to start moving any processing that can be done in the backend out of the frontend.  First up, syntax highlighting.

It's a bit more complicated to do it in node, because the library requires a DOM.  I use JSDOM to extract the bits it needs, feed them in one-by-one, and then replace the content with marked up versions.

I've added some basic tests, and we can extend this as we see things we like or don't like happening, and add little fixes.

NOTE: I'm not removing the frontend code fully yet, only reducing how much it runs.  The new code will check to see if it needs to run the highlighter or not, and skip it if it's already been run.

I need the work @c3ho is doing in #1033 to land first, since I need all post content to be highlighted in the backend when we store it.  Then I can do a new PR to remove the frontend code completely.

### DIFF
--- a/src/backend/utils/syntax-highlighter.js
+++ b/src/backend/utils/syntax-highlighter.js
@@ -1,0 +1,64 @@
+const hljs = require('highlight.js/lib/highlight');
+const jsdom = require('jsdom');
+const { logger } = require('./logger');
+
+const { JSDOM } = jsdom;
+
+// Tweak the language list here, see https://highlightjs.org/usage/
+hljs.registerLanguage('cpp', require('highlight.js/lib/languages/cpp'));
+hljs.registerLanguage('sql', require('highlight.js/lib/languages/sql'));
+hljs.registerLanguage('powershell', require('highlight.js/lib/languages/powershell'));
+hljs.registerLanguage('cs', require('highlight.js/lib/languages/cs'));
+hljs.registerLanguage('go', require('highlight.js/lib/languages/go'));
+hljs.registerLanguage('css', require('highlight.js/lib/languages/css'));
+hljs.registerLanguage('makefile', require('highlight.js/lib/languages/makefile'));
+hljs.registerLanguage('markdown', require('highlight.js/lib/languages/markdown'));
+hljs.registerLanguage('swift', require('highlight.js/lib/languages/swift'));
+hljs.registerLanguage('armasm', require('highlight.js/lib/languages/armasm'));
+hljs.registerLanguage('diff', require('highlight.js/lib/languages/diff'));
+hljs.registerLanguage('python', require('highlight.js/lib/languages/python'));
+hljs.registerLanguage('dockerfile', require('highlight.js/lib/languages/dockerfile'));
+hljs.registerLanguage('http', require('highlight.js/lib/languages/http'));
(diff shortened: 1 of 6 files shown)


test/syntax-highlighter.test.js
```diff
@@ -0,0 +1,48 @@
+/**
+ * @jest-environment jsdom
+ */
+
+const syntaxHighlighter = require('../src/backend/utils/syntax-highlighter');
+
+/**
+ * syntaxHighlighter() will markup code so it can be styled as code with CSS
+ */
+describe('syntax-highlighter tests', () => {
```
c3ho, line 10: should we add a test here for it not to highlight a non-registered language?

humphd: What would you like it to do?  Leave it as is?  I could probably check the return value of `language` and if it's not defined, just return the raw HTML again?

Or we could let it try to do the syntax highlighting, knowing it will get some cases wrong.  It might be easier to fix this when we start seeing it make mistakes, and have ready-made test cases.  Or do you have one in mind?

src/backend/utils/syntax-highlighter.js
```diff
@@ -0,0 +1,64 @@
...
+ * Given a String of HTML, find all <pre>...</pre> code blocks
+ * and apply syntax highlight markup.
+ */
+module.exports = function (html) {
+  try {
+    const dom = new JSDOM(html);
+    const doc = dom.window.document;
+
+    doc.querySelectorAll('pre code').forEach((code) => {
+      const { value, language } = hljs.highlightAuto(code.innerHTML);
```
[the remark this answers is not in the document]
humphd: For cases where you don't want it to guess, and guess wrong.  In my experience, it mostly gets it right, since I've eliminated languages that are unlikely to appear in our planet.

test/syntax-highlighter.test.js
c3ho: > Leave it as is? I could probably check the return value of language and if it's not defined, just return the raw HTML again?
I was thinking something along the lines of this

The second test sounds great too to be honest. But as you said, that's something we can add later on when find issues with the highlighting.

humphd:
I added some more test cases and error handling, but it's nearly impossible to write a test that sees code and doesn't try to make a (usually good) guess and highlight it.

Let's fix this when it becomes a problem, and use the specific test case as a guide.

test/syntax-highlighter.test.js
```diff
@@ -0,0 +1,72 @@
...
+    const original = undefined;
+    const result = syntaxHighlighter(original);
+    expect(result).toEqual(original);
+  });
+
+  test('null is returned untouched', () => {
+    const original = null;
+    const result = syntaxHighlighter(original);
+    expect(result).toEqual(original);
+  });
```
c3ho, line 27: Ahhh nice, I was thinking about these

humphd:
I'll land this now, but it won't start working for all posts until #1033 is in and I can trigger a complete re-build on staging and prod (we'll maybe need to ship 1.0.1 to do that).  Then I'll do a follow-up to rip out the frontend code completely.

humphd:
Filed #1041 to do this.

PR #1045, opened 2020-04-22
Enable HTTP2 in nginx

### LINKED ISSUE #1044: Use HTTP2 in nginx
In #1042, we're exploring performance improvements.  One obvious one is to switch to HTTP2 so we can multiplex requests, and get other improvements:

Screen Shot 2020-04-22 at 12 44 15 PM

I *think* this is the only change I need to make, but there are 2 other `listen` lines that need to be examined:

```
 # this should be fine as is, since we'll redirect you to HTTPS, and it will do HTTP2, but confirm for me.
 server {
    listen 80 default_server;
    server_name _;
    return 307 https://$host$request_uri;
  }
```

```
 # do we still need this? seems like we don't need this in production?
  server {
    listen 80;
    server_name login.telescope.cdot.systems;
    location / {
      proxy_pass http://telescope_production:3000/auth/login;
    }
  }
```

humphd:
This fixes #1044.  See my comments there for the help I need reviewing this.  Then we could try it on staging and see how it goes, compare to prod.

### DIFF
--- a/nginx-production.conf
+++ b/nginx-production.conf
@@ -16,7 +16,7 @@ http {
   }
 
   server {
-    listen 443 ssl;
+    listen 443 ssl http2;
     server_name telescope.cdot.systems;
 
     include /etc/letsencrypt/options-ssl-nginx.conf;
--- a/nginx-production.conf
+++ b/nginx-production.conf
@@ -81,12 +81,4 @@ http {
       proxy_pass http://telescope_production:3000;
     }
   }
-
-  server {
-    listen 80;
-    server_name login.telescope.cdot.systems;
-    location / {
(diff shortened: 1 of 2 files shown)


humphd:
Can we delete this bit?  Do we need it at all?

```
 # do we still need this? seems like we don't need this in production?
  server {
    listen 80;
    server_name login.telescope.cdot.systems;
    location / {
      proxy_pass http://telescope_production:3000/auth/login;
    }
  }
```

manekenpix:
@humphd yes, right now we're not using it.

humphd:
I've removed the login setup as well.

Issue #1042, opened 2020-04-22
Improve our app's startup performance


humphd:
We're not doing well on startup:

<img width="1080" alt="Screen Shot 2020-04-22 at 10 00 55 AM" src="https://user-images.githubusercontent.com/427398/79991497-609fff00-8480-11ea-844f-247ff8b4d928.png">

In the startup trace above, you can see that from 1s to 5s we are idle, waiting on data from 4 separate requests.

Step one to faster loading: figure out how to get that time down to as close to nothing as possible.  I suspect we'll need a bunch of things to happen together (it won't be one fix).

humphd:
First question I have: why does it take 5 seconds for our web server to respond to any requests?  What's causing all of these to be slow?  I could understand if it was the `/stats` call, since it does a lot of work in the backend.  But the `/health` one should be really fast.  They all take ~5 seconds, which is odd.

humphd:
Testing these routes on their own, they don't seem to be the bottleneck (I'm getting ~100ms response times from nginx and the node app:

```
time curl https://dev.telescope.cdot.systems/health
{"status":"ok","info":{"version":"1.0.0","sha":"56a665b3827b9bf9a070c8eebdfd9ae0caf21732","gitHubUrl":"https://github.com/Seneca-CDOT/telescope/commit/56a665b3827b9bf9a070c8eebdfd9ae0caf21732"}}
real	0m0.097s
user	0m0.012s
sys	0m0.006s
[humphd@mbp ~/repos/telescope (backend-syntax-highlight)]$ time curl https://dev.telescope.cdot.systems/stats/year
{"posts":136,"authors":22,"words":102986}
real	0m0.087s
user	0m0.011s
sys	0m0.006s
[humphd@mbp ~/repos/telescope (backend-syntax-highlight)]$ time curl https://dev.telescope.cdot.systems/posts
[...]

PR #1046, opened 2020-04-22
Enable gzip compression in nginx

humphd:
More work toward #1042, this alters our nginx config to add gzip compression for text types we serve.

For reference:

- https://docs.nginx.com/nginx/admin-guide/web-server/compression/
- https://www.digitalocean.com/community/tutorials/how-to-increase-pagespeed-score-by-changing-your-nginx-configuration-on-ubuntu-16-04
- http://nginx.org/en/docs/http/ngx_http_gzip_module.html

This should help further reduce the size of our transferred html, js, etc. on the network.

### DIFF
--- a/nginx-production.conf
+++ b/nginx-production.conf
@@ -1,6 +1,9 @@
 events {}
 
 http {
+  include       mime.types;
+  default_type  application/octet-stream;
+
   # levels        - Defines hierarchy levels
   # keys_zone     - Name for these settings
   # inactive      - Cached data that are not accessed during the time specified by the inactive parameter get removed from the cache regardless of their freshness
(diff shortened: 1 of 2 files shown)


humphd:
Added another commit to include mime type handling.  I noticed that right now we server `.js` files as `text/plain`.

humphd:
```
$ curl -H 'Accept-Encoding: gzip' -I https://dev.telescope.cdot.systems
HTTP/2 200
server: nginx/1.17.10
date: Thu, 23 Apr 2020 00:16:17 GMT
content-type: text/html
last-modified: Thu, 23 Apr 2020 00:07:14 GMT
vary: Accept-Encoding
etag: W/"5ea0dc32-9425"
strict-transport-security: max-age=63072000
content-encoding: gzip
```

This is working on staging :)

PR #1048, opened 2020-04-23
Move img and iframe lazy loading to server

humphd:
This moves the `` and `<iframe>` lazy loading out of the frontend and does it in the backend during feed processing.  I've refactored the backend HTML processing code to share some common bits (e.g., use the same JSDOM instance).  I've also added and updated tests.

I still want to test this more, but it's close I think.


humphd:
I tested this overnight, and found a crash case that I've fixed.  I think it's good to go now.

humphd:
NOTE: the build failure on zeit is due to an outage that GitHub had earlier, it can be ignored.

PR #1051, opened 2020-04-23
Fix #982: restore nameID to user info for logout requests

### LINKED ISSUE #982: Unable to logout on Seneca's IdP (staging or production)
**What happened**:
I can log onto https://dev.telescope.cdot.systems/ using my senstudents.com with no problem but I cannot log out. When I click on LOGOUT, I get redirected to this:
Annotation 2020-04-09 231359

I should be able to log in and log out because my account is legit.

Try logging in and logging out using your senstudents.com SSO account

**Environment**:

- OS: Windows
- Browser: Chrome/Firefox

humphd:
Fixes #982.  This adds back the `nameID` and `nameIDFormat` values we get from the Seneca IdP.  We need this to be able to properly do an IdP SLO Logout.

This is impossible to test until it lands on staging.  So I mostly need a code review here, and if someone could test it locally.

### DIFF
--- a/src/backend/data/admin.js
+++ b/src/backend/data/admin.js
@@ -6,8 +6,8 @@ const hash = require('./hash');
 const administrators = process.env.ADMINISTRATORS ? process.env.ADMINISTRATORS.split(' ') : [];
 
 class Admin extends User {
-  constructor(name, email, id) {
-    super(name, email, id);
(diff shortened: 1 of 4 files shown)


humphd:
#1054 adds a bit that I missed in this.

PR #1056, opened 2020-04-23
Fix #1036: reduce number of posts loaded on startup

### LINKED ISSUE #1036: Load fewer posts initially, increase startup performance
When the app loads, we pull down a lot of data (including posts), and it cause the app to take a long time to start.  Let's see how fast we can make startup.

One easy fix we could try is to dramatically reduce the number of posts we request on first load.  I think 3-5 is probably all you need, and then you can click "Load More" when you eventually want to see older posts.  We could also do something were we wait on the app to get fully loaded with 3-5 initial posts, then when the system is idle, trigger a call to get the next 10-15.

Later on, I'd also love to see if we can leverage https://swr.now.sh/.

We can break this up into small fixes.  To start, what if we try only loading the 5 most recent posts.

humphd:
Fixes #1036

As part of #1042, we need to reduce the size of our DOM.  The easiest way to do this is to load fewer posts.  We have code to "load more" now, so we don't need so many at once.

This drops it down to 10.  We might be able to go lower still, but let's see how this performs.

### DIFF
--- a/env.example
+++ b/env.example
@@ -102,7 +102,7 @@ FEED_QUEUE_DELAY_MS=60000
 FEED_QUEUE_PARALLEL_WORKERS=1
 
 # Max number of posts per page
-MAX_POSTS_PER_PAGE=30
+MAX_POSTS_PER_PAGE=10
 
 # If you need to do a release, GitHub requires a Personal Access token.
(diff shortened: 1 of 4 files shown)


humphd:
We had a test that depended on this being 30.  I've changed the test to not need a specific number.

PR #1043, opened 2020-04-22
adds: gatsby image optimization for telescope images

raygervais:
## Issue This PR Addresses
Fixes #1000 #894 

## Type of Change

- [ ] **Bugfix**: Change which fixes an issue
- [ ] **New Feature**: Change which adds functionality
- [ ] **Documentation Update**: Change which improves documentation
- [X] **UI**: Change which improves UI

## Description
Changes collection which is used by `DynamicBackgroundContainer.jsx` to use our [Telescope 1.0](https://unsplash.com/collections/9975402/telescope-1.0)  collection.

## Checklist

- [ ] **Quality**: This PR builds and passes our npm test and works locally
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not (if applicable)
- [ ] **Documentation**: This PR includes updated/added documentation to user exposed functionality or configuration variables are added/changed or an explanation of why it does not(if applicable)

PR #1064, opened 2020-04-24
Stop loading /posts twice on startup

### LINKED ISSUE #1060: Stop loading /posts twice on startup
Screen Shot 2020-04-23 at 6 53 02 PM

We seem to kick-off two nearly simultaneous requests to `/posts`:

1. The first happens [here](https://github.com/Seneca-CDOT/telescope/blob/56a665b3827b9bf9a070c8eebdfd9ae0caf21732/src/frontend/src/pages/index.js#L60) and goes to `https://dev.telescope.cdot.systems/posts`
2. The second happens [here](https://github.com/Seneca-CDOT/telescope/blob/56a665b3827b9bf9a070c8eebdfd9ae0caf21732/src/frontend/src/pages/index.js#L94) and goes to `https://dev.telescope.cdot.systems/posts?page=1`.  This one is only being run to get the `x-total-count` header (`HEAD` request), and could be paired with the first.

Let's reduce this down to a single request to `/posts?page=1` and process the `x-total-count` and post data all at once.

cc @c3ho, @Grommers00, @raygervais.

c3ho:
## Issue This PR Addresses
Fixes #1060 

## Type of Change

## Description
Created new state [update, setUpdate] to let when `getPostsCount()` should run (not at the beginning). Only the first `useEffect` (line 56) should run now at startup.

To test:

1. open up dev tools on chrome and go to `Performance` tab.
2. Press the record button and reload telescope page, stop as soon as posts load
3. Check if there are two fetch requests to telescopeUrl/posts

## Checklist

### DIFF
--- a/src/frontend/src/pages/index.js
+++ b/src/frontend/src/pages/index.js
@@ -44,6 +44,7 @@ export default function IndexPage() {
   const [currentNumPosts, setCurrentNumPosts] = useState(0);
   const [endOfPosts, setEndOfPosts] = useState(false);
   const [alert, setAlert] = useState(false);
+  const [shouldCheckForNewPosts, setShouldCheckForNewPosts] = useState(false);
   const { telescopeUrl } = useSiteMetaData();
   const savedCallback = useRef();
   const snackbarMessage = 'There is new content available!';
@@ -90,6 +91,9 @@ export default function IndexPage() {
   }
 
   const getPostsCount = useCallback(async () => {
+    if (!shouldCheckForNewPosts) {
+      return null;
+    }
     try {
       const res = await fetch(`${telescopeUrl}/posts`, { method: 'HEAD' });
       if (!res.ok) {
@@ -98,9 +102,11 @@ export default function IndexPage() {
       return res.headers.get('x-total-count');
     } catch (error) {
       console.log(error);
+    } finally {
+      setShouldCheckForNewPosts(false);
     }
     return null;
-  }, [telescopeUrl]);
+  }, [telescopeUrl, shouldCheckForNewPosts]);
 
   const callback = useCallback(async () => {
     setCurrentNumPosts(await getPostsCount());
@@ -124,6 +130,7 @@ export default function IndexPage() {
 
   useEffect(() => {
     function getCurrentNumPosts() {
+      setShouldCheckForNewPosts(true);
       savedCallback.current();
     }
     savedCallback.current = callback;


src/frontend/src/pages/index.js
```diff
@@ -90,17 +91,20 @@ export default function IndexPage() {
   }
 
   const getPostsCount = useCallback(async () => {
-    try {
-      const res = await fetch(`${telescopeUrl}/posts`, { method: 'HEAD' });
-      if (!res.ok) {
-        throw new Error(res.statusText);
+    if (update) {
```
humphd, line 94:
```js
if(!update) {
  return null;
}
...continue without indent
```

src/frontend/src/pages/index.js
```diff
@@ -90,17 +91,20 @@ export default function IndexPage() {
...
+      try {
+        const res = await fetch(`${telescopeUrl}/posts`, { method: 'HEAD' });
```
humphd, line 96: So how does this update logic work?  After the first load, you want to keep checking the number of posts, and if it changes, we know we have new ones?

If so, I wonder if "update" is the right name.  `shouldCheckForNewPost` or something (shorter) but better documenting?

humphd:
This is *way* better:

Screen Shot 2020-04-23 at 10 45 10 PM

Wait til we get @raygervais's work from #1043.  It's the largest blocking network call in startup now.

src/frontend/src/pages/index.js
c3ho: Doesn't this way make it kind of messy with two return null statements?

```
const getPostsCount = useCallback(async () => {
    if (!shouldCheckForNewPosts) {
      return null;
    }
    try {
      const res = await fetch(`${telescopeUrl}/posts`, { method: 'HEAD' });
      if (!res.ok) {
        throw new Error(res.statusText);
      }
      setShouldCheckForNewPosts(false);
      return res.headers.get('x-total-count');
    } catch (error) {
      console.log(error);
    }
    return null;
  }, [telescopeUrl, shouldCheckForNewPosts]);
```

src/frontend/src/pages/index.js
```diff
@@ -91,20 +91,21 @@ export default function IndexPage() {
...
-        console.log(error);
+    if (!shouldCheckForNewPosts) {
+      return null;
+    }
+    try {
+      const res = await fetch(`${telescopeUrl}/posts`, { method: 'HEAD' });
+      if (!res.ok) {
+        throw new Error(res.statusText);
       }
+      setShouldCheckForNewPosts(false);
```
humphd, line 102: If we error on line 100 above, we won't set this to `false`.  Is that OK?  Should we do this in a `finally` block, or you want it to remain `true?

src/frontend/src/pages/index.js
humphd: Don't believe what your other profs told you, [Return Early, Return Often](https://blog.humphd.org/vocamus-1421/).

src/frontend/src/pages/index.js
c3ho: You're right, a `finally` block is required to set it to `false`

c3ho:
Received two approvals. Rebasing then merging

PR #1072, opened 2020-04-24
Add cache-control headers for Gatsby's files

humphd:
We don't set proper `Cache-Control` headers on the files built by Gatsby.  This tries to fix that.  I'm not 100% sure that I've done it correctly yet, and would like to try on staging, then tweak.

The docs for Gatsby outline what needs to happen: https://www.gatsbyjs.org/docs/caching

### DIFF
--- a/nginx-production.conf
+++ b/nginx-production.conf
@@ -120,7 +120,6 @@ http {
     location ~ ^/(admin|user|health|auth) {
       proxy_cache_bypass 1;
       proxy_pass http://telescope_production:3000;
-
     }
 
     # Static content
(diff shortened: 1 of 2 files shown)


humphd:
That broke staging.  I've reverted.

manekenpix:
@c3ho and I debugged the PR, this is what we saw:

Nginx was giving us a warning:
```sh
nginx_1          | 2020/04/25 05:00:01 [warn] 8#8: duplicate MIME type "text/html" in /etc/nginx/nginx.conf:57
nginx_1          | nginx: [warn] duplicate MIME type "text/html" in /etc/nginx/nginx.conf:57
```
We removed `text/html` from line 38 to get rid of the warning.
More info about this [here](https://stackoverflow.com/questions/6475472/duplicate-mime-type-text-html)

The cache-control headers were missing a semicolon in line 159 (that's what caused `nginx` to crash). After adding the missing semicolon, `nginx` wasn't able to find telescope.
The reason (we think) is that we tell `nginx` to check a specific folder to serve static content:
```nginx
location / {
      # Directory from which static content is served
      root /var/www/data;

      # Try serving static content, and if not found continue with @proxy
      try_files $uri $uri/ @proxy;
    }
```
All the new added locations don't have
```nginx
root /var/www/data;
```
so `nginx` was forwarding the request to telescope, and since the `/` route [only works in development](https://github.com/Seneca-CDOT/telescope/blob/6b7803c94506eee848e3c67f4a04c4a2986c8924/src/backend/web/routes/index.js#L50) `nginx` wasn't able to serve anything.

A solution for this (@c3ho's suggestion) could be adding all the new locations as [nested locations](https://xintron.se/code/nginx-nested-locations-and-pagination) of `/`.
With this, all the new locations would be able to use:
```nginx
root /var/www/data;
```
This is what what we did to get it to work:
```nginx
# Non-cached routes /admin /user /health /auth
location ~ ^/(admin|user|health|auth) {
  proxy_cache_bypass 1;
  proxy_pass http://telescope_staging:3000;
}                                                                                                                              
                                                                                                                                     
# Specify cache behaviour for Gatsby's public build files.                                                                         
# See docs in https://www.gatsbyjs.org/docs/caching.                                                                           
                                                                                                                                     
# Static content                                                                                                               
location / {
  # Directory from which static content is served
  root /var/www/data;

  # 1. Don't cache HTML: https://www.gatsbyjs.org/docs/caching/#html
    location ~* \.(?:html)$ {
      expires epoch;
      add_header Cache-Control "public, max-age=0, must-revalidate";
    }
  # 2. Don't cache /page-data (including app-data.json): https://www.gatsbyjs.org/docs/caching/#page-data
    location /page-data {
      expires epoch;
      add_header Cache-Control "public, max-age=0, must-revalidate";
    }
  # 3. Don't cache the service worker /sw.js script, see:
  # https://www.gatsbyjs.org/docs/caching/#javascript-and-css
    location = /sw.js {
      expires epoch;
      add_header Cache-Control "public, max-age=0, must-revalidate";
    }
  # 4. Files in static/ can be cached forever: https://www.gatsbyjs.org/docs/caching/#static-files
    location /static {
      expires max;
      add_header Cache-Control "public, max-age=31536000, immutable";
    }
  # 5. Cache js and css forever, see
  # https://www.gatsbyjs.org/docs/caching/#javascript-and-css
    location ~* \.(?:js|css)$ {
      expires max;
      add_header Cache-Control "public, max-age=31536000, immutable";
    }
  # Try serving static content, and if not found continue with @proxy
  try_files $uri $uri/ @proxy;
}                                                                                                                                     

location @proxy {
  proxy_cache telescope_cache;
...
```
Also, some of the new added locations do the same, so I think we could try to join them together in one location using the OR operator as we do here:
```nginx
location ~ ^/(admin|user|health|auth) {
```

PR #1075, opened 2020-04-25
Add flag routes and function to backend

### LINKED ISSUE #1071: Add /feeds/:id/flag route to flag/unflag feeds and update backend
An admin user should be able to flag a feed, and have it not appear in the frontend, or get returned as one of the feeds in our various queries or other data routes.

We will need the following routes, accessible only to admin users:

1. `PUT /feeds/:id/flag` - would set the feed with `id` as flagged
2. `DELETE /feeds/:id/flag` - would unflag the feed with `id

In Redis, we can add a new set, `t:feeds:flagged` to manage these.  Given a feed that we want to flag (e.g., `id` of `NirlSYranl`), we'd add the key `NirlSYranl` to the set `t:feeds:flagged`.  When we delete a feed from Redis, we'd need to delete it from both `t:feeds` and perhaps `t:feeds:flagged` as well.

Our code for getting "all" feeds should be updated so that it uses [SDIFF](https://redis.io/commands/sdiff) to get the set of all feeds that ARE `t:feeds` but NOT in `t:feeds:flagged`.  This will automatically exclude all flagged feeds when we do processing.

Flagged feeds should not show up in the UI, and should not be returned as part of normal API requests.  It would be the same as if they didn't exist.

c3ho:
## Issue This PR Addresses
Fixes #1071 

## Type of Change

## Description
Adds feature to flag/unflag feeds by hitting the `/feed/:id/flag` route

Done:
Added functions to `storage.js` to flag/unflag
Added functions to `feed.js`
Added tests to `feed.test.js` and `feeds.test.js`
Added routes to `/feed`

**Two ways to test:**
Run my tests

_or..._ 
1. Find an admin account
2. Use `redis-cli smembers t:feeds` to get a list of `feedId` and pick one of them. Keep note of the # of feeds in `t:feeds`
3. While logged in as admin, hit the route `(telescopeUrl)/feeds/(feedId)/flag`. Please use the below screenshot as reference to test `PUT`/`DELETE` routes
image
4. Use `redis-cli smembers t:feeds` again after a `PUT` method. The count should be the previous # you got - 1
5. Verify the feed has moved with `redis-cli smembers t:feeds:flagged`. There should be one lonely feed in there with the `feedId` used in the PUT method
6. Use `redis-cli smembers t:feeds` again after a `DELETE` method. The count should be back to your starting count.
7. Verify the feed has moved with `redis-cli smembers t:feeds:flagged`. The lonely feed should have joined its friends again

## Checklist

### DIFF
--- a/src/backend/data/feed.js
+++ b/src/backend/data/feed.js
@@ -12,6 +12,9 @@ const {
   removePost,
   getPost,
   getPosts,
+  getFlaggedFeeds,
+  setFlaggedFeed,
+  unsetFlaggedFeed,
 } = require('../utils/storage');
 
 const { deletePost } = require('../utils/elastic');
@@ -112,6 +115,22 @@ class Feed {
   }
 
   /**
+   * Flags the feed, preventing posts from the feed to be displayed
+   * Returns a Promise<Boolean>
+   */
+  flag() {
+    return setFlaggedFeed(this.id);
+  }
+
+  /**
+   * Unflags the feed, allowing posts from the feed to be displayed
+   * Returns a Promise<Boolean>
+   */
+  unflag() {
+    return unsetFlaggedFeed(this.id);
+  }
+
+  /**
    * Creates a new Feed object by extracting data from the given feed-like object.
    * @param {Object} feedData - an Object containing the necessary fields.
    * Returns the newly created Feed's id as a Promise<String>
@@ -155,11 +174,20 @@ class Feed {
   }
 
   /**
-   * Returns all the feeds
+   * Returns all unflagged feeds
    * Returns a Promise<Feeds>
    */
   static async all() {
     const ids = await getFeeds();
+    return Promise.all(ids.map(Feed.byId));
+  }
+
+  /**
+   * Returns all flagged feeds
+   * Returns a Promise<Feeds>
+   */
+  static async flagged() {
+    const ids = await getFlaggedFeeds();
     return Promise.all(ids.map(Feed.byId));
   }
 
--- a/src/backend/utils/storage.js
+++ b/src/backend/utils/storage.js
@@ -1,7 +1,8 @@
 const { redis } = require('../lib/redis');
-
+const { logger } = require('./logger');
 // Redis Keys
 const feedsKey = 't:feeds';
+const flaggedFeedsKey = 't:feeds:flagged';
 const postsKey = 't:posts';
 
 // Namespaces
@@ -25,6 +26,10 @@ module.exports = {
    * Feeds
    */
   addFeed: async (feed) => {
+    // Check if feed being added already exists in flagged feeds set
+    // If it is, do nothing
+    if (await redis.sismember(flaggedFeedsKey, feed.id)) return;
+
     const key = createFeedKey(feed.id);
     await redis
       .multi()
@@ -54,6 +59,8 @@ module.exports = {
 
   getFeeds: () => redis.smembers(feedsKey),
 
+  getFlaggedFeeds: () => redis.smembers(flaggedFeedsKey),
+
   getFeed: (id) => redis.hgetall(feedNamespace.concat(id)),
 
   getFeedsCount: () => redis.scard(feedsKey),
@@ -69,12 +76,23 @@ module.exports = {
    */
   removeFeed: async (id) => {
     const key = createFeedKey(id);
-    await redis
-      .multi()
-      .hdel(key, 'id', 'author', 'url', 'user', 'link', 'etag', 'lastModified')
-      .srem(feedsKey, id)
-      .exec();
+    // Checks which set the feed is currently in
+    const redisKey = (await redis.sismember(feedsKey, id)) ? feedsKey : flaggedFeedsKey;
+    try {
+      await redis
+        .multi()
+        .hdel(key, 'id', 'author', 'url', 'user', 'link', 'etag', 'lastModified')
+        .srem(redisKey, id)
+        .exec();
+    } catch (error) {
+      logger.error({ error }, `Error removing Feed ${id} from Redis`);
+      throw new Error(`Error trying to remove feed from Redis`);
+    }
   },
+
+  setFlaggedFeed: (id) => redis.smove(feedsKey, flaggedFeedsKey, id),
+
+  unsetFlaggedFeed: (id) => redis.smove(flaggedFeedsKey, feedsKey, id),
 
   isInvalid: (id) => redis.exists(createInvalidFeedKey(id)),
 
--- a/src/backend/web/routes/feeds.js
+++ b/src/backend/web/routes/feeds.js
@@ -52,6 +52,27 @@ feeds.get('/:id', async (req, res) => {
   }
 });
 
+feeds.put('/:id/flag', protectAdmin(), async (req, res) => {
+  const { id } = req.params;
+  try {
+    const feed = await Feed.byId(id);
+    if (!feed) {
+      return res.status(404).json({
+        message: `Feed for Feed ID ${id} doesn't exist.`,
+      });
+    }
+    await feed.flag();
+    return res.status(200).json({
+      message: `Feed successfully flagged`,
+    });
+  } catch (error) {
+    logger.error({ error }, 'Unable to flag feed in Redis');
+    return res.status(503).json({
+      message: 'Unable to flag Feed',
+    });
+  }
+});
+
 feeds.post('/', protect(), async (req, res) => {
   const feedData = req.body;
   const { user } = req;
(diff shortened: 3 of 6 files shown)


src/backend/utils/storage.js
```diff
@@ -3,6 +3,7 @@ const { redis } = require('../lib/redis');
 // Redis Keys
 const feedsKey = 't:feeds';
 const postsKey = 't:posts';
+const flaggedFeedsKey = 't:feeds:flagged';
```
humphd, line 6: Move this up a line to sit with `feedsKey`

src/backend/utils/storage.js
```diff
@@ -76,6 +77,10 @@ module.exports = {
       .exec();
   },
 
+  flagFeed: (id) => redis.smove(feedsKey, flaggedFeedsKey, id),
+
+  unflagFeed: (id) => redis.smove(flaggedFeedsKey, feedsKey, id),
```
humphd, line 82: I like the idea of doing `smove`, but I have a question: what other functions touch `feedsKey` that need to be updated to also know about `flaggedFeedsKey`?  For example, if we delete a feed, it could be in one or the other, so we'd need to do it on both sets.  Are there other cases?

We'll also need a way to get all flagged feeds.

src/backend/data/feed.js
```diff
@@ -12,6 +12,9 @@ const {
   removePost,
   getPost,
   getPosts,
+  getFlaggedFeeds,
+  setFlaggedFeed,
+  setUnflaggedFeed,
```
humphd, line 17: `unsetFlaggedFeed`?

src/backend/data/feed.js
```diff
@@ -111,6 +114,22 @@ class Feed {
     return (await isDelayed(this.id)) === '1';
   }
 
+  /**
+   * Flags the feed, preventing posts from the feed to be displayed
+   * Returns a Promise<Boolean>
+   */
+  setFlagged() {
```
humphd, line 121: `flag() {...}`

src/backend/data/feed.js
```diff
@@ -111,6 +114,22 @@ class Feed {
...
+    return setFlaggedFeed(this.id);
+  }
+
+  /**
+   * Unflags the feed, allowing posts from the feed to be displayed
+   * Returns a Promise<Boolean>
+   */
+  setUnflagged() {
```
humphd, line 129: `unflag() {...}`

src/backend/data/feed.js
```diff
@@ -155,14 +174,23 @@ class Feed {
...
   static async all() {
     const ids = await getFeeds();
     return Promise.all(ids.map(Feed.byId));
   }
 
+  /**
+   * Returns all flagged feeds
+   * Returns a Promise<Feeds>
+   */
+  static async allFlagged() {
```
humphd, line 189: what about if we have `all()` and `flagged()`?

src/backend/utils/storage.js
```diff
@@ -69,13 +72,26 @@ module.exports = {
...
     const key = createFeedKey(id);
+    // If feed is in t:feeds
+    if (await redis.sismember(feedsKey, id)) {
+      await redis
+        .multi()
+        .hdel(key, 'id', 'author', 'url', 'user', 'link', 'etag', 'lastModified')
+        .srem(feedsKey, id)
+        .exec();
+    }
+    // If feed is in t:feeds:flagged
```
humphd, line 83: Should we wrap this in an `else` since it's only in one or the other, not both, right?

src/backend/web/routes/feeds.js
```diff
@@ -52,6 +52,27 @@ feeds.get('/:id', async (req, res) => {
   }
 });
 
+feeds.put('/:id/flag', protectAdmin(true), async (req, res) => {
```
humphd, line 55: Remove the `true` in `protectAdmin()`, that's only if we want to redirect a user vs. JSON route.

src/backend/web/routes/feeds.js
```diff
@@ -108,4 +129,21 @@ feeds.delete('/:id', protect(), async (req, res) => {
   }
 });
 
+feeds.delete('/:id/flag', protectAdmin(true), async (req, res) => {
```
humphd, line 132: Here too, remove `true` arg.

test/feed.test.js
```diff
@@ -218,6 +224,49 @@ describe('Post data class tests', () => {
...
+      // test setFlagged()
+      await Promise.all([feed.setFlagged(), feed2.setFlagged()]);
+      unFlaggedFeeds = await Feed.all();
+      expect(unFlaggedFeeds.length).toBe(1);
+
+      let flaggedFeeds = await Feed.allFlagged();
+      expect(flaggedFeeds.length).toBe(2);
+
+      // Flagged feeds should have same data as feed + feed2
+      expect(flaggedFeeds[0].id).toBe(feed.id);
```
humphd, line 253: Can you do this check without relying on the order?  I'm not sure we can always assume we'll get back the values in these positions.  A set has no order.

c3ho: I changed it to the below
```
expect(flaggedFeeds.some((flaggedFeed) => flaggedFeed.id === feed.id)).toBe(true);
expect(flaggedFeeds.some((flaggedFeed) => flaggedFeed.id === feed2.id)).toBe(true);
```

src/backend/utils/storage.js
```diff
@@ -69,13 +72,27 @@ module.exports = {
...
-      .srem(feedsKey, id)
-      .exec();
+    // If feed is in t:feeds
+    if (await redis.sismember(feedsKey, id)) {
+      await redis
+        .multi()
+        .hdel(key, 'id', 'author', 'url', 'user', 'link', 'etag', 'lastModified')
+        .srem(feedsKey, id)
+        .exec();
+    } else {
```
humphd, line 82: Follow-up question: if a feed isn't in either set, do we care?  That is, will we do the right thing, or crash?  Do we need another check for membership?

Probably this is fine, but maybe add a test?

c3ho: Will add a test. I don't think it will error, should just return `0` or `false`, but lets see

c3ho: I don't know if it is what you expected, but I added line 277 in `feed.test.js` and line 81 in `storage.test.js`

src/backend/utils/storage.js
```diff
@@ -26,31 +26,34 @@ module.exports = {
...
-        feed.link,
-        'etag',
-        feed.etag,
-        'lastModified',
-        feed.lastModified
-      )
-      .sadd(feedsKey, feed.id)
-      .exec();
+    // Check if feed being added already exists in flagged feeds set
+    if (!(await redis.sismember(flaggedFeedsKey, feed.id))) {
```
humphd, line 30: This is probably right, but one thing it will do is show that a feed was added successfully, but in reality we'll skip adding it.  Should we error?  I'm not sure we should.  But this will be a kind of "silent" no-op/failure.

c3ho: what if we return the redis operation values?

For example when you try to add a Feed with `addFeed(feed)` and it is successful
image

vs adding a Feed that exists in `t:feeds:flagged` set
image

test/feeds.test.js
```diff
@@ -279,3 +279,126 @@ describe('test DELETE /feeds/:id endpoint', () => {
...
+      .put(`/feeds/${feeds[0].id}/flag`)
+      .send()
+      .set('Accept', 'application/json');
+    expect(res.status).toEqual(403);
+  });
+
+  it('should respond with a 404 status trying to flag feed that does not exist when logged in as admin user', async () => {
+    loginAdmin('Johannes Kepler', 'user1@example.com');
+    const res = await request(app)
+      .put(`/feeds/${123456}/flag`)
```
humphd, line 337: Doesn't need to be a templated interpolation, just embed it in the string:

```js
.put(`/feeds/123456/flag`)
```

humphd: This got missed in your last update.

src/backend/utils/storage.js
```diff
@@ -25,35 +26,40 @@ module.exports = {
...
```
humphd, line 30: Nit: if you have to indent your entire function, flip your logic and return early.  I won't block on this, but get into the habit of keeping your code left-aligned.

src/backend/utils/storage.js
```diff
@@ -69,13 +75,23 @@ module.exports = {
...
+    // Checks which set the feed is currently in
+    const redisKey = (await redis.sismember(feedsKey, id)) ? feedsKey : flaggedFeedsKey;
+    try {
+      await redis
+        .multi()
+        .hdel(key, 'id', 'author', 'url', 'user', 'link', 'etag', 'lastModified')
+        .srem(redisKey, id)
+        .exec();
+    } catch (error) {
+      logger.error({ error }, `Error removing Feed ${id} from Redis`);
```
humphd, line 87: Should we not throw too?  I'm not sure if eating a failure is a good idea.  Log it, yes, but throw too.

c3ho: Added a throw on line 89, let me know if that is what you had in mind

c3ho:
I ran `npm test`, and this passed all tests locally. Not sure if I should rebase atm

manekenpix (approved):

src/backend/utils/storage.js
```diff
@@ -25,8 +26,12 @@ module.exports = {
    * Feeds
    */
   addFeed: async (feed) => {
+    // Check if feed being added already exists in flagged feeds set
+    // If it is, do nothing
+    if (await redis.sismember(flaggedFeedsKey, feed.id)) return;
+
     const key = createFeedKey(feed.id);
-    await redis
+    redis
```
humphd, line 34: Why remove the `await` here?

humphd (dismissed):
Looks great, one question.  Sorry for the slow review.

src/backend/utils/storage.js
c3ho: Whoops, it shouldn't be removed.

cindyledev (approved):
Will this ancient PR finally get merged now? :P

manekenpix (approved):

c3ho:
Thanks, going to squash and merge the PR since it received two approvals

Issue #1091, opened 2020-04-30
Improper entities in code blocks


humphd:
We're altering `<pre><code>...</code></pre>` blocks to use entities, and shouldn't.  Here's my recent blog post.  On the left, the original (notice the `=>` arrow function on line 2), and on the right, Telescope's version:

<img width="1512" alt="Screen Shot 2020-04-30 at 10 05 27 AM" src="https://user-images.githubusercontent.com/427398/80720099-6faf2e80-8aca-11ea-8431-d069befe1616.png">

This is probably our HTML sanitizer trying to protect us.

For writing a test case, here is what the HTML actually looks like:

```
<span class="hljs-string">', ...].map((number) =>;
    twilio.messages.create({
      body: '...'
</span>
```

smilegodly:
Is it alright if I work on this one ?

manekenpix:
> Is it alright if I work on this one ?

@smilegodly absolutely!

PR #1157, opened 2020-10-06
Fixing code block entities

smilegodly:
## Issue This PR Addresses
Fixes: https://github.com/Seneca-CDOT/telescope/issues/1091

## Type of Change

## Description
This fix cleans up the code blocks within our posts at runtime to rid the source code of any HTML entity. If there is a better way I should have gone about implementing this fix please let me know!

Before the fix. Notice the encoded HTML entities within the code `&lt;` , `&gt;`, `&amp;`
old

After the fix.
new

## Checklist

### DIFF
--- a/src/backend/utils/html/replace-entities.js
+++ b/src/backend/utils/html/replace-entities.js
@@ -1,7 +1,7 @@
 const entities = require('entities');
 
 function decode(codeElement) {
-  // decode twice for double encoded entities
+  // decode twice for double encoded html entities
   const result = entities.decodeHTML(codeElement.innerHTML);
   return entities.decodeHTML(result);
 }
@@ -17,5 +17,5 @@ module.exports = function (dom) {
     fixedCodeElement = decode(code);
     code.innerHTML = fixedCodeElement;
   });
-  return String(fixedCodeElement); // return result to the tests
+  return `<code>${fixedCodeElement}</code>`; // return decoded string to the tests
 };
--- a/src/backend/utils/html/replace-entities.js
+++ b/src/backend/utils/html/replace-entities.js
@@ -2,22 +2,22 @@ const entities = require('entities');
 
 function decode(codeElement) {
   // decode twice
-  console.log('replace-entities: INSIDE OF THE DECODE FUNCTION!');
-  let result = entities.decodeHTML(codeElement.innerHTML);
-  result = entities.decodeHTML(result);
-  return result;
+  const result = entities.decodeHTML(codeElement.innerHTML);
+  return entities.decodeHTML(result);
 }
 // This function is meant query code elements, decode the entities, and then change the dom with those changes
 module.exports = function (dom) {
-  console.log('replace-entities: INSIDE OF THE FUNCTION!');
   if (!(dom && dom.window && dom.window.document)) {
-    return;
+    return null;
   }
+
+  let fixedCodeElement = {};
 
   dom.window.document.querySelectorAll('code').forEach((code) => {
     console.log('replace-entities: INSIDE OF THE FOREACH FUNCTION!');
-    const fixedCodeElement = decode(code);
+    fixedCodeElement = decode(code);
     code.innerHTML = '';
     code.innerHTML = fixedCodeElement;
   });
+  return String(fixedCodeElement);
 };
--- a/src/backend/utils/html/replace-entities.js
+++ b/src/backend/utils/html/replace-entities.js
@@ -1,0 +1,7 @@
+const entities = require('entities');
+
+module.exports = function (dom, codeBlock) {
+  // decode twice
+  const result = entities.decodeHTML(codeBlock);
+  return entities.decodeHTML(result);
+};
--- a/src/backend/utils/html/replace-entities.js
+++ b/src/backend/utils/html/replace-entities.js
@@ -1,7 +1,23 @@
 const entities = require('entities');
 
-module.exports = function (dom, codeBlock) {
+function decode(codeElement) {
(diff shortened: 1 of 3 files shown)


smilegodly:
For some reason not working on the [Vercel preview](https://telescope-eo9rnzt5c.vercel.app/) but works for me locally using docker.

manekenpix:
> For some reason the fix is not being applied on the [Vercel preview](https://telescope-eo9rnzt5c.vercel.app/) but works for me locally using docker.

@smilegodly you made changes to the back-end, and `Vercel` uses telescope's back-end (on the staging server) to tests front-end.

smilegodly:
@manekenpix 
Ah okay, I understand.
Thank you!

src/backend/utils/html/replace-entities.js
```diff
@@ -0,0 +1,21 @@
+const entities = require('entities');
+
+function decode(codeElement) {
+  // decode twice for double encoded entities
+  const result = entities.decodeHTML(codeElement.innerHTML);
+  return entities.decodeHTML(result);
+}
+
+module.exports = function (dom) {
+  if (!(dom && dom.window && dom.window.document)) {
```
manekenpix, line 10: I wonder if it'd be possible to use the [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) here
```suggestion
  if (!(dom && dom.window?.document)) {
```

src/backend/utils/html/replace-entities.js
```diff
@@ -0,0 +1,21 @@
...
+    return null;
+  }
+
+  let fixedCodeElement = {};
+
+  dom.window.document.querySelectorAll('code').forEach((code) => {
+    fixedCodeElement = decode(code);
+    code.innerHTML = fixedCodeElement;
```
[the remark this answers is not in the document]
smilegodly: The reason I put the `fixedCodeElement` there was so that I could return that value back to the tester 

https://github.com/Seneca-CDOT/telescope/blob/9f5747786f85a8cc7525766fbeecece1c520cab9/src/backend/utils/html/replace-entities.js#L20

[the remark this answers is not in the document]
manekenpix: How about:
```js
code.innerHTML = decode(code);
return code.innerHTML;
```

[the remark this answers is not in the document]
manekenpix: I mean in your tests, the same way that [syntax-highlight does it](https://github.com/Seneca-CDOT/telescope/blob/0764c770d57641e2bd866fb80e9beaa3253873d6/test/syntax-highlight.test.js#L8).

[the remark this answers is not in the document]
smilegodly: > How about:
> 
> ```js
> code.innerHTML = decode(code);
> return code.innerHTML;
> ```
If we return the innerHTML of the element to the tests, I'm pretty sure that the decoded text that gets returned will become encoded again because its HTML.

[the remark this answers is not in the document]
manekenpix: Sorry, I meant in your test. You could avoid returning here in your code, and add a return in your test.
Check out how `syntax-highlight` does it:
[Here](https://github.com/Seneca-CDOT/telescope/blob/0764c770d57641e2bd866fb80e9beaa3253873d6/src/backend/utils/html/syntax-highlight.js#L38) is the tool (it doesn't return anything)
And [here](https://github.com/Seneca-CDOT/telescope/blob/0764c770d57641e2bd866fb80e9beaa3253873d6/test/syntax-highlight.test.js#L8) is the test (using return to test the code)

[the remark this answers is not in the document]
smilegodly: If we return `dom.window.document.body.innerHTML`  like in `syntax-highlight` for this code , the result being passed back would be encoded, which isn't useful as we expect decoded html for the test.

src/backend/utils/html/replace-entities.js
humphd: I don't think we should introduce this, as it depends on node 14.  Wait until it's baked a bit longer and found its way into an LTS.

manekenpix: :+1:

manekenpix (approved):

src/backend/utils/html/replace-entities.js
```diff
@@ -0,0 +1,21 @@
...
+  });
+  return String(fixedCodeElement); // return result to the tests
```
humphd, line 20: Should this be `return fixedCodeElement.outerHTML;`?

humphd: Or am I wrong?  I'm unclear on the `String()` conversion though.

smilegodly: I believe that the character entities `>` `<` `&` inside the code elements will return encoded if we do `return fixedCodeElement.outerHTML;` 

The `String()` conversion was just me being explicit. We can *probably* get rid of it and it would be fine.

humphd: Do you want to add a test that confirms that `typeof result === 'string'` and we'll know for sure?  I bet we don't need it

smilegodly: Great idea, I'll add those.

I'm still looking for a better way to return the decoded code elements to the tests without encoding it again.

humphd (approved):

manekenpix (approved):

Issue #1055, opened 2020-04-23
Increase contrast in post header


humphd:
<img width="724" alt="Screen Shot 2020-04-23 at 2 03 36 PM" src="https://user-images.githubusercontent.com/427398/80133483-4858de80-856b-11ea-87e3-38cf4601beb7.png">

Caused by:

<img width="354" alt="Screen Shot 2020-04-23 at 2 04 21 PM" src="https://user-images.githubusercontent.com/427398/80133521-59095480-856b-11ea-9240-43b75d6a3701.png">

This needs to get fixed so it's more accessible.

agarcia-caicedo:
I played around with some colours, and I think its better to just use two colours for the post header. Otherwise the only colours we could use would be a blue, white, and black.

![image](https://user-images.githubusercontent.com/49992710/81254998-8853a300-8ffa-11ea-86d7-d25f95045079.png)

![image](https://user-images.githubusercontent.com/49992710/81254926-622e0300-8ffa-11ea-8be4-828259e3b74f.png)
This one keeps the dark to #002944, but changes the blue to #5995C9

![image](https://user-images.githubusercontent.com/49992710/81255116-da94c400-8ffa-11ea-8ab6-59dc7f5df06d.png)

![image](https://user-images.githubusercontent.com/49992710/81255144-eb453a00-8ffa-11ea-97ba-9c819b456503.png)
This one keeps the light colour to #A4D4FF, but changes the blue to #2A5A83

![image](https://user-images.githubusercontent.com/49992710/81255408-8dfdb880-8ffb-11ea-96ab-a8c61ad97508.png)

![image](https://user-images.githubusercontent.com/49992710/81255494-c43b3800-8ffb-11ea-9cde-6970e7de4217.png)
This one keeps the blue at #3670A5, but changes the light to #AFFDFC

The blue doesn't work well with any dark colour, so I can't present a variation

cindyledev:
Gives me dark mode vibes #77 ;)

humphd:
@strawberries73 is going to give this a try.

strawberries73:
@humphd  Do I need to stick with the "blue" theme?

humphd:
Yes, this isn't about changing colour, only contrast.  @agarcia-caicedo has already done colour theming in other PRs.

humphd:
Confirmed this was fix with light theme changes, closing.  Sorry Jenn, we'll find you another bug.

PR #1159, opened 2020-10-06
Fixes #1088: Add Icons for Admin to delete/flag a feed

### LINKED ISSUE #1088: Add icons for admins to flag and delete feeds from timeline
If an admin is reading the timeline, and notice something problematic, they should be able to flag (#1075) or delete the feed right there.  I suggest we add two new icons to the upper-right corner of the blue post header bar.  Here's a rough sketch of what I mean:

Screen Shot 2020-04-28 at 12 33 56 PM

c3ho:
## Issue This PR Addresses
Fixes #1088

## Type of Change

## Description
Not sure if wiring up the buttons is part of this PR, I had use inline styling for the icons as I was not able to change the icon size or colour through the classes way. The PR can be checked in Vercel Deployment (I think). Otherwise it will require a running local backend + `npm run build` and logging in with user1 // user1pass. `npm run develop` will not work for this.

EDIT: I'm going to split this up into two parts, one for the icons and the next one for functionality(wiring up buttons + displaying a confirmation/popup message)

When not logged in / user is not admin
image

When logged in / user is admin
image

## Checklist

### DIFF
--- a/src/frontend/src/components/AdminButtons/AdminButtons.jsx
+++ b/src/frontend/src/components/AdminButtons/AdminButtons.jsx
@@ -1,0 +1,40 @@
+import React, { useContext } from 'react';
+import { makeStyles } from '@material-ui/core/styles';
+import { IconButton } from '@material-ui/core';
+import FlagIcon from '@material-ui/icons/Flag';
+import DeleteIcon from '@material-ui/icons/Delete';
+import { UserStateContext } from '../../contexts/User/UserContext';
+
+const useStyles = makeStyles(() => ({
+  iconDiv: {
+    float: 'right',
+  },
+  iconSpan: {
+    paddingLeft: '10px',
+    paddingRight: '10px',
+  },
+}));
+
+function AdminButtons() {
+  const classes = useStyles();
+  const user = useContext(UserStateContext);
+
+  return (
+    user.isAdmin && (
+      <div className={classes.iconDiv}>
+        <span className={classes.iconSpan}>
+          <IconButton size="small">
+            <FlagIcon fontSize="large" />
+          </IconButton>
+        </span>
+        <span className={classes.iconSpan}>
+          <IconButton size="small">
+            <DeleteIcon fontSize="large" />
+          </IconButton>
+        </span>
+      </div>
+    )
+  );
+}
+
+export default AdminButtons;
--- a/src/frontend/src/components/Post/Post.jsx
+++ b/src/frontend/src/components/Post/Post.jsx
@@ -5,6 +5,7 @@ import useSWR from 'swr';
 import { makeStyles } from '@material-ui/core/styles';
 import { Box, Grid, Typography, ListSubheader } from '@material-ui/core';
 import './telescope-post-content.css';
+import AdminButtons from '../AdminButtons';
 
 const useStyles = makeStyles((theme) => ({
   root: {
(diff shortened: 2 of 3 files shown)


agarcia-caicedo:
Based on the material-ui docs, I think the only way to change size and colour for icons is in-line, so no worries. Some of @humphd's follow up comments in #1088  kinda suggest the wiring to be part of this pr, but if this is not the case, I think this is fine. 

I do have some ux nits:
- The target areas for the buttons seems a bit too close together, so some more spacing between them would be nice
- There is contrast issues with the buttons, but that's more a result from our theme. (See #1055 )

humphd:
Can we get some screenshots in this PR?

humphd:
Sorry, hit close by mistake.

humphd:
These colours (especially the red) don't really match our theme, do they?  The trashcan looks like it's disabled, the flag is way to dominant.  Can we improve this @agarcia-caicedo?

agarcia-caicedo:
@humphd I think we can change it to the same blue as the title while we find a way to fix our contrast issues in the theme

agarcia-caicedo (approved):

src/frontend/src/components/Post/Post.jsx
```diff
@@ -84,6 +93,18 @@ const Post = ({ id, html, author, url, title, date, link }) => {
   return (
     <Container className={classes.root}>
       <ListSubheader className={classes.header}>
+        <div className={classes.iconDiv}>
```
humphd, line 96: This extra UI should only show for an admin user, right?  We need that logic, too.

humphd (changes requested):

src/frontend/src/components/Post/Post.jsx
```diff
@@ -81,9 +92,27 @@ const Post = ({ id, html, author, url, title, date, link }) => {
...
+          </IconButton>
+        </span>
+      </div>
+    );
+  }
+
   return (
     <Container className={classes.root}>
       <ListSubheader className={classes.header}>
+        {user.isAdmin ? displayPostButtons() : null}
```
humphd, line 115: This should be done with a new component, and let it decide to render or not based on the `user` prop:

```jsx
<AdminButtons user={user} />
```

Then create a new component `admin-buttons.jsx` and refactor your function above into a function component:

```jsx
function AdminButtons({ user }) {
    if(!user.isAdmin) {
      return null;
    }

    return (
      <div className={classes.iconDiv}>
        <span className={classes.iconSpan}>
          <IconButton size="small">
            <FlagIcon fontSize="large" />
          </IconButton>
        </span>
        <span className={classes.iconSpan}>
          <IconButton size="small">
            <DeleteIcon fontSize="large" />
          </IconButton>
        </span>
      </div>
    );
}
```

humphd (changes requested):

src/frontend/src/components/Post/Post.jsx
c3ho: Got it, I opted to not pass user as a prop since we have the `UserStateContext` that can be used

src/frontend/src/components/AdminButtons/AdminButtons.jsx
```diff
@@ -0,0 +1,41 @@
...
+}));
+
+function AdminButtons() {
+  const classes = useStyles();
+  const user = useContext(UserStateContext);
+  if (!user.isAdmin) {
+    return null;
+  }
+
+  return (
```
manekenpix, line 25: you can use [short circuit evaluation](https://dev.to/superhackerman/short-circuit-evaluation-with-react-3dn4) here and remove the if statement above.
```jsx
return (
  {user.isAdmin && ...
```

c3ho:
Aside from the conflict issues, is this good to go?

humphd (approved):

manekenpix (approved):

cindyledev (approved):

PR #1177, opened 2020-10-11
Fixes #1169: Fix search posts

### LINKED ISSUE #1169: 'Search' returns latest posts instead of results for searched input.
**What happened**:
I entered a string in the search box and it returned the latest posts instead of actual results

Peek 2020-10-08 11-58

I should have seen posts related to the entered string

Go to https://dev.telescope.cdot.systems/search, enter any string in the search box and compare results to latest posts in `/`

c3ho:
## Issue This PR Addresses
Fixes #1169 

## Type of Change

## Description
This PR allows users to search posts again.

I tried to change as little code as possible for this PR, such as assigning the passed results from search as props and then assigning it to `posts`. `Load More Posts` button will not show anymore if we're searching posts.

## Checklist

### DIFF
--- a/src/frontend/src/components/Posts/Posts.jsx
+++ b/src/frontend/src/components/Posts/Posts.jsx
@@ -34,7 +34,7 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-const Posts = () => {
+const Posts = (props) => {
   const classes = useStyles();
   const savedCallback = useRef();
   const [initNumPosts, setInitNumPosts] = useState(0);
@@ -46,6 +46,7 @@ const Posts = () => {
   const [posts, setPosts] = useState([]);
   const [numPages, setNumPages] = useState(1);
   const [endOfPosts, setEndOfPosts] = useState(false);
+  const { searchPostResults } = props;
   const snackbarMessage = 'There is new content available!';
 
   // Pagination
@@ -78,8 +79,12 @@ const Posts = () => {
         setLoading(false);
       }
     }
+    if (searchPostResults) {
+      setPosts(searchPostResults);
+    } else {
+      getPosts();
+    }
 
-    getPosts();
     // Disabling the eslint check as nextPageLink and posts will cause the page to not render properly
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [telescopeUrl, numPages]);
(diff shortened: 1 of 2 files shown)


src/frontend/src/components/Posts/Posts.jsx
```diff
@@ -78,8 +79,14 @@ const Posts = () => {
         setLoading(false);
       }
     }
+    if (searchPostResults) {
+      setPosts(searchPostResults);
+      console.log('hi');
```
humphd, line 84: Hi

humphd (changes requested):

src/frontend/src/components/Posts/Posts.jsx
```diff
@@ -78,8 +79,14 @@ const Posts = () => {
...
+    } else {
+      getPosts();
+      console.log('bye');
```
raygervais, line 87: Bye

raygervais (changes requested):

src/frontend/src/components/Posts/Posts.jsx
```diff
@@ -46,6 +46,7 @@ const Posts = () => {
   const [posts, setPosts] = useState([]);
   const [numPages, setNumPages] = useState(1);
   const [endOfPosts, setEndOfPosts] = useState(false);
+  const { searchPostResults } = props;
```
humphd, line 49: So it makes more sense to pass in the search results here vs. doing it in this component?

I would have thought doing:

```jsx
<Posts search="This is my search text" author="Calvin Ho" />
```

would have been more logical.  But maybe the code doesn't allow this easily?

c3ho: We have two separate components, depending on what type of search is being carried out: if by posts, we reuse the post component and just pass it everything,

If they search by author: we have an author results component that is displayed instead with no posts, so I'm not sure we'll need to pass the author information

humphd: OK, we can evolve this in follow up PRs then.  Let's keep this one focused on getting results into Posts.

src/frontend/src/components/Posts/Posts.jsx
```diff
@@ -212,7 +220,8 @@ const Posts = () => {
 
 Posts.propTypes = {
   className: PropTypes.string,
-  posts: PropTypes.array,
+  searchResults: PropTypes.array,
+  searchPostResults: PropTypes.array,
```
humphd, line 224: Why do we have two props here?  You only appear to use `searchResults` in the component.  Let's remove `searchPostResults`

humphd: Another question re: naming - if what we're passing into this component on props are actually `posts` then maybe that's the right name vs. `searchResults` (i.e.,. it's more accurate, generic, and reusable for the `Posts` component)?

```jsx
<!-- No posts are being passed, load them internally -->
<Posts />

<!-- We are passing posts to be rendered -->
<Posts posts={...} />
```

humphd (changes requested):

src/frontend/src/components/Posts/Posts.jsx
c3ho: I wanted to use `posts`, however we are using that variable already in [posts](https://github.com/Seneca-CDOT/telescope/blob/master/src/frontend/src/components/Posts/Posts.jsx#L46) to declare an empty array and append to as we get more posts. React also yells at me to use object destructuring, unless you want me to ignore it and use something like `const (variable) = props.posts`

humphd: Gotcha.  Let's call it `searchResults` then, and you can use destructing if you want, sure.  Remove the second prop in proptypes.

humphd (dismissed):

cindyledev (dismissed):

c3ho:
Merging as this previously received 2 approvals prior to rebase.

PR #1192, opened 2020-10-13
Fixes #1070: Fix scroll overlap bug with header bars

### LINKED ISSUE #1070: Fix scroll overlap bug with header bars
Follow-up from #1034, we need to deal with the two header bars overlapping and not playing nicely together:

scroll-bug

PedroFonsecaDEV:
## Issue This PR Addresses
Fixes #1070 

## Type of Change

## Description

To solve the bug, I deleted the component HideOnScroll and the last Toolbar and changed the AppBar position to "static" both for Desktop and Mobile headers components.

## Checklist


agarcia-caicedo:
@humphd, are we ok with the loss of the sticky header? If the best solution is to get rid of one, this is probably the better call, since the fab button can allow for quick (though no as quick) access to the header. 

If we are ok with removing the sticky header, I'm ok with approving this.

humphd:
I don't want to totally lose the main header, no.  If you're deep into the posts, you have to go a long way up to find those links.  The fab takes you up, but not all the way.

I think we should find a way to have it stay anchored at the top, and the posts occupy less than the full viewport height.  Can we think of a site that does something like this, which we can use as a model?  Here's one:

https://novotel.accor.com/gb/usa/index.shtml

scroll-demo

The way that you have a single header, scroll down, then pick-up a second header (search fields) that stays with you as you scroll--that's kind of what I want, I think.

What do others think?  @PedroFonsecaDEV could you get that working here?

agarcia-caicedo:
Another thing to keep in mind, is that this direction might not be ideal for mobile. So while we could keep both bars on desktop, we might have to do something different on mobile. For example, in the site provided by @humphd, the mobile version only keeps the one bar

image

PedroFonsecaDEV:
Sure. I will work on it. Let's keep in touch.

PedroFonsecaDEV:
I believe everything is rebased and good to go. 
Please check the Vercel build. 

These changes will fix the bug(s).
Thanks.

src/frontend/src/components/ScrollToTop/ScrollToTop.jsx
```diff
@@ -58,7 +58,7 @@ export default function BackToTop(props) {
   const classes = useStyles();
   return (
     <>
-      <Toolbar id="back-to-top-anchor" className={classes.anchor} />
+      <Toolbar id="delete-in-the-future" className={classes.anchor} />
```
agarcia-caicedo, line 61: If the toolbar does nothing and should be deleted, why not remove it now?

agarcia-caicedo (changes requested):

agarcia-caicedo:
Something to keep in mind, with this change, in certain screens, a little over a quarter of the screen gets taken over by the headers. This doesn't led to an optimal reading experience, so we might need to change the size of the posts header to take up less room in a follow up issue.

image

src/frontend/src/components/ScrollToTop/ScrollToTop.jsx
PedroFonsecaDEV: There a lot of components with a high level of coupling, a mess on the banner, a lot of useEffects causing troubles. CSS that are causing child elements living outside and extremely far away from it's parents. So In the FUTURE I will check these components. I left some lines to help me comprehend the old ”logic”.

PedroFonsecaDEV:
Like I sad the CSS is a mess. I think that is missing a responsive query to medium size in some components.

humphd:
In a follow-up, one thing we could do with the post header is collapse it down to just the title as you scroll through the post, so it's much more compact, then expand it to full when you go to the top.

I agree that holding that much space at the top of the viewport hostage isn't ideal, but I also love not losing track of where you are as you move through the posts.

agarcia-caicedo:
> Like I sad the CSS is a mess. I think that is missing a responsive query to medium size in some components.

By some screens I meant my laptop, which is small, but it is still meets the large breakpoint

PedroFonsecaDEV:
@agarcia-caicedo please feel free to file a new issue like "delete this line" as a good first issue, It will be good for newcomers.

@humphd I believe that I fixed the issue-1070(Fix scroll overlap bug with header bars ) created on Apr 24 with this pull request. 

Are we good to merge it and let me continue to work on other issues?

src/frontend/src/components/ScrollToTop/ScrollToTop.jsx
humphd: I agree with @agarcia-caicedo, this is not a great solution.  What tends to happen with planning artifacts in source code is that they get lost.

A better solution here is to file an issue with all the details, then you can link it into a comment with `<!-- TODO: remove this as part of issue #1345 -->` or something.

@PedroFonsecaDEV could you file that issue now for me?

PedroFonsecaDEV: Okay. Thanks

agarcia-caicedo (approved):

manekenpix (approved):

PR #1196, opened 2020-10-14
Add Elasticsearch Mocking

humphd:
I'm going to get my current open source students to start working on Telescope, and it's going to be painful to have them all do the setup.  Most people working on things in the frontend or backend can probably do what they need with no Redis or Elastic, no Docker, etc.

This adds `npm run quick-start`, which mocks Redis, Elastic, and uses `env.example` automatically.  I'm hoping people can do `npm install && npm run quick-start` and have a working Telescope.

This mostly works, but I can't get the feed parser to pull jobs from the queue.  Likely it's missing Pub/Sub features in the Redis mock or something else that Bull needs.  What I probably need to do is to seed the Redis instance with some initial data so you can load the server and get back some feeds and posts.  Maybe I should try another mock redis, like https://github.com/yeahoffline/redis-mock.

I'm putting this up now because I'm not sure when I'll get back to it.  Hopefully in the next few weeks.


manekenpix:
I found [an interesting issue](https://github.com/OptimalBits/bull/issues/962) where people tried different solutions to the `bull`+`mocked-redis` problem. Maybe one of them could work with the changes in this branch.

humphd:
It's not possible to mock Redis the way I want and have Bull work, so I'm going to abandon that plan.  People will have to install a Redis server somehow.

However, I'm going to keep the work I did on mocking Elasticsearch, since it works, and you can run our server using a mock instance.

I've also left the fallback env loading when there is no env, since it's kind of nice to not have to set one.

This might mess up our tests, though.  I need to see what CI thinks.

manekenpix (approved):
Tested using `MOCK_ELASTIC=1` (redis on docker, and no elasticsearch service).

c3ho (approved):
Added env vars and was able to start backend with just redis on docker

manekenpix (approved):

c3ho:
Whoops sorry if I accidentally wrote over this.

c3ho (approved):

PR #1217, opened 2020-10-20
Use swr and reworking Posts, Post, and Search loading

### LINKED ISSUE #1201: Initial load times for content are horrible
Our app loads fast, but the content takes forever.  On my machine it can take 25s before the main content appears, which includes:

- hero banner image
- stats
- posts

When I profile it, it appears to all be blocking on network I/O (nothing else is happening):

Screen Shot 2020-10-14 at 6 45 42 PM

If I hit those routes via curl, they are super fast.  So I'm not sure why they suddenly conspire to take 25s together.

humphd:
Fixes #1201. I've wanted to try using https://github.com/vercel/swr for a while now, and decided to do an experiment.  This PR is not ready, but I wanted to get it build on Vercel so I could see how it works.

This PR does a few things:

- moves away from us manually managing our data fetching, and let's `swr` do it instead.  The way it works is to give you back a cached version of the data while it goes and gets an updated one (then cache that for next time).  The 2nd, 3rd, etc. load is extremely fast
- I'm reworking how `Post` and `Posts` work, so that we delegate more of the loading and don't block for so long.  Also, we can use `swr` here too, since posts can get cached and re-used
- I notice we still have the syntax highlighting stuff in the frontend, which we don't need anymore.

As I say, this needs more stuff added back (I ripped out most of `Posts` to do this).


humphd:
I think this is ready for a proper review.  I'll explain a bit more about what it does.

I'm adding https://swr.vercel.app/, which is a fantastic React Hook to manage your data fetching and caching.  The idea is, use cached data if we have it AND go get new data, so you always get fast results.

I've refactored Posts and Post, and we now have Posts, Timeline, and Post.  I wanted to re-use the logic of how our "posts" are shown for both the main timeline and also for search results.

The Post component is now responsible for loading it's own content from the server.  It accepts a url to a post on its props, then it manages getting the data and rendering it.  This speeds things up by not having so many parallel promises that we have to block on in Posts.  I also stripped out the redundant client-side syntax highlighting we were doing--it's all done server side now.

The Posts component has been trimmed down a lot.  It uses [useSWRInfinite](https://swr.vercel.app/docs/pagination#use-swr-infinite) to manage loading posts page-by-page.  The results are now returned/stored as an array of arrays (e.g., pages of posts).  I manage going to the next page using the `setPage()` function it provides, which loads in new data and merges it.  The rendering of the posts themselves happens in the Timeline component.

The Timeline renders pages of posts, and optionally shows a Load More button.  It's used by the Posts component and SearchPage, both of which are responsible for the getting the post data to show.  If a `nextPage` function is passed on props, we show a Load More button, which is a component I've spit out.  I use a trick to render the button at the end of the array of posts, so that when you click it, it will get deleted and recreated at the end of the newly inserted posts (e.g., you don't lose your place anymore when clicking Load More).

There's more I could do, but I think this might be good enough to start with, and we can build on it.

NOTE: we need better UI error messages, which we can do in follow-ups.

manekenpix (approved):
A much needed refactor. Awesome!

cindyledev (approved):
wow, the posts load instantly now O_O

humphd:
I filed https://github.com/Seneca-CDOT/telescope/issues/1220 on the error and loading handling, which I didn't do enough of here.

raygervais (approved):
Love everything about this

c3ho (approved):

Issue #1171, opened 2020-10-09
/stats routes shouldn't block the web server


humphd:
From Slack, I'm sick of how slow our startup time is.  It seems like it's due to how we load the posts.  Let's figure out what's taking so long.

> I'm sick of how long it takes to load the app on startup
> 95% of the time it takes is idle, waiting on network
> We should think about switching our approach: use Gatsby to re-build a static version of the whole site every time we get newer posts
> this thing should load instantly
> it's taking 13s to do these GET requests
> what on earth is it doing for all that time?
> getting the /posts response takes 143ms, but getting a /post takes 13000ms!
> to get 2K of data

raygervais:
Sharing here some thoughts, if we leverage a data source such as https://www.gatsbyjs.com/plugins/gatsby-source-rest-api/ and retrofit the back-end to provide ALL posts URLS (instead of paged), this could greatly improve load performance at the cost of storage, build time, and a terrible developer experience.

humphd:
I thought I'd fixed the slow load times, but it looks like there are multiple angles to this.

I now suspect that we're blocking the node event loop in the web server.  Here's a demo that shows it happen.  I do a page load on staging, and then try to hit the `/health` route manually via my terminal.  You can see that the entire web server is locked up on a blocking call somewhere:

![slow-load](https://user-images.githubusercontent.com/427398/96893794-144a7180-1459-11eb-8ac1-3e8ecb584d16.gif)

My guess is that it's in the stats calculation, where we `Promise.all()` on getting ever post out of redis.  This should be easy to confirm/deny by simply removing the stats call and retesting.  If it's not that, we should figure out what we're doing in the web server that locks things up so badly.

c3ho:
Can this be closed?

humphd:
The bug is still hiding in our code, but we don't trigger it.  I think the stats stuff is the problem.  I'll morph this.

HyperTHD:
The status microservice should hopefully take care of this

humphd:
This is going away with the backend, so I'm going to close.

PR #1234, opened 2020-10-23
Remove call to /stats/year from frontend for perf

humphd:
Part of #1171.  This is a first-step toward fixing our horrible performance on page load.  I'm getting rid of the dynamic call to `/year/stats`, which I suspect of blocking our event loop while it resolves a billion promises.

We can add it back once we rewrite it to be efficient.


c3ho:
Wow that’s fast

agarcia-caicedo (approved):

manekenpix (approved):

PR #1256, opened 2020-11-01
Fixes #1224: Show all posts for an author sorted by date when searching by Author

### LINKED ISSUE #1224: Show all posts for an author sorted by date when searching by Author
When I search "by Author" I really want to see all the posts we have for this user.  It would be great to either replace what we have now with a timeline of their posts, sorted by date (newest to oldest), or add that to what we have now.

joelazwar:
## Issue This PR Addresses
Issue #1224 

## Type of Change

## Description

Changes search by author function to display the initial result (container that links to feed) and the addition of the timeline of their posts
joelauthorsearch

works with with multiple persons with shared names, but is limited to max posts per page. For best results, search for author's full name.

## Checklist

### DIFF
--- a/src/backend/utils/indexer.js
+++ b/src/backend/utils/indexer.js
@@ -82,13 +82,14 @@ const search = async (textToSearch, filter = 'post') => {
         fields: createFieldsFromFilter(filter),
       },
     },
+    sort: [{ published: { order: 'desc' } }],
   };
 
   const {
     body: { hits },
   } = await client.search({
     from: 0,
-    _source: ['id'],
+    _source: ['id', 'published'],
     size: ELASTIC_MAX_RESULTS || 100,
     index,
     type,
--- a/src/backend/utils/indexer.js
+++ b/src/backend/utils/indexer.js
@@ -89,7 +89,7 @@ const search = async (textToSearch, filter = 'post') => {
     body: { hits },
   } = await client.search({
     from: 0,
-    _source: ['id', 'published'],
+    _source: ['id'],
     size: ELASTIC_MAX_RESULTS || 100,
     index,
     type,
--- a/src/backend/utils/indexer.js
+++ b/src/backend/utils/indexer.js
@@ -67,6 +67,16 @@ const createFieldsFromFilter = (filter) => {
   }
 };
 
+const sortFromFilter = (filter) => {
+  switch (filter) {
+    case 'author':
+      return [{ published: { order: 'desc' } }];
+    case 'post':
+    default:
+      return undefined;
+  }
+};
+
 /**
  * Searches text in elasticsearch
  * @param textToSearch
@@ -82,7 +92,7 @@ const search = async (textToSearch, filter = 'post') => {
         fields: createFieldsFromFilter(filter),
       },
     },
-    sort: [{ published: { order: 'desc' } }],
+    sort: sortFromFilter(filter),
   };
 
   const {


src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -76,7 +77,14 @@ const SearchPage = () => {
             },
           };
         });
-      setResults({ type: 'author', searchResults: finalResults });
+      const timeline = data.getPosts.map((result) => {
+        return {
+          id: result.id,
+          url: `/posts/${result.id}`,
+          author: result.feed.author,
```
c3ho, line 84: Could this be a possible issue if we have two feed `authors` with the exact name?

c3ho (changes requested):

src/frontend/src/components/SearchPage/SearchPage.jsx
joelazwar: ahh true i didn't think of that. Maybe we could use the feed id instead then?

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -135,7 +141,10 @@ const SearchPage = () => {
       //  for each with feed guid as key
       if (results.type === 'author') {
         return results.searchResults.map((result) => (
-          <AuthorResult key={result.id} author={result.author} post={result.post} />
+          <div>
```
humphd, line 144: I think the fact that you need to wrap this in a `<div>` means we probably should break this out into its own component.

humphd (changes requested):
This is really interesting.  I love being able to get the posts for a user.  I'm not 100% what to think about mixing the Author "card" with the results, whether it's a good thing or not.

One thing, when we fix #1260 this will maybe bee moot, since we can always show results for a specific author.

src/frontend/src/components/SearchPage/SearchPage.jsx
joelazwar: Ahh true. Maybe I can just implement the timeline on the AuthorResult component itself. But regarding the author card, I also wasn't really sure if that was the best way to present it but I was also wanted to know what you thought of it since you mentioned replacing the results with the timeline or 'adding it to the what we have now'. I'll try and think of a better way to show it, but if you have any ideas as well I'm all ears.

humphd: Maybe we should consider removing the Author card and only showing results?

cc @c3ho, @cindyledev, @agarcia-caicedo, who might have thoughts.

c3ho: For disclosure I am a UI idiot, I am fine iwth removing the author card though.

manekenpix: The Author card was removed in #1295, so I recommend rebasing before so you can work on the most recent version of the code.

humphd:
@joelazwar Where does this stand?  At the very least we need a rebase to get the SearchPage changes in.

joelazwar:
@humphd Right, I haven't had the time to think of a new design lately, so I'll just stick with the timeline of posts as a result. I'll go ahead and rebase this now

joelazwar:
I rebased my branch but its practically more like a new branch I guess, since I undid all the changes (#1295 basically fixed the functionality) but I edited the query so that searched posts timeline is now sorted by date, however this is true for both author and posts filters. Let me know if that's problematic

src/backend/utils/indexer.js
```diff
@@ -82,13 +82,14 @@ const search = async (textToSearch, filter = 'post') => {
...
     },
+    sort: [{ published: { order: 'desc' } }],
   };
 
   const {
     body: { hits },
   } = await client.search({
     from: 0,
-    _source: ['id'],
+    _source: ['id', 'published'],
```
c3ho, line 92: Don't think we need the published here, we just need it to be sorted when we get the results which is what line 85 does

c3ho (changes requested):

src/backend/utils/indexer.js
```diff
@@ -82,6 +82,7 @@ const search = async (textToSearch, filter = 'post') => {
         fields: createFieldsFromFilter(filter),
       },
     },
+    sort: [{ published: { order: 'desc' } }],
```
humphd, line 85: Should we add a  numeric field type for `published` here to indicate that it's a date? https://www.elastic.co/guide/en/elasticsearch/reference/current/sort-search-results.html#_sorting_numeric_fields

I'm not sure if that does anything for us or if leaving it off is OK.

joelazwar: Hmmm I might be wrong, but I think that's only necessary for casting to a different type? The code seems to be working without it

humphd: OK, let's not over-complicate this then.

humphd (changes requested):
So this issue is fixing "Show all posts for an author sorted by date when searching by Author" but you seem to be doing it for all filter types.  Can we limit this to `filter='author'`?

joelazwar:
@humphd done [189961e](https://github.com/Seneca-CDOT/telescope/pull/1256/commits/189961e4aa6eb512f7b3abe5b12b8ac14b0c92f4)

src/backend/utils/indexer.js
```diff
@@ -82,6 +82,7 @@ const search = async (textToSearch, filter = 'post') => {
         fields: createFieldsFromFilter(filter),
       },
     },
+    sort: filter === 'author' ? [{ published: { order: 'desc' } }] : undefined,
```
humphd, line 85: Let's follow the same pattern as `createFieldsFromFilter` above, and move this to a separate function.

humphd (changes requested):

src/backend/utils/indexer.js
joelazwar: ok, made the seperate function.

humphd (approved):

src/backend/utils/indexer.js
```diff
@@ -67,6 +67,16 @@ const createFieldsFromFilter = (filter) => {
   }
 };
 
+const sortCase = (filter) => {
```
humphd, line 70: Can we rename this for consistency to `sortFromFilter`?

humphd (changes requested):

src/backend/utils/indexer.js
joelazwar: of course

humphd (approved):

manekenpix (approved):

c3ho:
Searching by author doesn't work for me, did a search using incognito on Chrome, no results. Post worked though

c3ho (approved):

PR #1274, opened 2020-11-02
Fixes #1223: Merge our two "sticky headers" on scroll

### LINKED ISSUE #1223: Merge our two "sticky headers" on scroll
We landed a fantastic PR the other day to support scrolling and having 2 sticky headers.  The only down side of it is that the two headers take up a lot of space at the top.

Today I noticed that GitHub does something similar, but collapses the two down into a single-height header:

scroll-header

I wonder if we should flip our top header into the "mobile" view and show the menu on the right vs. the words/icons:

Screen Shot 2020-10-20 at 9 44 51 PM

Then we could slide the post title into the middle, empty section between the logo and search/menu icons.  That would let us do both things in the same amount of space.

PedroFonsecaDEV:
## Issue This PR Addresses
Fixes: #1223

## Type of Change

## Description
Fixes: 1223
      *ScrollTop and ScrollDown becomes: ScrollAction (We had repeated code for up and down, now we have only one component)
      *BackToTop becomes: BackToTopButton - No more coupling with ScrollTop
      *Also updated to ES6 the code that I worked on. 
      *Merge our two "sticky headers" on scroll (#1223)

## Checklist

- [X] **Tests**: This PR includes thorough tests or an explanation of why it does not

### DIFF
--- a/src/frontend/src/components/Banner/Banner.jsx
+++ b/src/frontend/src/components/Banner/Banner.jsx
@@ -119,7 +119,7 @@ export default function Banner() {
         justify="center"
         className={classes.container}
       >
-        <Grid id="back-to-top-anchor" item xs={8}>
+        <Grid id="back-to-top-anchor-mobile" item xs={8}>
           <div className={classes.icon}>
             <ScrollAction>
               <Fab color="secondary" aria-label="scroll-down">
@@ -129,6 +129,7 @@ export default function Banner() {
           </div>
         </Grid>
       </Grid>
+      <div id="back-to-top-anchor"></div>
     </>
   );
 }
--- a/src/frontend/src/components/Post/Post.jsx
+++ b/src/frontend/src/components/Post/Post.jsx
@@ -21,10 +21,10 @@ const useStyles = makeStyles((theme) => ({
     padding: '1.4em',
     lineHeight: '1.3',
     zIndex: 1500,
-    [theme.breakpoints.down(1400)]: {
+    [theme.breakpoints.down(1440)]: {
       padding: '.7em',
     },
-    [theme.breakpoints.down(1020)]: {
+    [theme.breakpoints.down(1065)]: {
       position: 'static',
     },
   },
--- a/src/frontend/src/components/Post/Post.jsx
+++ b/src/frontend/src/components/Post/Post.jsx
@@ -18,12 +18,13 @@ const useStyles = makeStyles((theme) => ({
   header: {
     backgroundColor: theme.palette.primary.main,
     color: theme.palette.text.secondary,
-    padding: '3em',
+    padding: '1.4em',
     lineHeight: '1.3',
-    top: '62px',
-    [theme.breakpoints.between('xs', 'sm')]: {
-      paddingTop: '2em',
-      paddingBottom: '2em',
+    zIndex: 1500,
+    [theme.breakpoints.down(1350)]: {
+      padding: '.7em',
+    },
+    [theme.breakpoints.down(985)]: {
       position: 'static',
     },
   },
@@ -79,15 +80,14 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-function formatPublishedDate(dateString) {
+const formatPublishedDate = (dateString) => {
   const date = new Date(dateString);
   const options = { month: 'long', day: 'numeric', year: 'numeric' };
   const formatted = new Intl.DateTimeFormat('en-CA', options).format(date);
   return `Last Updated ${formatted}`;
-}
+};
 
 const Post = ({ postUrl }) => {
-  // id, html, author, url, title, date, link
   const classes = useStyles();
   // We need a ref to our post content, which we inject into a <section> below.
   const sectionEl = useRef(null);
@@ -143,7 +143,8 @@ const Post = ({ postUrl }) => {
         <Typography variant="h1" title={post.title} id={post.id} className={classes.title}>
           {post.title}
         </Typography>
-        <Typography variant="h3" className={classes.author}>
+        <Typography variant={'p'} className={classes.author}>
+          {' '}
           By{' '}
           <a className={classes.link} href={post.feed.link}>
             {post.feed.author}
@@ -151,7 +152,7 @@ const Post = ({ postUrl }) => {
         </Typography>
         <a href={post.url} rel="bookmark" className={classes.published}>
           <time className={classes.time} dateTime={post.updated}>
-            {formatPublishedDate(post.updated)}
+            {` ${formatPublishedDate(post.updated)}`}
           </time>
         </a>
       </ListSubheader>
--- a/src/frontend/src/components/Post/Post.jsx
+++ b/src/frontend/src/components/Post/Post.jsx
@@ -21,10 +21,10 @@ const useStyles = makeStyles((theme) => ({
     padding: '1.4em',
     lineHeight: '1.3',
     zIndex: 1500,
-    [theme.breakpoints.down(1350)]: {
+    [theme.breakpoints.down(1400)]: {
       padding: '.7em',
     },
-    [theme.breakpoints.down(985)]: {
+    [theme.breakpoints.down(1020)]: {
       position: 'static',
     },
   },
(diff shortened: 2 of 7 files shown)


manekenpix:
I really like this, and I have 2 questions:
- Could `ScrollAction` and `BackToTopButton` be just one component that would do scroll up or down depending on a passed prop?
- Should `ScrollAction` and `BackToTopButton` (or the combination of both if merged) use the same colour that we use in the header?

humphd (changes requested):
This is heading in a great direction, I love this!  Before I review the code, some UX issues I notice.

First, there is a pixel difference between the bottom of the nav bar and the post header when they collapse together:

Screen Shot 2020-11-02 at 7 54 39 PM

Next, I don't love how we've lost all the padding around the post header.

Screen Shot 2020-11-02 at 7 57 37 PM

I think the look is quite different, and not for the better.  Can we do something where it stays the way it is until it collapses into the top?

scrolling

You can't see it great in this GIF, but on my machine it bounces as it changes.

src/frontend/src/components/Banner/Banner.jsx
```diff
@@ -142,16 +119,17 @@ export default function Banner() {
         justify="center"
         className={classes.container}
       >
-        <Grid id="back-to-top-anchor" item xs={8}>
+        <Grid id="back-to-top-anchor-mobile" item xs={8}>
```
humphd, line 122: Why `-mobile` on an id?  Seems like something that should only happen in a class you apply, since this will be the same element in every size, no?

src/frontend/src/components/Post/Post.jsx
```diff
@@ -69,15 +70,14 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-function formatPublishedDate(dateString) {
+const formatPublishedDate = (dateString) => {
```
humphd, line 73: Why this change?

src/frontend/src/components/Post/Post.jsx
```diff
@@ -100,15 +100,16 @@ const Post = ({ postUrl }) => {
         <Typography variant="h1" title={post.title} id={post.id} className={classes.title}>
           {post.title}
         </Typography>
-        <Typography variant="h3" className={classes.author}>
+        <Typography variant={'p'} className={classes.author}>
+          {' '}
```
humphd, line 104: Why not `&nbsp;` for these spaces?

src/frontend/src/components/ScrollAction/ScrollAction.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const ScrollAction = (props) => {
```
humphd, line 4: `const ScrollAction = ({ children }) => {`

humphd (changes requested):

src/frontend/src/components/Banner/Banner.jsx
PedroFonsecaDEV: There are two different anchors. 
The desktop anchor places the post's title on the top of the navbar, the mobile anchor is placing the post's title under the navbar.

src/frontend/src/components/Post/Post.jsx
PedroFonsecaDEV: Upgrading to ES6 all the codes I'm working on. Is this ok?

src/frontend/src/components/ScrollAction/ScrollAction.jsx
PedroFonsecaDEV: Sure. Sorry.

src/frontend/src/components/Post/Post.jsx
humphd: In general, I would avoid a change like this if you're not updating that code, and do it in a dedicated "switch to ES6" type PR (i.e., isolate the changes from each other so the git history stays in tact with the change).  But it's not a big deal, just FYI.

agarcia-caicedo (commented):
My main problem with the design, is that it uses elevation in both an unexpected way and unlike how we were using it by relying on material design. 

Before even interacting with the design, the expectation is as follows [(Taken from the material design website)](https://material.io/design/environment/elevation.html#default-elevations)

image
image
image

The app bar either has the highest elevation going over everything else (like Telescope) , or it doesn't and you scroll away (like Github)

When the user does interact with the design, the app-bar already shows that it is elevated higher than both the hero image and the scroll down button, by the combination of using a shadow and by sliding on top of those components

image

When the user reaches the spot just before the app bar meets the blog card, there is no indication that the card title occupies a higher elevation than the app bar (which it needs to be able to show the title on the app bar). On the contrary, the app bar has a more prominent shadow that gives the impression of a higher elevation. 

image

However, when the user continues scrolling, it reacts in an unexpected way - The card title begins to merge with the app bar, but they visibly don't have the same elevation (or the card title being at a higher elevation) because of the shadow difference

image

As the user continues to scroll, the unexpected behaviour continues. There isn't an elevation difference between the card title and the card content. So the expectation would be that both would go over the app bar, but they don't. The card title stays on top of the app bar, but the content goes behind. 

In the end, this design choice both break the user's expectations based on their previous experiences with material design, and the expectations set up with this design.

### Other minor feedback unrelated to elevation. Mostly

- The author and the date are a bit too close, which could create an ux issue since the small gap could mean people click on the wrong link.
- At a certain width the title of the post cuts of the search button
image

- Before the card title joins the app bar, the sizing it's ok, but once it goes on the app bar, the font size its too big for its space. Rather than increase the padding, (big app bard are not ideal), the font size could be shrunk. Similar how Github does it. 

image
image

humphd:
@agarcia-caicedo have you got ideas on how to do this in a way that stays within the current design?  I'm trying you read your comment above so as to find a way forward.

I really love how this cleans up the view, but retains the post info.  But I get what you're saying too.

I guess what I want is my cake and eat it to :)

PedroFonsecaDEV:
I didn't change the font size or spacing, just the padding. So we are seeing the same font and the same size as we see in prod and dev. 
But I think the author and date are small too. Let's increase it.  
I will improve and fix everything.

PedroFonsecaDEV:
I'm still working on it. The last commit doesn't represent the final work.

PedroFonsecaDEV:
@humphd @manekenpix @cindyledev 
- All bugs are fixed. 
- Rebased.
- Good to go!

This is just the beginning for a better UX. I hope more people jump into our frontend with more improvements. 
Thanks.

humphd:
@PedroFonsecaDEV what about this:

Screen Shot 2020-11-02 at 7 57 37 PM

That still doesn't look right to me.  Should we tackle this in a follow-up or is it doable here?

Losing that padding really changes the look of the site.

I think the rest of this is good, though.

PedroFonsecaDEV:
@humphd Let's fix the padding in a follow-up PR.

humphd (dismissed):
Great.  I'll approve this conditional on you filing an issue to fix the padding in another PR.

This is a great fix, thanks for fighting with it.

PedroFonsecaDEV:
@humphd 
Thanks.
Issue: 
https://github.com/Seneca-CDOT/telescope/issues/1329

manekenpix (dismissed):

manekenpix:
@PedroFonsecaDEV it's better if you rebase, don't merge master into your branch
image

PedroFonsecaDEV:
PLEASE DON'T MERGE THIS PR. 
I WILL START ANOTHER ONE TOMORROW. 

Thanks.

manekenpix:
@PedroFonsecaDEV you don't need to submit another PR, just remove the last commit (`git reset HEAD~1`) and rebase. I'll add the `blocked` label.

PedroFonsecaDEV:
@humphd @manekenpix 

Rebased and good to go.

humphd (approved):
@PedroFonsecaDEV looks good.  Did you want to try squashing this too, so you only have 1 commit?

manekenpix (approved):

PR #1282, opened 2020-11-03
Fixes #1266: modifies sanitize.js to allow Telescope to parse img tags with data uris

### LINKED ISSUE #1266: Missing src attribute on img elements in posts when using data URIs
I notice a post today that is missing images.  The `` element is there, but it has no `src` attribute:

Screen Shot 2020-11-01 at 6 15 15 PM

The original post does have them: https://palak-chawla.blogspot.com/2020/10/release-02-pr-4.html, but they are using Data URIs:

Screen Shot 2020-11-01 at 6 17 40 PM

chrispinkney:
## Issue This PR Addresses
Fixes #1266 

## Type of Change

## Description
  * Modified `sanitize.js` to allow for `img` tags to use `data` uris as `src` so they can be successfully parsed and indexed by Telescope.
  * Adds a test to ensure `data` uris are sanitized properly.
  * Adds two more `http` and `https` schema tests for `img` tag with a data uri `src`.
  * Adds `blockquote` `test` and `cite` sanitation.

Hey everyone!

As per issue #1266, this PR allows images which have a `src` `data` uri type to be properly indexed. It passes all the tests on my end thus showing that data uris are being allowed by Telescope.

I tried to ensure that this works as expected by purging my redis db via `docker exec -it telescope_redis_1 redis-cli flushall` but it seems that once a post is indexed, it is how it is and forever shall be the way it was initially indexed. So instead of going back in time I disallowed all html attributes and waited for a post to come in. Once one arrived I noticed that it had no tags at all, just plaintext, so I reverted back to normal and allowed `img`s with a data uri.

While this isn't exactly a perfect sign that everything works as expected, given that `npm test sanitize-html` runs and returns that all tests pass, and specifically: `` returned as expected, I feel that Telescope should now properly parse `img`s with a data uri.

Let me know, thanks!

## Checklist

- [X] **Tests**: This PR includes thorough tests or an explanation of why it does not

### DIFF
--- a/src/backend/utils/html/sanitize.js
+++ b/src/backend/utils/html/sanitize.js
@@ -41,10 +41,12 @@ module.exports = function (dirty) {
       'tr',
       'ul',
     ],
+    allowedSchemesByTag: { img: ['http', 'https', 'data'] },
     allowedAttributes: {
       iframe: ['src'],
       img: ['src'],
       a: ['href'],
+      blockquote: ['cite'],
     },
     allowedIframeHostnames: ['www.youtube.com', 'player.vimeo.com', 'giphy.com'],
   });
--- a/src/backend/utils/html/sanitize.js
+++ b/src/backend/utils/html/sanitize.js
@@ -46,7 +46,6 @@ module.exports = function (dirty) {
       iframe: ['src'],
       img: ['src'],
       a: ['href'],
-      blockquote: ['cite'],
     },
     allowedIframeHostnames: ['www.youtube.com', 'player.vimeo.com', 'giphy.com'],
   });
--- a/test/sanitize-html.test.js
+++ b/test/sanitize-html.test.js
@@ -30,6 +30,11 @@ describe('Sanitize HTML', () => {
     expect(data).toBe('<p><a><img src="http://www.telescope.com/http_image.jpg" /></a></p>');
   });
 
+  test('protocoless urls should be accepted (i.e. not sanitized)', () => {
+    const data = sanitizeHTML('<img src="//www.telescope.com/image.jpg" />');
+    expect(data).toBe('<img src="//www.telescope.com/image.jpg" />');
+  });
+
   test('<a><img> should work, but inline js should not', () => {
     const data = sanitizeHTML(
       '<a href="https://www.telescope.com"><img border="0" alt="W3Schools" src="logo.gif" width="100" height="100"></a>'
@@ -91,15 +96,6 @@ describe('Sanitize HTML', () => {
     );
     expect(data).toBe(
       '<table><tbody><tr><td><a href="www.senecacollege.ca"><img src="https://1.bp.blogspot.com/11.JPG" /></a></td></tr><tr><td>The Final Product</td></tr></tbody></table>'
-    );
-  });
-
-  test('<blockquote> should return properly and retain its citation', () => {
-    const data = sanitizeHTML(
-      '<blockquote cite="http://www.worldwildlife.org/who/index.html"> For 50 years, WWF has been protecting the future of nature from Stone Cold Steve Austin. </blockquote>'
-    );
-    expect(data).toBe(
-      '<blockquote cite="http://www.worldwildlife.org/who/index.html"> For 50 years, WWF has been protecting the future of nature from Stone Cold Steve Austin. </blockquote>'
     );
   });
 });
--- a/test/sanitize-html.test.js
+++ b/test/sanitize-html.test.js
@@ -23,6 +23,7 @@ describe('Sanitize HTML', () => {
     expect(data).toBe('<p><a><img src="https://www.telescope.com/https_image.jpg" /></a></p>');
   });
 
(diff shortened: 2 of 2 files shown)


chrispinkney:
So this is my first solo rebase that I did because I had like 3 commits. I also pushed them by accident, so I `git rebase HEAD~3 -i` and squashed the other two, then force pushed this to my branch and made a PR. I think I did all the steps right so let me know if I didn't!

humphd:
@chrispinkney can you fix the title of this PR?  `fixes issue-1266` is not useful, as it doesn't reference what you're working on (the issue number is good, but describe the work too).

src/backend/utils/html/sanitize.js
```diff
@@ -41,10 +41,12 @@ module.exports = function (dirty) {
       'tr',
       'ul',
     ],
+    allowedSchemesByTag: { img: ['http', 'https', 'data'] },
     allowedAttributes: {
       iframe: ['src'],
       img: ['src'],
       a: ['href'],
+      blockquote: ['cite'],
```
humphd, line 49: Why is this here?

test/sanitize-html.test.js
```diff
@@ -6,6 +6,30 @@ describe('Sanitize HTML', () => {
...
+  });
+
+  test('img links over https should be accepted (i.e. not sanitized)', () => {
+    const data = sanitizeHTML(
+      '<p><a><img src="https://www.telescope.com/https_image.jpg" /></a></p>'
+    );
+    expect(data).toBe('<p><a><img src="https://www.telescope.com/https_image.jpg" /></a></p>');
+  });
+
+  test('img links over http should be accepted (i.e. not sanitized)', () => {
```
humphd, line 26: This is an interesting one, since Chrome is moving to block mixed content.  We may need to undo this later, since it will affect our security in the browser (we load over https but mixed content with http images will break us).

test/sanitize-html.test.js
```diff
@@ -69,4 +93,13 @@ describe('Sanitize HTML', () => {
       '<table><tbody><tr><td><a href="www.senecacollege.ca"><img src="https://1.bp.blogspot.com/11.JPG" /></a></td></tr><tr><td>The Final Product</td></tr></tbody></table>'
     );
   });
+
+  test('<blockquote> should return properly and retain its citation', () => {
```
humphd, line 97: This needs to be its own issue/pr

humphd (changes requested):
Can you also add a test for protocol-less URLs? (i.e., ``)

src/backend/utils/html/sanitize.js
chrispinkney: Just wanted to add an additional test, hm, it does look irrelevant though in hindsight. I'll remove it from this PR.

test/sanitize-html.test.js
chrispinkney: Neat, should I add a comment specifying this?

src/backend/utils/html/sanitize.js
humphd: It's a great change, just do it in its own issue/pr.

test/sanitize-html.test.js
humphd: Sure, maybe link to https://web.dev/what-is-mixed-content/

chrispinkney:
> 
> 
> Can you also add a test for protocol-less URLs? (i.e., ``)

I was going to add that too but there's already one in place [that does the same thing here](https://github.com/Seneca-CDOT/telescope/blob/b4151142d8dbda3f09343b91b072efb7003ca892/test/sanitize-html.test.js#L33).

chrispinkney:
> 
> 
> @chrispinkney can you fix the title of this PR? `fixes issue-1266` is not useful, as it doesn't reference what you're working on (the issue number is good, but describe the work too).

Is this better?

humphd (approved):

tonyvugithub:
@chrispinkney @humphd I wonder why the github action for macos failed. Should we retry with another push with a combined commit and see if all are passed? Right now one check has failed. 

This is the error in the macos build
`[HPM] Proxy rewrite rule created: "^/admin/build/log" ~> "/log"
[ 2020-11-03 23:27:02.081 ] ERROR (3824 on Mac-1604444996117.local): Something went wrong with search indexing`

humphd:
That failure is not actually causing the tests to fail (i.e, the test expects the failure).  What's causing the failure is some process hanging and blocking the tests from exiting correctly (see #1284):

```
A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks.

Test Suites: 26 passed, 26 total
Tests:       169 passed, 169 total
Snapshots:   0 total
Time:        15.689 s
Ran all test suites.
Error: The operation was canceled.
```

tonyvugithub:
@humphd I noticed that too. So I guess this PR is good to go ahead then if thats the case

tonyvugithub (approved):

PR #1283, opened 2020-11-03
Connect Pino logger to ElasticSearch

### LINKED ISSUE #1268: Connect Pino logger to Elasticsearch
Our [Pino logger](https://getpino.io/#/) supports [various transports](https://getpino.io/#/docs/transports), so you can log into various services and programs.  We're also using [Elasticsearch](https://www.elastic.co/elasticsearch/) for full-text search.

Let's combine these two and add support for https://github.com/pinojs/pino-elasticsearch.  Probably we need to make this optional (i.e., via `.env` configs) so that people can opt into it.  On staging and production, having our logs indexed in Elasticsearch will make it possible to build some cool admin log features.  I know I often want to quickly search through the logs for things, and this would make that trivial.  We'll be able to update and replace the `/admin/log` route.

I'm not 100% of the implications of doing this, and we'll need to research a bit.  For example, we don't want to fill our disk infinitely.  Can we evict old data from Elasticsearch beyond some time window (maybe keep logs for a month, 6 months, etc)?

c3ho:
## Issue This PR Addresses
Fixes #1268 

## Type of Change

## Description
Connecting our logger to elastic so we can index logs too and have them show up in `localhost:<elasticPort>`. Your console/terminal should not be spammed anymore by the logging if this is enabled, it should all be redirected to ElasticSearch now.

I should be able to remove the need to put `elasticStream` inside `destination = pino.destination()` and just use it directly
according to the response https://github.com/pinojs/pino/issues/845

To test:
1. Set `PINO_ELASTIC=1` in `.env` or add the variable to your `.env`
2. `npm install`
3. `npm start`
4. navigate to `http://localhost:9200/_cat/indices?v`

You should now see a new index called `logs`

This was from an earlier version I was playing with:
image

There's a new index called `an-index`

Visualization will come in my next PR.
## Checklist

### DIFF
--- a/env.example
+++ b/env.example
@@ -35,6 +35,19 @@ LOG_LEVEL=debug
 # LOG_FILE is used to set a destination path to write logs. Works in production mode only.
 LOG_FILE=
 
+# Used to set if Pino should log are indexed to ElasticSearch, LOG_ELASTIC=1 to enable, if LOG_FILE and LOG_ELASTIC is set,
+# LOG_ELASTIC will take priority
+LOG_ELASTIC=
+# Following are used to set up Pino-Elastic
+# Index name for logs to be stored under
+INDEX_NAME=logs
+# Determines consistency of the write, valid values are: one, quorum or all
+CONSISTENCY=one
+# Elasticsearch Version
+ELASTIC_VERSION=7
+# The number of bytes for each bulk insert
+FLUSH_BYTES=1000
+
 # FEED_URL url used to access feed list
 FEED_URL=https://wiki.cdot.senecacollege.ca/wiki/Planet_CDOT_Feed_List
 # Milliseconds to wait after attempting to fetch the feed list when the server is not available
--- a/src/backend/utils/logger.js
+++ b/src/backend/utils/logger.js
@@ -1,4 +1,5 @@
 const pino = require('pino');
+const pinoElastic = require('pino-elasticsearch');
 const expressPino = require('express-pino-logger');
 const os = require('os');
 const path = require('path');
@@ -32,7 +33,21 @@ if (process.env.LOG_FILE) {
   destination = pino.destination(process.stdout.fd);
 }
 
-const logger = pino(options, destination);
+const streamToElastic = pinoElastic({
+  // Name of the index the logs will be stored under
+  index: (process.env.INDEX_NAME || 'logs').toLowerCase(),
+  // Consistency of the write, valid values: 'one', 'quorum', 'all'
+  consistency: (process.env.CONSISTENCY || 'one').toLowerCase(),
+  // URL of Elasticsearch
+  node: `http://${process.env.ELASTIC_URL}:${process.env.ELASTIC_PORT}`,
+  // Elasticsearch version
+  'es-version': process.env.ELASTIC_VERSION || 7,
+  // The number of bytes for each bulk insert
+  'flush-bytes': process.env.FLUSH_BYTES || 1000,
+});
+
+const logger = pino({ options }, process.env.LOG_ELASTIC ? streamToElastic : destination);
+
 const expressLogger = expressPino({ logger });
 
 module.exports = expressLogger;
--- a/package.json
+++ b/package.json
@@ -92,6 +92,7 @@
     "passport": "0.4.1",
     "passport-saml": "1.3.5",
     "pino": "6.6.1",
+    "pino-elasticsearch": "5.3.0",
     "pino-pretty": "4.2.1",
     "sanitize-html": "1.27.4",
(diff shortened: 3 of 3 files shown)


src/backend/utils/logger.js
```diff
@@ -32,7 +33,18 @@ if (process.env.LOG_FILE) {
...
-const logger = pino(options, destination);
+const streamToElastic = pinoElastic({
+  index: (logTime) => {
+    // the logTime is a ISO 8601 formatted string of the log line
+    return `Telescope-${logTime.substring(5, 10)}`;
+  },
+  consistency: 'one',
+  node: `${process.env.ELASTIC_URL}:${process.env.ELASTIC_PORT}`,
+  'es-version': 7,
+  'flush-bytes': 1000,
```
manekenpix, line 44: I think it's better to store this number in a constant and use it here (avoid [magic numbers](https://codeburst.io/software-anti-patterns-magic-numbers-7bc484f40544))

src/backend/utils/logger.js
```diff
@@ -32,6 +33,20 @@ if (process.env.LOG_FILE) {
...
 }
 
+const streamToElastic = pinoElastic({
+  /* index: (logTime) => {
+    // the logTime is a ISO 8601 formatted string of the log line
+    return `Telescope-${logTime.substring(5, 10)}`;
+  }, */
+  index: 'an-index',
+  consistency: 'one',
+  node: `http://localhost:9200`,
```
manekenpix, line 43: You can use our `.env` here for the url.

env.example
```diff
@@ -35,6 +35,14 @@ LOG_LEVEL=debug
 # LOG_FILE is used to set a destination path to write logs. Works in production mode only.
 LOG_FILE=
 
+# Used to set if Pino should log are indexed to ElasticSearch, PINO_ELASTIC=1 to enable
+PINO_ELASTIC=
```
humphd, line 39: `LOG_ELASTIC` maybe.  People won't understand what Pino means, but they do understand logs.

env.example
```diff
@@ -35,6 +35,14 @@ LOG_LEVEL=debug
...
+# Following are used to set up Pino-Elastic
```
humphd, line 40: I have no idea how to set any of these.  Should they be config options or just something we do in code?  I think the latter, unless you think we need to tweak them a lot?

env.example
humphd: What if someone sets `LOG_FILE` and `LOG_ELASTIC`?  Should we document which one gets priority?

src/backend/utils/logger.js
```diff
@@ -32,7 +33,16 @@ if (process.env.LOG_FILE) {
   destination = pino.destination(process.stdout.fd);
 }
 
-const logger = pino(options, destination);
+const streamToElastic = pinoElastic({
+  index: (process.env.INDEX_NAME || 'logs').toLowerCase(),
```
humphd, line 37: Can you leave comments above each of these with some info on how to set them?

humphd (changes requested):

env.example
c3ho: the code has LOG_ELASTIC as priority
`const logger = pino({ options }, process.env.LOG_ELASTIC ? streamToElastic : destination);` but I added a line to the `env.example` to be more explicit
```
# Used to set if Pino should log are indexed to ElasticSearch, LOG_ELASTIC=1 to enable, if LOG_FILE and LOG_ELASTIC is set,
# LOG_ELASTIC will take priority
```

manekenpix:
I think we need better commit messages
image

c3ho:
@manekenpix will be rebasing after it passes everything =)

env.example
```diff
@@ -35,6 +35,19 @@ LOG_LEVEL=debug
...
+# Index name for logs to be stored under
+INDEX_NAME=
+# Determines consistency of the write, valid values are: one, quorum or all
+CONSISTENCY=
+# Elasticsearch Version
+ELASTIC_VERSION=
+# The number of bytes for each bulk insert
+FLUSH_BYTES=
+
```
manekenpix, line 50: I think we should have default values here like we have in all the other variables.

c3ho: I left default values in the other `logger.js` file. I have no idea which is the preferred one.

manekenpix: I think both should have default values, for consistency

c3ho: Sure, I have no issues with that

humphd (dismissed):

manekenpix (dismissed):

c3ho:
rebasing to get rid of the nasty commit messages then merging

humphd (approved):

PR #1287, opened 2020-11-04
fix the bug set LOG_LEVEL in .env is not working

### LINKED ISSUE #1264: LOG_LEVEL works via shell but not .env
I noticed today that I couldn't set my `LOG_LEVEL` via the `.env`, but could on the command line.

In both cases I used `LOG_LEVEL=debug` but got `INFO` vs. `DEBUG` when I did it via `.env`, whereas this worked on macOS:

```sh
$ LOG_LEVEL=debug npm start
```

tianlangwu:
## Issue This PR Addresses
fix #1264 

## Type of Change

## Description

Fix the bug that the setting of LOG_LEVEL in .env is not working properly. Even set `LOG_LEVEL=debug`, `npm start` still print as `LOG_LEVEL=info`. 

Remove `logger` dependency and replaced it to `Console.log()/error()` inside `lib/config.js`.

To test this, you can set `LOG_LEVEL=debug` in the .env, then run the server `npm start`.

```sh
[ 2020-11-04 17:55:58.805 ] INFO  (16544 on LAPTOP-LCOUFEIF): Loading all feeds into feed queue for processing
[ 2020-11-04 17:55:58.810 ] INFO  (16544 on LAPTOP-LCOUFEIF): Telescope listening on port 3000
[ 2020-11-04 17:55:59.335 ] INFO  (16544 on LAPTOP-LCOUFEIF): Connected to elasticsearch!
[ 2020-11-04 17:55:59.336 ] INFO  (16544 on LAPTOP-LCOUFEIF): Starting 1 instance of feed processor.
[ 2020-11-04 17:55:59.354 ] DEBUG (16544 on LAPTOP-LCOUFEIF): Job 0f20520bfe is active
[ 2020-11-04 17:56:00.692 ] INFO  (2132 on LAPTOP-LCOUFEIF): [HTTP 304 - Not Modified] Feed not downloaded: http://shavedyak.blogspot.com/feeds/posts/default/-/opensource
[ 2020-11-04 17:56:00.697 ] DEBUG (16544 on LAPTOP-LCOUFEIF): Job 0f20520bfe completed.
[ 2020-11-04 17:56:00.706 ] DEBUG (16544 on LAPTOP-LCOUFEIF): Job bb9952ec17 is active
```

## Checklist

### DIFF
--- a/src/backend/lib/config.js
+++ b/src/backend/lib/config.js
@@ -1,11 +1,10 @@
 const path = require('path');
 const dotenv = require('dotenv');
-const { logger } = require('../utils/logger');
 
 const result = dotenv.config();
 
-if (result.error && logger) {
-  logger.error(
+if (result.error) {
+  console.warn(
     '\n\n\t💡 It appears that you have not yet configured a .env file.' +
       '\n\t   Please refer to our documentation regarding environment configuration:' +
       '\n\t   https://github.com/Seneca-CDOT/telescope/blob/master/docs/CONTRIBUTING.md\n'
@@ -14,7 +13,7 @@ if (result.error && logger) {
   // Try and use the example env instead if we're not in production
   if (process.env.NODE_ENV !== 'production') {
     const envSamplePath = path.resolve(process.cwd(), 'env.example');
-    logger.info(`Attempting to use default env in ${envSamplePath} instead.\n`);
+    console.log(`Attempting to use default env in ${envSamplePath} instead.\n`);
     dotenv.config({ path: envSamplePath });
   }
 }


src/backend/utils/logger.js
```diff
@@ -3,7 +3,7 @@ const expressPino = require('express-pino-logger');
 const os = require('os');
 const path = require('path');
 
-require('../lib/config');
+require('dotenv').config();
```
humphd, line 6: We do this in [lib/config.js](https://github.com/Seneca-CDOT/telescope/blob/master/src/backend/lib/config.js) already, as well as some fallback logic.

I'd like to isolate the config loading to `lib/config.js` vs. doing it here.  Can you see what we're doing wrong in that file instead of this change?

humphd (changes requested):

src/backend/utils/logger.js
tianlangwu: Yea, I am going to look it

tianlangwu: In config.js file. It works as long as` const result = dotenv.config(); ` before `const { logger } = require('../utils/logger');`

humphd: Ah, interesting, because we have a circular dependency here.

Let's rip the `logger` dependency out of `lib/config.js` and use `console.log()` in that file instead.  Maybe we can avoid logging in there altogether.

src/backend/lib/config.js
```diff
@@ -1,11 +1,10 @@
 const path = require('path');
 const dotenv = require('dotenv');
-const { logger } = require('../utils/logger');
 
 const result = dotenv.config();
 
-if (result.error && logger) {
-  logger.error(
+if (result.error) {
+  console.error(
```
humphd, line 7: `console.warn` vs. `console.error` here please, so we don't include a stack

humphd (changes requested):

src/backend/lib/config.js
tianlangwu: thanks for the advice!

humphd (approved):

manekenpix (approved):
Tested locally :+1:

PR #1292, opened 2020-11-05
fixes issue-1180 fixes the post padding for mobile devices found in timeline.jsx

### LINKED ISSUE #1180: Remove outside post padding in mobile
After #1174 was pushed, posts on mobile have padding, which prevents them from taking up the full screen. 

What we have:
image

What we want:
image

The padding makes the paragraphs more narrow than they need to be, allowing for less words per line.

### LINKED ISSUE #1226: Fix timeline width in search results
In #1217 I updated how our search results show for posts, but I couldn't get the width right on the container (there's extra padding/margin that makes it narrower than it should be):

Screen Shot 2020-10-20 at 9 52 59 PM

I think the posts should be the same width as the search UI above it.

pyvelkov:
## Issue This PR Addresses
Fixes #1180 

## Type of Change

## Description
I traced how the padding was being applied through Chrome developer tools. I saw that the default root `padding` was being applied meaning the posts have `padding` to the left and right and were not being overridden as expected. From the previous fix that was made I saw that root style was being overridden with a `padding` of 0 in posts.jsx but since the change in #1217 , timeline.jsx was created and that is where the issue stemmed from. I saw that the Timeline function was returning the wrapped array in a container that had the `className={classes.root}`. However custom overriding of the root style was not being called in "useStyles". 

### Before
without fix

### After
with fix

<br /><br />
EDIT:
I believe this also fixed #1226 
fixed search

I also ran `npm run test` and everything passed. Everything looks alright on my end.

Let me know what you think and if this is what we are after in #1180 , my JS is terrible and I probably spent far longer than needed for something like this. I hope it is what we are after.

## Checklist

### DIFF
--- a/src/frontend/src/components/Post/Post.jsx
+++ b/src/frontend/src/components/Post/Post.jsx
@@ -6,7 +6,6 @@ import { makeStyles } from '@material-ui/core/styles';
 import { Box, Grid, Typography, ListSubheader } from '@material-ui/core';
 import './telescope-post-content.css';
 import AdminButtons from '../AdminButtons';
-import { autoDetection } from 'highlight.js';
 
 const useStyles = makeStyles((theme) => ({
   root: {
--- a/src/frontend/src/components/Post/Post.jsx
+++ b/src/frontend/src/components/Post/Post.jsx
@@ -6,6 +6,7 @@ import { makeStyles } from '@material-ui/core/styles';
 import { Box, Grid, Typography, ListSubheader } from '@material-ui/core';
 import './telescope-post-content.css';
 import AdminButtons from '../AdminButtons';
+import { autoDetection } from 'highlight.js';
 
 const useStyles = makeStyles((theme) => ({
   root: {
@@ -52,6 +53,7 @@ const useStyles = makeStyles((theme) => ({
     },
   },
   content: {
+    overflow: 'auto',
     padding: '2em',
     color: theme.palette.text.default,
   },
--- a/src/frontend/src/components/Posts/Timeline.jsx
+++ b/src/frontend/src/components/Posts/Timeline.jsx
@@ -9,6 +9,10 @@ import LoadMoreButton from './LoadMoreButton.jsx';
 import useSiteMetaData from '../../hooks/use-site-metadata';
 
 const useStyles = makeStyles((theme) => ({
+  root: {
+    padding: 0,
+    maxWidth: '785px',
+  },
   activeCircle: {
     borderRadius: '4rem',
     transition: 'all linear 250ms',


humphd:
I'm not sure if we should deal with this here or in another issue, but I also notice we have an overflow problem on mobile with content extending beyond the parent container(s):

width-bug

pyvelkov:
> I'm not sure if we should deal with this here or in another issue, but I also notice we have an overflow problem on mobile with content extending beyond the parent container(s):
> 
> width-bug

I don't mind doing this in another post but I can do it here. I sort of found a fix by applying style of`overflown` content to `auto` and it added a scroll like effect for posts like this.
scroll on certain posts

humphd:
Sure, push another commit and let's do it here.

pyvelkov:
> Sure, push another commit and let's do it here.

I just added the commit that adds the scrollbar for overflown posts. Let me know, Thanks :)

cindyledev (dismissed):
Thanks a lot for fixing this

src/frontend/src/components/Post/Post.jsx
```diff
@@ -6,6 +6,7 @@ import { makeStyles } from '@material-ui/core/styles';
 import { Box, Grid, Typography, ListSubheader } from '@material-ui/core';
 import './telescope-post-content.css';
 import AdminButtons from '../AdminButtons';
+import { autoDetection } from 'highlight.js';
```
c3ho, line 9: Is it just me? I don't see this being used anywhere in the code change.

c3ho (changes requested):

src/frontend/src/components/Post/Post.jsx
manekenpix: we have this:
```js
import 'highlight.js/styles/github.css';
```
in line 4. So what is this in line 9 for?

pyvelkov: oh shoot. I think I was testing a fix for the overflown content mentioned above. I will remove it. I apologize.

pyvelkov: Good catch @c3ho I will check for that in the future. Sorry about that again.

c3ho (approved):

manekenpix (approved):

c3ho:
@pyvelkov Feel free to squash and merge

pyvelkov:
> @pyvelkov Feel free to squash and merge

would you like me to do it, or..?

humphd:
@pyvelkov go for it!  With 2 reviews giving an approval, you can squash and merge.

Issue #552, opened 2020-01-19
Remove GraphQL implementation


manekenpix:
**What would you like to be added**:

Our backend will soon incorporate [GraphQL](https://graphql.org/) support (#532 ) and we'll need tests for queries requesting both feeds and posts. 

Here is [a module](https://www.apollographql.com/docs/apollo-server/testing/testing/) for testing developed by the Apollo-server community.

Another possibility could be [easygraphql-tester](https://github.com/EasyGraphQL/easygraphql-tester).

**Why would you like this to be added**:

We need to make sure that our GraphQL endpoint answers queries correctly.

c3ho:
Let's have this finished prior to 1.0

manekenpix:
I was planning to do some work on this, but after seeing comments and discussions on slack, maybe we should close this and start thinking about the new fron-tend. Thoughts?

humphd:
I think tests for the REST API and GraphQL endpoints would be good for any/all frontends we do, no?  I can't see a downside to you working on this.

Or, if you would rather, spec it out as bugs other students could do?

manekenpix:
It's just that it seems our REST API fits better with the way Telescope works, and maybe we're going to end up removing all our GraphQL code (we barely use it), that's why I wasn't sure about spending time on this.

humphd:
That's another way to go: rip it all out.  We should discuss with people and see what the consensus is.

I could go either way.

humphd:
Morphing this to be about removing GraphQL, since we don't need it for next.js.

c3ho:
![image](https://user-images.githubusercontent.com/18711727/97814904-948e8500-1c58-11eb-98c5-bafc33d4832d.png)

PR #1294, opened 2020-11-06
added more values in elastic search

### LINKED ISSUE #1280: Add more fields to Elasticsearch
When we store posts in Elasticsearch, we currently only include the posts' `text`:

https://github.com/Seneca-CDOT/telescope/blob/master/src/backend/utils/indexer.js#L12-L30

We can add more fields, and I think it would be good to include the `author` name.  This way we can convert our Author Search feature to also use Elastic vs. doing it manually.

Then we can modify our search to work against the text or the author name, as the user specifies, and limit the fields we return from Elastic.

StellaJung-Student:
## Issue This PR Addresses

Fixes #1280 

## Type of Change

## Description

I updated 3 more fields which are title, updated and author.

## Checklist

### DIFF
--- a/src/backend/data/post.js
+++ b/src/backend/data/post.js
@@ -76,6 +76,10 @@ class Post {
     return textParser(this.html);
   }
 
+  get author() {
+    return this.feed.author;
+  }
+
   /**
    * Parse an article object into a Post object.
    * @param {Object} article parsed via feedparser, see:
@@ -146,10 +150,7 @@ class Post {
       article.guid,
       feed
     );
-    await Promise.all([
-      post.save(),
-      indexPost(post.text, post.id, post.title, post.updated, post.feed.author),
-    ]);
+    await Promise.all([post.save(), indexPost(post)]);
     return post.id;
   }
 
--- a/src/backend/data/post.js
+++ b/src/backend/data/post.js
@@ -40,7 +40,7 @@ function ensureFeed(feed) {
 }
 
 class Post {
-  constructor(title, html, datePublished, dateUpdated, postUrl, guid, feed, author) {
+  constructor(title, html, datePublished, dateUpdated, postUrl, guid, feed) {
     // Use the post's guid as our unique identifier
     this.id = hash(guid);
     this.title = title;
@@ -49,7 +49,6 @@ class Post {
     this.updated = ensureDate(dateUpdated, datePublished);
     this.url = postUrl;
     this.guid = guid;
-    this.author = author;
 
     // We expect to get a real Feed vs. a feed id
     if (!(feed instanceof Feed)) {
@@ -145,12 +144,11 @@ class Post {
       // link is the url to the post
       article.link,
       article.guid,
-      feed,
-      article.author
+      feed
     );
     await Promise.all([
       post.save(),
-      indexPost(post.text, post.id, post.title, post.updated, post.author),
+      indexPost(post.text, post.id, post.title, post.updated, post.feed.author),
     ]);
     return post.id;
   }
@@ -171,8 +169,7 @@ class Post {
       postData.updated,
       postData.url,
       postData.guid,
-      feed,
-      postData.author
+      feed
     );
     await post.save();
     return post.id;
@@ -197,8 +194,7 @@ class Post {
       data.updated,
       data.url,
       data.guid,
-      feed,
-      data.author
+      feed
     );
     return post;
   }
--- a/src/backend/data/post.js
+++ b/src/backend/data/post.js
@@ -40,7 +40,7 @@ function ensureFeed(feed) {
 }
 
 class Post {
-  constructor(title, html, datePublished, dateUpdated, postUrl, guid, feed) {
+  constructor(title, html, datePublished, dateUpdated, postUrl, guid, feed, author) {
     // Use the post's guid as our unique identifier
     this.id = hash(guid);
     this.title = title;
@@ -49,6 +49,7 @@ class Post {
     this.updated = ensureDate(dateUpdated, datePublished);
     this.url = postUrl;
     this.guid = guid;
+    this.author = author;
 
     // We expect to get a real Feed vs. a feed id
     if (!(feed instanceof Feed)) {
@@ -144,9 +145,13 @@ class Post {
       // link is the url to the post
       article.link,
       article.guid,
-      feed
+      feed,
+      article.author
     );
-    await Promise.all([post.save(), indexPost(post.text, post.id)]);
+    await Promise.all([
+      post.save(),
+      indexPost(post.text, post.id, post.title, post.updated, post.author),
+    ]);
     return post.id;
   }
 
@@ -166,7 +171,8 @@ class Post {
       postData.updated,
       postData.url,
       postData.guid,
-      feed
+      feed,
+      postData.author
     );
     await post.save();
     return post.id;
@@ -191,7 +197,8 @@ class Post {
       data.updated,
       data.url,
       data.guid,
-      feed
+      feed,
+      data.author
     );
     return post;
   }
--- a/src/backend/utils/__mocks__/indexer.js
+++ b/src/backend/utils/__mocks__/indexer.js
@@ -4,9 +4,9 @@ const db = {
   values: [],
 };
 
-const indexPost = (text, postId) => {
+const indexPost = ({ text, id }) => {
   db.values.push({
-    id: postId,
+    id,
     text,
     score: 10,
   });
--- a/src/backend/utils/__mocks__/indexer.js
+++ b/src/backend/utils/__mocks__/indexer.js
@@ -27,7 +27,7 @@ const deletePost = (postId) => {
   return Promise.resolve();
 };
 
-const mockCreateFieldsFromFilter = (filter) => {
+const createFieldsFromFilter = (filter) => {
   switch (filter) {
     case 'author':
       return ['author'];
@@ -37,11 +37,11 @@ const mockCreateFieldsFromFilter = (filter) => {
   }
 };
 
-const search = (keyword = '', filter = 'post') => {
-  const fields = mockCreateFieldsFromFilter(filter);
+const search = (textToSearch = '', filter = 'post') => {
+  const fields = createFieldsFromFilter(filter);
   let filtered = db.values;
   fields.forEach((field) => {
-    filtered = filtered.filter((value) => value.body[field].includes(keyword));
+    filtered = filtered.filter((value) => value.body[field].includes(textToSearch));
   });
   return Promise.resolve({
     results: filtered.length,
--- a/src/backend/utils/__mocks__/indexer.js
+++ b/src/backend/utils/__mocks__/indexer.js
@@ -27,8 +27,22 @@ const deletePost = (postId) => {
   return Promise.resolve();
 };
 
-const search = (keyword = '') => {
-  const filtered = db.values.filter((value) => value.body.text.includes(keyword));
+const mockCreateFieldsFromFilter = (filter) => {
+  switch (filter) {
+    case 'author':
+      return ['author'];
+    case 'post':
+    default:
+      return ['text', 'title'];
+  }
+};
+
+const search = (keyword = '', filter = 'post') => {
+  const fields = mockCreateFieldsFromFilter(filter);
+  let filtered = db.values;
+  fields.forEach((field) => {
+    filtered = filtered.filter((value) => value.body[field].includes(keyword));
+  });
   return Promise.resolve({
     results: filtered.length,
     values: filtered.map((value) => ({ id: value.id })),
--- a/src/backend/utils/__mocks__/indexer.js
+++ b/src/backend/utils/__mocks__/indexer.js
@@ -4,10 +4,17 @@ const db = {
   values: [],
 };
 
-const indexPost = ({ text, id }) => {
+const indexPost = ({ text, id, title, published, author }) => {
   db.values.push({
+    index: 'posts',
+    type: 'post',
     id,
-    text,
+    body: {
+      text,
+      title,
+      published,
+      author,
+    },
     score: 10,
   });
   db.results = db.values.length;
@@ -21,7 +28,7 @@ const deletePost = (postId) => {
 };
 
 const search = (keyword = '') => {
-  const filtered = db.values.filter((value) => value.text.includes(keyword));
+  const filtered = db.values.filter((value) => value.body.text.includes(keyword));
   return Promise.resolve({
     results: filtered.length,
     values: filtered.map((value) => ({ id: value.id })),
--- a/src/backend/utils/indexer.js
+++ b/src/backend/utils/indexer.js
@@ -13,8 +13,11 @@ const type = 'post';
  * Indexes the text and id from a post
  * @param text from a post
  * @param postId same id used to store on redis the post object where the text is from
+ * @param title from a post
+ * @param updated from a post
+ * @param author from a post
  */
-const indexPost = async (text, postId) => {
+const indexPost = async (text, postId, title, updated, author) => {
   try {
     await client.index({
       index,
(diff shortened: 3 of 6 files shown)


src/backend/data/post.js
```diff
@@ -144,9 +145,13 @@ class Post {
...
       article.link,
       article.guid,
-      feed
+      feed,
+      article.author
     );
-    await Promise.all([post.save(), indexPost(post.text, post.id)]);
+    await Promise.all([
+      post.save(),
+      indexPost(post.text, post.id, post.title, post.updated, post.author),
```
manekenpix, line 153: You can find `author` inside of `feed` (line 148). 
```suggestion
      indexPost(post.text, post.id, post.title, post.updated, post.feed.author),
```
Check [here](https://github.com/Seneca-CDOT/telescope/blob/3e0592038bca898e7df11e90611104d39e3b09ee/src/backend/data/feed.js#L27) what's inside of a `feed` object.

src/backend/data/post.js
```diff
@@ -40,7 +40,7 @@ function ensureFeed(feed) {
 }
 
 class Post {
-  constructor(title, html, datePublished, dateUpdated, postUrl, guid, feed) {
+  constructor(title, html, datePublished, dateUpdated, postUrl, guid, feed, author) {
```
manekenpix, line 43: You can use `author` included in `feed` (line 148).

src/backend/data/post.js
```diff
@@ -49,6 +49,7 @@ class Post {
     this.updated = ensureDate(dateUpdated, datePublished);
     this.url = postUrl;
     this.guid = guid;
+    this.author = author;
```
manekenpix, line 52: You can use `author` included in `feed` (line 148).

src/backend/data/post.js
```diff
@@ -166,7 +171,8 @@ class Post {
       postData.updated,
       postData.url,
       postData.guid,
-      feed
+      feed,
+      postData.author
```
manekenpix, line 175: `author` is included in `feed` (line 148).

src/backend/data/post.js
```diff
@@ -191,7 +197,8 @@ class Post {
       data.updated,
       data.url,
       data.guid,
-      feed
+      feed,
+      data.author
```
manekenpix, line 201: `author` is included in `feed` (line 148).

src/backend/utils/indexer.js
```diff
@@ -56,7 +62,7 @@ const search = async (textToSearch) => {
       simple_query_string: {
         query: textToSearch,
         default_operator: 'and',
-        fields: ['text'],
+        fields: ['text', 'title', 'author'],
```
manekenpix, line 65: I wonder if we should choose between fields to address different filters.
```suggestion
        fields: filter === 'post' ? ['text'] : ['author'],
```

src/backend/utils/storage.js
```diff
@@ -121,7 +121,9 @@ module.exports = {
         'guid',
         post.guid,
         'feed',
-        post.feed
+        post.feed,
+        'author',
+        post.author
```
manekenpix, line 126: `author` is already stored in `feed`. We don't need this.

test/post.test.js
```diff
@@ -18,6 +18,7 @@ describe('Post data class tests', () => {
     url: 'https://user.post.com/?post-id=123',
     guid: 'https://user.post.com/?post-id=123&guid',
     id: hash('https://user.post.com/?post-id=123&guid'),
+    author: 'Unknown',
```
manekenpix, line 21: `author` is stored in a `feed` object. We don't need this

test/post.test.js
```diff
@@ -34,7 +35,16 @@ describe('Post data class tests', () => {
...
-    new Post(data.title, data.html, data.published, data.updated, data.url, data.guid, feed);
+    new Post(
+      data.title,
+      data.html,
+      data.published,
+      data.updated,
+      data.url,
+      data.guid,
+      feed,
+      data.author
```
manekenpix, line 46: `author` is stored in a `feed` object. We don't need this

test/post.test.js
```diff
@@ -49,6 +59,7 @@ describe('Post data class tests', () => {
     expect(post.updated).toEqual(data.updated);
     expect(post.guid).toEqual(data.guid);
     expect(post.feed).toEqual(feed);
+    expect(post.author).toEqual(data.author);
```
manekenpix, line 62: `author` is already stored in `feed`. We don't need this

test/post.test.js
```diff
@@ -60,7 +71,8 @@ describe('Post data class tests', () => {
       'Thu, 20 Nov 2014 18:59:18 UTC',
       data.url,
       data.guid,
-      feed
+      feed,
+      data.author
```
manekenpix, line 75: `author` is stored in a `feed` object. We don't need this

test/post.test.js
```diff
@@ -75,15 +87,25 @@ describe('Post data class tests', () => {
       new Date('Thu, 20 Nov 2014 18:59:18 UTC'),
       data.url,
       data.guid,
-      feed
+      feed,
+      data.author
```
manekenpix, line 91: `author` is stored in a `feed` object. We don't need this

test/post.test.js
```diff
@@ -75,15 +87,25 @@ describe('Post data class tests', () => {
...
-      new Post(data.title, data.html, data.published, data.updated, data.url, data.guid, f);
+      new Post(
+        data.title,
+        data.html,
+        data.published,
+        data.updated,
+        data.url,
+        data.guid,
+        f,
+        data.author
```
manekenpix, line 107: `author` is stored in a `feed` object. We don't need this

test/post.test.js
```diff
@@ -186,6 +208,7 @@ describe('Post data class tests', () => {
       expect(post.guid).toEqual('5df7bbd924511e03496b4734');
       expect(post.id).toEqual(hash(post.guid));
       expect(post.feed).toEqual(feed);
+      expect(post.author).toEqual('David Humphrey');
```
manekenpix, line 211: `author` is stored in a `feed` object. We don't need this

test/posts.test.js
```diff
@@ -87,7 +87,8 @@ describe('test /posts/:id responses', () => {
     new Date('2009-09-07T22:23:00.000Z'),
     'url',
     randomGuid,
-    feed1
+    feed1,
+    'unknown'
```
manekenpix, line 91: `author` is stored in a `feed` object. We don't need this

test/posts.test.js
```diff
@@ -99,7 +100,8 @@ describe('test /posts/:id responses', () => {
...
```
manekenpix, line 104: `author` is stored in a `feed` object. We don't need this

manekenpix (changes requested):

src/backend/utils/indexer.js
StellaJung-Student: I also wonder that we need the feature? Then we can do it.
I looked at the code and I realized that for author, the search used graphql~.
Please let me clear~
I thought the reason the fields are added in ES is checking more fields only for the post case.

src/backend/data/post.js
StellaJung-Student: I didn't know it~ I followed your suggestion and removed all my author parts.

src/backend/utils/indexer.js
manekenpix: Yes, for `author` we use graphql, but we're in the process of removing it and we will probably use ES instead.
Adding `author` here is fine, it's just that I think we should add it like in my suggestion so we can search by `text` **or** `author` depending on the selected filter in the front-end. The way you add it means that ES will try to find the text in the posts **and** and the authors, which is something that might be added in the future, but currently doesn't match the way the front-end works.

StellaJung-Student:
I heard my task is adding more fields so I put all fields for post type. 
If you don't want, I will modify it.

manekenpix:
@StellaJung-Student I think it's better if you add the `author` from `feed` to ES, instead of creating a new property in `post`.

src/backend/data/post.js
```diff
@@ -146,7 +146,10 @@ class Post {
       article.guid,
       feed
     );
-    await Promise.all([post.save(), indexPost(post.text, post.id)]);
+    await Promise.all([
+      post.save(),
+      indexPost(post.text, post.id, post.title, post.updated, post.feed.author),
```
manekenpix, line 151: Should we use `published` instead of `updated`? Both are useful, but I think `published` might make more sense?

cc @c3ho @humphd

manekenpix (approved):
LGTM :+1:

src/backend/data/post.js
humphd: What's your thinking here?  Doesn't `updated` give us a more accurate sense of the timing of this post?

src/backend/utils/indexer.js
```diff
@@ -13,15 +13,21 @@ const type = 'post';
  * Indexes the text and id from a post
  * @param text from a post
  * @param postId same id used to store on redis the post object where the text is from
+ * @param title from a post
+ * @param updated from a post
+ * @param author from a post
  */
-const indexPost = async (text, postId) => {
+const indexPost = async (text, postId, title, updated, author) => {
```
humphd, line 20: Let's just pass the entire `post` object in, and then you can pick off the properties you want in this function vs. having to get the order of the arguments correct above.

src/backend/utils/indexer.js
```diff
@@ -48,15 +54,16 @@ const deletePost = async (postId) => {
...
  */
-const search = async (textToSearch) => {
+const search = async (textToSearch, filter = 'post') => {
   const query = {
     query: {
       simple_query_string: {
         query: textToSearch,
         default_operator: 'and',
-        fields: ['text'],
+        fields: filter === 'post' ? ['text', 'title', 'author'] : ['author'],
```
humphd, line 66: We might add more filters later (e.g., by date, or just title, ...) so let's make a function for this instead:

```js
const createFieldsFromFilter = (filter) => {
  switch(filter) {
    case 'author':
      return ['author']
    case 'post':
    default:
      return ['text', 'title'];
    }
}
```

And then you can call it here:

```
fields: createFieldsFromFilter(filter),
```

humphd (changes requested):
Let's get this merged before we take #552, since it depends on some of this.

src/backend/data/post.js
manekenpix: Yes, you're right. I was thinking that maybe the original date was the one that we'd like to use, but `updated` makes more sense.

StellaJung-Student:
Thank you, @humphd 
I accepted your advice.
And for the author, I didn't know how to get nested property so I made a getter for author. Is it OK?

src/backend/utils/indexer.js
```diff
@@ -65,7 +88,7 @@ const search = async (textToSearch) => {
     body: { hits },
   } = await client.search({
     from: 0,
-    _source: ['id'],
+    // _source: ['id'],
```
humphd, line 91: Let's not do this.  Remove it if it's not needed.

src/backend/utils/indexer.js
```diff
@@ -74,7 +97,7 @@ const search = async (textToSearch) => {
 
   return {
     results: hits.total.value,
-    values: hits.hits.map(({ _id }) => ({ id: _id })),
+    values: hits.hits, // .map(({ _id }) => ({ id: _id })),
```
humphd, line 100: Get rid of code, don't comment it out.

src/backend/utils/indexer.js
```diff
@@ -45,18 +53,33 @@ const deletePost = async (postId) => {
   }
 };
 
+/**
+ * Creates fields from filter
```
humphd, line 57: Add a note of what some examples of `filter` are (e..g, `post`, `author`)

src/backend/utils/indexer.js
```diff
@@ -11,21 +11,29 @@ const type = 'post';
...
- * @param postId same id used to store on redis the post object where the text is from
+ * @param post
+ * @param post.text from a post
+ * @param post.id same id used to store on redis the post object where the text is from
+ * @param post.title from a post
+ * @param post.published from a post
+ * @param post.author from a post
  */
-const indexPost = async (text, postId) => {
+// post.text, post.id, post.title, post.updated, post.feed.author
```
humphd, line 21: Remove this

humphd (changes requested):
Looking really close, a few more things.

src/backend/utils/indexer.js
StellaJung-Student: Sorry, I thought I removed all of parts of test....

src/backend/utils/__mocks__/indexer.js
```diff
@@ -4,9 +4,9 @@ const db = {
   values: [],
 };
 
-const indexPost = (text, postId) => {
+const indexPost = ({ text, id }) => {
```
humphd, line 7: This doesn't match the signature of the updated `indexPost` you have below:

```js
const indexPost = async ({ text, id, title, published, author }) => {
```

humphd (changes requested):

src/backend/utils/__mocks__/indexer.js
StellaJung-Student: Thanks, you are right~, @humphd 
Do you want me to add those parameters as optional ones?
Or just follow the all parameters?

humphd: We need our mock version (the fake one we use in tests when there's no Elasticsearch service running) to work identically to the real one.  So yes, update the mock so it adds all that data from a `post` object.

humphd (changes requested):

src/backend/data/post.js
```diff
@@ -146,7 +150,7 @@ class Post {
       article.guid,
       feed
     );
-    await Promise.all([post.save(), indexPost(post.text, post.id)]);
+    await Promise.all([post.save(), indexPost(post)]);
```
humphd, line 153: Reading this file again, I wonder if we should really be doing `indexPost(post)` inside `post.save()` vs. here.  If a post gets update and needs to be save to the db.

It looks like we don't currently do that, but later on we might want to shuffled this around.  cc @manekenpix, @c3ho.

@StellaJung-Student I think this is fine, but I wanted to make a note.

src/backend/utils/__mocks__/indexer.js
```diff
@@ -21,7 +28,7 @@ const deletePost = (postId) => {
 };
 
 const search = (keyword = '') => {
```
humphd, line 30: This doesn't match your new signature either: `const search = async (textToSearch, filter = 'post') => {`

humphd (changes requested):

StellaJung-Student:
I am really not sure that I got the error when I import createFieldsFromFilter from indexer so I repeated it.
Is it because of that unit test should be run in the same file?
So, I made the mockCreateFieldsFromFilter.

src/backend/utils/__mocks__/indexer.js
```diff
@@ -20,8 +27,22 @@ const deletePost = (postId) => {
   return Promise.resolve();
 };
 
-const search = (keyword = '') => {
-  const filtered = db.values.filter((value) => value.text.includes(keyword));
+const mockCreateFieldsFromFilter = (filter) => {
```
humphd, line 30: This is fine to duplicate for now.  Rename it to `createFieldsFromFilter` like the other.

src/backend/utils/__mocks__/indexer.js
```diff
@@ -20,8 +27,22 @@ const deletePost = (postId) => {
...
+  switch (filter) {
+    case 'author':
+      return ['author'];
+    case 'post':
+    default:
+      return ['text', 'title'];
+  }
+};
+
+const search = (keyword = '', filter = 'post') => {
```
humphd, line 40: This still doesn't match what you're doing in the other file.  It needs to be identical:

```
const search = async (textToSearch, filter = 'post') => {
```

humphd (changes requested):

humphd (approved):
Good job.

manekenpix (approved):
:+1:

Issue #1225, opened 2020-10-21
Reposition search button for search UI


humphd:
Our frontend kicks off searches every time you start typing.  I'd love to see us:

- add a button so it doesn't happen until I click or press 'enter'
- and/or debounce this keyboard event so it doesn't start happening until I finish typing

If you type something like, it will fire and re-fire the query on every letter.  That's bad UI, and it overloads the network and React.

agarcia-caicedo:
There currently is a button, but I'm not sure about the code surrounding it, so I'm not sure if its currently doing what its supposed to or not.

humphd:
I didn't realize that icon was clickable.  Could we move it down so it's inline with the text box that it goes with?  It looks like decoration vs. usable UI to me.

TDDR:
Hello, I would like to take a stab at this one if you could assign me please.

c3ho:
@manekenpix is this being fixed as part of your PR?

manekenpix:
Yes, this is addressed in #1295

c3ho:
@TDDR sorry about that, do you mind picking up another issue?

manekenpix:
#1295 doesn't move its position though, so that's something that could be done. @TDDR

c3ho:
With #1295 , I'm going to change the title of this to repositioning it instead.

strawberries73:
@humphd I would like to work on this issue.

humphd:
@strawberries73 go ahead and assign to yourself.

PR #1295, opened 2020-11-06
Closes #552: Remove GraphQL implementation

### LINKED ISSUE #552: Remove GraphQL implementation

manekenpix:
## Issue This PR Addresses
Closes #552 
Progress towards #1224 
Progress towards #1225 

## Description
This removes the `graphql` implementation and documentation. Since removing `graphql` breaks the search functionality for `author`, I made some changes so the front-end uses the same endpoint to search both authors and posts, taking advantage of what was added in #1280.

Other changes:
- Entering input doesn't trigger a search, only hitting enter or clicking on the search icon (this is part of #1225)
- Searching for author returns posts by that author(s), instead of the author card. Sorting the returned posts could be addressed in #1224 
- The filter in the search page is now set to `post` by default.

## Checklist


c3ho:
I'll review this tomorrow. I saved a branch of this in case I need to remember some GraphQL stuff

cindyledev (approved):
I tested this locally using Docker on Windows.
I was able to search for posts using the search button. As mentioned, searching using the Author filter doesn't work.

This PR doesn't fully remove all GraphQL from Telescope. There are still bits of it in our 
[Banner](https://github.com/Seneca-CDOT/telescope/blob/master/src/frontend/src/components/DynamicImage/DynamicImage.jsx), my poorly implemented About page (it has graphql in multiple places unfortunately)
https://github.com/Seneca-CDOT/telescope/blob/master/src/frontend/src/templates/template.js
https://github.com/Seneca-CDOT/telescope/blob/master/src/frontend/gatsby-node.js.

I was surprised that none of the other components broke in this PR

manekenpix:
@cindyledev 
>I was able to search for posts using the search button. As mentioned, searching using the Author filter doesn't work
I just tested it searching by Author and it works. Can you tell what to do to reproduce your results?

>This PR doesn't fully remove all GraphQL from Telescope
This is intentional. The `graphql` pieces left in the front-end are necessary to build pages (I think `gatsby` needs it) or to fetch metadata. In both cases, they don't talk to the `graphql` server in the back-end, which is what was removed in this PR.

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -33,92 +30,36 @@ const useStyles = makeStyles(() => ({
...
+      if (!res.ok) {
+        throw new Error(res.statusText);
+      }
+      const searchResults = await res.json();
+      // ES values property contains an array of objects with a (feed) id property
+      const posts = searchResults.values.map((result) => ({
+        id: result.id,
+        url: `/posts/${result.id}`,
+      }));
+      setFetchLoading(false);
```
humphd, line 53: Do you want to move this to a `finally` block at the end so we always do it?

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -33,92 +30,36 @@ const useStyles = makeStyles(() => ({
...
+  const search = async () => {
+    const encodedSearchText = encodeURIComponent(searchText);
+    try {
+      setFetchLoading(true);
+      const res = await fetch(`${telescopeUrl}/query?text=${encodedSearchText}&filter=${filter}`);
+      if (!res.ok) {
+        throw new Error(res.statusText);
+      }
+      const searchResults = await res.json();
+      // ES values property contains an array of objects with a (feed) id property
```
humphd, line 48: This comment seems wrong, we pass  `post` ids, no?

Also, I think we should do this transform from `{id}` to `{id, url}` on the server, either in this PR or another.  Let's prepare the results to be useful right away from the REST api.  I actually thought we were doing this in another PR, which might still be in review?  cc @c3ho.

humphd (changes requested):
This is awesome, thanks for doing it, and sorry to throw away all the good work you, @c3ho and others did.  It was great to experiment with this.

Do we still need to do `scriptSrc: ["'self'", 'cdn.jsdelivr.net', "'unsafe-inline'"],` in the express app if we rip out the graphql playground (the cdn I mean)?

src/frontend/src/components/SearchPage/SearchPage.jsx
c3ho: I think we merged all new ES related PRs last week. I checked the [code](https://github.com/Seneca-CDOT/telescope/pull/1294/files) and we don't store `feedID` at least from what I see

manekenpix: I'll change this. I mainly deleted what's not needed anymore but I didn't really pay much attention to the code that's going to stay.
Also, I think the backend already sends `{id, url}` (I think it's like this since the very beginning of telescope) it does it [here](https://github.com/Seneca-CDOT/telescope/blob/464dd741fbd8e4519da6b888ed63ddd1c0d81452/src/backend/web/routes/posts.js#L68), so I'll remove the transform (sorry about that, again I just deleted stuff without putting too much attention on what's going to stay).

humphd: Awesome, thank you both for looking into this.  I'm reviewing this code again since we're touching it; sorry to make work for you @manekenpix.  We can file follow-ups on anything we find vs. delaying this surgery from being completed.

manekenpix: My bad, the endpoint returning `{id, url}` is `GET /posts`. I just changed the indexer to return the same thing, and removed the transformation from the front-end.

c3ho (approved):
Tested:
Confirmed search was not firing after `onChange`
Confirmed default search is `Post`
Confirmed search results by searching Telescope for `Post` and James for `Author` only authors with James in their name were displayed

humphd (approved):
Awesome.

PR #1328, opened 2020-11-11
Fixes #Issue 1279: Better errors for failures with Search and ElasticSearch

### LINKED ISSUE #1279: Better errors for failures with Search and Elasticsearch
In our search code we have some issues with how errors are handled.  Both the backend and frontend are doing the wrong thing.

When Elasticsearch fails, it will end up right here https://github.com/Seneca-CDOT/telescope/blob/master/src/backend/web/routes/query.js#L28
src/backend/web/routes/query.js:28

```js
    res.send(`There was an error while executing your query: ${error}`);
```

This is wrong, since we're sending a 200 response with a string vs. an error code with an error message.

In the frontend, because we always send a 200 response, it assumes the error message is a result.  In this line we will get back that "error" message as a String https://github.com/Seneca-CDOT/telescope/blob/master/src/frontend/src/components/SearchPage/SearchPage.jsx#L94

```js
          const searchResults = await res.json();
```

And just assume that's the results from Elasticsearch.  It will fail on the next line, 96, and that will throw, then it will land on 103, console log, and quit.

We should be doing something better on the server and frontend when things go wrong.

HyperTHD:
## Issue This PR Addresses
Fixes #1279 

## Type of Change

## Description

This PR attempts to address the bug involving an incorrect response status being sent in circumstances not already covered, such as when elasticSearch is shut downed. As this fix also affects the UI on the front-end, this PR also attempts to update the front-end UI to put informative message to the user that the post could not be retrieved. 

An example of the correct API response that should be sent:
PR

## Checklist

### DIFF
--- a/src/frontend/src/components/SearchPage/SearchResults.jsx
+++ b/src/frontend/src/components/SearchPage/SearchResults.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { makeStyles } from '@material-ui/core/styles';
 import { useSWRInfinite } from 'swr';
-import { Container } from '@material-ui/core';
+import { Container, Box, Grid, Typography, ListSubheader } from '@material-ui/core';
 
 import useSiteMetadata from '../../hooks/use-site-metadata';
 import Timeline from '../Posts/Timeline.jsx';
@@ -38,10 +38,18 @@ const SearchResults = ({ text, filter }) => {
   );
 
   if (error) {
-    // TODO: https://github.com/Seneca-CDOT/telescope/issues/1279
     return (
       <Container className={classes.searchResults}>
-        <p>Error loading search results</p>;
+        <Box className={classes.root} boxShadow={2} marginTop={10}>
+          <ListSubheader>
+            <Typography variant="h1" color="secondary" className={classes.title}>
+              <Grid container className={classes.error}>
+                {' '}
+                There was an error while processing your query
+              </Grid>
+            </Typography>
+          </ListSubheader>
+        </Box>
       </Container>
     );
   }
--- a/src/backend/web/routes/query.js
+++ b/src/backend/web/routes/query.js
@@ -11,7 +11,7 @@ router.get('/', validateQuery(), async (req, res) => {
     const { text, filter, page, perPage } = req.query;
     res.send(await search(text, filter, page, perPage));
   } catch (error) {
-    res.send(`There was an error while executing your query: ${error}`);
+    res.status(500).send(`There was an error while executing your query: ${error}`);
     logger.error({ error }, 'Something went wrong with search indexing');
   }
 });


HyperTHD:
The build tests failing was expected since I did not update the test files. With that in mind, I'm confused with why the return status was 500 in the test case with multiple characters since it should be checked in the if statement

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -63,8 +64,8 @@ const SearchPage = () => {
       );
     }
 
-    if (!results) {
-      return null;
+    if (results === undefined) {
```
c3ho, line 67:
```
if (!results) {
//code here
}```

**Review by c3ho — CHANGES_REQUESTED** [review]

**Comment by humphd** [comment]
@HyperTHD how is this coming?

**Comment by HyperTHD** [comment]
@c3ho So I added what you requested. Only problem now is because results is assigned asynchronously, that check will return true as soon as the page loads. I'm stumped on what to do at this point. I put some sample jsx that I can change easily. But my main concern is how to showcase the jsx upon the search being completed and not when the page loads the first time

File: src/frontend/src/components/SearchPage/SearchPage.jsx [hunk 2]
```diff
@@ -43,6 +44,7 @@ const SearchPage = () => {
       setResults(posts.values);
     } catch (error) {
       console.error('Something went wrong while fetching data', error);
+      setSearchError(true);
```
humphd, line 47: Set it to the `error` object vs. `true` so we have access to it, in case we ever need to use data within it.

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -57,19 +59,19 @@ const SearchPage = () => {
         </h1>
       );
     }
-    if (!results) {
-      return (
+    if (!results || searchError) {
```
humphd, line 62: I don't think these are the same thing.  If there's an error, it's worth retrying.  If there are no results, searching again won't produce results.  Let's split these up.

humphd (changes requested):

c3ho:
@humphd can we roll #1376 into this and turn the error message part into a new component that displays whatever passed as a prop: "No results" or "Posts failed to load" (something like this)?

humphd:
@c3ho yeah, we can do that.  I reworked this code a bunch in https://github.com/jossiey/telescope/pull/1 btw.

src/frontend/src/components/SearchPage/SearchPage.jsx
HyperTHD: While I agree that should be the case, we're talking about a 500 server error no? If an error occurs on the server-side, the user is not obligated to retry because it's something that's out of their control. That's what this issue is specifically targeting, which is why I believe they should be together. I can assign searchError to be the error object that can be used later on though.

humphd:
@HyperTHD what is happening here? Are you hoping to get this done on time for 0.3?  If so, we need to co-ordinate this with what's happening in #1384 (cc @joelazwar).  I think that one is pretty close to being done.

HyperTHD:
@humphd @c3ho Let me know if I need to do anything else or if I'm good to merge

humphd (approved):

c3ho (approved):

src/frontend/src/components/SearchPage/SearchResults.jsx
```diff
@@ -38,10 +38,18 @@ const SearchResults = ({ text, filter }) => {
...
   if (error) {
-    // TODO: https://github.com/Seneca-CDOT/telescope/issues/1279
     return (
       <Container className={classes.searchResults}>
-        <p>Error loading search results</p>;
+        <Box className={classes.root} boxShadow={2} marginTop={10}>
+          <ListSubheader>
+            <Typography variant="h1" color="secondary" className={classes.title}>
+              <Grid container className={classes.error}>
+                {' '}
```
c3ho, line 47: Not going to block on this, do we need this line though?

Issue #1296, opened 2020-11-06
Fix code within blog posts


smilegodly:
**What would you like to be added**:

Alot of code in Telescope posts have no code blocks and are just in plain text

![image](https://user-images.githubusercontent.com/35276477/98364888-8eedc200-1fff-11eb-97c5-95d9177016d8.png)

![image](https://user-images.githubusercontent.com/35276477/98364948-ad53bd80-1fff-11eb-9798-875a21615d62.png)

**Why would you like this to be added**:

Code should be indented nicely in the posts, with syntax highlighting

smilegodly:
This issue is similiar to https://github.com/Seneca-CDOT/telescope/issues/1250 except all the code in this issue are in  empty `<div>` tags

humphd:
This seems risky.  How do you identify code in a post if it isn't marked up?  I also see some people use tables, rows, and table cells for code.

I think it's going to be hard to get this right.

humphd:
Here's a post that we could maybe fix: https://egrinberg.medium.com/lab-6-dps909-11767c605e33

It uses a `<pre>` per line, so the file is `<pre>...</pre><pre>...</pre><pre>...</pre>...`.  We could probably collapse these together into a single `<pre>`.

smilegodly:
There are going to be lots of problematic edge cases like this for this issue. In the future more are likely to pop up.

A photo of the above mentioned case

![image](https://user-images.githubusercontent.com/35276477/98743067-77a33180-237d-11eb-9284-25b9d433bc35.png)

humphd:
It might be interesting to survey the top blog platforms, and see what the HTML they produce is like for "code" blocks.  Then we could write a parser that deals with these cases, and ignore the others (which will be rare or mistakes on the part of a user).

smilegodly:
> Here's a post that we could maybe fix: https://egrinberg.medium.com/lab-6-dps909-11767c605e33
> 
> It uses a `<pre>` per line, so the file is `<pre>...</pre><pre>...</pre><pre>...</pre>...`. We could probably collapse these together into a single `<pre>`.

I'm picking this issue back up again.

humphd:
I believe this is fixed.  I haven't seen any issues since we landed the changes to the parser in the fall.  Closing.

PR #1331, opened 2020-11-11
Fix #1250: syntax highlighting for all empty <pre>

### LINKED ISSUE #1250: Syntax highlighting for <pre> vs. <pre><code>
I noticed that we aren't syntax highlighting all code blocks:

Screen Shot 2020-10-27 at 1 29 09 PM

It looks like this uses a `<pre>` without a `<code>`.  I wonder if we should be looking for `<pre>` without `<code` as well?

smilegodly:
## Issue This PR Addresses
Attempts to fix #1250 

This PR adds to the number of code posts that get syntax highlighted. 
There are too many edge cases to cover here so I included the cases in this PR that have *most* of the syntax highlighting issues.

## Type of Change

## Description

Here are some before and after photos of the fix..

Before...

before

And after...

after

Here is another Before....

before2

And after...

after2

## Conclusions

This [Wordpress](https://nesabyte.wordpress.com/ ) blog seems to be the only problem case that wouldn't get fixed in this PR.

All of the code posts from the blog https://nesabyte.wordpress.com/ do not get fixed in this PR sadly, even though at a first glance, that blog seems to have the same edge cases as the above fixes.

I looked at all of the posts from the past 2 months and pretty much all of the problem edge cases that weren't getting fixed were coming from this particular [Wordpress](https://nesabyte.wordpress.com/) blog.

I think the correct move is to write a parser covering Wordpress, Medium, etc as mentioned in https://github.com/Seneca-CDOT/telescope/issues/1296

Here is one example of a case from the above mentioned blog that has the issue still. 

image

## Checklist

### DIFF
--- a/src/backend/utils/html/modify-pre.js
+++ b/src/backend/utils/html/modify-pre.js
@@ -1,12 +1,16 @@
 /**
- * Take a <pre></pre> and return <pre><code>...</code></pre>
+ * Take a <pre>...</pre> and return <pre><code>...</code></pre>
  */
 function insertCodeElement(pre, document) {
   const innerElem = document.createElement('code');
   innerElem.innerText = pre.innerText;
   pre.innerHTML = innerElem.outerHTML;
 }
-
+/**
+ * Given a parsed JSDOM Object. find all <pre></pre> tags without child code elements and modify the
+ * pre tag to have inner <code></code> tags which wraps around the previous content
+ * resulting in <pre><code>...</code></pre>
+ */
 module.exports = function (dom) {
   if (!(dom && dom.window && dom.window.document)) {
     return;
--- a/src/backend/utils/html/modify-pre.js
+++ b/src/backend/utils/html/modify-pre.js
@@ -1,14 +1,14 @@
 /**
  * Take a <pre>...</pre> and return <pre><code>...</code></pre>
  */
-function insertCodeElement(pre, document) {
-  const innerElem = document.createElement('code');
-  innerElem.innerHTML = pre.innerHTML;
-  pre.innerHTML = innerElem.outerHTML;
+function insert(pre, document) {
+  const codeElem = document.createElement('code');
+  codeElem.innerHTML = pre.innerHTML;
+  pre.innerHTML = codeElem.outerHTML;
 }
 /**
  * Given a parsed JSDOM Object. find all <pre></pre> tags without child code elements and modify the
- * pre tag to have inner <code></code> tags which wraps around the previous content
+ * Pre tag to have inner <code></code> tags which wraps around the previous content
  * resulting in <pre><code>...</code></pre>
  */
 module.exports = function (dom) {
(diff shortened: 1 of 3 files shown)


PedroFonsecaDEV (approved):
@smilegodly 
Good job.
See you.

src/backend/utils/html/modify-pre.js
```diff
@@ -0,0 +1,28 @@
+/**
+ * Take a <pre>...</pre> and return <pre><code>...</code></pre>
+ */
+function insert(pre, document) {
```
humphd, line 4: `insertCode` might be a better name for this.

src/backend/utils/html/modify-pre.js
```diff
@@ -0,0 +1,28 @@
...
+module.exports = function (dom) {
+  if (!(dom && dom.window && dom.window.document)) {
+    return;
+  }
+
+  dom.window.document.querySelectorAll('pre').forEach((pre) => {
+    // no child elements
+    if (pre.children.length <= 0) insert(pre, dom.window.document);
+    // has child <br> elements
+    else if (pre.children.length > 0 && pre.children[0].nodeName === 'BR') {
```
humphd, line 23: This first condition isn't necessary, since it must be greater than 0 if we didn't fall into the previous case.

test/modify-pre.test.js
```diff
@@ -0,0 +1,29 @@
...
+    expect(res).toEqual(og);
+  });
+
+  test('pre tags without inner <code></code> elements should be fixed', () => {
+    const og = '<pre>console.log("Hello World")</pre>';
+    const res = fixEmpties(og);
+    const fix = '<pre><code>console.log("Hello World")</code></pre>';
+    expect(res).toEqual(fix);
+  });
+});
```
humphd, line 29: Let's add a test for the `<pre>...<br>...<br></pre>` case you have above.

humphd (changes requested):

PedroFonsecaDEV (approved):

src/backend/utils/html/modify-pre.js
```diff
@@ -0,0 +1,28 @@
...
+  if (!(dom && dom.window && dom.window.document)) {
+    return;
+  }
+
+  dom.window.document.querySelectorAll('pre').forEach((pre) => {
+    // no child elements
+    if (pre.children.length <= 0) insertCode(pre, dom.window.document);
+    // has child <br> elements
+    else if (pre.children[0].nodeName === 'BR') {
+      pre.innerHTML = pre.innerHTML.replace(/<br>/g, '\n'); // trim
```
humphd, line 24: The one downside I can see to this is that it will be impossible to have a `<br>` in an HTML code block.  We have no way to tell a literal `<br>` from one used to add a new line.

smilegodly: That's a good point, the best choice probably is to delete the line and just live with the new line `<br>` tags in code.. atleast for now..

smilegodly: Only 1 post from 2013 uses `<br>` as a literal as pointed out by @humphd on Slack. If there happens to be unwanted side effects within the posts in the future because of this fix, we will address them in another issue.

smilegodly:
this PR should be ready to land

humphd (approved):

PR #1339, opened 2020-11-12
Fix #1313  Add query params validation for GET /posts?page=<num>&per_page=<num> route using Express Validator 

zg3d:
## Issue This PR Addresses 
https://github.com/Seneca-CDOT/telescope/issues/1313

## Type of Change

- [ ] **Bugfix**: Change which fixes an issue
- [x] **New Feature**: Change which adds functionality
- [ ] **Documentation Update**: Change which improves documentation
- [ ] **UI**: Change which improves UI

## Description

I added validation for the query parameters in the route /Posts. I just wanted feedback on what I have done will add tests if needed

## Checklist

- [x] **Quality**: This PR builds and passes our npm test and works locally
- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not (if applicable)
- [x] **Documentation**: This PR includes updated/added documentation to user exposed functionality or configuration variables are added/changed or an explanation of why it does not(if applicable)

PR #1342, opened 2020-11-13
Fix #Issue 1311 - validate GET /posts/:id route using Express Validator

### LINKED ISSUE #1311: Add query params validation for GET /posts/:id route using Express Validator and unit tests
Request to add query params validation for GET /posts/:id route using [Express-validator](https://express-validator.github.io/docs/) and also unit tests to ensure that it works

It would be better to validate the query before it reach the route handler. 

reference: https://express-validator.github.io/docs/check-api.html

All logic for the middleware should be place inside `backend/web/validation.js`

eunbeek:
## Issue This PR Addresses

Fixes #1311 
https://github.com/Seneca-CDOT/telescope/issues/1311

## Type of Change

## Description
 

The validation will check the length of the parameter 'Id' in posts. In this issue, I only applied it to GET /posts/:id but it could be applied to any posts url with id param.  Normal id length is 10 and the reason why I used the length check is the id value is encoded by hash algorithms and sliced to 10 characters. 

## Checklist


eunbeek:
Reference here!
https://github.com/express-validator/express-validator/issues/449

src/backend/web/validation.js
```diff
@@ -43,11 +44,27 @@ function validateNewFeed() {
...
+    try {
+      validationResult(req).throw();
+      next();
+    } catch (e) {
+      console.log(e);
+      res.status(400).json({ message: 'ID Length is invalid' });
+    }
+  });
+};
+
```
tonyvugithub, line 66: Can you return a middleware instead? Like below. You are on the right track
```
function validateNewFeed() {
    const feedData = req.body;
    const valid = ajv.validate(feedSchema, feedData);

    if (!valid) {
      res.status(400).send();
    } else {
      next();
    }
  };
}
```

src/backend/web/validation.js
```diff
@@ -43,11 +44,27 @@ function validateNewFeed() {
...
+      return res.status(400).send();
     } else {
-      next();
+      return next();
     }
   };
 }
 
-module.exports.validateNewFeed = validateNewFeed;
+const validateParam = (req, res, next) => {
```
tonyvugithub, line 54: Change the name to validatePostsIdParam please. It would reflect the purpose of your middleware better

src/backend/web/validation.js
tonyvugithub: so in the route you can just call validatePostsIdParam()

test/posts.test.js
```diff
@@ -166,4 +166,13 @@ describe('test /posts/:id responses', () => {
...
     expect(res.body).toEqual(JSON.parse(JSON.stringify(receivedPost1)));
   });
+
+  test('requests ID length Test', async () => {
+    const res = await request(app)
+      .get(`/posts/${addedPost1.id.slice(0, 4)}`)
+      .set('Accept', 'text/html');
+    expect(res.status).toEqual(400);
+    expect(res.get('content-length')).toContain('34');
+    //expect(res.text).toEqual(receivedPost1.html);
```
tonyvugithub: One of your tests should look something similar to this
 ```
test('should return 200 if it is valid', async () => {
    const id = <A valid id>;
    const res = await request(app).get(`/posts/${id}`);

   \\ Do test expecting here.
    expect(res.status).toEqual(200);
  });
```
I commented on the issue page with more details

**Review by tonyvugithub — CHANGES_REQUESTED** [review]
request changes. Read comment for details

File: src/backend/web/validation.js (hunk 1 above)
Comment (eunbeek) [reply src/backend/web/validation.js:66]: Thanks, I changed it back and the validation name also is changed to validatePostsIdParam.

File: test/posts.test.js (hunk 3 above)
Comment (eunbeek) [reply test/posts.test.js]: Now I have, 
1. request with over 10 characters scenario
2. request with less 10 characters scenario

I will keep thinking more case like you mentioned in issue. :)
Thanks!

File: src/backend/web/validation.js [hunk 4]
```diff
@@ -51,7 +51,7 @@ function validateNewFeed() {
   };
 }
 
-const validateParam = (req, res, next) => {
+const validatePostsIdParam = (req, res, next) => {
   param('id').isLength({ min: 10, max: 10 })(req, res, (err) => {
```
tonyvugithub, line 55: Hi, One little twist. Can you follow the same pattern as in the following commit 
https://github.com/Seneca-CDOT/telescope/pull/1339/commits/803fb2e64aecbc39459910907ed3f071f1c23b26
This one the person is doing the same thing as you do. See what I mean by return a middleware. Then what you left is the unit tests.

tonyvugithub (changes requested):

src/backend/web/validation.js
eunbeek: Hey, @tonyvugithub  So Need to change his format like this?

const validatePostsIdParamRules = [
  param('id','Id Length is invalid').isLength({ min: 10, max: 10 })
]; 

const validatePostsIdParam = () => {
    await Promise.all(validatePostsIdParamRules.map((rule)=>rule.run(req)));

    const result = validationResult(req);
    if(result.isEmpty()){
      return next();
    }
    const errors = result.array();
    return res.status(400).send(errors);
    
  };
};

I changed it with his style. Is it right?

Thanks :)

test/posts.test.js
```diff
@@ -172,14 +172,18 @@ describe('test /posts/:id responses', () => {
...
       .get(`/posts/${addedPost1.id.concat('1234')}`)
       .set('Accept', 'text/html');
     expect(res.status).toEqual(400);
-    expect(res.get('content-length')).toContain('34');
+  });
+
+  test('requests ID with valid length but not exist Test', async () => {
+    const res = await request(app).get(`/posts/1234567890`).set('Accept', 'text/html');
+    expect(res.status).toEqual(404);
+    expect(res.get('Content-length')).toEqual('46');
```
tonyvugithub, line 187: @eunbeek : This looks more like it. One last thing I want to ask from you. Can you squash all of your commit into a single commit? Then resubmit with --force again. I don't want to history to be dirty.

tonyvugithub (changes requested):

test/posts.test.js
```diff
@@ -166,4 +166,24 @@ describe('test /posts/:id responses', () => {
     expect(res.get('Content-type')).toContain('application/json');
     expect(res.body).toEqual(JSON.parse(JSON.stringify(receivedPost1)));
   });
+
+  test('requests ID with 6 characters Test', async () => {
+    const res = await request(app)
+      .get(`/posts/${addedPost1.id.slice(4)}`)
```
humphd, line 172: Simplify this to manually write something that is 6 characters long:

```js
      .get(`/posts/123456)
```

test/posts.test.js
```diff
@@ -166,4 +166,24 @@ describe('test /posts/:id responses', () => {
...
+      .set('Accept', 'text/html');
+    expect(res.status).toEqual(400);
+  });
+
+  test('requests ID with 14 characters Test', async () => {
+    const res = await request(app)
+      .get(`/posts/${addedPost1.id.concat('1234')}`)
```
humphd, line 179: Same thing here, manually do it:

```js
      .get(`/posts/12345678901234)
```

humphd (changes requested):

test/posts.test.js
eunbeek: #1359 I reopen this PR again for messed commit!

PR #1343, opened 2020-11-13
Fixes #1270: let URL in sync with keywords to make searches sharable

### LINKED ISSUE #1270: Make searches sharable via URL
Today I needed to go and look for blog posts students have written on Pandas.  I did a search, and got back my results.  Now I want to share that search and the results with a colleague.  There is no way to do this without telling them how to recreate the search.

A better solution would be to use the React routing and History API to push some state into the URL, so you get something like `https://telescope.cdot.systems/search?q=pandas` or something.  As I type, it should update the URL so it stays in sync with my search, but not in an annoying way so that we break the Back button.

When the search page is rendered, we should read out params from the URL if present, too.

jossiey:
## Issue This PR Addresses
Fixes #1270 

## Type of Change

## Description
* URL synchronizes with keyword when typing in search page. After being submitted, URL shows https://telescope.cdot.systems/search?q=keyword&f=author
* When someone browses to the URL https://telescope.cdot.systems/search?q=keyword&f=author, keyword and author are shown on the search box, click submit button to do a search.

## Checklist

### DIFF
--- a/src/frontend/src/components/SearchBar/SearchBar.jsx
+++ b/src/frontend/src/components/SearchBar/SearchBar.jsx
@@ -90,20 +90,23 @@ function CustomizedInputBase(props) {
   const classes = useStyles();
   const { searchText, onChangeHandler, onFilterChangeHandler, filter, onFormSubmit } = props;
 
+  const prepareParamsAndNavigate = (text, filter) =>
+    navigate(`/search?text=${encodeURIComponent(text)}&filter=${encodeURIComponent(filter)}`);
+
   const onFilterChange = (event) => {
     onFilterChangeHandler(event.target.value);
-    navigate(`/search?text=${searchText}&filter=${event.target.value}`);
+    prepareParamsAndNavigate(searchText, event.target.value);
   };
 
   const onTextChange = (event) => {
     onChangeHandler(event.target.value);
-    navigate(`/search?text=${event.target.value}&filter=${filter}`);
+    prepareParamsAndNavigate(event.target.value, filter);
   };
 
   const onSubmit = (event) => {
     event.preventDefault();
     onFormSubmit();
-    navigate(`/search?text=${searchText}&filter=${filter}`);
+    prepareParamsAndNavigate(searchText, filter);
   };
 
   const searchOptions = ['post', 'author'];
--- a/src/frontend/src/components/SearchBar/SearchBar.jsx
+++ b/src/frontend/src/components/SearchBar/SearchBar.jsx
@@ -90,8 +90,11 @@ function CustomizedInputBase(props) {
   const classes = useStyles();
   const { searchText, onChangeHandler, onFilterChangeHandler, filter, onFormSubmit } = props;
 
-  const prepareParamsAndNavigate = (text, filter) =>
-    navigate(`/search?text=${encodeURIComponent(text)}&filter=${encodeURIComponent(filter)}`);
+  const prepareParamsAndNavigate = (text, filter) => {
+    navigate(`/search?text=${encodeURIComponent(text)}&filter=${encodeURIComponent(filter)}`, {
+      replace: true,
+    });
+  };
 
   const onFilterChange = (event) => {
     onFilterChangeHandler(event.target.value);
--- a/src/frontend/src/components/SearchBar/SearchBar.jsx
+++ b/src/frontend/src/components/SearchBar/SearchBar.jsx
@@ -92,17 +92,18 @@ function CustomizedInputBase(props) {
 
   const onFilterChange = (event) => {
     onFilterChangeHandler(event.target.value);
+    navigate(`/search?text=${searchText}&filter=${event.target.value}`);
   };
 
   const onTextChange = (event) => {
     onChangeHandler(event.target.value);
-    navigate(`/search?q=${event.target.value}`);
+    navigate(`/search?text=${event.target.value}&filter=${filter}`);
   };
 
   const onSubmit = (event) => {
     event.preventDefault();
     onFormSubmit();
-    navigate(`/search?q=${searchText}&f=${filter}`);
+    navigate(`/search?text=${searchText}&filter=${filter}`);
   };
 
   const searchOptions = ['post', 'author'];
--- a/src/frontend/src/components/SearchBar/SearchBar.jsx
+++ b/src/frontend/src/components/SearchBar/SearchBar.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { makeStyles } from '@material-ui/core/styles';
-
+import { navigate } from 'gatsby';
 import PropTypes from 'prop-types';
 import SearchIcon from '@material-ui/icons/Search';
 
@@ -96,11 +96,13 @@ function CustomizedInputBase(props) {
 
   const onTextChange = (event) => {
     onChangeHandler(event.target.value);
+    navigate(`/search?q=${event.target.value}`);
   };
 
   const onSubmit = (event) => {
     event.preventDefault();
     onFormSubmit();
+    navigate(`/search?q=${searchText}&f=${filter}`);
   };
 
   const searchOptions = ['post', 'author'];
--- a/src/frontend/src/components/SearchBar/SearchBar.jsx
+++ b/src/frontend/src/components/SearchBar/SearchBar.jsx
@@ -1,6 +1,5 @@
 import React from 'react';
 import { makeStyles } from '@material-ui/core/styles';
-import { navigate } from 'gatsby';
 import PropTypes from 'prop-types';
 import SearchIcon from '@material-ui/icons/Search';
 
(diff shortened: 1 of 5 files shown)


src/frontend/src/components/SearchBar/SearchBar.jsx
```diff
@@ -96,11 +96,13 @@ function CustomizedInputBase(props) {
 
   const onTextChange = (event) => {
     onChangeHandler(event.target.value);
+    navigate(`/search?q=${event.target.value}`);
   };
 
   const onSubmit = (event) => {
     event.preventDefault();
     onFormSubmit();
+    navigate(`/search?q=${searchText}&f=${filter}`);
```
humphd, line 105: Why deviate from `text` and `filter`?  This seems like it will cause confusion between frontend and backend.

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -26,10 +26,15 @@ const useStyles = makeStyles(() => ({
 const SearchPage = () => {
   const { telescopeUrl } = useSiteMetadata();
   const classes = useStyles();
-
-  const [searchText, setSearchText] = useState('');
+  let [keyword] = useQueryParam('q', StringParam);
+  let [filters] = useQueryParam('f', StringParam);
```
humphd, line 30: It's just `filter` vs. `filters` (singular).

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -26,10 +26,15 @@ const useStyles = makeStyles(() => ({
 const SearchPage = () => {
   const { telescopeUrl } = useSiteMetadata();
   const classes = useStyles();
-
-  const [searchText, setSearchText] = useState('');
+  let [keyword] = useQueryParam('q', StringParam);
```
humphd, line 29: Probably this should be `text` or `searchText`

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -26,10 +26,15 @@ const useStyles = makeStyles(() => ({
...
+  let [filters] = useQueryParam('f', StringParam);
+  if (!keyword) {
```
humphd, line 31: This shouldn't be necessary.  In the UI rendering below, substitute default values if they are missing.

src/frontend/src/components/SearchPage/SearchPage.jsx
```diff
@@ -26,10 +26,15 @@ const useStyles = makeStyles(() => ({
...
+    keyword = '';
+    filters = 'post';
+  }
+  const [searchText, setSearchText] = useState(`${keyword}`);
```
humphd, line 35: Let's not mix the query param hook and `useState`.  We shouldn't need to use state at all, just the values coming from `userQueryParam`

humphd (changes requested):

src/frontend/src/components/SearchBar/SearchBar.jsx
jossiey: Cause "/search?q={...}" is the standard query string, I thought it could be easily to be understood by others. I could make it consistence with our backend.

src/frontend/src/components/SearchPage/SearchPage.jsx
jossiey: I tried to combine them like this " const [searchText, setSearchText] = useState(`${keyword}`|| '' );" But when keyword is missing, "undefined" is shown in the searchbox instead, not ' '. I don't know why it happens.

src/frontend/src/components/SearchPage/SearchPage.jsx
humphd: You can set your defaults for these two values like this:

```js
const [searchText = ''] = useQueryParam('text', StringParam);
const [filter = 'post'] = useQueryParam('filter', StringParam);
```

src/frontend/src/components/SearchPage/SearchPage.jsx
humphd: See my comment above.  We don't need state here, since the URL's query params are the state.

src/frontend/src/components/SearchPage/SearchPage.jsx
humphd: Remove these

humphd (changes requested):

src/frontend/src/components/SearchPage/SearchPage.jsx
jossiey: I can do the initial like this?  Learned it. Thank you so much.

humphd: I had to look it up too.  I was surprised that `useQueryParam()` doesn't provide a mechanism to do it.  This is using what modern JS provides vs. what React/useQueryParam give.

jossiey: It is really helpful. I spent a lot of time last night trying to make my code streamline and effective. If you didn't mention it, I could never know this approach.

jossiey:
Modified the code as required.

src/frontend/src/components/SearchBar/SearchBar.jsx
```diff
@@ -92,15 +92,18 @@ function CustomizedInputBase(props) {
 
   const onFilterChange = (event) => {
     onFilterChangeHandler(event.target.value);
+    navigate(`/search?text=${searchText}&filter=${event.target.value}`);
```
humphd, line 95: Let's write a function above these that does a few things:

```js
const prepareParamsAndNavigate = (text, filter) =>
  navigate(`/search?text=${encodeURIComponent(text)}&filter=${encodeURIComponent(filter)}`);
```

Now in all these cases where you call `navigate` you can pass the args to `prepareParamsAndNavigate(text, filter)` and have it fix them up for URL inclusion and navigate.

humphd (changes requested):
Awesome.  One refactoring we can make, and this is probably good to go.

src/frontend/src/components/SearchBar/SearchBar.jsx
jossiey: Good suggestion. Need to improve coding skill. lol

humphd (changes requested):
Tested this on the live deployment of this PR.  If I go to https://telescope-1l33gyz50.vercel.app/search/?text=abcdef&filter=post it populates the UI from the query params, but doesn't run the search.  Let's get that added to this.

src/frontend/src/components/SearchBar/SearchBar.jsx
```diff
@@ -90,17 +90,23 @@ function CustomizedInputBase(props) {
   const classes = useStyles();
   const { searchText, onChangeHandler, onFilterChangeHandler, filter, onFormSubmit } = props;
 
+  const prepareParamsAndNavigate = (text, filter) =>
+    navigate(`/search?text=${encodeURIComponent(text)}&filter=${encodeURIComponent(filter)}`);
```
humphd, line 94: One other change to make here: let's pass `true` for `options.replace` to `navigate()` so that it doesn't add history state, but replaces what's there (i.e., then the back button will work like you expect), see: 

https://www.gatsbyjs.com/docs/gatsby-link/#replace-history-during-programmatic-navigation

humphd (changes requested):

src/frontend/src/components/SearchBar/SearchBar.jsx
jossiey: Let me do a research and try it again.

src/frontend/src/components/SearchBar/SearchBar.jsx
```diff
@@ -90,17 +90,26 @@ function CustomizedInputBase(props) {
   const classes = useStyles();
   const { searchText, onChangeHandler, onFilterChangeHandler, filter, onFormSubmit } = props;
 
+  const prepareParamsAndNavigate = (text, filter) => {
+    navigate(`/search?text=${encodeURIComponent(text)}&filter=${encodeURIComponent(filter)}`, {
```
humphd, line 94: Testing this, I think that using navigate is wrong, since it's continuing to accumulate history despite the `replace:true`.  To see what I mean, type something like `react native` and then hit the Back button.  You have to do it for every letter that you typed.

I think we can drop the changes to this file completely, and just use what you have in the SearchPage, since `useQueryParams` lets you set the text and filter values already.

humphd (changes requested):

src/frontend/src/components/SearchBar/SearchBar.jsx
jossiey: Let me clarify this problem. My understanding is when click back button on `http://telescope/search?text=abcde&filter=post`, it should go back to `http://telescope/search/` or other previous history page, not go back to `http://telescope/search?text=abcd&filter=post`. Gatsby navigate( {replace: true}) supposed to do that. But actually, it could not achieve the effect correctly. I also find the problem, but I thought maybe my understanding was wrong. Now I got it. Let me try what else I can do to get there.

jossiey: I tried to remove all navigate and modification from SearchBar.jsx, problem is still there. I think when we use gatsby useQueryParams, it records all typed letter steps to history.

humphd:
I added some other reviewers, since this now includes code that I wrote.

To test this:

1. go to https://telescope-cpoilmarj.vercel.app/ on Vercel
1. click on Search
1. note that you can type search text, but nothing happens until you click the search button
1. clicking search will run the search and also update the URL.
1. copy the URL and open it in a new tab, you should get the same search running
1. change the search text, click search button again, notice the URL (it should update)
1. try the Back button.  It should work as you expect.

manekenpix (approved):
Tested used instructions in @humphd comments.
All 7 steps work as expected.

PedroFonsecaDEV (approved):
It's working as @humphd described.

PR #1345, opened 2020-11-13
Fixed triple-reporting lint issues

### LINKED ISSUE #1336: GitHub Actions is triple-reporting lint issues
We do GitHub Actions builds on multiple platforms, and each one reports the same errors inline in our PRs.  This is annoying, and it would be better to only have lint issues reported once:

Screen Shot 2020-11-12 at 11 49 40 AM

yzwdroid:
## Issue This PR Addresses
Fixes #1336

## Type of Change

## Description
only run `test-ci` when the os is `ubuntu`

## Checklist

### DIFF
--- a/.github/workflows/node-js-ci.yml
+++ b/.github/workflows/node-js-ci.yml
@@ -33,10 +33,10 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
-      - run: npm run prettier-check
+      - run: npm install && npm run prettier-check
   Lint:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
-      - run: npm run eslint
+      - run: npm install && npm run eslint
--- a/.github/workflows/node-js-ci.yml
+++ b/.github/workflows/node-js-ci.yml
@@ -27,4 +27,16 @@ jobs:
 
       - run: npm install
       - run: cp env.example .env
-      - run: npm run test-ci
+      - run: npm run test
+  Prettier:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+      - run: npm run prettier-check
+  Lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+      - run: npm run eslint
--- a/.github/workflows/node-js-ci.yml
+++ b/.github/workflows/node-js-ci.yml
@@ -26,7 +26,7 @@ jobs:
         with:
           # npm cache files are stored in `~/.npm` on Linux/macOS
           path: ~/.npm
-          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package.json') }}
           restore-keys: |
             ${{ runner.os }}-build-${{ env.cache-name }}-
             ${{ runner.os }}-build-
@@ -40,7 +40,7 @@ jobs:
       - run: npm install
       - run: cp env.example .env
       - run: npm run test
-  Prettier:
+  Prettier-Check:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/node-js-ci.yml
+++ b/.github/workflows/node-js-ci.yml
@@ -19,8 +19,20 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      - name: Cache node modules
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-node-modules
+        with:
+          # npm cache files are stored in `~/.npm` on Linux/macOS
+          path: ~/.npm
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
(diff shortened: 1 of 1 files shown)


humphd (changes requested):
The problem I see with this approach is that we no longer run tests on the other platforms, in which case there is no point having them listed in the matrix.

@raygervais do you have any insight into other ways to reduce the lint spam without shutting off the build matrix?

humphd:
@yzwdroid this needs a rebase, and an update on where we stand.  Are you hoping to get this done in time for 0.3?  If so, we need to finalize and merge this today.

yzwdroid:
I am thinking if we have a better solution for this. I already have a PR merged for release 0.3.  Thanks.
I'll do the rebase now.

humphd (changes requested):
Alright, I did some research and here's a better way, let's separate the lint and test scripts, and do something like [Svelte is doing](https://github.com/sveltejs/svelte/actions/runs/375353097/workflow).  Here is their workflow file:

```yml
name: CI
on: [push, pull_request]
jobs:
  Tests:
    runs-on: ${{ matrix.os }}
    timeout-minutes: 10
    strategy:
      matrix:
        node-version: [8, 10, 12, 14]
        os: [ubuntu-latest, windows-latest, macOS-latest]
    steps:
    - uses: actions/checkout@v1
    - uses: actions/setup-node@v1
      with:
        node-version: ${{ matrix.node-version }}
    - run: npm install
    - run: npm test
      env:
        CI: true
  Lint:
    runs-on: ubuntu-latest
    timeout-minutes: 2
    steps:
    - uses: actions/checkout@v1
    - uses: actions/setup-node@v1
    - run: 'npm i && npm run lint'
  Unit:
    runs-on: ${{ matrix.os }}
    timeout-minutes: 5
    strategy:
      matrix:
        os: [ubuntu-latest, windows-latest, macOS-latest]
    steps:
    - uses: actions/checkout@v1
    - uses: actions/setup-node@v1
    - run: 'npm i && npm run test:unit'
```

To do this, we'll need to get rid of [`test-ci`](https://github.com/Seneca-CDOT/telescope/blob/master/package.json#L48) and do these three things separately:

1. `npm run prettier-check` to make sure the PR has proper formatting.  Do this only on linux
1. `npm run eslint` to make sure the PR passes lint. Do this only on linux
1. `npm run test` to make sure the PR passes tests.  Do this on all OSes

yzwdroid:
Thanks for the help @humphd . I changed the condition as you mentioned.

humphd (changes requested):
You haven't done what I asked.  Can you split up the jobs so that we have separate prettier, eslint, and test phases that can run in parallel?

yzwdroid:
I am a little confused about this approach. If we run all the jobs async, then we need to do `npm install` three times in the ubuntu server. Is that slow down a little bit?

humphd:
Not if we also add caching, see https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows.

yzwdroid:
I'll add the cache, and commit it later.

.github/workflows/node-js-ci.yml
```diff
@@ -28,4 +28,16 @@ jobs:
       - run: npm install
       - run: cp env.example .env
       - run: npm run build
```
humphd, line 30: Let's get rid of this build step too, since we aren't using the build in CI.

humphd (changes requested):

.github/workflows/node-js-ci.yml
humphd: See https://github.com/Seneca-CDOT/telescope/pull/1444/commits/b615e5c12d143c6138f0c4bb2bd27d82355a6eea, where I'm doing that for another PR too.

yzwdroid:
After I add the cache, the build time reduced from 3m 30 s to  2m 54s on Ubuntu from my repo [https://github.com/yzwdroid/telescope/runs/1449300001?check_suite_focus=true](https://github.com/yzwdroid/telescope/runs/1449300001?check_suite_focus=true). I'm not sure if I have done it right.

.github/workflows/node-js-ci.yml
```diff
@@ -19,13 +19,36 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
+      - name: Cache node modules
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-node-modules
+        with:
+          # npm cache files are stored in `~/.npm` on Linux/macOS
+          path: ~/.npm
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
```
humphd, line 29: We don't use `package-lock.json` in most cases.  What does `${{ hashFiles('**/package-lock.json') }}` do and can we live without it?

.github/workflows/node-js-ci.yml
```diff
@@ -19,13 +19,36 @@ jobs:
...
         uses: actions/setup-node@v1
         with:
           node-version: ${{ matrix.node-version }}
 
       - run: npm install
       - run: cp env.example .env
-      - run: npm run build
-      - run: npm run test-ci
+      - run: npm run test
+  Prettier:
```
humphd, line 43: `Prettier-Check`

humphd (changes requested):

.github/workflows/node-js-ci.yml
yzwdroid: we recache the node modules when this file changed.

humphd: Right, but when we don't have it?

humphd:
Let's get this rebased for the conflict that's in node-js-ci.yml.

yzwdroid:
I rebased the current commit, Is there any code to be changed?

humphd (approved):

manekenpix (approved):

Issue #1323, opened 2020-11-11
Add test for post.js to call the constructor for a Post() 


joey2031:
**What would you like to be added**: In post.test.js add a test for src/backend/data/post.js that tries to call the constructor for a Post() and pass it invalid dates, and make sure we are throwing.

**Why would you like this to be added**: We have tests for many functions throughout Telescope, however, we are still missing many such as this one

c3ho:
Feel free to create a PR and mention the backend test issue.

joey2031:
@c3ho Will do, ill start working on it and if I run into any issues ill bring it up in Slack, Thanks!

joey2031:
@c3ho So I made quite a bit of progress for this issue I just don't know why the test keeps failing when I pass (invalidDate, validDate) I know it shouldn't throw since the fallbackDate is valid.  I tested the ensureDate function in a separate project and made some adjustments to it. In the other project, I called it in a try-catch and it didn't throw so I'm not sure what's wrong when I put it in Telescope. Thanks!


Fails at line 115
![image](https://user-images.githubusercontent.com/36198917/99121961-63547400-25cb-11eb-9a44-278488cac511.png)

```javascript
 test('Post constructor should throw if invalid dates is passed', () => {
    const createPostWithDates = (datePublished, dateUpdated) =>
      new Post(data.title, data.html, datePublished, dateUpdated, data.url, data.guid, feed);
    expect(() =>
      createPostWithDates(
        new Date('Thu, 20 Nov 2014 18:59:18 UTC'),
        new Date('Fri, 28 Nov 2014 18:59:18 UTC')
      )
    ).not.toThrow();
    expect(() =>
      createPostWithDates(
        new Date('Thu, 45 Nov 2014 18:59:18 UTC'),
        new Date('Thu, 35 Dec 2014 18:59:18 UTC')
      )
    ).toThrow();
    expect(() =>
      createPostWithDates( // this test fails.
        new Date('Thu, 40 Nov 2014 18:59:18 UTC'),
        new Date('Thu, 20 Nov 2014 18:59:18 UTC')
      )
    ).not.toThrow();
    expect(() =>
      createPostWithDates(
        new Date('Sat, 1 Nov 2014 18:59:18 UTC'),
        new Date('Thu, 35 Dec 2014 18:59:18 UTC')
      )
    ).not.toThrow();
  });
```

Modified ensureDate function
```javascript
function ensureDate(date, fallbackDate) {
  if (
[...]

PR #1350, opened 2020-11-13
Test fails when we have (invalidDate, validDate)

joey2031:
## Type of Change

## Checklist

This code has some errors but I am making a PR for @c3ho to fetch and test to see what's wrong. https://github.com/Seneca-CDOT/telescope/issues/1323#issuecomment-727067374

### DIFF
--- a/src/backend/data/post.js
+++ b/src/backend/data/post.js
@@ -17,15 +17,20 @@ const { indexPost } = require('../utils/indexer');
 function ensureDate(date, fallbackDate) {
   if (
     date &&
-    (Object.prototype.toString.call(date) === '[object String]' ||
-      (Object.prototype.toString.call(date) === '[object Date]' && !Number.isNaN(date)))
+    ((Object.prototype.toString.call(date) === '[object String]' &&
+      !Number.isNaN(Date.parse(date))) ||
+      (Object.prototype.toString.call(date) === '[object Date]' && !Number.isNaN(Date.parse(date))))
   ) {
     return new Date(date);
   }
-  if (Object.prototype.toString.call(fallbackDate) === '[object Date]') {
+  if (
+    (Object.prototype.toString.call(fallbackDate) === '[object String]' &&
+      !Number.isNaN(Date.parse(fallbackDate))) ||
+    (Object.prototype.toString.call(fallbackDate) === '[object Date]' &&
+      !Number.isNaN(Date.parse(fallbackDate)))
+  ) {
     return new Date(fallbackDate);
   }
-
   throw new Error(`post has an invalid date: ${date}'`);
 }
 
--- a/src/backend/data/post.js
+++ b/src/backend/data/post.js
@@ -17,20 +17,15 @@ const { indexPost } = require('../utils/indexer');
 function ensureDate(date, fallbackDate) {
   if (
     date &&
-    ((Object.prototype.toString.call(date) === '[object String]' &&
-      !Number.isNaN(Date.parse(date))) ||
-      (Object.prototype.toString.call(date) === '[object Date]' && !Number.isNaN(Date.parse(date))))
+    (Object.prototype.toString.call(date) === '[object String]' ||
+      (Object.prototype.toString.call(date) === '[object Date]' && !Number.isNaN(date)))
   ) {
     return new Date(date);
   }
-  if (
-    (Object.prototype.toString.call(fallbackDate) === '[object String]' &&
-      !Number.isNaN(Date.parse(fallbackDate))) ||
-    (Object.prototype.toString.call(fallbackDate) === '[object Date]' &&
-      !Number.isNaN(Date.parse(fallbackDate)))
-  ) {
+  if (Object.prototype.toString.call(fallbackDate) === '[object Date]') {
     return new Date(fallbackDate);
   }
+
   throw new Error(`post has an invalid date: ${date}'`);
 }
 
--- a/test/post.test.js
+++ b/test/post.test.js
@@ -92,7 +92,34 @@ describe('Post data class tests', () => {
     expect(() => createPostWithFeed({})).toThrow();
     expect(() => createPostWithFeed(feed)).not.toThrow();
   });
-
+  test('Post constructor should throw if invalid dates is passed', () => {
+    const createPostWithDates = (datePublished, dateUpdated) =>
+      new Post(data.title, data.html, datePublished, dateUpdated, data.url, data.guid, feed);
+    expect(() =>
+      createPostWithDates(
+        new Date('Thu, 20 Nov 2014 18:59:18 UTC'),
+        new Date('Fri, 28 Nov 2014 18:59:18 UTC')
+      )
+    ).not.toThrow();
+    expect(() =>
+      createPostWithDates(
+        new Date('Thu, 45 Nov 2014 18:59:18 UTC'),
+        new Date('Thu, 35 Dec 2014 18:59:18 UTC')
+      )
+    ).toThrow();
+    expect(() =>
+      createPostWithDates(
+        new Date('Thu, 35 Dec 2014 18:59:18 UTC'),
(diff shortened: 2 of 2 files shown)


c3ho:
This is weird, I broke off the ensureDate() function and moved it into a separate file in the project. With the test values you provided, they didn't error and even returned the expected date:

image

joey2031:
@c3ho Some questions:
1. What file in Telescope did you move the function to?
2. I see you used the original ensureDate function and not the modified one I sen should I di that?
3. Does this mean we have a fix to this? Just move the function to another file? Or are w still in need of a fix? 
Thank you so much, have a great day!

c3ho:
I just moved the code out to a file called test.js in the root directory of the project and called node test.js on it. The code is mostly what you have, but I just didn’t use the Number.isNan since that’s going to have the same result as Date.Parse

joey2031:
> I just moved the code out to a file called test.js in the root directory of the project and called node test.js on it. The code is mostly what you have, but I just didn’t use the Number.isNan since that’s going to have the same result as Date.Parse

But we still don't know why is not working when it's in post.js? It needs to be in post.js right?
```
function ensureDate(date, fallbackDate) {
  if (
    date &&
    ((Object.prototype.toString.call(date) === '[object String]' &&
      !Date.parse(date)) ||
      (Object.prototype.toString.call(date) === '[object Date]' && !Date.parse(date)))
  ) {
    return new Date(date);
  }
  if (
    (Object.prototype.toString.call(fallbackDate) === '[object String]' &&
      !Date.parse(fallbackDate)) ||
    (Object.prototype.toString.call(fallbackDate) === '[object Date]' &&
      !Date.parse(fallbackDate))
  ) {
    return new Date(fallbackDate);
  }
  throw new Error(`post has an invalid date: ${date}'`);
}
```

c3ho:
Yep, still needs to be inside Post.js and I have no idea why it’s throwing.

manekenpix:
@joey2031 next time, you don't have to create a PR so other contributors can see your code, just push it to your repo and they can add it as a remote, as explain in week 5's lecture [Understanding Git Remotes](https://www.youtube.com/watch?v=RI1qCMOblA8).

c3ho:
@manekenpix sorry I asked him to 😅

src/backend/data/post.js
```diff
@@ -17,15 +17,20 @@ const { indexPost } = require('../utils/indexer');
 function ensureDate(date, fallbackDate) {
   if (
     date &&
-    (Object.prototype.toString.call(date) === '[object String]' ||
-      (Object.prototype.toString.call(date) === '[object Date]' && !Number.isNaN(date)))
+    ((Object.prototype.toString.call(date) === '[object String]' &&
```
humphd, line 20: None of this needs to be modified, it's only the test that needs to be modified.

test/post.test.js
```diff
@@ -92,7 +92,34 @@ describe('Post data class tests', () => {
     expect(() => createPostWithFeed({})).toThrow();
     expect(() => createPostWithFeed(feed)).not.toThrow();
   });
-
+  test('Post constructor should throw if invalid dates is passed', () => {
+    const createPostWithDates = (datePublished, dateUpdated) =>
+      new Post(data.title, data.html, datePublished, dateUpdated, data.url, data.guid, feed);
+    expect(() =>
+      createPostWithDates(
+        new Date('Thu, 20 Nov 2014 18:59:18 UTC'),
```
humphd, line 100: All of these tests are passing a `Date` instance, whether valid or invalid.  To hit that final `throw` that you're trying to test, you need to pass `null` or `undefined` or something that falls through those ifs.

humphd (changes requested):

humphd:
I think part of your confusion here is that when you use `new Date(...)` it will return a `Date` instance, but it may or may not be valid:

```js
> let badDate = new Date(undefined);
> badDate
Invalid Date
> badDate instanceof Date
true
> isNaN(badDate)
true
```

So you get a `Date` instance, but its unusable.

test/post.test.js
joey2031: So by a valid date you mean is the date object null or not? Not something like Dec. 40 2018?

humphd: Stay focused on what this issue is all about: you are trying to get that "throw" line to be hit in the tests.  You need a test case that causes it to happen.  To do that, you need to stop passing a Date object, or you will always fall into the existing `if` statements.

joey2031: Ohhh I think I'm understanding, when we said throw if the date is invalid I thought u meant to check for something like Sept 40 2020. Turns out the Date() object already does that so all we need to worry about is if a Date object is being passed. ill keep working on this! Thanks for the explanation, it really helped!
image

c3ho: If all you're doing is writing tests you can just checkout master, create a new branch and just pull the updated test file from this branch into your new branch. You'll have to change one of your tests though (the one that has to throw), you can pass it two integers and make it fail that way.

joey2031: Sorry just saw this, I fixed everything and pushed to the issue-1323 branch. Would you like me to make a new branch eg. issue-1323-fixed and make the pr for there? As of now everything is updated on the issue-1323 branch

test/post.test.js
```diff
@@ -92,7 +92,22 @@ describe('Post data class tests', () => {
...
+        new Date('Fri, 28 Nov 2014 18:59:18 UTC')
+      )
+    ).not.toThrow();
+    expect(() => createPostWithDates(new Date('Thu, 20 Nov 2014 18:59:18 UTC'), 23)).not.toThrow();
+    expect(() => createPostWithDates(23, new Date('Thu, 20 Nov 2014 18:59:18 UTC'))).toThrow();
```
c3ho, line 105: I was thinking this line wouldn't throw since we have a fallbackDate

test/post.test.js
c3ho: No worries, that's fine

test/post.test.js
joey2031: @c3ho thats right,... I tried to use .not.toThrow() but it fails.
```
function ensureDate(date, fallbackDate) {
  if (
    date &&
    (Object.prototype.toString.call(date) === '[object String]' ||
      (Object.prototype.toString.call(date) === '[object Date]' && !Number.isNaN(date)))
  ) {
    return new Date(date);
  }
  if (Object.prototype.toString.call(fallbackDate) === '[object Date]') {
    return new Date(fallbackDate);
  }

  throw new Error(`post has an invalid date: ${date}'`);
}
```
image

joey2031: I just noticed if I dod this:       createPostWithDates('zzz', new Date('Thu, 20 Nov 2014 18:59:18 UTC'))
then it throws but when I put a number it fails.


image
image

c3ho: 'zzz' is a string clears the first if statement so it returns `new Date('zzz')`

joey2031: Right but why would it throw with     
expect(() => createPostWithDates('zzz', new Date('Thu, 20 Nov 2014 18:59:18 UTC'))).toThrow();
but not
expect(() => createPostWithDates(23, new Date('Thu, 20 Nov 2014 18:59:18 UTC'))).toThrow();
For the purpose of this issue should I just change it to that and upload the final code? If so do you want me to make a new branch or push to the [issue-1323](https://github.com/joey2031/telescope/tree/issue-1323) branch? Thanks.

c3ho: It should be this branch, change it to following:
`expect(() => createPostWithDates('zzz', new Date('Thu, 20 Nov 2014 18:59:18 UTC'))).not.toThrow();`

joey2031: @c3ho I'm going to make a new branch form master copy and paste my code and make a PR. Thank you so much for all of your help!

PR #1360, opened 2020-11-15
Fixes issue #1323 Added tests to make sure we throw when an invalid date is passed

### LINKED ISSUE #1323: Add test for post.js to call the constructor for a Post()

joey2031:
## Issue This PR Addresses: 
Fixes #1323 

## Type of Change

## Description

I added some tests to post.test.js make sure we throw when an invalid date is passed. **Use to show line 29 needs a test
TelescopeComplete

## Checklist

### DIFF
--- a/test/post.test.js
+++ b/test/post.test.js
@@ -93,7 +93,7 @@ describe('Post data class tests', () => {
     expect(() => createPostWithFeed(feed)).not.toThrow();
   });
 
-  test('Post constructor should throw if invalid dates is passed', () => {
+  test('Post constructor should throw if a string or date is not passed', () => {
     const createPostWithDates = (datePublished, dateUpdated) =>
       new Post(data.title, data.html, datePublished, dateUpdated, data.url, data.guid, feed);
     expect(() =>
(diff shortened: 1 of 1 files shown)


test/post.test.js
```diff
@@ -93,6 +93,27 @@ describe('Post data class tests', () => {
     expect(() => createPostWithFeed(feed)).not.toThrow();
   });
 
+  test('Post constructor should throw if invalid dates is passed', () => {
```
c3ho, line 96: I think we should change this to:
`test('Post constructor should throw if a string or date is not passed`

What do you think? Since it will fail if either of the values are not a string/date

test/post.test.js
```diff
@@ -93,6 +93,27 @@ describe('Post data class tests', () => {
...
+    const createPostWithDates = (datePublished, dateUpdated) =>
+      new Post(data.title, data.html, datePublished, dateUpdated, data.url, data.guid, feed);
+    expect(() =>
+      createPostWithDates(
+        new Date('Thu, 20 Nov 2014 18:59:18 UTC'),
+        new Date('Fri, 28 Nov 2014 18:59:18 UTC')
+      )
+    ).not.toThrow();
+    expect(() =>
+      createPostWithDates(new Date('Thu, 20 Nov 2014 18:59:18 UTC'), '23')
```
c3ho, line 106: should we just use 'string' here instead of '23', I want to avoid the situation where someone might miss the `' '` and think this is actually an integer

test/post.test.js
```diff
@@ -93,6 +93,27 @@ describe('Post data class tests', () => {
...
+    ).not.toThrow();
+    expect(() =>
+      createPostWithDates('23', new Date('Thu, 20 Nov 2014 18:59:18 UTC'))
```
c3ho, line 109: same as above

c3ho (changes requested):

test/post.test.js
joey2031: True! I didn't about that, will fix

test/post.test.js
joey2031: I think that makes sense, ill fix that.

c3ho (approved):

humphd (approved):

Issue #1349, opened 2020-11-13
Add infinite scrolling to search results


humphd:
Once we have #1222 finished we'll be able to page through search results.  When that happens, I want to see something similar to the work in #1332 happen in our timeline of search results.  Ideally we should be able to leverage the same component.

cc @klee214, who worked on the original infinite scrolling and can probably help mentor/review this work.

joelazwar:
@humphd id like to work on this issue!

humphd:
Awesome, thank you @joelazwar.  This will depend on #1222, so we need someone for that too.

joelazwar:
@humphd Hmmm in that case I can work on that issue first. Hope it's fine if I assign that issue for myself

humphd:
@joelazwar sounds good

PR #1384, opened 2020-11-17
Fixes #1222/1349 added paging functionality for search results

### LINKED ISSUE #1222: Allow paging through search results
In #1217 I'm adding some data fetching paging logic using https://swr.vercel.app.  It could also be used for our search front-end (e.g., SearchPage.jsx) to allow us to page through the results we get back from our backend.

To do this, we need to pass the data to our `<Timeline>` component in "pages" like this:

```jsx
<Timeline pages={arrayOfArraysOfPosts} nextPage={functionToGetNextPage />
```
 
The data we get back from the server for a query looks like this:

```js
{
  results: 1436
  values: [
    {
      id: '0f8021a157',
      text: 'full text of post....................'
      score: 4.913031
    },
    ...
  ]
}
```

See #1221, where this might change a bit.  We know the number of `results`, and could request chunks of it using the `from` and `size` options in our elasticsearch query.  We'd have to update our `/query` route to accept params with this info, and then do something like this https://swr.vercel.app/docs/pagination#example-2-cursor-or-offset-based-paginated-api.

joelazwar:
## Issue This PR Addresses
Fixes #1222 and #1349 

## Type of Change

## Description

Search results now infinitely load (through scrolling) in increments of 10 (MAX_POSTS_PER_PAGE)

infiniteresults

## Checklist

### DIFF
--- a/env.example
+++ b/env.example
@@ -68,7 +68,7 @@ MOCK_ELASTIC=
 ELASTIC_URL=http://127.0.0.1
 ELASTIC_PORT=9200
 # Max number of results per query
-ELASTIC_MAX_RESULTS=50
+ELASTIC_MAX_RESULTS=10
 # Delay to check connectivity with Elasticsearch in ms
 ELASTIC_DELAY_MS=10000
 
--- a/env.example
+++ b/env.example
@@ -68,7 +68,7 @@ MOCK_ELASTIC=
 ELASTIC_URL=http://127.0.0.1
 ELASTIC_PORT=9200
 # Max number of results per query
-ELASTIC_MAX_RESULTS_PER_PAGE=10
+ELASTIC_MAX_RESULTS_PER_PAGE=5
 # Delay to check connectivity with Elasticsearch in ms
 ELASTIC_DELAY_MS=10000
 
--- a/env.example
+++ b/env.example
@@ -68,7 +68,7 @@ MOCK_ELASTIC=
 ELASTIC_URL=http://127.0.0.1
 ELASTIC_PORT=9200
 # Max number of results per query
-ELASTIC_MAX_RESULTS=10
+ELASTIC_MAX_RESULTS_PER_PAGE=10
 # Delay to check connectivity with Elasticsearch in ms
 ELASTIC_DELAY_MS=10000
 
--- a/src/backend/utils/indexer.js
+++ b/src/backend/utils/indexer.js
@@ -2,7 +2,7 @@ require('../lib/config');
 
 const { setIntervalAsync, clearIntervalAsync } = require('set-interval-async/dynamic');
 
-const { ELASTIC_MAX_RESULTS } = process.env;
+const { ELASTIC_MAX_RESULTS_PER_PAGE } = process.env;
 const { client } = require('../lib/elastic');
 const { logger } = require('./logger');
 
@@ -83,7 +83,12 @@ const sortFromFilter = (filter) => {
  * @param filter
  * @return all the results matching the passed text
  */
-const search = async (textToSearch, filter = 'post', page = 0, perPage = ELASTIC_MAX_RESULTS) => {
+const search = async (
+  textToSearch,
+  filter = 'post',
+  page = 0,
+  perPage = ELASTIC_MAX_RESULTS_PER_PAGE
+) => {
   const query = {
     query: {
       simple_query_string: {
--- a/src/backend/utils/indexer.js
+++ b/src/backend/utils/indexer.js
@@ -2,7 +2,7 @@ require('../lib/config');
 
 const { setIntervalAsync, clearIntervalAsync } = require('set-interval-async/dynamic');
 
-const { ELASTIC_MAX_RESULTS_PER_PAGE } = process.env;
+const { ELASTIC_MAX_RESULTS_PER_PAGE = 5 } = process.env;
 const { client } = require('../lib/elastic');
 const { logger } = require('./logger');
 
--- a/src/backend/utils/indexer.js
+++ b/src/backend/utils/indexer.js
@@ -81,7 +81,7 @@ const calculateFrom = (page, perPage) => {
   const ES_MAX = 10000; // 10K is the upper limit of what ES will return without issue for searches
   const wanted = page * perPage;
   // Don't exceed 10K, and if we will, return an offset under it by one page size
-  return wanted + perPage < ES_MAX ? wanted : ES_MAX - perPage;
+  return wanted + perPage <= ES_MAX ? wanted : ES_MAX - perPage;
 };
 
 /**
--- a/src/backend/utils/indexer.js
+++ b/src/backend/utils/indexer.js
@@ -2,7 +2,7 @@ require('../lib/config');
 
 const { setIntervalAsync, clearIntervalAsync } = require('set-interval-async/dynamic');
 
-const { ELASTIC_MAX_RESULTS } = process.env;
+const { MAX_POSTS_PER_PAGE } = process.env;
 const { client } = require('../lib/elastic');
 const { logger } = require('./logger');
 
@@ -83,7 +83,7 @@ const sortFromFilter = (filter) => {
  * @param filter
  * @return all the results matching the passed text
  */
-const search = async (textToSearch, filter = 'post') => {
+const search = async (textToSearch, filter = 'post', page, perPage = MAX_POSTS_PER_PAGE) => {
   const query = {
     query: {
       simple_query_string: {
@@ -98,9 +98,9 @@ const search = async (textToSearch, filter = 'post') => {
   const {
     body: { hits },
   } = await client.search({
-    from: 0,
+    from: page * perPage,
+    size: perPage,
     _source: ['id'],
-    size: ELASTIC_MAX_RESULTS || 100,
     index,
     type,
     body: query,
--- a/src/backend/utils/indexer.js
+++ b/src/backend/utils/indexer.js
@@ -2,7 +2,7 @@ require('../lib/config');
 
 const { setIntervalAsync, clearIntervalAsync } = require('set-interval-async/dynamic');
 
-const { MAX_POSTS_PER_PAGE } = process.env;
+const { ELASTIC_MAX_RESULTS } = process.env;
 const { client } = require('../lib/elastic');
 const { logger } = require('./logger');
 
@@ -83,7 +83,7 @@ const sortFromFilter = (filter) => {
  * @param filter
  * @return all the results matching the passed text
  */
-const search = async (textToSearch, filter = 'post', page = 0, perPage = MAX_POSTS_PER_PAGE) => {
+const search = async (textToSearch, filter = 'post', page = 0, perPage = ELASTIC_MAX_RESULTS) => {
   const query = {
     query: {
       simple_query_string: {
--- a/src/backend/utils/indexer.js
+++ b/src/backend/utils/indexer.js
@@ -83,7 +83,7 @@ const sortFromFilter = (filter) => {
  * @param filter
  * @return all the results matching the passed text
  */
-const search = async (textToSearch, filter = 'post', page, perPage = MAX_POSTS_PER_PAGE) => {
+const search = async (textToSearch, filter = 'post', page = 0, perPage = MAX_POSTS_PER_PAGE) => {
   const query = {
     query: {
       simple_query_string: {
--- a/src/backend/utils/indexer.js
+++ b/src/backend/utils/indexer.js
@@ -77,6 +77,13 @@ const sortFromFilter = (filter) => {
   }
 };
 
+const calculateFrom = (page, perPage) => {
+  const ES_MAX = 10000; // 10K is the upper limit of what ES will return without issue for searches
+  const wanted = page * perPage;
+  // Don't exceed 10K, and if we will, return an offset under it by one page size
+  return wanted + perPage < ES_MAX ? wanted : ES_MAX - perPage;
+};
+
 /**
  * Searches text in elasticsearch
  * @param textToSearch
@@ -103,7 +110,7 @@ const search = async (
   const {
     body: { hits },
   } = await client.search({
-    from: page * perPage,
+    from: calculateFrom(page, perPage),
     size: perPage,
     _source: ['id'],
     index,
--- a/src/backend/web/routes/query.js
+++ b/src/backend/web/routes/query.js
@@ -8,8 +8,8 @@ const router = express.Router();
 
 router.get('/', validateQuery(), async (req, res) => {
   try {
-    const { text, filter } = req.query;
-    res.send(await search(text, filter));
+    const { text, filter, page, perPage } = req.query;
+    res.send(await search(text, filter, page, perPage));
   } catch (error) {
     res.send(`There was an error while executing your query: ${error}`);
     logger.error({ error }, 'Something went wrong with search indexing');
--- a/src/backend/web/validation.js
+++ b/src/backend/web/validation.js
@@ -73,6 +73,13 @@ const queryValidationRules = [
     .isIn(['post', 'author'])
     .withMessage('invalid filter value')
     .bail(),
+
+  query('perPage', 'per_page needs to be empty or a integer').custom(
+    (value) => !value || Number.isInteger(+value)
+  ),
+  query('page', 'page needs to be empty or an integer').custom(
+    (value) => !value || Number.isInteger(+value)
+  ),
 ];
 
 const postsIdParamValidationRules = [
--- a/src/backend/web/validation.js
+++ b/src/backend/web/validation.js
@@ -74,12 +74,17 @@ const queryValidationRules = [
     .withMessage('invalid filter value')
     .bail(),
 
-  query('perPage', 'per_page needs to be empty or an integer').custom(
-    (value) => !value || Number.isInteger(+value)
-  ),
-  query('page', 'page needs to be empty or an integer').custom(
-    (value) => !value || Number.isInteger(+value)
-  ),
+  check('perPage')
+    .optional({ checkFalsy: true })
+    .isInt({ min: 1, max: 10 })
+    .withMessage('perPage should be empty or a number between 1 to 10')
+    .bail(),
+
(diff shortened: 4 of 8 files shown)


src/backend/utils/indexer.js
```diff
@@ -2,7 +2,7 @@ require('../lib/config');
 
 const { setIntervalAsync, clearIntervalAsync } = require('set-interval-async/dynamic');
 
-const { ELASTIC_MAX_RESULTS } = process.env;
+const { MAX_POSTS_PER_PAGE } = process.env;
```
humphd, line 5: This can't be changed without updating our `env` variables in the different sample files we have, e.g., https://github.com/Seneca-CDOT/telescope/blob/master/env.example#L71

src/backend/utils/indexer.js
```diff
@@ -98,9 +98,9 @@ const search = async (textToSearch, filter = 'post') => {
   const {
     body: { hits },
   } = await client.search({
-    from: 0,
+    from: page * perPage,
```
humphd, line 101: What if `page` is `undefined`?  We better deal with the case that it's not passed to this function and use `0`.  Also, what if you go past the end of the result list?  We need to add a check to make sure we don't request a page beyond the end of the results.

src/backend/web/routes/query.js
```diff
@@ -8,8 +8,8 @@ const router = express.Router();
 
 router.get('/', validateQuery(), async (req, res) => {
   try {
-    const { text, filter } = req.query;
-    res.send(await search(text, filter));
+    const { text, filter, page, perPage } = req.query;
```
humphd, line 11: These need to get added to our `validateQuery` validation logic

src/frontend/src/components/SearchPage/SearchResults.jsx
```diff
@@ -23,15 +23,18 @@ const useStyles = makeStyles(() => ({
...
 
   // We only bother doing the request if we have something to search for.
   const shouldFetch = () => text.length > 0;
-  const { data, loading, error } = useSWR(shouldFetch() ? url : null, async (u) => {
-    const res = await fetch(u);
-    const results = await res.json();
-    return results.values;
-  });
+  const { data, size, setSize, loading, error } = useSWRInfinite(
+    (index) => (shouldFetch() ? url + index : null),
```
humphd, line 31: Your URL will need to include size info, so we ask for the right offset.  We can't send `&page=` (i.e., with no value) all the time.

humphd (changes requested):

src/backend/utils/indexer.js
joelazwar: i only set this as a default if the perPage query wasnt given/undefined. sorry if misunderstood the problem. whats wrong with using the process env constant?

src/backend/utils/indexer.js
joelazwar: set page to default to 0 if not passed the arg. I wrote the logic for checking if in the last page of the result in searchResults.jsx hope you can take a look since it can probably be improved 😂 its the only thing i could come up with.

src/backend/web/routes/query.js
joelazwar: added, let me know if I did it right.

src/frontend/src/components/SearchPage/SearchResults.jsx
joelazwar: by size info/offset do you mean the number of results per fetch? also im under the impression that page= will have a value because of the index argument in the getKey function for swr infinite. sorry I'm still learning how to use the swr react hook so i might be a little confused, what would you suggest I change here?

src/backend/utils/indexer.js
humphd: The problem is that you've left the `ELASTIC_MAX_RESULTS` orphaned in the env.  If you don't use it anymore, it needs to get deleted.  I think re-using this variable from the env is fine, but you might also want to update the docs in the env files to indicate that this will apply to posts as well as search results.

src/backend/web/validation.js
```diff
@@ -73,6 +73,13 @@ const queryValidationRules = [
     .isIn(['post', 'author'])
     .withMessage('invalid filter value')
     .bail(),
+
+  query('perPage', 'per_page needs to be empty or a integer').custom(
```
humphd, line 77: `an` vs. `a` integer.

src/frontend/src/components/SearchPage/SearchResults.jsx
```diff
@@ -54,7 +57,18 @@ const SearchResults = ({ text, filter }) => {
 
   return (
     <Container className={classes.searchResults}>
-      {data && data.length ? <Timeline pages={[data]} /> : <h1>No results search</h1>}
+      {data && data.length ? (
+        <Timeline
+          pages={data.map((d) => d.values)}
```
humphd, line 62: Can we do this map when we get the results vs. here?  That is, let's have `data` be in the format we expect so it doesn't have to be altered.

humphd: Also, `pages` expects an array of arrays.

src/frontend/src/components/SearchPage/SearchResults.jsx
```diff
@@ -54,7 +57,18 @@ const SearchResults = ({ text, filter }) => {
...
+          nextPage={
+            data[0].values.length * data.length < data[0].results
```
humphd, line 64: This is ugly.  I feel like the logic here to advance pages could probably be moved into the backend vs. doing it here.  Let's just try to get the next page, and ES can return the one it wants.

src/frontend/src/components/SearchPage/SearchResults.jsx
humphd: I see, you're actually doing it correctly, but it's not very easy to read.  Let's do this via a function call:

```js
const prepareUrl = (index) => ${telescopeUrl}/query?text=${encodeURIComponent(text)}&filter=${filter}&page=${index + 1}`;
...
const { data, size, setSize, loading, error } = useSWRInfinite(
    (index) => (shouldFetch() ? prepareUrl(index) : null),
```

humphd (changes requested):

src/frontend/src/components/SearchPage/SearchResults.jsx
joelazwar: yeap, I had data take both the results, and values property from the fetch to try and write logic for last page detecting but I'll change it back since we'll move that logic to the backend

src/frontend/src/components/SearchPage/SearchResults.jsx
joelazwar: Right, I thought we could add logic to determine whether or not to keep fetching from the front end. But I'll write something in the backend.

joelazwar: sorry just to clarify, are you saying to omit the functionality for now? or do you still want me to write a logic in the backend for it

src/backend/utils/indexer.js
joelazwar: ahhh okay. Maybe we can keep `ELASTIC_MAX_RESULTS` but change its default value in the env for this feature?

src/backend/utils/indexer.js
humphd: This stuff gets complicated, and I don't have a ton of experience tuning ES, so probably we should do the simple/naive thing for now, and fix it later in follow-ups.  There are extensive docs on this in https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html.

Based on that, what about something like:

```
from: calculateFrom(page, perPage),
size: perPage
```

where `calculateFrom` is something like:

```js
const calculateFrom = (page, perPage) => {
  const ES_MAX = 10000; // 10K is the upper limit of what ES will return without issue for searches
  const wanted = page * perPage;
  // Don't exceed 10K, and if we will, return an offset under it by one page size
  return wanted + perPage < ES_MAX ? wanted : ES_MAX - perPage
};
```

This code is 100% untested, but you can use it as a starting point :)

src/backend/web/validation.js
```diff
@@ -73,6 +73,13 @@ const queryValidationRules = [
     .isIn(['post', 'author'])
     .withMessage('invalid filter value')
     .bail(),
+
+  query('perPage', 'per_page needs to be empty or an integer').custom(
```
humphd, line 77: Let's also set a max on this of 10. So it can be nothing, or a number between 1 and 10.

src/backend/web/validation.js
```diff
@@ -73,6 +73,13 @@ const queryValidationRules = [
...
+    (value) => !value || Number.isInteger(+value)
+  ),
+  query('page', 'page needs to be empty or an integer').custom(
```
humphd, line 80: This can be empty or a number between 1 and 1000 (i.e., 10 * 1000 = 10,000).

src/backend/web/validation.js
```diff
@@ -73,6 +73,13 @@ const queryValidationRules = [
...
+    (value) => !value || Number.isInteger(+value)
+  ),
```
humphd, line 82: Also, you need to add more tests for the validation on these routes to make sure that good values for `page` and `per_page` give 200s, and values outside the valid range return 400s.

humphd (changes requested):
This is getting closer and closer!

src/backend/web/validation.js
joelazwar: page technically starts from 0, but that could be changed if you want

src/backend/web/validation.js
joelazwar: They should give the right response codes now according to the values

env.example
```diff
@@ -68,7 +68,7 @@ MOCK_ELASTIC=
 ELASTIC_URL=http://127.0.0.1
 ELASTIC_PORT=9200
 # Max number of results per query
-ELASTIC_MAX_RESULTS=50
+ELASTIC_MAX_RESULTS_PER_PAGE=10
```
humphd, line 71: This can drop to 5, since the infinite scroll will mean you don't notice not having tons of results.

src/backend/utils/indexer.js
humphd: This wasn't fixed.  Please don't resolve if you don't actually fix a review comment.

src/frontend/src/components/SearchPage/SearchResults.jsx
```diff
@@ -54,7 +58,11 @@ const SearchResults = ({ text, filter }) => {
 
   return (
     <Container className={classes.searchResults}>
-      {data && data.length ? <Timeline pages={[data]} /> : <h1>No results search</h1>}
+      {data && data.length ? (
+        <Timeline pages={data} nextPage={() => setSize(size + 1)} />
```
humphd, line 62: I guess if ES is 0-based, we don't need the `+ 1`.

humphd (changes requested):

src/backend/utils/indexer.js
joelazwar: sorry about that, the comment the previous comment didnt load until i refreshed the page

src/frontend/src/components/SearchPage/SearchResults.jsx
joelazwar: oh I thought the size + 1 was to increment the index no?

joelazwar: btw I changed the page query range limit from 0-999 because of this.

src/backend/utils/indexer.js
joelazwar: I implemented the code that you wrote and tested it, it works correctly. Let me know if this is resolved for now.

src/backend/utils/indexer.js
```diff
@@ -77,13 +77,25 @@ const sortFromFilter = (filter) => {
   }
 };
 
+const calculateFrom = (page, perPage) => {
+  const ES_MAX = 10000; // 10K is the upper limit of what ES will return without issue for searches
+  const wanted = page * perPage;
+  // Don't exceed 10K, and if we will, return an offset under it by one page size
+  return wanted + perPage < ES_MAX ? wanted : ES_MAX - perPage;
```
humphd, line 84: I guess this needs to be `<= ES_MAX` to work for the `999` and `10` case.

src/backend/web/validation.js
```diff
@@ -73,6 +73,18 @@ const queryValidationRules = [
     .isIn(['post', 'author'])
     .withMessage('invalid filter value')
     .bail(),
+
+  check('perPage')
```
humphd, line 77: I don't see the validation tests for these.  Add test cases for:

- perPage is empty `&perPage=`
- perPage is some garbage string `&perPage=invalid`
- perPage is a number smaller than 1-10 `&perPage=0`
- perPage is a number larger than 1-10 `&perPage=11`
- perPage is valid `&perPage=3`

same idea for `page`

humphd (changes requested):

src/backend/web/validation.js
joelazwar: @humphd sorry bit of an express validation noob here, I though the `check('perPage')` and `check('page')` validation chains do that? most of the test cases give out the correct responses, page or perPage out of range gives 400, querying anything but an int will give a 400 as well. Only the first one doesn't seem to work

humphd: This is for our web route tests, to make sure that our web server rejects any params that aren't legal.  See https://github.com/Seneca-CDOT/telescope/blob/master/test/query.test.js which is where you need to add tests, since you're changing how our web routes work, you need new tests so we don't break it.

joelazwar: I've coded the validation tests but just one question. What should the response be for empty params e.g. `&perPage=` and `&page=` should it be valid 200 or invalid 400. Right now the query accepts the value as '' (empty string?) when passed like that so i need to code validation logic for that as well.

humphd: I think passing `&perPage=` and `&page=` is probably a user bug (i.e., 400) and we should error.  If you set these params in the URL, we'll expect you to give proper values.  If you leave them off, we'll do sane defaults.

joelazwar: Ok, I'm trying to get it to reject `&perPage=` and `&page=` since it express validation still accepts those values right now, will push a commit soon

test/query.test.js
```diff
@@ -53,4 +53,66 @@ describe('Testing query route', () => {
     const res = await request(app).get(`/query?text=&filter=`);
     expect(res.status).toBe(400);
   });
+
+  test('Testing page with empty param value', async () => {
```
humphd, line 57: Change this title to better reflect what the test is doing, namely, testing an empty `page` value.

test/query.test.js
```diff
@@ -53,4 +53,66 @@ describe('Testing query route', () => {
...
+    expect(res.status).toBe(400);
+  });
+
+  test('Testing with valid page in range', async () => {
+    const searchTerm = encodeURIComponent('I Love Telescope');
+    const res = await request(app).get(`/query?text=${searchTerm}&filter=post&page=1`);
+    expect(res.status).toBe(200);
+  });
+
+  test('Testing page with empty param value', async () => {
```
humphd, line 87: Fix this title too.

humphd (changes requested):
A couple of minor things.

test/query.test.js
joelazwar: whoops overlooked this one, on it

src/frontend/src/components/SearchPage/SearchResults.jsx
```diff
@@ -54,7 +58,11 @@ const SearchResults = ({ text, filter }) => {
...
+      ) : (
+        <h1>No results search</h1>
```
humphd, line 64: "No search results" would be more accurate, let's fix.

humphd (changes requested):
Last one...

humphd (approved):
Heroic!  This is such a great fix, thank you for sticking with it right to the end.

HyperTHD (dismissed):
Looks amazing. You got the ok from me!

src/backend/utils/indexer.js
```diff
@@ -2,7 +2,7 @@ require('../lib/config');
 
 const { setIntervalAsync, clearIntervalAsync } = require('set-interval-async/dynamic');
 
-const { ELASTIC_MAX_RESULTS } = process.env;
+const { ELASTIC_MAX_RESULTS_PER_PAGE } = process.env;
```
humphd, line 5: We need to set a default here so we don't blow up if the env is missing this.  I'd suggest:

```js
const { ELASTIC_MAX_RESULTS_PER_PAGE = 10 } = process.env;
```

humphd (changes requested):
That victory was too good to last.

src/backend/utils/indexer.js
joelazwar: i think we decided on 5 yesterday, so I'll set it to 5?

humphd: Correct, I'm wrong with 10.

HyperTHD (approved):
ES break works on my end with this functionality so that's amazing. Looking through it, everything appears to be in order. If @humphd is satisfied, I approve of this.

env.example
```diff
@@ -68,7 +68,7 @@ MOCK_ELASTIC=
 ELASTIC_URL=http://127.0.0.1
 ELASTIC_PORT=9200
 # Max number of results per query
-ELASTIC_MAX_RESULTS=50
+ELASTIC_MAX_RESULTS_PER_PAGE=5
```
humphd, line 71: Let's make this same change to https://github.com/Seneca-CDOT/telescope/blob/master/env.production and https://github.com/Seneca-CDOT/telescope/blob/master/env.staging.

humphd (changes requested):
I forgot about the changes to the prod/staging envs.

env.example
joelazwar: done

humphd (approved):
💯

Issue #999, opened 2020-04-14
On initial load do not show "load more button"


Grommers00:
**What happened**:
Currently, before it loads posts, the button says "No more posts add yours!"

**What should have happened**:
It shouldn't show anything till after the first posts load. 

**How to reproduce it (as precise as possible)**:
Before the first posts, load you can see it.

**Anything else we need to know?**:
Nope!

**Environment**:

- OS:
- Browser:

humphd:
I think it's a matter of dealing with the initial load case vs. an empty set of posts vs. a set of posts.  I'd set the state to `null` on load and then you can treat `[]` and `[post, post, ...]` differently.

PR #1391, opened 2020-11-18
Fixes #1227: This PR fixes the ESLint spamming warnings for methods in GitHub

### LINKED ISSUE #1227: Fix eslint warnings spamming our PRs
Our lovely GitHub Actions code is adding inline eslint warnings to our PRs:

Screen Shot 2020-10-20 at 9 55 10 PM

Let's fix the warnings so we don't have these.

ceterisp:
## Issue This PR Addresses

Fixes #1227

## Type of Change

### Problem
ESlint warning showing up when PR is made.

### Solution
Added an eslint-disable comment that disables the async/await preference and continues to use callback

## Checklist


src/backend/web/app.js
```diff
@@ -76,6 +76,7 @@ app.use('/', router);
  * Error Handler, Pass to front-end
  */
 /* eslint-disable no-unused-vars */
+// eslint-disable-next-line promise/prefer-await-to-callbacks
```
manekenpix, line 79: I think it'd be better to actually fix the problem instead of just hiding the warning. We should definitely use async/await instead of callbacks.

manekenpix (changes requested):

manekenpix:
Please add a proper title "Fixes #999: The problem this PR is fixing", and complete the "Issue this PR addresses" section as indicated in the template.

c3ho:
I thought the direction for this PR was discussed in slack already. This PR is doing the exact opposite of what I saw.

src/backend/web/app.js
abuZayed15: I thought we can't use async and await in `app.use` so I suppressed the warning. I'll research a bit more to figure out how to solve this. Thanks for your comment

abuZayed15:
@manekenpix I changed the title so that it sounds less vague. Thanks
@c3ho So, I thought the way `app.use` method is created, it won't work with async/await. Callbacks are what I have used in the past for `app.use` so I suppressed the warning. I created this PR as a way to get feedback. Now that I have, I can go back and look to incorporate async/await with `app.use`. Thanks

humphd:
This is actually a bug we should probably fix.  I don't think that express callbacks work cleanly with `async`, since you need to explicitly call `next` in the `catch` for the error case.  There are some wrappers that help make async-callbacks in the express middleware chain.

We should file a bug to investigate this, see http://expressjs.com/en/advanced/best-practice-performance.html#use-promises and a million other blog posts.  @abuZayed15 do you want to file that?

abuZayed15:
@humphd Yes, I will file this bug. I'm still unclear of what the expected solution is, but I am willing to take some time and explore possible solutions

abuZayed15:
So, during my research I only found examples of async and await being used in `app.get` middlewares. After diving deeper into the issue, I found out the issue is with ESlist asking for change callbacks to async/await is because of an issue with eslint. Apparently, they don't recognize `err` as a variable until I replaced it with `_err`. This caused the warning to go away and the jest tests passed. For this particular problem, could we file an issue to eslint?

About the wrapper, unfortunately, I wasn't sure what I was wrapping around the middleware. The example of `const wrap` used worked for `app.get` but didn't seem to work with `app.use`. Again, it could be because I wasn't sure what it is doing. I might not have resolved the core problem (the bug) of this issue but the eslint warnings are gone now. It's a very minor change but please let me know if this is what you are looking for. Thanks!

birtony:
@ceterisp thank you for looking into this issue. This PR got a bit outdated, so we have revisited [the issue](https://github.com/Seneca-CDOT/telescope/issues/1227) and fixed it in https://github.com/Seneca-CDOT/telescope/pull/1573.

Issue #1356, opened 2020-11-14
Add Tests for `useTelescopeContributor` hook


TDDR:
This issue is part of #1232 . 

There is a new hook that has been added that interacts with GitHub's API. We need a suite of tests to make sure that:
- A single, random contributor is returned from the hook
- An error object that is either `undefined` or an actual `Error` object
- Write tests for failures of the API.

jiyoungsin:
Hey, I made the tests you requested. However, I am having a difficult time understanding why Jest cannot parse the import file. I have created a branch on my [local repository](https://github.com/jiyoungsin/telescope/tree/issue-1356) with a possible solution.
Just a reminder that this repository is failing the tests "npm test use-telescope-contributor". If you have any idea on why Its having a hard time understanding our code please let me know 👍

humphd:
The code you're trying to test doesn't live in the backend, it's in the frontend, which has separate tests (cc @tonyvugithub).  See https://github.com/Seneca-CDOT/telescope/blob/master/src/frontend/test/Spinner.test.js for example.

You'll need to write a hook-based test that mocks a network response from the GitHub API returning user info in the form this hook expects to receive it.

https://blog.logrocket.com/a-quick-guide-to-testing-react-hooks-fa584c415407/ and other resources on testing hooks might help.

jiyoungsin:
Thank you for the insight about frontend and backend having separate tests, I was not aware of this. 

I have moved all the code I wrote over to the front end tester and trying to make it work. I am trying to use 
```js
  test('Write tests for failures of the API', () => {
    expect(() => useTelescopeContributor()).toThrow(Error);
  });
  test('A single, random contributor is returned from the hook', () => {
    const { result } = renderHook(() => useTelescopeContributor());
    expect(typeof result.contributor).toBe('object');
  });
```
To check if they are of type Object, however, no matter what I try I always get an undefined with the error message below. Anyone have any ideas?
```js
 expect(received).toBe(expected) // Object.is equality

    Expected: "object"
    Received: "undefined"

       5 |   test('A single, random contributor is returned from the hook', () => {
       6 |     const { result } = renderHook(() => useTelescopeContributor());
    >  7 |     expect(typeof result.contributor).toBe('object');
         |                                       ^
       8 |   });
       9 |   // test('An error object that is either undefined or an actual Error object', () => {
      10 |   //   const { result } = renderHook(() => useTelescopeContributor());

[...]

PR #1431, opened 2020-11-20
Alternative fix to useTelescopeContributor

jiyoungsin:
## Issue This PR Addresses

2. If your PR addresses an issue but does not close it
    USAGE: #1356  I would like some feedback on this PR. Initially, I tried to create some tests for the useTelescopeContributor function. After some long consideration, I thought this PR would be a good way to get rid of the test completely. 

## Type of Change

## Description

The useTelescopeContributor function is being invoked without an await or promise resolution. Additionally the component cannot be awaited as it is nested into a useEffect. The changes that have been made are, allowing the API request to load first, after the component will be rendered. What the current code does, on line 41, it reads property avatar_url of null which throws an error. Additionally, I don't think it needs a test as the code seems straight forward.

## Checklist


humphd (changes requested):
No, this is not right.  This hook is working perfectly.  What we need is a test that allows the hook to run in a testing environment, and can simulate the side effect we need here.  We don't need to rewrite this code, or embed await.

jiyoungsin:
Thank you for the response, I will close this PR now.

Issue #1432, opened 2020-11-20
Mouse hover on search textField makes border disappeared


PedroFonsecaDEV:
**What happened**:
@VietnameZe pointed out this issue:
Mouse hover on search textField makes border disappeared
**What should have happened**:
The border should not change.
**How to reproduce it (as precise as possible)**:
Mouse hover on search input.
**Anything else we need to know?**:

**Environment**:

- OS:
- Browser:

PR #1433, opened 2020-11-20
Fix Issue1432: Mouse hover on search textField makes border disappeared

PedroFonsecaDEV:
## Issue This PR Addresses

#1432 - Mouse hover on search textField makes border disappeared
## Type of Change

## Description
Before we had the search text field's border disappearing when the mouse hovers on it, and the size of the inputs in the search bar didn't match. Follow screenshots: 

Before:

After:

## Checklist

### DIFF
--- a/src/frontend/gatsby/src/components/SearchInput/SearchInput.jsx
+++ b/src/frontend/gatsby/src/components/SearchInput/SearchInput.jsx
@@ -8,7 +8,7 @@ const useStyles = makeStyles((theme) => ({
   input: {
     fontSize: '1.6rem',
     '&:hover': {
-      border: '2px solid',
+      border: '1px solid',
       borderColor: theme.palette.primary.main,
     },
     '&:focus': {
--- a/src/frontend/gatsby/src/components/SearchInput/SearchInput.jsx
+++ b/src/frontend/gatsby/src/components/SearchInput/SearchInput.jsx
@@ -9,7 +9,7 @@ const useStyles = makeStyles((theme) => ({
     fontSize: '1.6rem',
     '&:hover': {
       border: '2px solid',
-      borderColor: theme.palette.background.default,
+      borderColor: theme.palette.primary.main,
     },
     '&:focus': {
       border: '2px solid',
--- a/src/frontend/gatsby/src/components/SearchInput/AuthorSearchInput.jsx
+++ b/src/frontend/gatsby/src/components/SearchInput/AuthorSearchInput.jsx
@@ -1,26 +1,25 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { FormControl, TextField } from '@material-ui/core';
 
 function AuthorSearchInput({ text, onChange, classes }) {
   return (
     <>
-      <input
-        className={classes.input}
-        list="search-suggestions"
-        placeholder="How to Get Started in Open Source"
-        inputProps={{ 'aria-label': 'search telescope' }}
-        value={text}
-        onChange={onChange}
-      />
-      <datalist id="search-suggestions" />
+      <FormControl fullWidth={true}>
+        <TextField
+          className={classes.input}
+          variant="outlined"
+          placeholder="How to Get Started in Open Source"
+          value={text}
+          onChange={onChange}
+        ></TextField>
+      </FormControl>
     </>
   );
 }
-
 AuthorSearchInput.propTypes = {
   text: PropTypes.string,
   onChange: PropTypes.func,
   classes: PropTypes.object,
 };
-
 export default AuthorSearchInput;
--- a/src/frontend/gatsby/src/components/SearchInput/AuthorSearchInput.jsx
+++ b/src/frontend/gatsby/src/components/SearchInput/AuthorSearchInput.jsx
@@ -1,19 +1,19 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
 import React from 'react';
 import PropTypes from 'prop-types';
-import { FormControl, TextField } from '@material-ui/core';
 
 function AuthorSearchInput({ text, onChange, classes }) {
   return (
     <>
-      <FormControl fullWidth={true}>
-        <TextField
-          className={classes.input}
-          variant="outlined"
-          placeholder="How to Get Started in Open Source"
-          value={text}
-          onChange={onChange}
-        ></TextField>
-      </FormControl>
+      <input
+        className={classes.input}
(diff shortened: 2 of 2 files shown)


src/frontend/src/components/SearchInput/AuthorSearchInput.jsx
```diff
@@ -1,17 +1,19 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { FormControl, TextField } from '@material-ui/core';
 
 function AuthorSearchInput({ text, onChange, classes }) {
   return (
     <>
-      <input
-        className={classes.input}
-        list="search-suggestions"
```
MSTEWARDSON, line 9: Removing list when you change it over to a TextField here would effect how the backend support for author autocomplete is handled later on.

MSTEWARDSON (commented):
The style changes are good, my only concern is that the reason AuthorSearchInput's return was a input type was so it could have a data-list attached to it. This would be needed later for a future PR to add in the backend support for author autocomplete. Your proposed changes to a TextField would remove this purpose, if you can find another way to attach the data-list to it while its a TextField then that would be fine.

PedroFonsecaDEV:
Sure.  That's why is important to ask a review from someone who wrote the last change in the code. So there will be a route to give us suggestions? Nice. 
I saw the list hang in there, but I didn't understand it.
We can handle it. I will work on this latter today. @MSTEWARDSON Thanks for the heads up.

PedroFonsecaDEV:
@MSTEWARDSON There is an Autocomplete component [here](https://material-ui.com/api/autocomplete/#autocomplete-api).
But it may require refactoring the SearchPage component's tree.
For now, let's fix this bug.

MSTEWARDSON:
@PedroFonsecaDEV Originally I was planning to use that component but I couldn't get it working the way I wanted it too, so it was switch to an input type. 

For the bug itself, your looking to make the border around the search-bar hide on hover right? In SearchInput the style input class had a on hover I added. I see you have commented it out for now. Couldn't we just add the '&hover' styles back?

PedroFonsecaDEV:
@MSTEWARDSON If you check the dev version, you will notice the two inputs don't have the same size, border, etc.
Today we have a problem that needs a fix. We can't compromise our front-end waiting for something that is still an idea.
You need to create a branch to start working on the author autocomplete if you want help push your branch to your GitHub, and we can have a look together.
As I said, I think that we need to improve the search bar components' tree.
But for today, let's have this bug fixed.
Nothing here is static; we will always be changing a lot of things.

MSTEWARDSON:
@PedroFonsecaDEV Ok so for now why don't we take in your changes here to fix the border disappearing/ converting the Author and post component types from input to TextField. I agree we shouldn't compromise the front-end for a future PR, that PR can be adjusted to work with the TextFields.

src/frontend/src/components/SearchInput/PostSearchInput.jsx
```diff
@@ -1,15 +1,20 @@
...
-    <input
-      className={classes.input}
-      placeholder="How to Get Started in Open Source"
-      inputProps={{ 'aria-label': 'search telescope' }}
-      value={text}
-      onChange={onChange}
-    />
+    <>
+      <FormControl fullWidth={true}>
+        <TextField
```
PedroFonsecaDEV, line 9: @MSTEWARDSON We are already using TextField.

src/frontend/src/components/SearchInput/AuthorSearchInput.jsx
```diff
@@ -1,18 +1,19 @@
...
-        placeholder="How to Get Started in Open Source"
-        inputProps={{ 'aria-label': 'search telescope' }}
-        value={text}
-        onChange={onChange}
-      />
-      <datalist id="search-suggestions"></datalist>
+      <FormControl fullWidth={true}>
+        <TextField
```
PedroFonsecaDEV, line 9: @MSTEWARDSON We are already using TextField.

MSTEWARDSON (dismissed):
I understand the changes needed and accept them.

src/frontend/src/components/SearchInput/AuthorSearchInput.jsx
```diff
@@ -1,18 +1,19 @@
...
-      <input
-        className={classes.input}
-        list="search-suggestions"
-        placeholder="How to Get Started in Open Source"
-        inputProps={{ 'aria-label': 'search telescope' }}
-        value={text}
-        onChange={onChange}
-      />
-      <datalist id="search-suggestions"></datalist>
+      <FormControl fullWidth={true}>
```
humphd, line 8: I don't understand why you're removing the `<datalist>`.  Can you not fix the border without adjusting the components?

humphd (changes requested):

src/frontend/src/components/SearchInput/AuthorSearchInput.jsx
PedroFonsecaDEV: The hover on the border can be "fixed," but the size of the inputs(Author/Post and search field) will be different because of the "native" CSS from Material-UI. 

Material-UI has an autocomplete component to our author autocomplete feature; To it works, we will need to refactor some components. But let's do it when implementing the author autocomplete.

PedroFonsecaDEV:
I blocked it until I rebase and so on.

PedroFonsecaDEV:
Now, this PR fixes the border on hover without changing the tags. 
I believe that my old proposal of using material UI components is the best approach for now and for author autocomplete. 
But this PR solves the issue as requested.

cc @humphd @MSTEWARDSON

src/frontend/gatsby/src/components/SearchInput/SearchInput.jsx
```diff
@@ -9,7 +9,7 @@ const useStyles = makeStyles((theme) => ({
     fontSize: '1.6rem',
     '&:hover': {
       border: '2px solid',
```
humphd, line 11: Can we fix this so it uses the same width in the hover and non-hover cases, so it doesn't push the contents if the input when you hover?  That extra 1px moves everything inward vs. just changing the colour.

humphd (changes requested):
While you're touching this file, can you do one small fix?

PedroFonsecaDEV:
@MSTEWARDSON @MLJBrackett @humphd @tianlangwu

humphd (approved):

tianlangwu (approved):

Issue #1442, opened 2020-11-22
Error: Unable to parse elastic URL "http://elasticsearch:9200" and/or PORT "9200"


humphd:
I'm not able to run `docker-compose up --build` locally.  When I do, I get this error in Telescope:

```
telescope_1      | > node src/backend
telescope_1      |
telescope_1      | {"level":50,"time":1606086074314,"pid":17,"hostname":"a8951a77ddcd","error":{"input":"http://elasticsearch:9200:9200","code":"ERR_INVALID_URL"},"msg":"Unable to parse elastic URL \"http://elasticsearch:9200\" and/or PORT \"9200\""}
telescope_1      | /telescope/src/backend/lib/elastic.js:40
telescope_1      |     throw new Error(message);
telescope_1      |     ^
telescope_1      |
telescope_1      | Error: Unable to parse elastic URL "http://elasticsearch:9200" and/or PORT "9200"
telescope_1      |     at createElasticClient (/telescope/src/backend/lib/elastic.js:40:11)
telescope_1      |     at Object.<anonymous> (/telescope/src/backend/lib/elastic.js:48:11)
telescope_1      |     at Module._compile (internal/modules/cjs/loader.js:1063:30)
telescope_1      |     at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
telescope_1      |     at Module.load (internal/modules/cjs/loader.js:928:32)
telescope_1      |     at Function.Module._load (internal/modules/cjs/loader.js:769:14)
telescope_1      |     at Module.require (internal/modules/cjs/loader.js:952:19)
telescope_1      |     at require (internal/modules/cjs/helpers.js:88:18)
telescope_1      |     at Object.<anonymous> (/telescope/src/backend/utils/indexer.js:6:20)
telescope_1      |     at Module._compile (internal/modules/cjs/loader.js:1063:30)
```

It doesn't like http://elasticsearch:9200 as a URL (error code is `ERR_INVALID_URL`), which it's picking up from the `docker-compose.yml` file vs. my `.env`.

I think what's happening here is this:

```
> const { Client } = require('@elastic/elasticsearch');
> new Client({ node: 'http://elasticsearch:9200:9200' })
Uncaught TypeError [ERR_INVALID_URL]: Invalid URL: http://elasticsearch:9200:9200
    at onParseError (internal/url.js:258:9)
    at new URL (internal/url.js:334:5)
[...]

PR #1443, opened 2020-11-23
Error: Unable to parse elastic URL "http://elasticsearch:9200" and/or PORT "9200" #1442

zg3d:
## Issue This PR Addresses
#1442 

## Type of Change

## Description
Tested the docker build and it was succes full

## Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,7 +16,7 @@ services:
       - REDIS_URL=redis://redis:${REDIS_PORT}
       # variables for elasticsearch
       - ES_HOST=elasticsearch
-      - ELASTIC_URL=http://elasticsearch:${ELASTIC_PORT}
+      - ELASTIC_URL=http://elasticsearch
     depends_on:
       - redis
       - login
--- a/src/backend/lib/elastic.js
+++ b/src/backend/lib/elastic.js
@@ -2,6 +2,7 @@ const { ELASTIC_URL, ELASTIC_PORT } = process.env;
 const { Client } = require('@elastic/elasticsearch');
 const Mock = require('@elastic/elasticsearch-mock');
 const { logger } = require('../utils/logger');
+const parseUrl = require('../utils/url-parser');
 
 function MockClient(options) {
   const mock = new Mock();
@@ -32,7 +33,7 @@ const ElasticConstructor = useMockElastic ? MockClient : Client;
 
 function createElasticClient() {
   try {
-    const elasticUrl = `${ELASTIC_URL}:${ELASTIC_PORT}` || 'http://localhost:9200';
+    const elasticUrl = parseUrl(ELASTIC_URL, ELASTIC_PORT) || 'http://localhost:9200';
     return new ElasticConstructor({ node: elasticUrl });
   } catch (error) {
     const message = `Unable to parse elastic URL "${ELASTIC_URL}" and/or PORT "${ELASTIC_PORT}"`;
--- a/src/backend/utils/url-parser.js
+++ b/src/backend/utils/url-parser.js
@@ -1,0 +1,12 @@
+// parsesUrl to make sure there is only a single port
+const parseUrl = (url, port) => {
+  try {
+    const urlObj = new URL(url);
+    urlObj.port = port;
+    return urlObj.href;
+  } catch (e) {
+    return null;
+  }
+};
+
+module.exports = parseUrl;
--- a/test/url-parser.test.js
+++ b/test/url-parser.test.js
@@ -1,0 +1,48 @@
+const urlParser = require('../src/backend/utils/url-parser');
+
+describe('url-parser tests', () => {
+  const testUrlWithPort = 'http://127.0.0.1:9200';
+  const testUrlWithoutPort = 'http://127.0.0.1';
+  const testport = '9200';
+  const expectedReturn1 = 'http://127.0.0.1:9200/';
+  const expectedReturn2 = 'http://127.0.0.1/';
+
+  test('urlParser with double ports', () => {
+    const result = urlParser(testUrlWithPort, testport);
+    expect(result).toBe(expectedReturn1);
+  });
+
+  test('urlParser with 1 string port', () => {
+    const result = urlParser(testUrlWithoutPort, testport);
+    expect(result).toBe(expectedReturn1);
+  });
+
+  test('urlParser with invalid url', () => {
+    const result = urlParser('', testport);
+    expect(result).toBe(null);
+  });
+
+  test('urlParser with url with no port and no port', () => {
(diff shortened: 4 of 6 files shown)


src/backend/lib/elastic.js
```diff
@@ -30,9 +30,20 @@ const useMockElastic = process.env.MOCK_ELASTIC;
...
+    return urlObj.port === '' ? `${url}:${port}` : url;
+  }
+
+  return null;
+};
+
 function createElasticClient() {
   try {
-    const elasticUrl = `${ELASTIC_URL}:${ELASTIC_PORT}` || 'http://localhost:9200';
+    const elasticUrl = parseUrl(ELASTIC_URL, ELASTIC_PORT) || 'http://localhost:9200';
```
manekenpix, line 46: Take a look at what we do to parse [redis' url](https://github.com/Seneca-CDOT/telescope/blob/7531ea5bbb3eb9f40d16b744e418ead6ca9afb69/src/backend/lib/redis.js#L18), you might be able to do something similar here.

zg3d: @manekenpix  is that not deprecated now. I believe node URL has deprecated it. https://nodejs.org/api/url.html

humphd: Yeah, we can switch to the global `URL` constructor everywhere.  @zg3d want to fix that for Redis while you're doing this surgery?  Also, check if we do it anywhere else.

manekenpix: @zg3d I completely forgot :+1:

zg3d: Ya, I can do that.

docker-compose.yml
```diff
@@ -16,7 +16,7 @@ services:
       - REDIS_URL=redis://redis:${REDIS_PORT}
       # variables for elasticsearch
       - ES_HOST=elasticsearch
-      - ELASTIC_URL=http://elasticsearch:${ELASTIC_PORT}
+      - ELASTIC_URL=http://elasticsearch
```
humphd, line 19: Why this change?  If I want to run es natively, or over the network, why not allow that?

src/backend/utils/url-parser.js
```diff
@@ -0,0 +1,17 @@
+// parsesUrl to make sure there is only a single port
+const parseUrl = (url, port) => {
+  let result;
```
humphd, line 3: Drop this variable, we'll return early.

src/backend/utils/url-parser.js
```diff
@@ -0,0 +1,17 @@
...
+  try {
+    const urlObj = new URL(url);
+    urlObj.port = port;
+    if (urlObj.port !== '') {
```
humphd, line 7: Why bother with this?  Consider:

```js
parseUrl('http://localhost'); // no port (undefined or null), `port = undefined` or `port = null` will both give ''
parseUrl('http://localhost', 3000); // port as number, `port = 3000` will give '3000'
parseUrl('http://localhost', '3000'); // port as string, `port = '3000'` will give '3000'
parseUrl('http://localhost', 'invalid'); // port is invalid, `port = 'invalid'` will give ''
```

Given that `new URL(...).port = ...` won't pollute the port (it will just ignore bad data), shouldn't we always set `urlObj.port = port` and then return `urlObj.href` immediately?

src/backend/utils/url-parser.js
```diff
@@ -0,0 +1,17 @@
...
+      result = urlObj.href;
+    }
+  } catch (e) {
+    result = null;
```
humphd, line 11: `return null;`

test/url-parser.test.js
```diff
@@ -0,0 +1,27 @@
...
+    expect(result).toBe(expectedReturn);
+  });
+
+  test('urlParser with invalid url', () => {
+    const result = urlParser('', testport);
+    expect(result).toBe(null);
+  });
+
+  test('urlParser with url with no port and no port', () => {
+    const result = urlParser(testUrlWithoutPort, '');
```
humphd, line 24: Add test cases for:

- port is undefined
- port is null
- port is a number
- port is a string
- port is an invalid string
- port is a number outside the port range (e.g., 65535 + 1)

humphd (changes requested):
Nice, a few things I see.

docker-compose.yml
zg3d: Was this not the cause of the issue according to @manekenpix

src/backend/utils/url-parser.js
zg3d: is this about the try block ?

zg3d: @humphd  I don't understand this comment, is this not what I am already doing

humphd: No, you only return the `.href` in the case that the port is not nothing, meaning you now have 3 different return types:

1. `string` when port is set and we get back `.href`
2. `undefined` when port is `''`
3. `null` when there's a parsing error.

Let's reduce that to 2 and return either `null` or the URL string.  I don't think you need this `if` check, just return `urlObj.ref` in all cases within this try.

humphd (approved):
Awesome, great work.  One tip on the way you're writing these tests.  You have 6 variables at the top of the describe block that are shared between all the tests.  This isn't a great pattern, and you'd be better off moving all of these into the test functions that use them (maximum isolation between tests, no dependencies, etc).

In your case this isn't going to break, since you used `const` with strings; but if you had non-immutable types (e.g., object references) this could be problematic, if one test updated the data before another used it.

humphd:
Tagging @c3ho since he needs this fix for #1450.

c3ho:
image

I don't know if its me, but I can't get telescope to work with `docker-compose up --build` Telescope keeps crashing on Windows 10

humphd:
@c3ho what's in the log file it generated?

c3ho:
One of the earlier runs:
image

Terminal message in the most recent run:
image

I wiped docker images prior to testing this and still doesn't work. Maybe my elastic container is taking too long to start and Telescope crashes from it?

humphd:
Or maybe this PR isn't parsing the URL in your env correctly?  What are you using for env with these elasitic params?

c3ho:
image

Should it matter though? I thought we're overwriting it with the docker-compose envs when we run `docker-compose up --build`?

manekenpix:
@c3ho change `ELASTIC_DELAY_MS` in your `.env` file, give it more time. The same was happening to me and the problem was that it was taking ES more than `ELASTIC_DELAY_MS` to be ready to store posts.

manekenpix (approved):

c3ho (approved):
Got it to work after @manekenpix suggestion.

PR #1450, opened 2020-11-24
Fixed logging format

### LINKED ISSUE #1448: Logging no longer formatted for pretty display locally
We used to have really nice logging, pretty-printed for local debugging, but now I get raw JSON:

```
{"level":30,"time":1606175748870,"pid":88703,"hostname":"brightness.local","msg":"Starting 1 instance of feed processor."}
```

It might be something about how me `.env` is setup after the Kibana change?  Not sure, but I miss my pretty logs.

cc @c3ho

c3ho:
## Issue This PR Addresses
Fixes #1448 

## Type of Change

## Description
Now when logging to console, the log messages will look pretty again.

Could someone also test the pino-elastic logging as well for this and make sure it is also working? I tested it, but it was inconclusive for me.

Steps to test fix:
1. in the `.env` file set `LOG_LEVEL` to `info`
1. `docker-compose up redis elasticsearch`
1. `npm start`

Check info messages in terminal tab. They should look pretty now (just like before)

To regression test:
1. in the `.env` file set `LOG_LEVEL` to `info` and set `LOG_ELASTIC` to 1
1. `docker-compose up redis elasticsearch`
1. `npm start`, there should be no logging appearing in terminal tab anymore as they're directed to elasticsearch now
1. navigate to `http://localhost:9200/_cat/indices?v` You should see two rows and under the `indexes` column there should be 2 indexes  `logs` and `posts`
1. Remember to set `LOG_ELASTIC` to blank again in your `.env` file after testing.

## Checklist

### DIFF
--- a/src/backend/utils/logger.js
+++ b/src/backend/utils/logger.js
@@ -1,16 +1,28 @@
+const {
+  ELASTIC_URL,
+  ELASTIC_PORT,
+  INDEX_NAME,
+  ELASTIC_VERSION,
+  FLUSH_BYTES,
+  CONSISTENCY,
+  LOG_ELASTIC,
+  LOG_LEVEL,
+  NODE_ENV,
+} = process.env;
 const pino = require('pino');
 const pinoElastic = require('pino-elasticsearch');
 const expressPino = require('express-pino-logger');
 const os = require('os');
 const path = require('path');
+const parseUrl = require('./url-parser');
 
 require('../lib/config');
 
 // Deal with log levels we don't know, or which are incorrectly formatted
-let logLevel = (process.env.LOG_LEVEL || 'info').toLowerCase();
+let logLevel = (LOG_LEVEL || 'info').toLowerCase();
 if (!pino.levels.values[logLevel]) {
   // Use `debug` by default in development mode, `info` otherwise.
-  logLevel = process.env.NODE_ENV === 'development' ? 'debug' : 'info';
+  logLevel = NODE_ENV === 'development' ? 'debug' : 'info';
 }
 
 const options = {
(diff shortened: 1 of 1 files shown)


src/backend/utils/logger.js
```diff
@@ -39,14 +39,14 @@ const streamToElastic = pinoElastic({
   // Consistency of the write, valid values: 'one', 'quorum', 'all'
   consistency: (process.env.CONSISTENCY || 'one').toLowerCase(),
   // URL of Elasticsearch
-  node: `http://${process.env.ELASTIC_URL}:${process.env.ELASTIC_PORT}`,
+  node: `${process.env.ELASTIC_URL}:${process.env.ELASTIC_PORT}`,
```
humphd, line 42: We probably need the same treatment that @zg3d is doing in #1443 to build a proper URL + port without duplication of port.  Once that land, you can rebase and use his function.

humphd (changes requested):

src/backend/utils/logger.js
c3ho: Newest commit should address these

humphd (dismissed):
Looks good.

manekenpix (approved):

c3ho:
@humphd Is it safe to assume we have should have an almost even amount of docs in `logs` and `posts` index generated when we use `LOG_LEVEL=info`?

I'm thinking a doc should be created for each feed that we process(or skip)

humphd:
@c3ho sorry, I don't follow what you're asking.  Also, do you want to merge this?

c3ho:
@humphd sorry I'm holding off on merging this because I'm testing the options part. It seems to not like `prettyPrint` as I'm unable to get the `logs` index to show.

What I wanted to ask was for example we have the following:

image

with this PR, we should have a new index called `logs`, if I have `LOG_LEVEL=info` set. Shouldn't the `docs.count` # be almost the same if not more than the `docs.count` # for posts? I'm asking this because the `docs.count` for logs is only 8 after running it for 2-3 hours. I was expecting it to be a much higher account with all the lines of `INFO` we're getting in the terminal

humphd:
Yeah, that sounds odd to me (that it's so low).  What do you see in logs when you get those 8?

c3ho:
🤷 I'm still trying to figure out how to use kibana so I can look inside those logs

c3ho:
Nevermind, I got ~7k hits on logs just now.

humphd:
It might be caching results and not flushing them to the db until some threshold is reached.

c3ho:
I think that's the case with the `flush-bytes` setting.

c3ho:
@humphd I put another ternary to check `LOG_ELASTIC` in `.env` if set we'll use `{ log: logLevel }` instead of options (so pino-elastic actually works), if we're not logging to elastic we'll use `options` and keep `prettyPrint`

humphd (approved):

manekenpix (approved):

Issue #1366, opened 2020-11-15
Slack integrations


humphd:
**What would you like to be added**:

Let's see what we can do to add some notifications to our #telescope Slack channel.  Maybe there are existing tools that we can leverage for this, or maybe we can build our own (possibly as a separate repo).

**Why would you like this to be added**:

We have a number of situations that would be good to know about when the occur.  For example:

- dev or prod go down (e.g., something broke one of the servers)
- dev or prod get deployed (e.g., new build available)

Probably there are other things, but these are the main ones that occur to me.

PedroFonsecaDEV:
I would like to try it.

humphd:
@PedroFonsecaDEV please assign yourself.

sirinoks:
Good move to assign me

sirinoks:
Research:
[Github webhooks event: release](https://docs.github.com/en/developers/webhooks-and-events/events/github-event-types#releaseevent)
Making a [Slack bot](https://api.slack.com/bot-users)

What I need to find is an event when a build occurs on master, and when it breaks.

sirinoks:
Look into /tools 
Autodeployment server

humphd:
A build on the staging/production boxes is done when this function is called: https://github.com/Seneca-CDOT/telescope/blob/master/tools/autodeployment/builds.js#L17-L31

If the `code` is `=== 0` then it passed, otherwise it failed, so pseudo-code would be:

```js
if(code !==0) {
  slackBot.sendMessage("Build failed...");
}
```

JerryHue:
@sirinoks, what has been your progress related to this? I know that you were researching for any Slack APIs that would let you do this, but I don't know what you concluded with.

sirinoks:
> @sirinoks, what has been your progress related to this? I know that you were researching for any Slack APIs that would let you do this, but I don't know what you concluded with.

There has been no progress since Feb 15th. I still need to figure out the automation stuff and I haven't looked there.

cindyledev:
> > @sirinoks, what has been your progress related to this? I know that you were researching for any Slack APIs that would let you do this, but I don't know what you concluded with.
> 
> There has been no progress since Feb 15th. I still need to figure out the automation stuff and I haven't looked there.

Perhaps you can talk to @DukeManh , I remember he added a Slackbot but it was in the wrong channel so it was removed

sirinoks:
[...]

PR #1472, opened 2020-12-01
Issue 1366 - Slack integrations [ Docker Events ]

PedroFonsecaDEV:
## Issue This PR Addresses
This is the beginning of the implementation related to #1366.
 

## Type of Change

This

## Description
I created a new workspace to test it, and it's working. This is just a test, a draft.

This feature will send a message to our slack channel whenever our servers are down and when they back to work again. So we can have a live check on our servers(prod and dev). This feature doesn't rely on user commands; it will always be running.
This one is covering just production.

We will have slack commands, but for now, I believe this feature is more important. 

## Checklist


humphd (changes requested):
Nice work, this is going to be very cool.

I agree that this polling technique (doing a request every 5s to determine server health) is not ideal.  What about a backwards approach?  Two ideas:

First, we have code in Telescope to manage shutdowns, both graceful and error-based, see https://github.com/Seneca-CDOT/telescope/blob/736eb97c754047fc8b159d11a0216106fcb86965/src/backend/lib/shutdown.js.  We could modify this code to `POST` to the autodeployment server with details about why we are shutting down.  Then this code could be converted to a pure webhook, along the same lines as we do now from GitHub.

Second, I wonder if we could do this directly from pm2/docker-container.  I'm not sure of the details, but I suspect that we can hook into events in our process manager and do something similar when the Telescope app goes down.

@manekenpix will have opinions as well, which I'd like to hear.

tools/autodeployment/server.js
```diff
@@ -7,6 +7,7 @@ const mergeStream = require('merge-stream');
 const fs = require('fs');
 
 const { buildStart, buildStop, handleStatus } = require('./info');
+const { prodStatus } = require('./slack_app');
```
manekenpix, line 10: We use the same code for `prod` and `dev`, so you can name this `status` or `telescopeStatus`.

tools/autodeployment/slack_app/index.js
```diff
@@ -0,0 +1,54 @@
+require('dotenv').config();
+const axios = require('axios').default;
+
+const { SLACK_MESSAGE_HOOK } = process.env;
+
+const prodStatus = () => {
+  let alarm = false;
+
+  setInterval(async () => {
```
manekenpix, line 9: If we end up keeping this, I think it's better to use [setInterval Async](https://www.npmjs.com/package/set-interval-async), like we do in the [wiki-feed-parser](https://github.com/Seneca-CDOT/telescope/blob/38be67c1c0692df619e21174f3657eb661a47464/src/backend/utils/wiki-feed-parser.js#L4).

tools/autodeployment/slack_app/index.js
```diff
@@ -0,0 +1,54 @@
...
+      prodResponse = await axios.get('https://telescope.cdot.systems/posts', {
+        method: 'HEAD',
+        timeout: 3000,
+      });
+    } catch (error) {
+      if (!alarm) {
+        const downDate = new Date().toTimeString();
+        axios.post(SLACK_MESSAGE_HOOK, {
+          response_type: 'in_channel',
+          text: `\n:rotating_light: PRODUCTION IS DOWN. :rotating_light:\n`,
```
manekenpix, line 22: Here, instead of hardcoding `Prod` or `Dev`, you can use [DEPLOY_TYPE](https://github.com/Seneca-CDOT/telescope/blob/38be67c1c0692df619e21174f3657eb661a47464/tools/autodeployment/env.example#L12)'s value to know which server is running this code.

manekenpix (changes requested):

manekenpix:
I like the first approach suggested by @humphd, but I'd still keep the automated check (not every 5 secs, but maybe every 5 or 10 min) so we get an alert in those weird cases in which, after a successful build, the front-end works but the back-end crashes at launch.

PedroFonsecaDEV:
@humphd What about check if our backend is up in the root(OS)? Something like cron jobs checking the `backend container`. What do you think?

humphd:
Apparently we can get events from docker-compose itself:

- https://docs.docker.com/compose/reference/events/
- https://docs.docker.com/engine/reference/commandline/events/#object-types

PedroFonsecaDEV:
@humphd okay cool. I will try pm2/docker.
Thanks.

humphd:
Another thought I had this morning: dev could monitor prod, and vice versa, like siblings checking in on each other periodically.  This would be in addition to the internal checks, basically seeing if the backend is still online.

manekenpix:
A bit of research on `docker events`: a bash scripts that processes `docker events` and filters the one including `start` or `stop`.
I think it's a good start for testing. This approach can be used to send messages to slack depending on emitted events, 

```sh
#!/bin/bash
docker events |  while read line; do  if [[ ${line} = *"start"* || ${line} = *"stop"* ]];then echo " Event: ${line}" ;fi ; done
```

PedroFonsecaDEV:
@humphd @manekenpix 
I like both ideas. I found this PM2 tool: 
https://pm2.keymetrics.io/docs/usage/docker-pm2-nodejs/

PedroFonsecaDEV:
I followed all the advice, and the partial result is good. 
Probably I will land a PR later today. 

Thanks.
@humphd @manekenpix

PedroFonsecaDEV:
@humphd @manekenpix 

I've been talking with Josue for a few days; he explained to me a lot about how Telescope builds on CDOT machine and about Telescope's architecture, and he also helped me with a lot of things. We discussed many options for this new feature, and I decided to use an event-driven approach to listen to our docker system.
This PR is totally different from the last one. Now we use the auto-deployment server to listen to docker events using bash scripts. Here you will find a sketch of my proposal. All the error messages are just for demonstration purposes. 
Once we agree, I will continue to work on the feature. 
Thanks

tools/autodeployment/env.example
```diff
@@ -16,3 +16,7 @@ UNSPLASH_CLIENT_ID=
 
 # Path to certificates
 PATH_TO_CERTS=
+
+# Slack "Incoming WebHook" to send messages to #telescope channel.
+# For more information: https://api.slack.com/messaging/webhooks
+SLACK_SEND_MESSAGE=
```
humphd, line 22: Let's add an example of what this should look like in addition to the docs.

tools/autodeployment/src/docker_events/index.js
```diff
@@ -0,0 +1,60 @@
+const shell = require('shelljs');
+const fetch = require('node-fetch');
+const Events = require('events');
+
+const { SLACK_SEND_MESSAGE } = process.env;
+
+const dockerEvents = new Events.EventEmitter();
+
+const message = {
```
humphd, line 9:
```js
const message = {
  down: {
    response_type: 'in_channel',
      text: `\n:rotating_light: PRODUCTION IS DOWN. :rotating_light:\n`,
      ...etc,
  },
  up: {
    ...values for up message
  }
};
```

Now you can use `message.up` and `message.down` below.

tools/autodeployment/src/docker_events/index.js
```diff
@@ -0,0 +1,60 @@
...
+  text: `\nPRODUCTION IS BACK\n`,
+  attachments: [
+    {
+      text: `PROD IS UP!\n`,
+    },
+  ],
+};
+
+const waitingDocker = () => {
+  shell.chmod('700', './waitDocker.sh');
```
humphd, line 30: This seems odd/dangerous to do in a script.  Can we not do it as part of installing the autodeployment server once?

tools/autodeployment/src/docker_events/failCheck.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
```
humphd, line 2: Let's add a comment block to the top of this file explaining what you're doing and why.

tools/autodeployment/src/docker_events/waitDocker.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
```
humphd, line 2: Same thing here, let's document what this is.

humphd (changes requested):

PedroFonsecaDEV:
@humphd 
Okay. working on it.

PedroFonsecaDEV:
I will break the #1472 into different PRs.

Here we have just the docker event listener, and it is not ready yet. It still needs to improve. I'm still working on the following "bugs":

1. Docker doesn't emit an event when we close it (Docker -> Quit Docker Desktop), and `docker events` runs only if the Docker app is running. So if docker crashes or closes for some reason, the `docker events` will quit without any error code; without sent us an Event. (Maybe enforce PM2 to restart Docker if it's crashed or not responding??). 

About PM2, I believe that we need to enforce it to keep the auto-deployment server running(restart the server if needed). 

There will be a second PR landing in the next hours covering the "siblings check". as David suggested:
>humphd commented 10 days ago
Another thought I had this morning: dev could monitor prod, and vice versa, like siblings checking in on each other periodically. This would be in addition to the internal checks, basically seeing if the backend is still online.

Thanks.

tools/autodeployment/src/docker_events/index.js
```diff
@@ -0,0 +1,108 @@
...
+};
+
+const dockerListener = () => {
+  const event = shell.exec('./docker-listener.sh', {
+    silent: true,
+    async: true,
+  });
+  event.stdout.on('data', (data) => {
+    const containerEvent = JSON.parse(data);
+    if (containerEvent.Actor.Attributes.name.includes('telescope')) {
```
PedroFonsecaDEV, line 98: @manekenpix, this line is for filter containers. Because we could have more than one project running Docker in the machine, let's think of how it would work on CDOT machines. 
This line works on my local machine because all the docker's containers start with the word "telescope"; but I think this is different on CDOT machines.

tools/autodeployment/src/docker_events/index.js
```diff
@@ -0,0 +1,108 @@
...
+  } = containerEvent;
+
+  if (
+    (containersDown.has(name) && eventsDown.includes(Action)) ||
+    (containersUp.has(name) && eventsBack.includes(Action))
+  )
+    return;
+
+  if (eventsDown.includes(Action)) {
+    containersUp.has(name) ? containersUp.delete(name) : null;
```
PedroFonsecaDEV, line 82: @humphd 
How can I improve my coding style here?
I'm trying to avoid callbacks hell, and then I had to disable linter (/* eslint-disable no-unused-expressions */) to code these lines where I use a ternary operator with null.

humphd:
@PedroFonsecaDEV from triage today, we are thinking that this is something to return to after the next.js port.  Is that your feeling?  Or should we work on this in the 1.6 time frame?  Happy to go either way.

PedroFonsecaDEV:
@humphd agree with you. We should focus on porting to Next.

izhuravlev:
Thank you very much to everyone committing to this PR, your work is very appreciated. 

We will be coming back to this PR later, so we are closing it for now.

PR #1524, opened 2020-12-11
remove empty whitespace-only blocks from posts

### LINKED ISSUE #1189: Remove empty whitespace-only blocks from posts
I feel like @agarcia-caicedo mentioned this in another issue, but I can't find it now.  A number of blog platforms add extra whitespace blocks to posts, for example:

Screen Shot 2020-10-12 at 4 56 27 PM

We could try doing some work in our HTML post-processing step to remove things like empty `<p>`, multiple `<br>`, `empty `<div></div>` or `<div><p></p></div>`

1jz:
## Issue This PR Addresses

Fixes #1189 

## Type of Change

## Description

Removes redundant elements wrapping \<br> elements and replaces repetitive \<br> elements with a single \<br>

Also removes empty \<p> elements.

### DIFF
--- a/src/backend/utils/html/replace-redundant-breaklines.js
+++ b/src/backend/utils/html/replace-redundant-breaklines.js
@@ -1,0 +1,5 @@
+const redundantBreaklineRegex = /(<br>((\s?)+)){2,}|((<p>|<div>)(\s+)?(<br>)(\s+)?(<\/p>|<\/div>)((\s?)+)){1,}/g;
+
+module.exports = function (html) {
+  return html.replace(redundantBreaklineRegex, '<br>\n').trim();
+};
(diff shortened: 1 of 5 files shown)


src/backend/utils/html/replace-redundant-breaklines.js
```diff
@@ -0,0 +1,5 @@
+const redundantBreaklineRegex = /(<br>((\s?)+)){2,}|((<p>|<div>)(\s+)?(<br>)(\s+)?(<\/p>|<\/div>)((\s?)+)){1,}/g;
```
humphd, line 1: Regex to deal with the DOM makes me nervous.  Can you not use CSS Selectors for this?  The DOM already has a parsing engine built into it without us inventing a new one via regex.

There's a famous StackOverflow answer about the perils of this: https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags?answertab=active#tab-top

humphd (changes requested):

Metropass (changes requested):
I know this might be a bit pedantic, but I think we should also check for the 0-dot braille for security reasons. Someone could possibly embed something malicious, and use the character as a means to bypass this check.  I'll make a mock function for it

PR #1530, opened 2020-12-13
Issue 1366 - Slack integrations [ Siblings Check ]

PedroFonsecaDEV:
## Type of Change

## Description
This PR is part of the solution of #1366.
Siblings Checkers: DEV and PROD checking each other periodically.

We also have an internal checker:
Slack Integration: [Docker Events](https://github.com/Seneca-CDOT/telescope/pull/1472)

## Checklist


PedroFonsecaDEV:
Screen Shot 2020-12-12 at 10 18 02 PM

tools/autodeployment/env.example
```diff
@@ -16,3 +16,14 @@ UNSPLASH_CLIENT_ID=
...
+# Slack "Incoming WebHook" to send messages to #telescope channel.
+# For more information: https://api.slack.com/messaging/webhooks
+# Example:
+# await fetch(SLACK_SEND_MESSAGE, {
+#       method: 'post',
+#       headers: { 'Content-Type': 'application/json' },
+#       body: "text": "This is the message"
+#     });
+# This will send the following message to our Slack channel: "This is the message"
+SLACK_SEND_MESSAGE=
```
manekenpix, line 29: I'm not sure of what this means. Is `SLACK_SEND_MESSAGE` the text that will be sent to the channel?
Also, I don't think you need to show an example here, the URL to webhooks should be enough.

tools/autodeployment/server.js
```diff
@@ -5,9 +5,14 @@ const createHandler = require('github-webhook-handler');
 const shell = require('shelljs');
 const mergeStream = require('merge-stream');
 const fs = require('fs');
+const { firstCheck } = require('./src/sibling-checker');
 
 const { buildStart, buildStop, handleStatus } = require('./info');
 
+// Server health check
+
+firstCheck();
```
manekenpix, line 14: You could just call it `healthCheck()`

tools/autodeployment/src/sibling-checker/index.js
```diff
@@ -0,0 +1,81 @@
...
+      timeout: 1500,
+    });
+    if (result.status !== 200) {
+      clearIntervalAsync(timerOne);
+      secondCheck();
+    }
+  }, 120000);
+};
+
+const secondCheck = () => {
```
manekenpix, line 28: can this have a name that tells you what it's doing instead of `secondCheck`?

tools/autodeployment/src/sibling-checker/index.js
```diff
@@ -0,0 +1,81 @@
...
+const fetch = require('node-fetch');
+const { setIntervalAsync, clearIntervalAsync } = require('set-interval-async/dynamic');
+
+const server =
+  process.DEPLOY_TYPE === 'production'
+    ? 'https://telescope.cdot.systems/health.'
+    : 'https://dev.telescope.cdot.systems/health.';
+
+let timerOne;
+let timerTwo;
```
manekenpix, line 12: These could be `healthTimer` and `whateverNameYouHaveForThisTimer`.

tools/autodeployment/src/sibling-checker/index.js
```diff
@@ -0,0 +1,81 @@
...
+
+const firstCheck = () => {
+  timerOne = setIntervalAsync(async () => {
+    const result = await fetch(server, {
+      method: 'head',
+      timeout: 1500,
+    });
+    if (result.status !== 200) {
+      clearIntervalAsync(timerOne);
+      secondCheck();
```
manekenpix, line 23: instead of this, why don't have the 2 intervals running independently? Or maybe one interval calling the 2 checks sequentially.

manekenpix (changes requested):
It looks good, just a few nits. We'll need a teams/zoom session to test this.

tools/autodeployment/src/sibling-checker/index.js
```diff
@@ -0,0 +1,81 @@
+/* eslint-disable no-use-before-define */
+/* eslint-disable no-unused-expressions */
+const fetch = require('node-fetch');
+const { setIntervalAsync, clearIntervalAsync } = require('set-interval-async/dynamic');
+
+const server =
+  process.DEPLOY_TYPE === 'production'
+    ? 'https://telescope.cdot.systems/health.'
+    : 'https://dev.telescope.cdot.systems/health.';
```
manekenpix, line 9: We could do here what @humphd suggested. We can make the servers check each other, so if the machine is down, we'd still get the status, but I think that you'd need to add a case in which the server never answers.

tools/autodeployment/server.js
PedroFonsecaDEV: Sure. It will make more sense in this scope.

tools/autodeployment/env.example
PedroFonsecaDEV: Example requested by @humphd:
https://github.com/Seneca-CDOT/telescope/pull/1472#discussion_r533473388

`SLACK_SEND_MESSAGE` is an URL that receives a fetch/post with the message as the body and publishes it on our slack channel. It's almost like an API KEY, everyone with this URL will be able to send messages to our slack channel.

tools/autodeployment/src/sibling-checker/index.js
PedroFonsecaDEV: I will sketch this feature. I think it will be better to understand it. But basically, we have the `firstCheck` running every 2 min. If the `firstCheck` fails, it will trigger the `secondCheck`  that stops the `firstCheck` and performs a fetch every 10sec. If the `secondCheck` fails for the first time, the message "SERVER DOWN" is sent to us on SLACK. Once the message is sent, the secondCheck continues to run (the message informing us about the crash is sent just once). When the server is back, `secondCheck` sends the message "SERVER IS BACK", stop itself and triggers the `firstCheck`. 

Because we rely on the network, sometimes the `firstCheck` can emit a false-negative when it fails due to network "issues" (timeout, etc.). That's why I created this system to double-check the server and try to avoid false-negative messages on our slack channel.

PedroFonsecaDEV:
@manekenpix 
> It looks good, just a few nits. We'll need a teams/zoom session to test this.

Sure. Let's do it.

PedroFonsecaDEV:
@manekenpix @humphd 
SlackIntegration-siblings-checker

tools/autodeployment/env.example
manekenpix: I still think the example should go [in the docs](https://github.com/Seneca-CDOT/telescope/blob/master/docs/staging-production-deployment.md), where all the `env` vars are explained.

PedroFonsecaDEV: Sure, you are right. We should add them to this doc too.

humphd: If this is a URL, let's label it as such.  I think @manekenpix's confusion is partly due to the name.  What about `SLACK_WEBHOOK_URL` instead?

tools/autodeployment/src/sibling-checker/index.js
```diff
@@ -0,0 +1,81 @@
+/* eslint-disable no-use-before-define */
+/* eslint-disable no-unused-expressions */
```
humphd, line 2: Are these really necessary?  Can we not change your code to fit our linting style?

tools/autodeployment/src/sibling-checker/index.js
```diff
@@ -0,0 +1,81 @@
...
+  timerOne = setIntervalAsync(async () => {
+    const result = await fetch(server, {
+      method: 'head',
+      timeout: 1500,
+    });
+    if (result.status !== 200) {
+      clearIntervalAsync(timerOne);
+      secondCheck();
+    }
+  }, 120000);
```
humphd, line 25: All these magic numbers should probably be something we can configure via env.

tools/autodeployment/src/sibling-checker/index.js
```diff
@@ -0,0 +1,81 @@
...
+          type: 'mrkdwn',
+          text: `* ${process.DEPLOY_TYPE.toUpperCase()} IS ${status === 200 ? 'BACK' : 'DOWN'}
+          *\n\n`,
+        },
+        accessory: {
+          type: 'image',
+          image_url:
+            status === 200
+              ? 'https://data.whicdn.com/images/186936913/original.jpg'
+              : 'https://64.media.tumblr.com/090e86896948354262566ddb906e514e/tumblr_mxrsfhStH61r62zwpo1_500.jpg',
```
humphd, line 64: Should we pull images we depend on into our repo so we can rely on the URLs not going down in the future?

tools/autodeployment/src/sibling-checker/index.js
```diff
@@ -0,0 +1,81 @@
...
+          alt_text: 'alt text for image',
```
humphd, line 65: This needs to be fixed to something useful.

humphd (changes requested):
Looking better and better.  I agree, let's get on a call early this week and test it together.  Let's co-ordinate on Slack when these changes are done and we can test for real.

tools/autodeployment/src/sibling-checker/index.js
PedroFonsecaDEV: Sure, this is just a test message.

tools/autodeployment/src/sibling-checker/index.js
PedroFonsecaDEV: Same as above. I create a simple message with an image to test the module.

tools/autodeployment/src/sibling-checker/index.js
PedroFonsecaDEV: Regarding the `/* eslint-disable no-unused-expressions */` it'solved. 

But regarding `/* eslint-disable no-unused-expressions */`
I have tried until I run out of ideas.

Because the `firstCheck` calls the `secondCheck` and vice versa. 

```js
const a = () => {
b(); -> LINTER: "b" was used before it was defined.
}
const b = () => {
a()
}
```

Any idea about this is welcome.

tools/autodeployment/server.js
```diff
@@ -5,9 +5,14 @@ const createHandler = require('github-webhook-handler');
 const shell = require('shelljs');
 const mergeStream = require('merge-stream');
 const fs = require('fs');
+const { firstCheck: healthCheck } = require('./src/sibling-checker');
 
 const { buildStart, buildStop, handleStatus } = require('./info');
 
+// Server health check
+
+healthCheck();
```
PedroFonsecaDEV, line 14: @manekenpix

Metropass:
This is very cool, might I add a suggestion?

I know this is bare bones at the moment, but I was wondering if we can write the status of the server into a .txt file

It might be a bit easier on the eyes, and it'll help archive the logs from the server

PedroFonsecaDEV:
> This is very cool, might I add a suggestion?
> 
> I know this is bare bones at the moment, but I was wondering if we can write the status of the server into a .txt file
> 
> It might be a bit easier on the eyes, and it'll help archive the logs from the server

Could you please elaborate more?

PedroFonsecaDEV:
I'm updating the images because I'm not sure about the license of the previous images.

docs/staging-production-deployment.md
```diff
@@ -63,6 +63,13 @@ Copy `env.example` in the `autodeployment` directory to `.env` and add the appro
   - deployment type (`staging` or `production`)
   - unsplash client id
 
+#### 3.1- Autodeployment server - Module: Slack Integration
+
+- env variables
+  - slack webhook url (#telescope's url Slack Incoming Webhooks)
+  - interval one (120000)
+  - interval two (10000)
```
manekenpix, line 71:
```suggestion
# Explanation here about what these intervals are for
  - interval one (120 * 1000)
  - interval two (10 * 1000)
```
Splitting the numbers like this^ helps you understand what they mean:

tools/autodeployment/src/sibling-checker/index.js
manekenpix: `timerOne` and `timerTwo` doesn't tell you anything about what their purpose is or why they exist.

tools/autodeployment/src/sibling-checker/index.js
```diff
@@ -0,0 +1,80 @@
...
+const server =
+  process.env.DEPLOY_TYPE === 'production'
+    ? 'https://telescope.cdot.systems/health.'
+    : 'https://dev.telescope.cdot.systems/health.';
+
+let timerOne;
+let timerTwo;
+let messageSent = false;
+
+const firstCheck = () => {
```
manekenpix, line 14: Are you going to change the name of the checks?

tools/autodeployment/src/sibling-checker/index.js
```diff
@@ -0,0 +1,80 @@
...
+  timerOne = setIntervalAsync(async () => {
+    const result = await fetch(server, {
+      method: 'head',
+      timeout: 1500,
```
manekenpix, line 18: Magic number

tools/autodeployment/src/sibling-checker/index.js
```diff
@@ -0,0 +1,80 @@
+/* eslint-disable no-use-before-define */
+const fetch = require('node-fetch');
+const { setIntervalAsync, clearIntervalAsync } = require('set-interval-async/dynamic');
+
+const server =
+  process.env.DEPLOY_TYPE === 'production'
+    ? 'https://telescope.cdot.systems/health.'
+    : 'https://dev.telescope.cdot.systems/health.';
```
manekenpix, line 8: What I meant with the sibling checker approach is that, in this section where you decide what URL to use, you could use `dev.telescope` for the production machine and vice versa.
This way, the checks would be run against *the other machine*, and we'd be able to get an answer even if the targeted machine is down (if the machine checking is also the machine being checked, you won't get any messages when the machine is down).

manekenpix (changes requested):
Some more minor nits.
I'm still not confident about this. There must be an easier way to do this, instead of this circular check. I think a sequential check should be enough (or more than one check, but run sequentially).
I'm definitely open to approve this is we can't find a better solution.

tools/autodeployment/src/sibling-checker/index.js
PedroFonsecaDEV: You are right, this is the idea. I need to invert the URLs!

tools/autodeployment/src/sibling-checker/index.js
PedroFonsecaDEV: Okay.

docs/staging-production-deployment.md
PedroFonsecaDEV: Sure.

yuanLeeMidori:
@PedroFonsecaDEV Hi, what's the progress of this PR?

PedroFonsecaDEV:
> @PedroFonsecaDEV Hi, what's the progress of this PR?

Hey @yuanLeeMidori 

As mentioned, all the Slack Integration PR's are on hold until we finish the new UI.

izhuravlev:
Thank you very much to everyone committing to this PR, your work is very appreciated. 

We will be coming back to this PR later, so we are closing it for now.

PedroFonsecaDEV:
Sorry, I will reopen this PR for testing purposes.

PedroFonsecaDEV:
Okay tested. Sorry.

PR #1513, opened 2020-12-10
[next] Fixes issue#1481: re-implements the LogoIcon component

chrispinkney:
Fixes #1481:
* Added LogoIcon component and the `.svg` Telescope logo.

## Type of Change

- [ ] **Bugfix**: Change which fixes an issue
- [X] **New Feature**: Change which adds functionality
- [ ] **Documentation Update**: Change which improves documentation
- [ ] **UI**: Change which improves UI

## Description

Re-implements the `LogoIcon` component in next.js using a webp version of the logo (because, why not? _but seriously I can change it back if required, I was just curious about the webp format_).

Simply import and call LogoIcon while specifying a size and we're off to the races:

```TS
import LogoIcon from './LogoIcon';
<LogoIcon height={320} width={320} />
```

## Checklist

- [ ] **Quality**: This PR builds and passes our npm test and works locally
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [X] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not (if applicable)
- [ ] **Documentation**: This PR includes updated/added documentation to user exposed functionality or configuration variables are added/changed or an explanation of why it does not(if applicable)

Issue #1316, opened 2020-11-11
[Meta] Port existing GatsbyJS code to Next.js


humphd:
## Co-ordination of the GatsbyJS to Next.js Port

In #1314 we are beginning the transition from [GatsbyJS](https://www.gatsbyjs.com/) to [Next.js](https://nextjs.org/) for the front-end.  We will run parallel versions of the front-end while the transition occurs, with the current GatsbyJS version being the one we ship.

A new branch, [next](https://github.com/Seneca-CDOT/telescope/tree/next), has been created to facilitate this work in parallel to the `master` branch.  **All PRs for the Next.js port should happen against the `next` vs. `master` branch**.

We are going to be introducing [TypeScript](https://www.typescriptlang.org/) in the front-end.

This issue will remain open for the duration of the port, and we'll use it to organize what is done and what is still happening.

## How to Participate

We'd love to have you help with the work!  If you'd like to take on some part of the port, please let us know by commenting in this issue, and mentioning which of the tasks you'd like to take.

When you have chosen a task, this comment will get updated to indicate that you're working on it.  You should file an issue for the work, and we'll include that issue number in this comment.  When it's done, we'll indicate that in this comment.

When doing your work, please consult the file(s) in [the PR that removed the Gatsby code](https://github.com/Seneca-CDOT/telescope/pull/1314/files) as well as the [current front-end in `master`](https://github.com/Seneca-CDOT/telescope/tree/master/src/frontend), which may have evolved.

When you work on a task, you should try to decide if the old code is still needed.  Some things might only have been necessary when using GatsbyJS.  You should also make sure that we aren't missing something specific to Next.js.

Please see [the GatsbyJS to Next.js migration guide](https://nextjs.org/docs/migrating/from-gatsby)

## Source Code Note

In https://github.com/Seneca-CDOT/telescope/commit/6195af3fb3b9a277ca52e24a72c95e765be57791 we merged the `next` branch with `master`, giving us two frontends in the same tree:

- `src/frontend/gatsby` is our current Gatsby.js based frontend.  Please don't break it, as we are still running it on https://dev.telescope.cdot.systems (our staging box).
- `src/frontend/next` is our new Next.js based frontend.  You should add new code here.

See https://github.com/Seneca-CDOT/telescope/blob/master/src/frontend/README.md for details on how to use the two front-ends.

## Live Preview

[...]

PR #1531, opened 2020-12-15
[next] 1358 - MainPage (index.js) 

### LINKED ISSUE #1358: [next] Main page, see src/frontend/src/pages/index.js
[Next.js migration. Port-MAINPAGE](https://github.com/Seneca-CDOT/telescope/issues/1316)
Next.js migration.

PedroFonsecaDEV:
## Issue This PR Addresses

Fixes #1358 - [next] Port MainPage.

## Type of Change
PORT TO NEXT.JS 

## Description
The MainPage.tsx is done; I had to create some basic components to have it. 

It seems to me that we don't need the PageBase component now. 

## Checklist


humphd:
Just starting to look at this, and wondering if you have tips on how to test this?  @PedroFonsecaDEV

src/frontend/next/src/pages/index.tsx
```diff
@@ -1,60 +1,25 @@
+import { FC } from 'react';
 import Head from 'next/head';
-import styles from '../styles/Home.module.css';
```
humphd, line 2: It looks like we haven't got CSS figured out yet, is that happening in a follow-up?

src/frontend/next/src/pages/index.tsx
```diff
@@ -1,60 +1,25 @@
...
 
-const Home = () => {
+import Banner from '../components/Banner/Banner';
+import BackToTopButton from '../components/BackToTopButton/BackToTopButton';
+import Posts from '../components/Posts/Posts';
+
+const MainPage: FC = () => {
   return (
-    <div className={styles.container}>
+    <>
```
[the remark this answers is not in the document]
HyperTHD: It's a React Fragment, which works in a similar area. refer [here](https://stackoverflow.com/questions/47761894/why-are-fragments-in-react-16-better-than-container-divs) for it's possible uses

PedroFonsecaDEV:
@humphd 
> Just starting to look at this, and wondering if you have tips on how to test this? @PedroFonsecaDEV

I use the Netlify deployment to test it here on GitHub.

src/frontend/next/src/pages/index.tsx
PedroFonsecaDEV: File: [telescope/src/frontend/gatsby/src/pages/index.js](https://github.com/Seneca-CDOT/telescope/blob/master/src/frontend/gatsby/src/pages/index.js)

```js
import React from 'react';

import PageBase from './PageBase';
import BackToTopButton from '../components/BackToTopButton';

export default function IndexPage() {
  return (
    <PageBase title="Home">
      <Banner />
      <BackToTopButton />
      <main className="main">
        <Posts />
      </main>
    </PageBase>
  );
}
```

We don't have a CSS in this component. The `PageBase` component holds the global `CSS` and `SEO`. 
With Next.js, we may have to break the `PageBase`. Reading this [Next.js doc](https://nextjs.org/docs/basic-features/built-in-css-support) I understood that the "global CSS" would be defined on the file: [next/src/pages/_app.tsx](https://github.com/Seneca-CDOT/telescope/blob/master/src/frontend/next/src/pages/_app.tsx). But I would appreciate feedback on this approach.  

Note about `SEO`: I'm not sure if we could add our `SEO` to `_app.tsx` or if we need to add it to each page.

src/frontend/next/src/pages/index.tsx
```diff
@@ -1,60 +1,25 @@
+import { FC } from 'react';
 import Head from 'next/head';
-import styles from '../styles/Home.module.css';
 
-const Home = () => {
+import Banner from '../components/Banner/Banner';
+import BackToTopButton from '../components/BackToTopButton/BackToTopButton';
+import Posts from '../components/Posts/Posts';
+
+const MainPage: FC = () => {
```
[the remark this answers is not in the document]
PedroFonsecaDEV: Sure.
In our Gatsby we export this [component](https://github.com/Seneca-CDOT/telescope/blob/master/src/frontend/gatsby/src/pages/index.js#L8) as `IndexPage`. 
and
In our [Port](https://github.com/Seneca-CDOT/telescope/issues/1316) issue we refer to this component as `Main Page`.

I will change it to `Home`.

src/frontend/next/src/pages/index.tsx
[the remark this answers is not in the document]
PedroFonsecaDEV: [React Fragments docs](https://reactjs.org/docs/fragments.html)

src/frontend/next/src/pages/index.tsx
tonyvugithub: @PedroFonsecaDEV : We can defined global CSS in MUI theme as well. I think that would be a better approach to have global styling in the theme.

tonyvugithub:
I am still leaning toward keeping the PageBase component for quick migration. The least adjustment we need to make the better. How you guys think?

src/frontend/next/src/pages/index.tsx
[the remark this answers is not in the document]
chrispinkney: Thanks guys!

src/frontend/next/src/pages/index.tsx
humphd: For global styling, the work happening in #1549 is probably the right approach.

PedroFonsecaDEV:
@tonyvugithub 
> I am still leaning toward keeping the PageBase component for quick migration. The least adjustment we need to make the better. How you guys think?

I'm working on the `pageBase` problem (https://github.com/Seneca-CDOT/telescope/issues/1548). Probably I will break It into small PRs. And the first one is gonna be the `pageBase` solution.

src/frontend/next/src/components/BackToTopButton/BackToTopButton.tsx
```diff
@@ -0,0 +1,5 @@
+import { FC } from 'react';
+
+const BackToTopButton: FC = () => <h3>BackToTopButton Component</h3>;
+
+export default BackToTopButton;
```
chrispinkney, line 5: In a few of our current PRs we're adding an `index.ts` file for easier imports, [see here](https://github.com/Seneca-CDOT/telescope/pull/1513/files#diff-23afde8fc396f80b6fe6e1c7acf70eebf18273d03d482bc0266ae5671cadda4cR1) for an example, can you convert `BackToTopButton`, `Banner`, and `Posts` into that? The imports will need to get changed after too 😄

tonyvugithub:
I wonder if we can merge this now or wait after we merge BackTopTopBUtton, Header and Posts to merge it. What would be a better approach here. This looks good to go though

src/frontend/next/src/components/BackToTopButton/BackToTopButton.tsx
humphd: This is a great point to raise, @chrispinkney.  Here are my thoughts:

- JSX (and for TypeScript, TSX) is a non-standard extension of JavaScript
- Some people, especially outside the React ecosystem, don't like using `.js` for a file that isn't standard JavaScript (or `.ts` for TypeScript). 
- Transpilers (i.e., Babel, which translates JSX to pure JS) doesn't care what you call it, and at runtime, there is no-such thing as JSX: it all gets converted to pure functions.
- The general advice with transpiled code is to import it **without** the extension: `import Foo from './foo'` without any extension, and let the module system figure it out at build- vs. run-time.
- We are using React and next.js, so the vast majority of files are likely to contain JSX (i.e., it's not an edge case we need to worry about).

I think that another reason we used the `index.js` approach to do the final export was that we also wanted to put CSS files in the same scope:

```
/Foo
    index.js    <---- takes care of combining the JSX and CSS, and provides the default export for Foo
    Foo.jsx
    Foo.css
```

Later, we moved to Material UI, and with it, CSS-in-JS.  Next.js doesn't really like global CSS anyway (they either want you to use CSS Modules or CSS-in-JS).

In summary, I think we can ditch the `index.js` export files completely.  If we follow the patterns in the [next.js TypeScript example](https://github.com/vercel/next.js/tree/canary/examples/with-typescript), in this PR I would suggest we do this:

`src/frontend/next/src/components/BackToTopButton/BackToTopButton.tsx` becomes ` src/frontend/next/src/components/BackToTopButton.tsx` (i.e., drop the directory and use `tsx`).

humphd (changes requested):

src/frontend/next/src/components/BackToTopButton/BackToTopButton.tsx
humphd: Filed https://github.com/Seneca-CDOT/telescope/issues/1553 for follow-up.

humphd: Filed https://github.com/Seneca-CDOT/telescope/issues/1554 for follow-up.

humphd (approved):
I think we can merge this, and then replace things when the other components are done.  It would be good to get a main page that works.

manekenpix (approved):
LGTM :+1: 

image

PR #1552, opened 2021-01-18
Ported ScrollAction component

### LINKED ISSUE #1469: [next] Port ScrollAction from Gatsby to NextJS
This is an issue to port ScrollToTop from Gatsby to NextJS
Part of #1316, Next.js migration.

tonyvugithub:
## Issue This PR Addresses
Fixed #1469 
## Type of Change

## Description
I have modified the ScrollAction Component to be written in TS.
## Checklist

### DIFF
--- a/src/frontend/next/src/components/ScrollAction/ScrollAction.tsx
+++ b/src/frontend/next/src/components/ScrollAction/ScrollAction.tsx
@@ -1,0 +1,26 @@
+import { ReactChildren, MouseEvent } from 'react';
+
+type ScrollActionProps = {
+  children: ReactChildren;
+};
+
+const ScrollAction = ({ children }: ScrollActionProps) => {
+  const handleClick = (event: MouseEvent<HTMLElement>) => {
+    const mobile = window.innerWidth <= 1065;
+    const anchor = mobile
+      ? (event.currentTarget.ownerDocument || document).querySelector('#back-to-top-anchor-mobile')
+      : (event.currentTarget.ownerDocument || document).querySelector('#back-to-top-anchor');
+
+    if (anchor) {
+      anchor.scrollIntoView({ behavior: 'smooth', block: 'start' });
+    }
+  };
+
+  return (
+    <div onClick={handleClick} role="presentation">
+      {children}
+    </div>
+  );
+};
+
+export default ScrollAction;


src/frontend/next/src/components/ScrollAction/ScrollAction.tsx
```diff
@@ -0,0 +1,30 @@
+import { FC, ReactChildren, MouseEvent } from 'react';
+
+type ScrollActionProps = {
+  children: ReactChildren;
+};
+
+const ScrollAction: FC<ScrollActionProps> = ({ children }) => {
```
humphd, line 7: Do we not need to type `{ children } : ScrollActionProps` here?

src/frontend/next/src/components/ScrollAction/ScrollAction.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const handleClick = (event: MouseEvent<HTMLElement>) => {
+    const mobile = window.innerWidth <= 1065;
+    const anchor = mobile
+      ? ((event.target as HTMLElement).ownerDocument || document).querySelector(
```
humphd, line 11: Can you not use `event.currentTarget` here so you don't have to cast?

src/frontend/next/src/components/ScrollAction/ScrollAction.tsx
```diff
@@ -0,0 +1,30 @@
...
+          '#back-to-top-anchor-mobile'
+        )
+      : ((event.target as HTMLElement).ownerDocument || document).querySelector(
```
humphd, line 14: Same question here.

humphd (changes requested):

src/frontend/next/src/components/ScrollAction/ScrollAction.tsx
HyperTHD: The line `FC<ScrollActionProps>` ensures that any props passed into this component must be of type "ScrollActionProps". So it's already taken cared of here

src/frontend/next/src/components/ScrollAction/ScrollAction.tsx
```diff
@@ -0,0 +1,30 @@
+import { FC, ReactChildren, MouseEvent } from 'react';
+
+type ScrollActionProps = {
```
PedroFonsecaDEV, line 3: What are the cons and pros of `type` and `interface`?

src/frontend/next/src/components/ScrollAction/ScrollAction.tsx
PedroFonsecaDEV: I thought `const ScrollAction: FC<ScrollActionProps>` was seting the return type to `ScrollActionProps`.

PedroFonsecaDEV: TS is so confused.

tonyvugithub: @PedroFonsecaDEV This part means ScrollAction is a functional component and it takes props in the form of ScrollActionProps

src/frontend/next/src/components/ScrollAction/ScrollAction.tsx
tonyvugithub: Yeah that was what I tried initially, but since I also tried to keep everything the same so I cast it. Should I use currentTarget instead?

src/frontend/next/src/components/ScrollAction/ScrollAction.tsx
tonyvugithub: @humphd : We don't need to put ScrollActionProps behind the props because it's already taken care of inside FC

src/frontend/next/src/components/ScrollAction/ScrollAction.tsx
tonyvugithub: @PedroFonsecaDEV : in term of what we need here, they are pretty much doing the same thing. Interface more like about the structure of an object while type is more like an alias of a specific type. Interface can be extended with new class while type cannot. In that sense I would say type is more limited than interface. However, @humphd mentioned that in the next docs they recommend using type over interface so maybe he can answer you better here. 
More about it here: https://blog.logrocket.com/types-vs-interfaces-in-typescript/

humphd: Interfaces can be extended, and are good for exporting public types.  For internal code, where things won't change once defined, a type is good because it gives a name to a shape/type.  See https://github.com/typescript-cheatsheets/react#types-or-interfaces

chrispinkney (approved):

humphd (approved):

PR #1525, opened 2020-12-11
Port hooks to NextJS

HenryZerocool:
## Issue This PR Addresses

Fixes #1502 

## Type of Change

- [ ] **Bugfix**: Change which fixes an issue
- [ ] **New Feature**: Change which adds functionality
- [ ] **Documentation Update**: Change which improves documentation
- [ ] **UI**: Change which improves UI
- [x] **Next**: Porting to NextJS

## Description

Port hooks folder and images folder to nextJS

## Checklist

- [ ] **Quality**: This PR builds and passes our npm test and works locally
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not (if applicable)
- [ ] **Documentation**: This PR includes updated/added documentation to user exposed functionality or configuration variables are added/changed or an explanation of why it does not(if applicable)

PR #1555, opened 2021-01-18
Fix #1502: Port hooks to NextJS

### LINKED ISSUE #1502: [next] Port hooks/ to NextJS
https://github.com/Seneca-CDOT/telescope/tree/master/src/frontend/gatsby/src/hooks
Port all hooks in hooks/ to NextJS using TS
Porting reason

humphd:
## Issue This PR Addresses

Fixes #1502.

## Type of Change

## Description

This completes the work begun in https://github.com/Seneca-CDOT/telescope/pull/1525.  Some notes:

- I needed to change the way that the Logo worked.  I've moved the SVG icons to `/public` and changed the name from `LogoIcon` to `Logo`, since it now contains multiple exports (i.e., I expose the URLs to use for badged and unbadged versions of the icon).
- I switched the GitHub Contributors hook to use [useSWR](https://swr.vercel.app/) instead of doing the `fetch()` manually. I want to move to use this everywhere we do data fetching, so figured I'd do it now.
- I've moved the Gatsby site metadata to a file called `src/config.ts` per the [next.js migration instructions on site metadata](https://nextjs.org/docs/migrating/from-gatsby#site-configuration).  We don't really need the `useSiteMetadata()` hook anymore, but I kept it to make porting go faster.  When we're done getting everything ported, we can switch to importing `src/config` directly.

I've tested all the hooks manually locally.  They are a bit hard to test at the moment, since nothing uses them.  If you want to try locally, add them to `src/frontend/next/src/pages/index.tsx` and call them in the `Home` function, so you can print the result.

cc @chrispinkney, @jossiey.  Chris, I made changes to your logo, and Jossie, you will need to update your work to use the new Logo (i.e., `Logo` vs. `LogoIcon`). 

cc @manekenpix, @raygervais I need some ideas on the best way to pass `API_URL` from our top level `process.env` into the frontend of next.js.  Reading https://nextjs.org/docs/basic-features/environment-variables, it seems like you have to prefix variables with `NEXT_PUBLIC_*` in order for them to appear in the client.  I tried a few hacky ways of doing this, but didn't find something I like.  We need a clean solution so we can set the API_URL in our env and have it work in next on the browser side.


src/frontend/next/src/components/Logo.tsx
```diff
@@ -0,0 +1,14 @@
+const logoUrl = '/logo.svg';
```
chrispinkney, line 1: Ah, nice!

src/frontend/next/src/config.ts
```diff
@@ -0,0 +1,7 @@
+export default {
+  title: `Telescope`,
+  description: `A tool for tracking blogs in orbit around Seneca's open source involvement`,
+  author: `SDDS Students and professors`,
+  // TODO: we need a way to pass in process.env.API_URL...
```
chrispinkney, line 5: Can we make a .js file in the frontend folder with the `API_URL` in it and just export it? Not very elegant though.
Or something like [this](https://github.com/facebook/create-react-app/issues/4283#issuecomment-380232039), maybe? 

Did Gatsby not have to proxy `.env` variables from backend to frontend?

chrispinkney (approved):

chrispinkney:
I'd love some more feedback, my knowledge of hooks is pretty barren.

src/frontend/next/src/config.ts
humphd: We have our `.env` files in the project root.  Here we define `API_URL`, so the value is known.

From the root, we run the `develop:next` or `build:next` commands, which in turn call `npm run ...` with the appropriate command within the `src/frontend/next` directory.  So step 1 is: get the `env` passed down to the `npm run` process that spans the call to next's dev/build script.

An easier way to do this might be to have the next.js build reach up to the root and load the `.env` config when it begins the build.  I did a version of this, by adding some code to `next.config.js`, which would `require()` the `dotenv` module, then use `process.env` to copy `API_URL` to `NEXT_PUBLIC_API_URL`.  This works, but it's a bit dirty.

Next.js has a built-in way to define `env` files for dev and prod, as well as local overrides.  However, this is separate from our root config, which isn't what we want (i.e., having to configure the app in multiple places is a good way to screw something up).  Not to mention that our Docker config needs to be able to override and pass down various variables in staging and production.

So maybe we have to use my "dirty" solution, and have the next.js build load the top-level config.  If anyone has a cleaner solution, I'm all ears.

humphd: I've made a new commit to show the method I'm discussing, see https://github.com/Seneca-CDOT/telescope/pull/1555/commits/05801b488b298e26d59c63d5c0c42d698b52c958.  This works.  Is it good enough?

manekenpix (approved):
Tested using the latest changes (`next.config.js` solution, I was reading about this right when this change was pushed).

image

chrispinkney (approved):

chrispinkney:
Awesome!

Code_2021-01-18_20-04-18

PR #1561, opened 2021-01-19
[next] Fix folder structure after PR #1559

PedroFonsecaDEV:
## Issue This PR Addresses
After PR #1559 our folder structure wasn't following our new folder structure rules. 

I removed the folder `components/SearchInput/PostSearchInput.tsx`
So now we have: `components/PostSearchInput.tsx`

## Type of Change

## Checklist


HyperTHD (approved):
I'd like to discuss how we should deal with grouping of components that are similar in nature to each other and rely on each other with the SearchInput component being the main point of contention

humphd:
I think that we can use folders under `components/` when it makes sense to group things.  For example, let's say we have `<Search />` and it has a bunch of children that only get used by it (e.g., `<SearchInput />` and `<SearchResults />` etc.).  In this case, I would do:

```
components/
    Search/
        index.tsx          <-- default export the Search component from here
        SearchInput.tsx
        SearchResults.tsx
```

Not everything needs to be in the global `components/` dir at the root level.

humphd (commented):
I'm not sure we need to break this out.  It's fine to leave it grouped under SearchInput I think, since it's a child component.

I'd close this PR, unless people disagree.

chrispinkney:
I'm currently working on an issue that relies on having the SearchInput folder, as in the Gatsby frontend we have multiple components inside of the SearchInput folder that relate to each other. I think it's fine to keep a folder for these, but I do think it's strange to sometimes have folders and sometimes not. I don't think there's a perfect way to do this though, and I'm fine with the way it is now.

chrispinkney (commented):
See my comment, curious to see what others think too :)

PR #1529, opened 2020-12-12
Port BackToTopButton from Gatsby to NextJS

joelazwar:
## Issue This PR Addresses

Fixes #1480 

## Type of Change

- [ ] **Bugfix**: Change which fixes an issue
- [ ] **New Feature**: Change which adds functionality
- [ ] **Documentation Update**: Change which improves documentation
- [ ] **UI**: Change which improves UI
- [x] **Next**: Porting to NextJS

## Description

Ports BackToTopButton Component from Gatsby to NextJS.

Note: This component is dependent on the ScrollAction component (#1469)

## Checklist

- [ ] **Quality**: This PR builds and passes our npm test and works locally
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not (if applicable)
- [ ] **Documentation**: This PR includes updated/added documentation to user exposed functionality or configuration variables are added/changed or an explanation of why it does not(if applicable)

PR #1564, opened 2021-01-19
Fix #1480 Port BackToTopButton from Gatsby to NextJS

### LINKED ISSUE #1480: Port BackToTopButtom from Gatsby to Next.js(TS)
What would you like to be added:
This is an issue to port BackToTopButton from Gatsby to NextJS

Why would you like this to be added:
Part of #1316

humphd:
## Issue This PR Addresses

Fixes #1480

## Type of Change

## Description

This continues the work from #1529.

I had to do a few more things to get this working:

1. I needed to stop eslint from checking for default props, and we'll do it instead via TypeScript.  It's not possible otherwise to use default values in TS.
2. I had to rework how the colors were being done, to use classes and the `theme.palette` instead of the `color` prop, which won't work with theme names.
3. I removed the `window` param from the scroll manager, and always use it [by default](https://material-ui.com/components/app-bar/#usescrolltrigger-options-trigger).
4. I had to change how `ScrollAction` was defined for `children` to work with typings

However, when I test this, my elements have a width/height of 0, so it displays nothing.  I can't figure out what's wrong, and would appreciate some help.


tonyvugithub (approved):

humphd:
Rebased to fix conflicts.

tonyvugithub (approved):

chrispinkney (approved):

cindyledev:
If you get rid of the `<Zoom> `, the button shows up. We may need to look into the MUI Zoom core component and figure out why it doesn't seem to be working as intended here

humphd:
@chrispinkney did you file the new bug on this?  If not, please do, then add @cindyledev's note to it.

PR #1569, opened 2021-01-20
Added more node versions to GH actions for testing

c3ho:
## Type of Change

## Description
Added extra node versions (13.x and 14.x) to Github Actions. As the current node LTS is 14.15.4, we should be updating our testing to account for that. I didn't add 15.x because of previous mentions of it being problematic. 

It's actually up in the air to decide if we still want to test for 12.x and 13.x @humphd 

## Checklist

### DIFF
--- a/.github/workflows/node-js-ci.yml
+++ b/.github/workflows/node-js-ci.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        node-version: [12.x]
+        node-version: [12.x, 13.x, 14.x]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/node-js-ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        node-version: [12.x]
+        node-version: [12.x, 13.x, 14.x]
```
humphd, line 18: I think only the even numbered releases are what people should be using for LTS.  I'd suggest 12, 14.

cc @manekenpix

humphd (changes requested):

.github/workflows/node-js-ci.yml
manekenpix: I agree with @humphd 
Take a look at [this](https://nodejs.org/en/about/releases/)

image

chrispinkney (approved):
Cool, curious about this.

manekenpix (approved):

PR #1606, opened 2021-01-25
Fixes #1576, Added FRONTEND env variable, updated routes/index.js

### LINKED ISSUE #1576: Use an env var to determine which front-end gets served

We need a way to specify which front-end gets served by telescope.
Currently, the path to the folder where the front-end build lives is [hard-coded](https://github.com/Seneca-CDOT/telescope/blob/6cfd7dfce743ed80bf38a7364c5a5511e4b50001/src/backend/web/routes/index.js#L50), and it's the `Gatsby` front-end.
We could have a variable in our `env.example` (and eventually in our staging and production `env` files) to choose between the Next and the Gatsby front-end.
Something like:
```js
const { FRONTEND } = process.env;

if (FRONTEND === 'gatsby')
  router.use(express.static(path.join(__dirname, '../../../frontend/gatsby/public'))); 

else if (FRONTEND === 'next')
  router.use(express.static(path.join(__dirname, '../../../frontend/next/out'))); 

else if (FRONTEND === '...')
...

```

This will be specially useful for testing `login` locally using Next.

EDIT: path to `next` folder changed following [this](https://github.com/Seneca-CDOT/telescope/issues/1576#issuecomment-764203256) comment.

Metropass:
## Issue This PR Addresses
Closes #1576 
Created a `FRONTEND` variable for all of the .env files, and added a conditional if within `backend/routes/index.js`
Currently, the only two values that can be accepted for `FRONTEND` are: `gatsby` and `next`
`gatsby` is the default value as of this Pull Request

## Type of Change

## Checklist


env.example
```diff
@@ -4,6 +4,16 @@ NODE_ENV=development
...
 PORT=3000
 
+
+# FRONTEND is the Front-End env variable that defines what Front-End
+# gets served for Telescope.
+# Currently, FRONTEND only has two supported values:
+#   'gatsby' - The Gatsby Front-End
+#   'next' - The NextJs Front-End
+# Our default value for FRONTEND is 'gatsby'
+FRONTEND='gatsby'
```
humphd, line 14: Remove the surrounding quotes, all values are strings:

```
FRONTEND=gatsby
```

netlify.toml
```diff
@@ -8,3 +8,4 @@
   API_URL="https://dev.telescope.cdot.systems"
   # We need to expose the API_URL to the front-end too
   NEXT_PUBLIC_API_URL="https://dev.telescope.cdot.systems"
+  FRONTEND='gatsby'
```
humphd, line 11: Should be `next` on Netlify.

src/backend/web/routes/index.js
```diff
@@ -17,7 +17,7 @@ const user = require('./user');
 const query = require('./query');
 
 const router = express.Router();
-
+const { FRONTEND } = process.env.FRONTEND;
```
humphd, line 20: In the rest of this code, we seem to be doing `process.env.FRONTEND` inline, instead of pulling things out.  Can you match that below?

src/backend/web/routes/index.js
```diff
@@ -45,9 +45,15 @@ if (process.env.NODE_ENV === 'development') {
...
     // Allow proxying the Gatsby dev server through our backend if PROXY_FRONTEND=1 is set in env
     router.use('/', createProxyMiddleware({ target: 'http://localhost:8000', changeOrigin: true }));
-  } else {
     // Or serve the static files in the Gatsby build directory
+  } else if (FRONTEND === 'gatsby') {
     router.use(express.static(path.join(__dirname, '../../../frontend/gatsby/public')));
+  } else if (FRONTEND === 'next') {
+    router.use(express.static(path.join(__dirname, '../../../frontend/next/out')));
+  } else {
+    throw new Error(
```
humphd, line 54: Let's test that this does what we want (e.g., we should probably crash right now).

src/backend/web/routes/index.js
```diff
@@ -45,9 +45,15 @@ if (process.env.NODE_ENV === 'development') {
   if (process.env.PROXY_FRONTEND) {
     // Allow proxying the Gatsby dev server through our backend if PROXY_FRONTEND=1 is set in env
     router.use('/', createProxyMiddleware({ target: 'http://localhost:8000', changeOrigin: true }));
-  } else {
     // Or serve the static files in the Gatsby build directory
+  } else if (FRONTEND === 'gatsby') {
```
humphd, line 49: Let's make sure that case doesn't affect these (e.g., `.toLowerCase()` on the frontend values we compare against).

humphd (changes requested):

env.example
Metropass: Will do, thanks!

src/backend/web/routes/index.js
manekenpix: Careful here: using [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#object_destructuring), you have to remove the name of the variable in the rhs.

```ts
const { FRONTEND } = process.env;
```

Metropass: Will do! Everything should be fixed along with the rest of your suggestions

src/backend/web/routes/index.js
```diff
@@ -17,7 +17,6 @@ const user = require('./user');
 const query = require('./query');
 
 const router = express.Router();
-
```
humphd, line 20: Watch out for spurious whitespace changes like this, as it churns git history unnecessarily.

humphd (changes requested):
Looks good.  @Metropass, please update the team tomorrow about this change so people can update their local .env, since it will throw right now if it's missing.

Question related to that: should we default to Gatsby if it's missing instead of throwing?  That is:

- gatsby: use gatsby
- next: use next
- {not set}: use gatsby
- anything else: throw

HyperTHD:
> Looks good. @Metropass, please update the team tomorrow about this change so people can update their local .env, since it will throw right now if it's missing.
> 
> Question related to that: should we default to Gatsby if it's missing instead of throwing? That is:
> 
> * gatsby: use gatsby
> * next: use next
> * {not set}: use gatsby
> * anything else: throw

I say for now, it should default to gatsby since the next transition is not complete yet. Whenever that's complete, unless we have any sort of reason to keep gatsby, we can just have it serve next, and throw if anything else other than that is used

src/backend/web/routes/index.js
```diff
@@ -45,13 +45,13 @@ if (process.env.NODE_ENV === 'development') {
     // Allow proxying the Gatsby dev server through our backend if PROXY_FRONTEND=1 is set in env
     router.use('/', createProxyMiddleware({ target: 'http://localhost:8000', changeOrigin: true }));
     // Or serve the static files in the Gatsby build directory
-  } else if (process.env.FRONTEND.toLowerCase() === 'gatsby') {
+  } else if (process.env.FRONTEND.toLowerCase() === 'gatsby' || process.env.FRONTEND === '') {
```
humphd, line 48: Question: will `process.env.FRONTEND.toLowerCase()` break if `FRONTEND` isn't defined?  That is, should this be:

```
else if(!process.env.FRONTEND || process.env.FRONTEND === '' || process.env.FRONTEND.toLowerCase() === 'gatsby') {...}
```

Metropass: Oh, that's a good point, I'll make that change real quick

manekenpix:
is this ready for review?

Metropass:
@manekenpix Should be ready now!

chrispinkney:
@Metropass This might be a stupid question but how can I test this? Also does this mean I have to copy a new example `.env` file after this merges?

env.example
```diff
@@ -4,6 +4,16 @@ NODE_ENV=development
 # PORT is the port used by the web server
 PORT=3000
```
chrispinkney, line 5: Does [this](https://github.com/Seneca-CDOT/telescope/pull/1649/files#diff-657de79094f16fa5feacfb48eb9d358b1060fb9b7d3256138b7f32b9c8c0fb22R8) change affect this?

Metropass:
> 
> 
> @Metropass This might be a stupid question but how can I test this? Also does this mean I have to copy a new example `.env` file after this merges?

It's not a stupid question at all!
Just 

> 
> 
> @Metropass This might be a stupid question but how can I test this? Also does this mean I have to copy a new example `.env` file after this merges?

It's not a stupid question at all!
Just have the FRONTEND variable set to which ever Front-End service you'd like to run `eg, FRONTEND=gatsby`, and `index.js` should take care of the rest.

env.example
Metropass: It does, thanks for bring it up. 
It seems like I need to make a conditional statement within the package.json
@humphd cc'ing just for clarity, Are we allowed to make conditionals in our packages for Telescope?

manekenpix: I don't think they're related. `PORT=3000` is for telescope (telescope serving the back end), The change @chrispinkney is talking about ([this](https://github.com/Seneca-CDOT/telescope/pull/1649/files#diff-657de79094f16fa5feacfb48eb9d358b1060fb9b7d3256138b7f32b9c8c0fb22R8)) is for next's production server (a separate server to serve the production build, but telescope has nothing to do with it).

env.example
```diff
@@ -4,6 +4,16 @@ NODE_ENV=development
...
```
chrispinkney, line 6: Haha sorry, I know it's pedantic, but can you get rid of the extra newlines on line 6 and 15?

chrispinkney (approved):
Otherwise LGTM!

HyperTHD (changes requested):
This PR looks okay. We're switching to next and not gatsby so we can make this exclusively next instead. You'll need to rebase and fix conflicts here too.

chrispinkney:
> 
> 
> This PR looks okay. We're switching to next and not gatsby so we can make this exclusively next instead. You'll need to rebase and fix conflicts here too.
> 
> Read #1733 for context

Hm good point. In fact AFAIK we'll be removing Gatsby entirely when Next finally lands. Not sure! Something to think about I guess @Metropass

humphd:
This was started nearly a month ago, and we needed it then, but at this point, I don't think it's useful, since we're moving to a single front-end.  Closing.

PR #1527, opened 2020-12-12
automate release

yzwdroid:
## Type of Change

- [x] **Bugfix**: Change which fixes an issue
- [ ] **New Feature**: Change which adds functionality
- [ ] **Documentation Update**: Change which improves documentation
- [ ] **UI**: Change which improves UI

## Description
Automate the release process by github action.  To release a new version, we just add a tag when we push or merge a pr. 

- only the commits follow the pattern will be added to the release note.
[https://github.com/lob/generate-changelog#usage](https://github.com/lob/generate-changelog#usage)
```
type(category): description [flags]

Where type is one of the following:

    breaking
    build
    ci
    chore
    docs
    feat
    fix
    other
    perf
    refactor
    revert
    style
    test
```

Screenshot_20201211_234114

## Checklist

- [x] **Quality**: This PR builds and passes our npm test and works locally
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [x] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not (if applicable)
- [ ] **Documentation**: This PR includes updated/added documentation to user exposed functionality or configuration variables are added/changed or an explanation of why it does not(if applicable)

PR #1628, opened 2021-02-01
[CI/CD] Fixes #1301: Created automated release workflow

### LINKED ISSUE #1301: Add github action to automate the release process.
I'd like to add a GitHub action to automate the release process. When there is a release tag like `1.5.0` push to the master branch, a release will be created with the release code and description.

automate the release process.

birtony:
## Issue This PR Addresses
Fixes #1301

## Type of Change

## Description
Rebased and updated https://github.com/Seneca-CDOT/telescope/pull/1527. This PR adds an automated release workflow for GitHub Actions. It would allow us to create a new release by running [`npm version`](https://docs.npmjs.com/cli/v6/commands/npm-version) (**DON'T RUN THIS CODE UNLESS YOU WANT TO MAKE A PRODUCTION RELEASE**):
```
npm version minor -m "Release 1.6.0" 
git push upstream master
git push upstream 1.6.0
```
The workflow would also run the same checks as our `node-js-ci` workflow (test, lint, prettier) and generate a changelog to be included in the release.

- commits that follow the pattern will be added to the corresponding sections in the release note. All other commits will be added under 'Other Changes` section.
[https://github.com/lob/generate-changelog#usage](https://github.com/lob/generate-changelog#usage)
```
type(category): description [flags]

Where type is one of the following:

    breaking
    build
    ci
    chore
    docs
    feat
    fix
    other
    perf
    refactor
    revert
    style
    test
```

### [Example of Generated Changelog](https://github.com/birtony/telescope/releases/tag/1.5.1):
image

### [Example of Workflow Status in GitHub Actions](https://github.com/birtony/telescope/actions/runs/528534877):
image
image

### How to Test
Create some releases **on your forked telescope repository**! 

- You can test the workflow by merging it into your forked repository and pushing a new tag to it:
> `git merge upstream/issue1301-create-release-workflow`
> `npm version minor -m "Release 1.6.0"`
> `it push upstream master`
> `git push upstream 1.6.0`

- Using [GitHub CLI](https://cli.github.com/) might make it easier to checkout to `issue1301-create-release-workflow` directly by executing the following:
> `gh pr checkout 1628`

- You can also create a number of dummy commits in your forked repository before creating a new release to see how the changelog is generated. 

- We currently use [a fairly simple regex to check the pattern of the semantic version](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet). [The recommended regex is outlined on semver repository](https://github.com/semver/semver/blob/master/semver.md#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string), but it cannot be used in YAML files ([some special characters might not be supported in YAML, such as the colon (`:`), for instance](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet)).

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,34 @@
+on:
+  push:
+    # Sequence of patterns matched against refs/tags
+    tags:
+      - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
+
+name: Create Release
+
+jobs:
+  build:
+    name: Create Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Changelog
+        uses: scottbrenner/generate-changelog-action@master
+        id: Changelog
+        env:
+          REPO: ${{ github.repository }}
+      - name: Create Release
+        id: create_release
+        uses: actions/create-release@latest
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # This token is provided by Actions, you do not need to create your own token
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: Release ${{ github.ref }}
+          body: |
+            ${{ steps.Changelog.outputs.changelog }}
+          draft: false
+          prerelease: false
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,7 +4,7 @@ on:
   push:
     # Sequence of patterns matched against refs/tags
     tags:
-      - '[0-9].[0-9]+.[0-9]+' # Push events to any matching semantic tag. For example, 1.10.1 or 2.0.0.
+      - '[0-9]+.[0-9]+.[0-9]+' # Push events to any matching semantic tag. For example, 1.10.1 or 2.0.0.
         # For more details, see https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet
         # and https://docs.npmjs.com/about-semantic-versioning
 
@@ -60,9 +60,6 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
-        # Update version in package.json according to the tag
-      - name: Check package version
-        uses: technote-space/package-version-check-action@v1
       - name: Changelog
         uses: scottbrenner/generate-changelog-action@master
         id: Changelog
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -53,6 +53,7 @@ jobs:
       - uses: actions/setup-node@v1
       - run: npm install && npm run eslint
   release:
+    needs: [build, Prettier-Check, Lint]
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,20 +1,67 @@
+name: Create Release
+
 on:
   push:
     # Sequence of patterns matched against refs/tags
     tags:
-      - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
-
-name: Create Release
+      - '[0-9].[0-9]+.[0-9]+' # Push events to any matching semantic tag. For example, 1.10.1 or 2.0.0.
+        # For more details, see https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet
+        # and https://docs.npmjs.com/about-semantic-versioning
 
 jobs:
   build:
-    name: Create Release
+    runs-on: ${{matrix.os}}
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        node-version: [12.x, 14.x]
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Cache node modules
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-node-modules
+        with:
+          # npm cache files are stored in `~/.npm` on Linux/macOS
+          path: ~/.npm
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package.json') }}
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+
+      - name: Test
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - run: npm install
+      - run: cp env.example .env
+      - run: npm run test
+  Prettier-Check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+      - run: npm install && npm run prettier-check
+  Lint:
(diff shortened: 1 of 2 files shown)


birtony:
Currently, GitHub does not seem to provide an easy way to postpone the execution of one workflow until the other one succeeds. In this PR I have used https://github.com/lewagon/wait-on-check-action to try to integrate a workaround. However, I could not achieve it work properly for every case. In addition, it does not seem to be a too efficient solution as it [makes a call to check the status of the other workflow](https://docs.github.com/en/rest/reference/checks#get-a-check-run) every time the `interval` expires. It would be great not to have to run rest/lint/prettier on the exact same version of code twice, but it looks like it might be the best solution for now. Please, feel free to share your opinions below.

CC @manekenpix @humphd @PedroFonsecaDEV @HyperTHD

humphd:
I would do the simple thing and just duplicate the lint/test steps in the release. It doesn't matter that it's duplicated.  Passing the tests twice is more proof that production isn't going to get code that can't be built.

birtony:
> I would do the simple thing and just duplicate the lint/test steps in the release. It doesn't matter that it's duplicated. Passing the tests twice is more proof that production isn't going to get code that can't be built.

Yeah, that makes sense. I will proceed to incorporate that logic into this PR.

chrispinkney (approved):
I think this is good, or at least it seems [like it's functioning as expected](https://github.com/chrispinkney/telescope/releases/tag/1.6.0).

🤤 🤤 🤤 🤤

.github/workflows/release.yml
```diff
@@ -0,0 +1,82 @@
+name: Create Release
+
+on:
+  push:
+    # Sequence of patterns matched against refs/tags
+    tags:
+      - '[0-9].[0-9]+.[0-9]+' # Push events to any matching semantic tag. For example, 1.10.1 or 2.0.0.
```
humphd, line 7: We have a tag `1.0.0-alpha` that wouldn't work with this.  What about:

```
- '[0-9]+.[0-9]+.[0-9]*?''
```

Not sure if that's enough, but I'm trying to add things like `-alpha` on the end.

.github/workflows/release.yml
```diff
@@ -0,0 +1,82 @@
...
+      - uses: actions/setup-node@v1
+      - run: npm install && npm run eslint
+  release:
+    needs: [build, Prettier-Check, Lint]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
humphd, line 62: Should we do this on 23 above too?

.github/workflows/release.yml
```diff
@@ -0,0 +1,82 @@
...
+        # Update version in package.json according to the tag
+      - name: Check package version
+        uses: technote-space/package-version-check-action@v1
```
humphd, line 65: I think we should have people do this with `npm version` locally, so they can update package.json and also tag before they push.  The the tag will match the commit that bumps the package version (vs. adding a new commit afterward).

.github/workflows/release.yml
```diff
@@ -0,0 +1,82 @@
...
+      - name: Changelog
+        uses: scottbrenner/generate-changelog-action@master
```
humphd, line 67: I like the idea of a beautiful changelog, but I think it's unrealistic with students who struggle to use git perfectly.  Can we not just use the entire git history of the branch unmodified?

humphd (changes requested):

.github/workflows/release.yml
birtony: If commits do not follow the format, they would still be included in the `Other Changes` section:
image
Would you still prefer not to add any sections to our changelog? @humphd

humphd: Oh, that's perfect!

.github/workflows/release.yml
birtony: You're right, the package.json update commit would now be the next after the tagged one:
image
I will try to see if we can do anything about it with GitHub Actions and switch to `npm release` in case we cannot.

.github/workflows/release.yml
birtony: That's a good question. I will need to do some research to answer this

.github/workflows/release.yml
birtony: True. It's a pity GitHub does not allegedly support full regex syntax. I will try to use the regex specified under server docs first to see if it might actually work. If not, we should at least modify it slightly

birtony: As suspected, [the recommended regex](https://github.com/semver/semver/blob/master/semver.md#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string) is not supported in YAML. How about we use `[0-9]+.[0-9]+.[0-9]+-?[a-z]+`? @humphd

.github/workflows/release.yml
birtony: @humphd so after looking up the docs the answer seems to be no. [`fetch-depth: 0`](https://github.com/actions/checkout/blob/main/README.md#checkout-v2) is needed for the `checkout` action to fetch all history for all branches and tags. In the `release` job we want that because we need to generate a changelog of everything that happened to the codebase since the last tag. On line 23 we want to fetch the latest version of code and run tests on it, no need to run tests on every version of code that existed since the last tag. 

P.S. Took a look at [this PR](https://github.com/actions/checkout/pull/258) and t[his issue](https://github.com/actions/checkout/issues/204) to get some context about the `fetch-depth`

.github/workflows/release.yml
birtony: Apparently, neither `'[0-9]+.[0-9]+.[0-9]*?'` nor `'[0-9]+.[0-9]+.[0-9]+-?[a-z]+'` are correct. Regex works very crooky in YAML. Needs to investigate more... @humphd

birtony: For some reason `[0-9]+.[0-9]+.[0-9]+-?[a-z]*` [does not get triggered by tag like `1.5.4`](https://github.com/birtony/telescope/releases/tag/1.5.4). It seems to be a valid expression in YAML though, as [it works for tags like `1.5.4-alpha`](https://github.com/birtony/telescope/releases/tag/1.5.4-alpha). @humphd any idea why is it happening?

humphd: I think we could start with a simple check for x.y.z style semver, and file an issue to fix this later.  It doesn't need to block 1.6.

humphd: No idea.  Let's do the most simple check possible on the tag and get this in, then refine it in follow-ups.

birtony: Agreed.

.github/workflows/release.yml
birtony: Could not find a way to match the tagged commit with the one that updates the version in `package.json` with GitHub Actions. Will get rid of `technote-space/package-version-check-action@v1` and switch to use `npm version` locally instead.

birtony:
Updated the workflow to remove the package version bump action. We should use `npm version` to generate a release as described in the updated description to this pr.

humphd (approved):
Do you want to add `preversion` to our `package.json` script to run `npm test`?  I think that's a logical approach.

HyperTHD (approved):
Perfection
LGTM. Can't wait to see this in action later this week

birtony:
@humphd @HyperTHD agreed. Let land this one first and take care of that in a subsequent PR. I will take care of it.

PR #1631, opened 2021-02-01
Fix #1629: improve next.js auth handling

### LINKED ISSUE #1629: Improve our next.js authentication handling

We should improve our user context and auth handling to match recommendations in https://nextjs.org/docs/authentication and https://github.com/vercel/next.js/tree/canary/examples/with-iron-session.

Deep linking into authenticated pages doesn't currently work, and I think it should be possible.  See /myfeeds page as an example, and going directly to it (first render with no user info yet) vs. via the home page (user info is fetched by the time we render).

humphd:
## Issue This PR Addresses

Fixes #1629

## Type of Change

## Description

This follows the suggestions of the next.js docs on doing statically rendered site authentication.  I've also extended it a bit to add the concept of a user vs. an admin.

First, we have 4 types that a "user" can be at runtime:

- `null` - we either don't know anything about them, or don't have the info (yet)
- `User` - we know whether they are logged in or not, nothing more (e.g., we don't have their user info yet)
- `AuthenticatedUser` - we know that they are logged in for sure, and their name, email, id
- `AdminUser` - we know that they are authenticated AND that they are an admin

In components/pages, you decide which one you want to use, based on your needs.  So if you only care about whether or not the user is logged in, you use `useUser()`:

```
const user = useUser();

// User might be null
if(!user) { 
  return null;
}

return user.isLoggedIn ? 'Logged In' : 'Not Logged In';
```

If you need to use their username or email, then you need `useAuthenticatedUser()` instead:

```
const authenticated = useAuthenticatedUser();

// User might be null
if(!authenticated) { 
  return null;
}

return authenticated.name;
```

Same idea for admin with `useAdminUser()`.

For `useAuthenticatedUser()` and useAdminUser()` you can also pass `true` to them to have the app redirect on failure to the login page.  This is useful for pages that need to have an authenticated user or admin user.

I've re-written the `/myfeeds` page to use this logic, and only render it once we have an Authenticated user.  To allow for passing the authenticated user down, I've refactored it a bit to move the body of the page into its own component that receives the user on props.

### DIFF
--- a/src/frontend/next/src/components/UserProvider.tsx
+++ b/src/frontend/next/src/components/UserProvider.tsx
@@ -1,9 +1,11 @@
 import { createContext, ReactNode, useContext } from 'react';
 
-import { User } from '../interfaces';
+import { User, AuthenticatedUser, AdminUser } from '../interfaces';
 import useAuthorization from '../hooks/use-authorization';
+import config from '../config';
 
-const UserContext = createContext<User>({ isLoggedIn: false });
+const { loginUrl } = config;
+const UserContext = createContext<null | User | AuthenticatedUser | AdminUser>(null);
 
 type Props = {
   children: ReactNode;
@@ -16,4 +18,63 @@ const UserProvider = ({ children }: Props) => {
 
 export default UserProvider;
 
-export const useUser = () => useContext(UserContext);
+/**
+ * User Hooks
+ * ----------
+ *
+ * We provide three different user hooks, which work in a similar way,
+ * but are meant to be used in different circumstances.
+ *
+ * useUser() - when all you care about is if the user is logged in or not.
+ *
+ * useAuthenticatedUser(shouldRedirect) - when you need an authenticated user.
+ * You will get null or an AuthenticatedUser in response.  If shouldRedirect
+ * is true, it will redirect to the Telescope login page.
+ *
+ * useAdminUser(shouldRedirect) - when you need an authenticated admin user.
+ * You will get null or an AdminUser in response.  If shouldRedirect
+ * is true, it will redirect to the Telescope login page.
+ */
+export const useUser = (): null | User => useContext(UserContext);
+
+export const useAuthenticatedUser = (shouldRedirect = false): null | AuthenticatedUser => {
+  const user = useUser();
+
+  // Wait until we have user info from the server
+  if (!user) {
+    return null;
+  }
+
+  if (!user.isLoggedIn) {
+    if (shouldRedirect) {
+      window.location.href = loginUrl;
+    }
+    // Caller wants an authenticated user, which we don't have,
+    // so return nothing.
+    return null;
+  }
+
+  // User's authentication status is known, return that
+  return user as AuthenticatedUser;
+};
+
+export const useAdminUser = (shouldRedirect = false): null | AdminUser => {
+  const user = useAuthenticatedUser();
+
+  // Wait until we have user info from the server
+  if (!user) {
+    return null;
+  }
+
+  if (!user.isAdmin) {
+    if (shouldRedirect) {
+      window.location.href = loginUrl;
+    }
+    // Caller wants an admin user, which we don't have,
+    // so return nothing.
+    return null;
+  }
+
+  // User's authentication status is known, return that
+  return user as AdminUser;
+};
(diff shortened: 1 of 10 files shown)


src/frontend/next/src/config.ts
```diff
@@ -1,9 +1,15 @@
...
   description: `A tool for tracking blogs in orbit around Seneca's open source involvement`,
   author: `SDDS Students and professors`,
-  // This comes via the top level .env and its API_URL value,
-  // and gets set in next.config.js at build time.
-  telescopeUrl: process.env.NEXT_PUBLIC_API_URL,
+  telescopeUrl,
+  loginUrl: `${telescopeUrl}/auth/login`,
+  logoutUrl: `${telescopeUrl}/auth/logout`,
+  userFeedsUrl: `${telescopeUrl}/user/feeds`,
+  feedsUrl: `${telescopeUrl}/feeds`,
```
birtony, line 13: Good stuff!

src/frontend/next/src/components/UserProvider.tsx
```diff
@@ -16,4 +18,63 @@ const UserProvider = ({ children }: Props) => {
...
+  const user = useUser();
+
+  // Wait until we have user info from the server
+  if (!user) {
+    return null;
+  }
+
+  if (!user.isLoggedIn) {
+    if (shouldRedirect) {
+      window.location.href = loginUrl;
```
birtony, line 50: That should be very useful

src/frontend/next/src/pages/myfeeds.tsx
```diff
@@ -1,208 +1,30 @@
...
     <>
       <Head>
         <title>My Feeds</title>
       </Head>
-      {renderMyFeeds()}
+
+      <Banner />
+      <BackToTopButton />
+      <main className="main">
+        <MyFeeds user={authenticated} />
```
birtony, line 24: This looks better than the function we used to have

birtony (dismissed):

yuanLeeMidori:
@humphd I can't get the admin userinfo in the AdminButton component.

image

humphd:
@birtony and others reviewing this, the easiest way to test this:

1. run the login, redis, elastic servers via docker
2. set `PROXY_FRONTEND=1` in your .env
3. `cd src/frontend/next` and then `npm run dev` to start the frontend in dev mode
4. start the backend `npm start`
5. Modify src/frontend/next/src/pages/index.tsx to include the Login component:

```tsx
import Head from 'next/head';
import BackToTopButton from '../components/BackToTopButton';

import Login from '../components/Login';  // <--- add this

const Home = () => {
  return (
    <>
      <Head>
        <title>Telescope</title>
        <meta property="og:title" content="Telescope" key="title" />
      </Head>

      <Banner />
      <BackToTopButton />
      <main className="main">
        <Login />                       // <---- add this
        <Posts />
      </main>
    </>
  );
};

export default Home;
```

Now go to `http://localhost:3000/`.  You should see a Login button on the right.  Try going to `/myfeeds` in the address bar.  It should redirect you to the login page.  Login as username=user1 and password=user1pass.  You should go back to the home page and see the login button shows your name.

You should also see the admin buttons on the blog posts.  Try going to `/myfeeds` again, and it should load.

humphd:
@yuanLeeMidori here's what I see when I do that:

login

src/frontend/next/src/components/UserProvider.tsx
```diff
@@ -16,4 +18,63 @@ const UserProvider = ({ children }: Props) => {
...
+    }
+    // Caller wants an admin user, which we don't have,
+    // so return nothing.
+    return null;
+  }
+
+  // User's authentication status is known, return that
+  return user as AdminUser;
```
HyperTHD, line 79: Since an AdminUser is basically an Authenticated user with an isAdmin value of true, wouldn't it make sense to just have one function but check that isAdmin value? You could remove the duplicate function by just adding an AdminUser to the list of return types here.

humphd: I could.  The reason I thought to split it out into a 3rd function is that it lets you signal intent in the components that use it:  Consider:

```
// if we only have useAuthenticatedUser()
const user = useAuthenticatedUser();  // you can't tell what I want here
if(user?.isAdmin) {
  // user is admin
}

// if we also have useAdminUser()
const user = useAdminUser; // indicating that I want an admin
if(user) {
  // user is admin
}
```

yuanLeeMidori:
@humphd thanks for the update, I'll need to see what's wrong on my end. Thanks!

yuanLeeMidori (approved):

birtony (approved):

Issue #1660, opened 2021-02-05
Fix inconsistent theming colors throughout all pages


tonyvugithub:
**What happened**:
The theme colors are not applied to all pages
**What should have happened**:
It should be consistent.

PedroFonsecaDEV:
@tonyvugithub, let's block this to work on the new theme colors (no more gatsby UI, let's work on the new UI).
I will post the colors on GitHub.

PR #1674, opened 2021-02-09
Fixed #1641: Moved ToggleThemeButton to Header and persist preferred theme value to localStorage

### LINKED ISSUE #1641: Put the toggle theme button in the header and serialize the prefered theme value to localStorage with hook
We need to think of a better way to integrate the theme switching option. Maybe a switch button in the header? Also we need to think of a way to retain the chosen theme so when users revisit the site, the same colors appear to them.
This will provide users option to choose their preferred theme and increase their engagement to the app. 
Reference for dark theme can be found #1637

tonyvugithub:
## Issue This PR Addresses
Fixed #1641 

## Type of Change

## Description

- Move ToggleThemeButton to Header
- Persist preferred theme value to localStorage

image
image

image
image

## Checklist

### DIFF
--- a/src/frontend/next/src/pages/_app.tsx
+++ b/src/frontend/next/src/pages/_app.tsx
@@ -1,17 +1,26 @@
 import { useEffect, useState } from 'react';
 import { AppProps } from 'next/app';
 import { ThemeProvider } from '@material-ui/core/styles';
+
 import Header from '../components/Header';
 import UserProvider from '../components/UserProvider';
 import '@fontsource/roboto';
 import '@fontsource/spartan';
 
-import '../styles/globals.css';
 import { darkTheme, lightTheme } from '../theme';
+import usePreferredTheme from '../hooks/use-preferred-theme';
+import { ThemeContext } from '../components/ThemeProvider';
+
+import '../styles/globals.css';
 
 // Reference: https://github.com/mui-org/material-ui/blob/master/examples/nextjs/pages/_app.js
 const App = ({ Component, pageProps }: AppProps) => {
-  // This hook is for ensuring the styling is sync between client and server
+  // Use the preferred theme for this user and the browser (one of 'dark' or 'light').
+  const [preferredTheme, setPreferredTheme] = usePreferredTheme();
+  // Set our initial theme to be whatever the preferred theme is, or the light theme if no preference,
+  const [theme, setTheme] = useState(preferredTheme === 'dark' ? darkTheme : lightTheme);
+
+  // This hook is for ensuring the styling is in sync between client and server
   useEffect(() => {
     // Remove the server-side injected CSS.
     const jssStyles = document.querySelector('#jss-server-side');
@@ -20,23 +29,26 @@ const App = ({ Component, pageProps }: AppProps) => {
     }
   }, []);
 
-  const [theme, setTheme] = useState(lightTheme);
-
+  // Switch the active theme, and also store it for next load
   const toggleTheme = () => {
-    if (theme.palette.type === 'light') {
+    if (theme === lightTheme) {
       setTheme(darkTheme);
+      setPreferredTheme('dark');
     } else {
       setTheme(lightTheme);
+      setPreferredTheme('light');
     }
   };
 
   return (
-    <ThemeProvider theme={theme}>
-      <UserProvider>
-        <Header />
-        <Component {...pageProps} theme={theme} toggleTheme={toggleTheme} />
-      </UserProvider>
-    </ThemeProvider>
+    <ThemeContext.Provider value={{ theme, themeName: theme.palette.type, toggleTheme }}>
+      <ThemeProvider theme={theme}>
+        <UserProvider>
+          <Header />
+          <Component {...pageProps} />
+        </UserProvider>
+      </ThemeProvider>
+    </ThemeContext.Provider>
   );
 };
 
--- a/src/frontend/next/src/components/Header/DesktopHeader.tsx
+++ b/src/frontend/next/src/components/Header/DesktopHeader.tsx
@@ -5,11 +5,12 @@ import SearchIcon from '@material-ui/icons/Search';
 
 import Logo from '../Logo';
 import Login from '../Login';
+import ThemeToggleButton from '../ThemeToggleButton';
 
 const useStyles = makeStyles((theme) => ({
   root: {
     flexGrow: 1,
-    backgroundColor: 'primary',
+    backgroundColor: theme.palette.primary.main,
     justifyContent: 'center',
     height: '7.8em',
   },
@@ -72,6 +73,7 @@ export default function DesktopHeader() {
             </Button>
           </Link>
           <Login />
+          <ThemeToggleButton />
         </Toolbar>
       </AppBar>
     </>
--- a/src/frontend/next/src/components/Header/MobileHeader.tsx
+++ b/src/frontend/next/src/components/Header/MobileHeader.tsx
@@ -7,11 +7,12 @@ import SearchIcon from '@material-ui/icons/Search';
 
 import Logo from '../Logo';
 import Login from '../Login';
+import ThemeToggleButton from '../ThemeToggleButton';
 
 const useStyles = makeStyles((theme) => ({
   root: {
     flexGrow: 1,
-    backgroundColor: 'primary',
+    backgroundColor: theme.palette.primary.main,
     justifyContent: 'center',
     height: '7em',
     [theme.breakpoints.down(1065)]: {
@@ -27,7 +28,7 @@ const useStyles = makeStyles((theme) => ({
   },
   searchIcon: {
     fontSize: '2.5rem',
-    color: theme.palette.background.default,
+    color: theme.palette.primary.contrastText,
   },
   menuIcon: {
     fontSize: '2.5rem',
@@ -56,7 +57,6 @@ const useStyles = makeStyles((theme) => ({
     backgroundColor: theme.palette.background.default,
   },
   item: {
-    color: theme.palette.background.default,
     fontFamily: 'Roboto, sans-serif',
     textDecoration: 'none',
     fontSize: '1.5rem',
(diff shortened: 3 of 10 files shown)


src/frontend/next/src/hooks/use-prefer-theme-colors.ts
```diff
@@ -0,0 +1,33 @@
...
+    // If there is no value store in local storage then refer to the prefer-color-scheme
+    const osPreferredTheme = isDarkThemePreferred ? 'dark' : 'light';
+
+    // Set the theme value to localStorage for the first time
+    localStorage.setItem('theme', osPreferredTheme);
+    return preferredTheme;
+  }
+
+  // If run on server, return null;
+  return null;
```
humphd, line 32: Why not return 'light' by default?

src/frontend/next/src/hooks/use-prefer-theme-colors.ts
```diff
@@ -0,0 +1,33 @@
...
+  const isDarkThemePreferred = useMediaQuery('(prefers-color-scheme: dark)');
+  const preferredTheme = useMemo(() => (isDarkThemePreferred ? darkTheme : lightTheme), [
+    isDarkThemePreferred,
+  ]);
+
+  // This portion of code only run on client
+  if (typeof window !== 'undefined') {
+    // First give priority to value stored in localStorage
+    const storedSelectedTheme = localStorage.getItem('theme');
+    if (storedSelectedTheme) {
```
humphd, line 16: This if isn't necessary

src/frontend/next/src/hooks/use-prefer-theme-colors.ts
```diff
@@ -0,0 +1,33 @@
...
+        return darkTheme;
+      }
+      return lightTheme;
+    }
+
+    // If there is no value store in local storage then refer to the prefer-color-scheme
+    const osPreferredTheme = isDarkThemePreferred ? 'dark' : 'light';
+
+    // Set the theme value to localStorage for the first time
+    localStorage.setItem('theme', osPreferredTheme);
```
humphd, line 27: What will this do on the server?

src/frontend/next/src/contexts/themeContext.ts
```diff
@@ -0,0 +1,8 @@
+import { createContext } from 'react';
+
+const ThemeContext = createContext({
+  themeType: 'light',
+  toggleTheme: () => {},
```
humphd, line 5: Why not provide a way to switch?  What is the point of this?

src/frontend/next/src/pages/_app.tsx
```diff
@@ -18,24 +20,44 @@ const App = ({ Component, pageProps }: AppProps) => {
     }
   }, []);
 
-  const [theme, setTheme] = useState(lightTheme);
+  const preferredTheme = usePreferredThemeColors();
+
+  const [theme, setTheme] = useState(preferredTheme || lightTheme);
```
humphd, line 25: If you always return one of `'light'` or `'dark'` this null check goes away.

humphd (changes requested):

src/frontend/next/src/contexts/themeContext.ts
tonyvugithub: @humphd : the `ThemeToggleButton` uses `themeType` for conditional rendering the icon and use the `toggleTheme` to switch the value. I just don't want to pass down to 3 levels of components from App -> Header -> DesktopHeader/MobileHeader -> ThemeToggleButton. So I create a context to simply use it in the` 
ThemeToggleButton`

src/frontend/next/src/hooks/use-prefer-theme-colors.ts
tonyvugithub: @humphd : On server there is no localStorage, but this hook still get compile, so this is to avoid what should be running on client not to run on server.

src/frontend/next/src/hooks/use-prefer-theme-colors.ts
tonyvugithub: @humphd : I was having multiple thoughts, but you are right, I can return 'light' here

src/frontend/next/src/hooks/use-prefer-theme-colors.ts
tonyvugithub: thats why I have `if (typeof window !== 'undefined')` to check whether we are on server or not

foxviet:
Awesome! But I have some suggestions. Color for links & visited links should be more visible. Also the toggle button I would move it to the very right of the nav, in case when reader scrolled down let's say 10 pages, they wanna change the theme and... hit the 'search' button by mistake, then they have to hit 'back' OR hit 'home' and scrolldown again to get to the page 10 
+ some extra space to the login button.

src/frontend/next/src/hooks/use-prefer-theme-colors.ts
tonyvugithub: @humphd : My idea is whatever the value stored in localStorage should take priority. If it is the first time the user loaded the app, there is nothing in localStorage then we look at the `prefer-color-scheme` and then store that value to localStorage as initial value

foxviet:
It also not working properly on search page

tonyvugithub:
@foxviet : What you mean by not working on search page? Can you provide me another screenshot to demonstrate it? Thanks

foxviet:
@tonyvugithub here's what I see

tonyvugithub:
@foxviet this is related to the styling of the searchPage and not really in the scope of this PR

tonyvugithub:
@foxviet : This will be addressed in this issue https://github.com/Seneca-CDOT/telescope/issues/1660

src/frontend/next/src/contexts/themeContext.ts
humphd: I understand the use of the context, I mean something different.  Above you have code like this:

```jsx
<IconButton onClick={toggleTheme} className={classes.themeToggleButton}>
```

It's using the `toggleTheme` function you define here, but that function is doing nothing.  Why does it not literally toggle the theme?

src/frontend/next/src/hooks/use-prefer-theme-colors.ts
```diff
@@ -0,0 +1,33 @@
...
+    return preferredTheme;
+  }
+
+  // If run on server, return null;
```
humphd, line 31: Update this comment to reflect new code.

src/frontend/next/src/pages/_app.tsx
```diff
@@ -18,24 +20,44 @@ const App = ({ Component, pageProps }: AppProps) => {
     }
   }, []);
 
-  const [theme, setTheme] = useState(lightTheme);
+  const preferredTheme = usePreferredThemeColors();
+
+  const [theme, setTheme] = useState(preferredTheme);
```
humphd, line 25: You shouldn't need to set this via state if you have it from the hook above.

src/frontend/next/src/pages/_app.tsx
```diff
@@ -18,24 +20,44 @@ const App = ({ Component, pageProps }: AppProps) => {
...
+  const [mounted, setMounted] = useState(false);
+
+  // This hook is to ensure that right after the component mounted aka on client, display the client version of the app
+  useEffect(() => {
+    setMounted(true);
+  }, []);
 
   const toggleTheme = () => {
```
humphd, line 33: Do this in your context code, where I'm commenting about that empty function, then you don't need to do this here, and you can expose your hook to get the current value at all times vs. using state.  See how I do the UserContext.

humphd (changes requested):

humphd:
Let's get this in for 1.7, as it blocks #1733.

src/frontend/next/src/pages/_app.tsx
tonyvugithub: @humphd : Hi Dave, I tried to find a way to accommodate what you mentioned in the comment but it's a bit challenging for me. My approach here is to follow the **Updating Context from a Nested Component** from the react page https://reactjs.org/docs/context.html#dynamic-context.

I'm not too familiar with context. Can you guide me through your comment?

src/frontend/next/src/pages/_app.tsx
tonyvugithub: @humphd : How do I trigger rerender of the whole app if this was the case?

src/frontend/next/src/contexts/themeContext.ts
tonyvugithub: I tried to do move toggleTheme function from _app to here but found some difficulty to rerender the app.

src/frontend/next/src/pages/_app.tsx
```diff
@@ -1,17 +1,28 @@
...
 import '@fontsource/roboto';
 import '@fontsource/spartan';
 
 import '../styles/globals.css';
+
 import { darkTheme, lightTheme } from '../theme';
+import usePreferredTheme from '../hooks/use-preferred-theme';
+import { ThemeContext } from '../components/ThemeProvider';
+
+import '../styles/globals.css';
```
HyperTHD, line 16: This is already imported, so you can remove it

HyperTHD (approved):

src/frontend/next/src/pages/_app.tsx
PedroFonsecaDEV: Line 9 x line 14.

PedroFonsecaDEV (changes requested):
Sorry guys, I will request @humphd to review this PR.

src/frontend/next/src/pages/_app.tsx
tonyvugithub: I removed one already. Initially there were two duplicate imports . Hence for the lint error

tonyvugithub:
@PedroFonsecaDEV : Actually, he was working on my old commit and push to my commit. This is the combined version of our works altogether. He would prefer others to review since he wrote part of the code himself.

birtony (approved):
Love this stuff

PR #1696, opened 2021-02-12
Fixes #1189, Braille Spaces & Empty Paragraphs removal

### LINKED ISSUE #1189: Remove empty whitespace-only blocks from posts

Metropass:
## Issue This PR Addresses
Fixes #1189, and also added a check for the 0-dot braille character (Unicode Hex: 2800).
For the `<br>` removal, I had to split the work in half, as there's an issue with DOM not properly cutting the extra `/n` values. 

## Type of Change

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,6 +75,8 @@
   "dependencies": {
     "@elastic/elasticsearch": "7.9.0",
     "@elastic/elasticsearch-mock": "0.3.0",
+    "@next/mdx": "10.0.6",
+    "@octokit/webhooks": "8.0.2",
     "@wordpress/wordcount": "2.11.0",
     "body-parser": "1.19.0",
     "bull": "3.18.0",
--- a/src/backend/utils/html/index.js
+++ b/src/backend/utils/html/index.js
@@ -6,6 +6,8 @@ const syntaxHighlight = require('./syntax-highlight');
 const replaceCodeEntities = require('./replace-entities');
 const fixEmptyPre = require('./modify-pre');
 const toDOM = require('./dom');
+const removeEmptyParagraphs = require('./remove-empty-paragraphs');
+const removeZeroDotBraille = require('./remove-zero-dot-braille');
 
 const { JSDOM } = jsdom;
 
@@ -42,7 +44,10 @@ module.exports = function process(html) {
   lazyLoad(dom);
   // Replace <code> elements with encoded entities to use characters
   replaceCodeEntities(dom);
-
+  // Remove empty <p> elements
+  removeEmptyParagraphs(dom);
   // Return the resulting HTML
+  removeZeroDotBraille(dom);
+
   return dom.window.document.body.innerHTML;
 };
--- a/src/backend/utils/html/remove-zero-dot-braille.js
+++ b/src/backend/utils/html/remove-zero-dot-braille.js
@@ -1,0 +1,13 @@
+module.exports = function (dom) {
+  const regexValue = new RegExp('u{2800}');
+  if (!(dom && dom.window && dom.window.document)) {
+    return;
+  }
+
+  dom.window.document.querySelectorAll('p').forEach((p) => {
+    const paragraphInnerHTML = p.innerHTML;
+    if (!paragraphInnerHTML.replace(regexValue, '').trim()) {
+      p.remove();
+    }
+  });
+};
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -75,6 +75,8 @@
   "dependencies": {
     "@elastic/elasticsearch": "7.9.0",
     "@elastic/elasticsearch-mock": "0.3.0",
+    "@next/mdx": "10.0.6",
+    "@octokit/webhooks": "8.0.2",
```
manekenpix, line 79: I don't think these are used in this PR.

manekenpix (changes requested):

package.json
Metropass: It's not, but I had to include them, otherwise it'll fail the tests for missing packages 🤷

manekenpix: how can you fail tests if those packages are not used in master?

manekenpix: oh, I see that `@next/mdx` is already included in master, and here it seems you're updating it to version `10.0.6`.

manekenpix: Also, I ran our tests without `@octokit/webhooks` and it didn't fail.
Should we update `@next/mdx` in  a different PR maybe?

cc: @humphd

humphd: Yes, cc @chrispinkney, who is managing this.  Let's not do it here, please revert this and we'll do it in a new PR.

src/backend/utils/html/index.js
```diff
@@ -42,7 +44,10 @@ module.exports = function process(html) {
   lazyLoad(dom);
   // Replace <code> elements with encoded entities to use characters
   replaceCodeEntities(dom);
-
+  // Remove empty <p> elements
+  removeEmptyParagraphs(dom);
   // Return the resulting HTML
```
humphd, line 49: This comment doesn't match what you're doing on the next line.  Move this comment to 51 above the final return statement, and properly document what you're doing with this code.

src/backend/utils/html/remove-zero-dot-braille.js
```diff
@@ -0,0 +1,13 @@
+module.exports = function (dom) {
+  const regexValue = new RegExp('u{2800}');
```
humphd, line 2: To include unicode in JS Regex, you need to use `\u2800`.  Also, let's do it inline below, and use a global replace to get all instances:

```js
    if (!paragraphInnerHTML.replace(/\u2800/g, '').trim()) {
```

test/remove-zero-dot-braille.test.js
```diff
@@ -0,0 +1,10 @@
+const toDom = require('../src/backend/utils/html/dom');
+const removeBraille = require('../src/backend/utils/html/remove-zero-dot-braille');
+
+test('should remove <p>⠀⠀⠀⠀</p> (braille)', () => {
+  const htmlData = toDom('<div><p></p></div>');
```
humphd, line 5: Be more explicit, use Unicode hex entity: `&#x2800;`

test/remove-empty-paragraphs.test.js
```diff
@@ -0,0 +1,28 @@
...
+
+  test('should remove <p>   </p> (spaces)', () => {
+    const htmlData = toDom('<div><p>   </p></div>');
+    removeEmptyParagraphs(htmlData);
+
+    const expectedHtml = '<div></div>';
+    expect(htmlData.window.document.body.innerHTML).toEqual(expectedHtml);
+  });
+
+  test('should remove <p>     </p> (tabs)', () => {
```
humphd, line 21: Can you be more explicit with the inclusion of the tabs here? `&#9;` will do it.

humphd (changes requested):
Left some feedback.

Also, a question: why are you focused on braille-dots specifically?  There are lots of invisible whitespace characters.  Why not get rid of them all with something like https://github.com/junosuarez/clean-whitespace instead?

Metropass:
Closed. Added a better functioning PR at #1787

PR #1731, opened 2021-02-13
Fix #1730 - reduce dependabot frequency, scope

### LINKED ISSUE #1730: Deal with Dependabot spamming CI
I got an email today that I'd been billed for excess hours by Netlify due to the number of CI jobs started by Dependabot.  It also looks like Vercel is now failing all jobs too, which I suspect is them rate limiting us too.

I don't think this current setup is going to work.  We'll run into limits with all our CI if we keep doing this.

Let's figure out a solution soon to reduce these jobs, group them, whatever, and do it soon.

cc @birtony

humphd:
Fixes #1730 by reducing the scope and frequency of dependabot PRs, to try and save our CI from getting disabled.  Specifically:

- move from daily to weekly
- reduce the scope to security fixes
- reduce the number of PRs from 10 to 5 for each of the areas (that's still 20 PRs across all package.json files we have).  That might still be too much, I'm not sure.

We can increase things later if this works.  The current level is not going to work.


cindyledev (approved):

humphd:
I've removed the `now.json` file from our repo for Vercel, and am reconfiguring the Vercel build via the web to use our `src/frontend/next` dir (i.e., we won't build Gatsby anymore for PRs).

However, I have a bug in it, and can't re-build any more for a few hours (rate limited).  So I can't test this again right now :(

cindyledev (approved):

cindyledev:
Since we're not using Netlify anymore, we should remove the Netlify badge on our README.md

humphd:
@cindyledev nice catch, removed the badge.

cindyledev (approved):

manekenpix (approved):

humphd:
This wasn't enough, I've removed it completely https://github.com/Seneca-CDOT/telescope/commit/7a05bc0ca5edfcc426e9a7813890b44dc13e98ab

PR #1782, opened 2021-02-18
Fixes #1252, Added Comments for Logger Discussion

### LINKED ISSUE #1252: Reduce log spam
Our logging is great, but too noisy by default.  It's nice to log all kinds of things re: feed parsing and network errors, but they fill the log with spam that doesn't matter when you're developing or trying to debug a problem on production.

Our logger, Pino, has these levels:

```
  trace: 'DEBUG',
  debug: 'DEBUG',
  info: 'INFO',
  warn: 'WARNING',
  error: 'ERROR',
  fatal: 'CRITICAL',
```

For production, we probably only care about error/critical, and for normal development, probably only info and up.

Let's audit our logging calls and change the level we use.  Especially for the places where we do feed logging, let's make that `debug` or something vs. `info`.

Metropass:
## Issue This PR Addresses
Fixes #1252 

## Type of Change

## Description

So, I wanted to make this PR to discuss what needs to be changed for the Backend's logger. 
I've done an audit and added comments to the code on what I personally think needs to be changed, and why. 
I would really like some feedback/discussions so that what we change will be useful, and won't omit any data when debugging Telescope.

* `shutdown.js`
[There ](https://github.com/Seneca-CDOT/telescope/blob/c7a532038950ea2d039a8de8682046300b30db10/src/backend/lib/shutdown.js#L18)seems to be a lot of `error` functions being invoked when the service is shutdown. We parse the error into the logger, and mention that the shutdown was not graceful & dirty. Is it possible to move this into debug? 

* `processor.js`
-[Inside](https://github.com/Seneca-CDOT/telescope/blob/c7a532038950ea2d039a8de8682046300b30db10/src/backend/feed/processor.js#L145), we have a lot of warnings being sent to the logger. I think that the `info` calls are appropriate, but should we change the warnings into debug, and return a generic message for production? 

- [On Line 135](https://github.com/Seneca-CDOT/telescope/blob/c7a532038950ea2d039a8de8682046300b30db10/src/backend/feed/processor.js#L135), we send some data into `info` mentioning that a certain feed as been skipped. It might be redundant to print this all the time, as the larger the ignored feed list grows, the more messages this will output. Should we change this into debug?

- [Line 110](https://github.com/Seneca-CDOT/telescope/blob/c7a532038950ea2d039a8de8682046300b30db10/src/backend/feed/processor.js#L110), we only check for an instance of `ArticleError`, to let out service continue. Should we add a debug call to this, for developers to see what type of error is thrown?

These were the log calls that I think should be changed, I'm a little bit iffy on changing any calls within `backend/web`, as they're appropriate (in terms of logging any errors that occur).

I would love to get feedback on what needs to be changed.

## Checklist

### DIFF
--- a/src/backend/feed/processor.js
+++ b/src/backend/feed/processor.js
@@ -207,7 +207,7 @@ module.exports = async function processor(job) {
         }`
       );
     } else {
-      logger.error({ error }, `Unable to process feed ${feed.url}`);
+      logger.debug({ error }, `Unable to process feed ${feed.url}`);
       throw error;
     }
   }
--- a/src/backend/feed/worker.js
+++ b/src/backend/feed/worker.js
@@ -31,9 +31,10 @@ exports.start = async function () {
   try {
     await waitOnReady();
     logger.info('Connected to elasticsearch!');
-
     const concurrency = getFeedWorkersCount();
-    logger.info(`Starting ${concurrency} instance${concurrency > 1 ? 's' : ''} of feed processor.`);
+    logger.debug(
+      `Starting ${concurrency} instance${concurrency > 1 ? 's' : ''} of feed processor.`
+    );
     feedQueue.process(concurrency, path.resolve(__dirname, 'processor.js'));
     return feedQueue;
   } catch (error) {
--- a/src/backend/lib/shutdown.js
+++ b/src/backend/lib/shutdown.js
@@ -11,7 +11,7 @@ async function stopQueue() {
     await feedQueue.close();
     logger.info('Feed queue shut down.');
   } catch (error) {
-    logger.error({ error }, 'Unable to close feed queue gracefully');
+    logger.debug({ error }, 'Unable to close feed queue gracefully');
   }
 }
 
@@ -24,7 +24,7 @@ async function stopWebServer() {
     await serverClose();
     logger.info('Web server shut down.');
   } catch (error) {
-    logger.error({ error }, 'Unable to close web server gracefully');
+    logger.debug({ error }, 'Unable to close web server gracefully');
   }
 }
 
@@ -33,7 +33,7 @@ async function cleanShutdown() {
     await Promise.all([stopQueue(), stopWebServer()]);
     logger.info('Completing shut down.');
   } catch (error) {
-    logger.error({ error }, 'Failed to perform clean shutdown');
+    logger.debug({ error }, 'Failed to perform clean shutdown');
   }
 }
 


HyperTHD:
Move all your comments to the side instead of directly below the catch to avoid all these linting errors

Metropass:
> Move all your comments to the side instead of directly below the catch to avoid all these linting errors

I'm aware there's some errors, I'm not trying to pass the tests atm until we figure out what to change

yuanLeeMidori:
@Metropass These are my thoughts,

- showdown
I agree it is not that necessary for error, maybe warn?
- processor
I'm not sure about this, the not-found or error log should be different from the general ones?
- line 135
I agree with you, skipping feeds might be more useful in debug level instead of info
- line 110
adding it to debug sounds good
--- 

I'm still looking for possibility if there is any error log can be moved to critical and warning. I'll append them in this PR if I have progress.

src/backend/feed/processor.js
```diff
@@ -107,6 +107,7 @@ function articlesToPosts(articles, feed) {
     articles.map(async (article) => {
       try {
         await Post.createFromArticle(article, feed);
+        //Add a debug log to see what is thrown?
```
humphd, line 110: We end up logging this in the caller on line 214, and that one could probably be turned into a warning vs. error.

src/backend/feed/processor.js
```diff
@@ -132,10 +133,12 @@ module.exports = async function processor(job) {
   let info;
   const [invalid, delayed] = await Promise.all([feed.isInvalid(), feed.isDelayed()]);
   if (invalid) {
+    //Debug?
     logger.info(`Skipping resource at ${feed.url}. Feed previously marked invalid`);
```
humphd, line 137: I think info is right here, since we'd want to see skipped feeds in production.

src/backend/feed/processor.js
```diff
@@ -132,10 +133,12 @@ module.exports = async function processor(job) {
...
     return;
   }
   if (delayed) {
+    //Debug?
     logger.info(`Skipping resource at ${feed.url}. Feed previously marked for delayed processing`);
```
humphd, line 142: Same, leave this one info

src/backend/feed/processor.js
```diff
@@ -144,6 +147,7 @@ module.exports = async function processor(job) {
     // If we get no new version info, there's nothing left to do.
     if (!info.shouldDownload) {
       // Log some common cases we see, with a general message if none of these
+      //A lot of warnings & info, should we change it?
```
humphd, line 150: These are all useful for debugging, I'd leave them.

src/backend/feed/worker.js
```diff
@@ -33,6 +33,7 @@ exports.start = async function () {
     logger.info('Connected to elasticsearch!');
 
     const concurrency = getFeedWorkersCount();
+    //Change into debug?
```
humphd, line 36: Yeah, debug, agree.

src/backend/lib/shutdown.js
```diff
@@ -11,6 +11,7 @@ async function stopQueue() {
     await feedQueue.close();
     logger.info('Feed queue shut down.');
   } catch (error) {
+    //Change to debug?
```
humphd, line 14: Agree, debug here.

src/backend/lib/shutdown.js
```diff
@@ -24,6 +25,7 @@ async function stopWebServer() {
     await serverClose();
     logger.info('Web server shut down.');
   } catch (error) {
+    // Change to debug?
```
humphd, line 28: Yeah, debug for sure.

src/backend/lib/shutdown.js
```diff
@@ -33,6 +35,7 @@ async function cleanShutdown() {
     await Promise.all([stopQueue(), stopWebServer()]);
     logger.info('Completing shut down.');
   } catch (error) {
+    //Change to debug? Shutsdown, but dirty
```
humphd, line 38: Agree, debug.

humphd (changes requested):

humphd (commented):
You've got both branches in here, please separate this work from the other.

package.json
```diff
@@ -82,6 +82,7 @@
     "body-parser": "1.19.0",
     "bull": "3.20.1",
     "bull-board": "1.3.1",
+    "clean-whitespace": "0.1.2",
```
chrispinkney, line 85: What's the point of this? Ah, from the merge?

Metropass: Yup, it's cleaned out now :)

chrispinkney (dismissed):
Lookin' good!

humphd (dismissed):

yuanLeeMidori (approved):

PedroFonsecaDEV (approved):

tonyvugithub (approved):

HyperTHD (approved):

birtony (approved):

Issue #1642, opened 2021-02-03
Add a User service, with database for user, feed info, drop planet feed list


humphd:
**What would you like to be added**:

Add a new database to our system to hold user, feed, and other related data that needs to exist long term.  Our new data design would be:

- database: long term storage of users and all information related to users, for example, their Seneca email, preferred display name, RSS feeds, URLs to blogs, social media, GitHub, etc.
- redis: ephemeral storage (e.g., cache) for posts, shared session data, etc
- elasticsearch: indexing of data for search, for example posts.

**Why would you like this to be added**:

We currently use the [CDOT Planet Feed List](https://wiki.cdot.senecacollege.ca/wiki/Planet_CDOT_Feed_List) as our primary data source.  We've augmented that with Redis, but we don't have a great backup strategy or dedicated DevOps team, so it's possible to lose our data (#1365).

It should be possible to use this new database and replace the planet feed list dependency completely.

**Considerations**:

I'm not sure which database we should use.  Here are some thoughts:

- students coming to this course have all worked with MongoDB in the web stream.  Because it's a database that people know, it might be a logical choice.
- it would be easy to add whatever database we choose to our network of apps (e.g., run it in docker).  But should we consider running this in a cloud we don't manage?  For example, [MongoDB Atlas has a free tier](https://www.mongodb.com/pricing), but it's not really designed for production use.  There is also [Fauna](https://fauna.com/), [Firebase](https://firebase.google.com/pricing), and lots more that have a free tier.  Our data needs are always going to be modest, since we only need to store user info and a few URLs for hundreds of users.  Maybe we can fit in a free tier db and then not worry about data loss?
- whatever db we choose, it would be nice if it was really easy for a dev to either mock locally, or have it be easy to sign-up for a free instance on their own to use in dev.
- if we do decide to add the db to our network and run it ourselves, we need to consider how to do backups.

PedroFonsecaDEV:
@humphd, so we need to find a cloud to hold our long term storage? Or a SQL running in our machines would solve the problem facing a power outage?

chrispinkney:
Let's just ask Seneca to provide us with a backup diesel generator in case of power-outages. 😄 

I'm using Firestore (nosql) in my capstone project, it's great and extremely generous with its free tier (50k reads per day, 20k writes/deletes per day.) Would love to discuss this further sometime.

PedroFonsecaDEV:
[...]

Issue #1778, opened 2021-02-18
Improve monorepo tooling with Rushjs or Turborepo


humphd:
We are evolving our app to use microservices, and thinking about having a collection of smaller apps (front-end, services) vs. a monolithic, single application.  At the same time, we're also trying to embrace the idea of having a single monorepo (all code in one place), so that we don't have to duplicate setup for things like prettier, eslint, CI, etc. for each new repo.

Until now we've been using a series of `npm` scripts to manage having multiple `package.json` files, and it's worked OK, but the number of these is about to grow as we split the back-end into services.  Our app is eventually going to look like this:

```
/
    src/
        /web
            package.json
        /api
            /image
                package.json
            /user
                package.json
            ...
package.json
```

Here we have a single `package.json` at the root, and then sub-dirs for each major "app" within the project (`/web` is the next.js front-end and all the microservices live in `/api`).

There are a number of ways to manage these separate apps-with-an-app from the stand-point of the package manager:

- [yarn workspaces](https://classic.yarnpkg.com/en/docs/workspaces/)
- [pnpm](https://pnpm.js.org)
- [Lerna](https://lerna.js.org/)
- [npm7 workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces)

We need to pick one of these and migrate our repo to use it.

Here are some considerations:

- Telescope is developed by students, who are new to most of the tech we use.  When we pick something, let's make sure we can easily teach it to new devs.
- It should work for our workflows.  If we have to bend our workflows a lot to accommodate the tool, maybe it's not a good fit?
- It needs to work on every OS

The first part of this issue is to research the various solutions for doing a monorepo with JS, and then we need to pick one and implement it.

HyperTHD:
Reading the docs for all 4, I'm going to rule out yarn and lerna. Lerna imo fails the first two considerations since it's a new technology that has to be introduced with it's own unique setup that would take alot of work to migrate over. Yarn is a bit easier to teach to new devs than lerna, but has it's own disadvantages, primarily the package layout being different between devs and consumers and having packages being installed from npm instead of from the linked package.json of another workspace, possibly breaking all microservices that use such dependency.

That lives us with npm7 and pnpm. The latter seems to be built on top of npm while npm7 is strictly npm that we're familiar with. I'll continue doing research on both, but early reviews is making me lean towards npm7. One advantage for npm7 is that future students of the course should already have it if they got it from installing the latest LTS version of Node.js.

[...]

PR #1796, opened 2021-02-20
Authentication/Authorization microservice

humphd:
## Overview

Our current authentication/authorization system is designed for a monolithic architecture, where the whole app (back-end and front-end) are served from the same origin (i.e., `https://domain:port` is an origin, https://telescope.cdot.systems:443 is our prod origin).

It works like this:

- A user in our front-end wants to login to their Seneca SSO account
- We redirect them to a `/login` route
- This in turn creates a SAML formatted (special XML authentication format) POST message that gets sent to Seneca's server.
- The user is redirected to a login page on another machine not owned by us. They enter their username/password, and are redirected back to `/login/callback` in our server, and a session cookie is set with their user info (we get that back with this request).
- We redirect the user back to the web app they started in, and any time they do network requests to our server, if the route is protected, we check the session for their authentication cookie

## Moving to Distributed Auth

We want to split the app up and run it on different origins.  For example, microservices, Vercel, localhost, etc.  That distributed architecture breaks our security model, since we can't share that cookie across our network of connected apps and services.

We need a way to let a user authenticate, then come back to another part of our system and prove that they are authenticated.  A common way to do this type of thing is with [OAuth2](https://oauth.net/2/).  OAuth2 is for **authorization** vs. **authentication**.  It lets systems decide if a user is authorized to do something, and assumes authentication happened somewhere else.

I spent a long time looking into OAuth2 for our needs, but it's overly complicated for our (current) use cases. Because we aren't authorizing third-party apps (e.g., like GitHub or Google do with external apps that can connect to your account), and because we already have an in-house authentication system connected to our authorization needs, we can build a simpler flow.

## A Plan

This is the beginnings of a new auth service.  It does a few things:

- uses our SAML-based single-sign-on flow for user authentication.  I've ported over our existing code and expanded it so that it allows Seneca users to login
- adds an oauth2 inspired authorization flow using JWT access tokens

Here's a rough sketch of what it will look like:

Screen Shot 2021-02-19 at 7 11 49 PM

1. A user goes to our front-end app, perhaps https://telescope.cdot.systems
2. The front-end app wants to access a resource on one of our secure microservices, maybe user information.
3. The user needs to login, so clicks the Login link in the front-end app.  A small bit of state (e.g., random string) is put into `localStorage` for later.
4. They are redirected to our auth service: `api.telescope.cdot.systems/auth/login?redirect_uri=https://telescope.cdot.systems/&state=a3f1b3413`.  The URL contains two things: 1) `redirect_uri` containing a URL pointing back to the entry point of the front-end app; 2) some random `state`.  The latter is used as a ride-along value on all the redirects that are about to take place, and lets the client know at the end that nothing was tampered with in between.
5. The auth service receives the request (/`login?redirect_uri=https://telescope.cdot.systems/&state=a3f1b3413`) and stores the `redirect_uri` and `state` in the session. It then prepares a SAML message for this user to authenticate, and redirects them to the SSO identify provider server.
6. The SSO identity provider receives the request to login, and shows the user a login page, where they enter their username and password.  This either works or fails, and it both cases, they are sent back to the auth server with details about what happened
7. The auth server receives the result of the SSO login attempt at `/login/callback` and examines whether or not the user was authenticated.  If they were, we create an access token (JWT) and the request is redirected back to the original app at the `redirect_uri`: https://telescope.cdot.systems?access_token=...jwt-token-here...&state=...original-state-here...
8. The frontend app examines the query string onload, and sees the `access_token` and `state`.  It confirms the `state` is what it expects (e.g., compares to what's in `localStorage`).  The token is then used with all subsequent API requests to our microservices.
9. A request is made to the secure microservice.  The token is included in the HTTP Headers: `Authorization: bearer <jwt token here>`
10. The secure microservice gets the request, and pulls the bearer token out of the headers.  It validates it, verifies it, and decides whether or not the user is allowed to get what they want.  A 200 or 401 is returned.

Easy, right?  I'm skipping some details, but that's the main thrust of what's happening here.

## Running and Testing

You can run and test this (yes, I even wrote unit and end-to-end browser tests!).  Both require you to run a fake SAML SSO Login server via docker.

### Tests

For running the tests:

1. `cd src/api`
2. `docker-compose -f docker-compose-tests.yml up`
3. `cd auth`
4. `npm run test:unit`
5. `npm run test:e2e`

If you change line 10 of `src/api/auth/test/e2e/e2e.test.js` to `browser = await chromium.launch({ headless: false, slowMo: 1000 });` you can see the browsers start and run the e2e tests visually.  It's possible these will fail on your machine due to timing/resource limits.  I am still getting the feel for how far I can push the [test runner](https://playwright.dev/).

### Manually

You can play with this manually too.  I made a tiny static HTML web page that lets you use it.  You need to run the auth server and login container together in docker:

1. `cd src/api`
2. `docker-compose -f docker-compose-api.yml up --build auth login`
3. `cd auth`
4.  `npm run test:manual`

Now you can navigate to the various servers:

1. http://localhost:8888 is the web app (like running Telescope's front-end)
2. http://auth.docker.localhost/authorize (this is the auth server, not much to see, no UI)
3. http://login.docker.localhost/simplesamle (the SSO server, like Seneca's SSO)

To try logging in:

1. Go to http://localhost:8888 and click `Login`
2. You get redirected to the login server, something like http://login.docker.localhost/simplesaml/module.php/core/loginuserpass.php?AuthState=_a09df419f1488d4ca3bb879e96f4a42a2399d22095%3Ahttp%3A%2F%2Flogin.docker.localhost%2Fsimplesaml%2Fsaml2%2Fidp%2FSSOService.php%3Fspentityid%3Dhttp%253A%252F%252Fauth.docker.localhost%252Fsp%26cookieTime%3D1613785032
3. Enter your username and password: `user1` and `user1pass`
4. You'll get redirected back to the web app, something like http://localhost:8888/?access_token=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJodHRwOi8vYXV0aC5kb2NrZXIubG9jYWxob3N0IiwiYXVkIjoiaHR0cDovL2xvY2FsaG9zdDo4ODg4LyIsInN1YiI6InVzZXIxQGV4YW1wbGUuY29tIiwiaWF0IjoxNjEzNzg1MDk1LCJleHAiOjE2MTM3ODg2OTV9.53s-dCQX5xPAryTvpEpV6Vo7NAT0jHSf4sUJkh0EXxU&state=R4j4vVKPceGESO368-plF
5. Notice the `access_token` and `state` on the URL.  Copy the access_token (`eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJodHRwOi8vYXV0aC5kb2NrZXIubG9jYWxob3N0IiwiYXVkIjoiaHR0cDovL2xvY2FsaG9zdDo4ODg4LyIsInN1YiI6InVzZXIxQGV4YW1wbGUuY29tIiwiaWF0IjoxNjEzNzg1MDk1LCJleHAiOjE2MTM3ODg2OTV9.53s-dCQX5xPAryTvpEpV6Vo7NAT0jHSf4sUJkh0EXxU`)
6. Go to https://jwt.io/ and scroll down to the Encoded section.  Paste it in to decode it:

Screen Shot 2021-02-19 at 8 40 00 PM

That is the authorization info we'll pass around to different servers.  We'll sign it cryptographiclly, so it can't be faked from some other source.  We'll also add more claims to it (e.g., user vs. admin, name).

I need to add more production-level implementation to this, but it's probably ready to land and people can play with it.  I also haven't connected the tests to CI properly yet. 

I also think I want to integrate this with the work in #1642 that @chrispinkney is doing (i.e., it's tightly bound to the idea of authorization, so it might make sense to do it all in one service).

I wanted to get this up now because the other microservices are going to need it, and I didn't want to block progress for others during 1.8.

I'll do a proper demo of all this next week during our calls.  Let me know if you have questions or thoughts.


humphd:
@chrispinkney I've added you to the reviewers list too, since you blogged that you're using an oauth2 flow in your systems project, and since you're going to connecting code to mine with the user service.

src/api/auth/src/authorization.js
```diff
@@ -0,0 +1,32 @@
...
+
+passport.use(
+  new JwtStrategy(
+    {
+      jwtFromRequest: ExtractJwt.fromAuthHeaderAsBearerToken(),
+      secretOrKey: secret,
+      issuer,
+    },
+    function (payload, done) {
+      // TODO: check that payload.sub is a user we allow in our system
```
Metropass, line 19: Maybe for debugging purposes, should we should add the logger?  (once the TODO has finished)

src/api/auth/src/hash.js
```diff
@@ -0,0 +1,20 @@
...
+const crypto = require('crypto');
+
+/**
+ * Hash function used to generate our unique data ids.
+ * We use sha256 and encode in hex, so it's safe to use
+ * in URLs. For example:
+ *
+ *   6Xoj0UXOW3FNirlSYranli5gY6dDq60hs24EIAcHAEc=
+ *
+ * but truncate to only use the first 10 characters
```
Metropass, line 12: This is a good approach, iirc, it's also has better performance than using the entire digest size

src/api/env.production
```diff
@@ -12,11 +12,71 @@ COMPOSE_PROJECT_NAME=telescope_api
...
+# Our apps's Entity ID, which is also the URL to our metadata.
+SAML_ENTITY_ID=https://api.telescope.cdot.systems/v1/auth/sp
+
+# SECRET = cookie session SECRET. If left empty, one will be set automatically
+SECRET=secret-sauce
+
+# ADMINISTRATORS is a list (space delimited) of users who have administrator
+# rights. Use the user's nameID (user2@example.com) or hashed version of
+# nameID (2b3b2b9ce8).  Either will work.
+ADMINISTRATORS=user1@example.com
```
manekenpix, line 59: Shouldn't we set this with real user IDs?

humphd: Yeah, when we get @chrispinkney's user stuff integrated. For now this is parity with the old stuff.

humphd:
Reviewers, this is basically ready now.  I had to do more than I wanted to in order to get all the tests to work, since I'm inventing a bunch of new ways to write tests with this PR.

In addition to the info in my initial comment, here are some other things to understand:

1. I'm removing CircleCI.  You can't easily mount volumes in CircleCI without using real VMs, and I don't want to get into another CI problem with too many builds on real hardware.  Since we have amazing test coverage in GitHub Actions now, we don't need CircleCI.  I'll remove the integration from the repo once this lands.
2. I've refactored and expanded our GitHub Actions `node-js-ci.yml` workflow so it supports both our current unit tests, and the new end-to-end tests with docker that I'm adding (see below).
3. I had to work on our eslint infrastructure a bit.  The `import/no-extranous-dependencies` rule isn't happy with devDependencies only being present in the root directory.  I don't want to install them twice, and this is a known limitation of the plugin with monorepos.  Eventually this will go away, when we solve #1778.  For now I've just disabled the error globally.
4. I've reworked our Jest setup so it's monorepo aware.  The idea is to have a `jest.config.base.js` configuration (I removed it from `package.json`) and use that to define our various projects testing setup (i.e., each microservice, web front-end, and legacy tests).  This mostly means telling Jest which folders to look in for test cases, where to collect coverage info, which folders to ignore, etc.  We can use my current layout to create tests for other services.  Note the use of `<rootDir>/...` in this and other config files, which is a Jest way of dealing with paths deep in a tree so you can point back to the root.  The files `jest.config.js` and `jest.config.e2e.js` define our top-level configurations for running unit and e2e (require microservices and other services running in docker) tests.  The file `jest-playwright.config.js` is related to the auth e2e tests (see below).  The way that Jest is meant to be used now is from the root of the project, and it's aware of all the sub-projects and how to test them.  You can run `npm run jest` and run unit tests or `npm run jest:e2e` and run the end-to-end tests.
5. I had to add a `babel.config.js` to get coverage info to work with the new monorepo-jest approach and our TSX files.
6. I've refactored the auth service `Dockerfile` to be more like what @raygervais did with mult-stage builds in our main `Dockerfile`.  I think I've done it right, but maybe he'll review and tell me?
7. The `test/e2e` directory has a new kind of test: end-to-end tests.  These are meant to be run against a real backend and in a browser.  I've used https://playwright.dev/ to write these, which lets you script a browser and easily write interactive tests.  It's amazing.  These tests require that some or all of our backend services are running (see the `npm run start:services` script I added previously) and then use static web pages to simulate our React app.  I've written login and logout tests that are the same as what you would do manually.
8. Speaking of manual testing, I also added `test/manual` and `npm run test:manual` in auth to give a simple way to run a frontend you can test against the auth backend service in docker.
9. I've stubbed out the env files for dev/staging/prod with values from our SSO setup, but they aren't really going to work until I talk to Seneca ITS and get a new SAML app setup with our new staging/prod domains for the microservices.  We'll update in the future.

I think that's most of it.  I know this is big and I apologize for doing so much in one PR.  A lot of this needed to get added quickly so everyone else could do their work.  However there is still a lot to do on this auth service, so I'm going to pause now and do the rest in follow up.

manekenpix:
I get this when I try to run `e2e` tests as suggested:

```
# e2e tests, docker required
npm run services:start
npm run jest:e2e
npm run services:stop
```
Output:

```sh
Jest dev-server output:
npx: installed 23 in 1.858s
[Jest Process Manager] Starting up http-server, serving src/api/auth/test/e2e
Available on:
[Jest Process Manager]   http://127.0.0.1:8888
  http://192.168.2.17:8888
  http://172.21.0.1:8888
Hit CTRL-C to stop the server
Error: Host system is missing dependencies!
  Missing libraries are:
      libwebp.so.6
      libenchant.so.1
      libicui18n.so.66
      libicuuc.so.66
```

humphd:
@manekenpix playwright needs various libs to run the libraries.  They have a docker container that lists them:

https://github.com/microsoft/playwright/blob/master/utils/docker/Dockerfile.bionic

I wonder how we should deal with this?  Via update to the docs for installing the dev env?

It's interesting that it runs OK in CI on GitHub but not your distro.

raygervais:
As long as we can run the Docker tests, I believe that is good enough. No point adding additional dependencies to the development environment which increase it's size. Thoughts?

humphd:
@manekenpix is going to test more to see what those missing libs are all about.  We do need to figure out at least a docs update to tell people that we use them for e2e tests.

In theory, running e2e tests only in CI is probably OK (e.g., not everyone needs to do it locally)

birtony (approved):

birtony (dismissed):

manekenpix (approved):

humphd:
I'm going to try and land this tonight, and filed follow-ups on various things that need to happen after this goes in: #1868, #1869, #1870, #1871, #1872, #1873

PR #1802, opened 2021-02-20
adds: initial service architecture for Elastic search service port  

raygervais:
## Issue This PR Addresses

Closes #1737

## Type of Change

- [ ] **Bugfix**: Change which fixes an issue
- [X] **New Feature**: Change which adds functionality
- [X] **Documentation Update**: Change which improves documentation
- [ ] **UI**: Change which improves UI

## Description

Adds support for `search.docker.localhost`, which ports the original ELK query controller to a dedicated service. The following is added as a warning to not work further on the original codebase for querying: 

```javascript
/*
 * NOTE: This has been ported to ~/src/api/search
 * Future updates will remove the codebase below
 * and point towards `service.docker.localhost` for querying.
 */
 ```

image

## Checklist

- [ ] **Quality**: This PR builds and passes our npm test and works locally
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [X] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not (if applicable)
- [ ] **Documentation**: This PR includes updated/added documentation to user exposed functionality or configuration variables are added/changed or an explanation of why it does not(if applicable)

PR #1812, opened 2021-02-21
Change traefik to use /v1/service-name URLs vs. subdomains

humphd:
From Slack today:

>humphd  12:25 PM @RayGervais @Josue can we settle on a URL architecture for the web services?  We've been using sub-domains in Traefik, but I think that's going to add overhead for devops with needing CNAMEs and SSL.  Instead of image.api.telescope.cdot.systems/* what about api.telescope.cdot.systems/image/*.  In other words: api-domain/service-name/route*?  So we'd have https://api.telescope.cdot.systems/image/ and https://api.telescope.cdot.systems/auth and https://api.telescope.cdot.systems/search etc.  We perhaps should also consider versioning it? https://api.telescope.cdot.systems/v1/image and https://api.telescope.cdot.systems/v1/auth etc.  Thoughts?
>
>RayGervais  12:29 PM I really like the proposed semantic and version, all for changing to that and updating search / image as reference. Having versioning (with proper docs, granted) in an API is a great experience

This PR changes the image service routing to use http://api.telescope.localhost/v1/image (e.g. the gallery is at http://api.telescope.localhost/v1/image/gallery) in development.

I've done a few things to accomplish this:

- renamed the `docker-compose-api.yml` file to `docker-compose.yml` so you don't have to always include the filename in development.
- updated the Traefik version to 2.4.5
- removed port 443 for development, we'll only use 80 locally
- started adding common middleware, which any of our services can re-use.  First one is `gz_compress`, which I have the image service use.  We can add more later this same way.
- I updated the image service so that the URL isn't `/v1/image/image` (i.e., I have it on the root so `/v1/image` does what you want)
- I fixed a few small things in the image service, and updated the Satallite dep to 1.6.0

I started to make the same changes to the production docker-compose but got stuck, and wanted to ask some questions (@josue, @raygervais):

- How do we want to deal with the domain name in the routes being different in staging vs production?  Have 2 files (docker-compose-staging.yml and docker-compose-produciton.yml) or pass in an environment somehow to docker-compose (you can't do this globally).
- How do we want to name the servers we don't expose publicly in production?  We can still use some kind of `.localhost` domain (e.g., private.docker.localhost) for the Traefik routes to things like Elasticsearch.
- I got rid of the broken `depends_on` syntax using conditions, since that got removed from Docker 

This will affect #1796 and #1802


humphd:
I did some more reading and now I understand how to do this.  I've updated this PR, and I'll try to explain how it works.

Our microservices live in `src/api`, and within that we have multiple docker-compose files.  The "base" layer is what you need for development: Traefik + microservices + any dependent servers (redis, elastic, ...).  It's called `docker-compose.yml` and it's not optimized for production.

There is also a `production.yml` and it is an "overlay" or "override" file that you use on top of the base stack for staging and production.  If you want to understand how docker-compose works in production, this is helpful: https://docs.docker.com/compose/production/. Basically, you do:

```
docker-compose -f docker-compose.yml -f production.yml up -d
```

That will apply the config from `docker-compose.yml`, and then override/extend it with everything in `production.yml`.  You can see the full config that it would use if you do:

```
docker-compose -f docker-compose.yml -f production.yml config
```

Next, we have the problem of how to swap things between staging and produciton, without needing more duplicate files.  Again, docker-compose can handle this no problem, [from the docs](https://docs.docker.com/compose/environment-variables/):

> If you have multiple environment variables, you can substitute them by providing a path to your environment variables file. By default, the docker-compose command will look for a file named .env in the project directory (parent folder of your Compose file).
>
> By passing the file as an argument, you can store it anywhere and name it appropriately, for example, .env.ci, .env.dev, .env.prod. Passing the file path is done using the --env-file option:
>
> docker-compose --env-file ./config/.env.dev up 

So we have 3 env files available: env.development, env.staging, and env.production.  You pick the one you want and copy it to `src/api/.env` before you run docker-compose, or you include the `--env-file` flag to specify the file directly.

This setup should make it pretty easy for us to use the same docker-compose stack in dev and production without having to change too much, except a few env variables, and most of those can get hard-coded into the production docker-compose.

humphd:
I did an experiment to pull the Satellite repo into a directory in our tree, and see if I could get the Docker stuff to work:

1. created `src/api/satellite` and copied the contents of the satellite repo in there
2. modified how the image service imports satellite using a file path: `npm install --save ../satellite` which I didn't know worked, but gives you a special dependency of "file:../satellite: and it works just fine!
3. broke the Dockerfile for satellite and image up into two stages, and had image pull from satellite's latest tag
4. modified the `docker-compose.yml` file to build the satellite container for us

Here's the diff

```diff
diff --git a/src/api/docker-compose.yml b/src/api/docker-compose.yml
index 99ac5de..112537b 100644
--- a/src/api/docker-compose.yml
+++ b/src/api/docker-compose.yml
@@ -23,6 +23,13 @@ services:
     labels:
       - 'traefik.enable=true'
 
+  # satellite microservice base
+  satellite:
+    container_name: 'satellite'
+    build:
+      context: ./satellite
+      dockerfile: Dockerfile
+
   # image service
   image:
     container_name: 'image'
@@ -35,6 +42,7 @@ services:
       - '${IMAGE_PORT}'
     depends_on:
       - traefik
+      - satellite
     labels:
       # Enable Traefik on the image service container
       - 'traefik.enable=true'
diff --git a/src/api/image/Dockerfile b/src/api/image/Dockerfile
index 05d9123..b349d3f 100644
--- a/src/api/image/Dockerfile
+++ b/src/api/image/Dockerfile
@@ -1,9 +1,7 @@
-FROM node:lts-alpine as base
+ARG TAG=latest
+FROM api_satellite:${TAG}
 
-# https://snyk.io/blog/10-best-practices-to-containerize-nodejs-web-applications-with-docker/
-RUN apk add dumb-init
-
-WORKDIR /app
+WORKDIR ${ROOT_DIR}/image
 
 COPY --chown=node:node . .
 
diff --git a/src/api/image/package.json b/src/api/image/package.json
index 7cc197a..a5681bf 100644
--- a/src/api/image/package.json
+++ b/src/api/image/package.json
@@ -15,7 +15,7 @@
   },
   "homepage": "https://github.com/Seneca-CDOT/telescope#readme",
   "dependencies": {
-    "@senecacdot/satellite": "^1.6.0",
+    "@senecacdot/satellite": "file:../satellite",
     "celebrate": "^13.0.4",
     "got": "^11.8.1",
     "http-errors": "^1.8.0",
```

This works, but I don't think it's right.  People online (and in the docker-compose repo) are saying that you shouldn't have build-only services in docker-compose.  Doing what I did, satellite builds, runs, and exits immediately (which is fine, but also technically wrong).

The alternative here is to have people build the satellite container themselves locally.  This is fine, but it requires some knowledge of how docker works.  We could write scripts, but they would have to be in node so they work cross platform, and you'd have to know to re-build satellite if the code ever changed (i.e., it wouldn't be automatic).

I want something that doesn't require a ton of extra steps or knowledge to use.  Maybe there's a better solution than this.

For now, I'll stick with the separate external npm dependency.

humphd:
I forgot to include the satellite Dockerfile, for reference:

```dockerfile
FROM node:lts-alpine as base

# https://snyk.io/blog/10-best-practices-to-containerize-nodejs-web-applications-with-docker/
RUN apk add dumb-init

ARG TAG=latest
ARG GIT_SHA1=latest

ENV MONOREPO_SHA1 ${GIT_SHA1}
ENV MONOREPO_VERSION ${TAG}
ENV ROOT_DIR /app

WORKDIR ${ROOT_DIR}/satellite

COPY --chown=node:node . .
```

humphd:
I'm just going to keep updating this until someone reviews it...eventually it will be 10,000 lines long.

This new commit uses another interesting aspect of docker-compose I didn't know about: you can [specify the docker-compose file(s) to use via an environment variable](https://docs.docker.com/compose/reference/envvars/#compose_file).  I've done that for staging and production so that you only need to use `docker-compose up`, which is nice.  NOTE: I've also set a custom path separator so it will work on all platforms.

humphd:
Still no reviews, so I'm back with more!

This commit adds npm scripts to make it easier for people to run the services via docker-compose:

- `npm run services:start` or `npm run services:start image auth search` runs all or the listed services
- `npm run services:logs image search` shows logs for the given service(s)
- `npm run services:stop` gracefully stops and removes service containers
- `npm run services:clean` removes any containers that weren't already

src/api/env.production
```diff
@@ -0,0 +1,22 @@
...
+COMPOSE_PROJECT_NAME=telescope_api
+
+# Compose files to use together on production. NOTE: we specify separator below
+# so it will work on Windows and Unix, see
+# https://docs.docker.com/compose/reference/envvars/#compose_file
+COMPOSE_PATH_SEPARATOR=;
+COMPOSE_FILE=docker-compose.yml;production.yml
+
+# The host where all the microservices run (e.g., http://api.telescope.localhost)
+API_HOST=api.telescope.cdot.systems
```
manekenpix, line 16: Since both `env.staging` and `env.production` are identical except for `API_HOST`, we could only have `env.production` and modify the value of `API_HOST` doing something similar to what we do with the deployment server, something like [this](https://github.com/Seneca-CDOT/telescope/blob/cc8151f3c6e5435bed4c2c6237aa4063d5c0f636/tools/autodeployment/deploy.sh#L30) or [this](https://github.com/Seneca-CDOT/telescope/blob/cc8151f3c6e5435bed4c2c6237aa4063d5c0f636/tools/autodeployment/deploy.sh#L42)

humphd: I wondered about that, too. I felt like having a second file was less hacky than scripts that switch out strings, but I'm happy to do whatever.

manekenpix:
If `docker-compose` looks for `.env` at launch, should we remove [this](https://github.com/Seneca-CDOT/telescope/blob/cc8151f3c6e5435bed4c2c6237aa4063d5c0f636/docker-compose-production.yml#L19) from our backend's `docker-compose.yml` and `docker-compose-production.yml`?

manekenpix (dismissed):

humphd:
I didn't want to mess with the current docker-compose.yml, but eventually we'll need to rewrite it, and can figure out our overall env strategy then.  It will probably make sense to merge the env files for everything at the root, but I left the api separate for now while it evolves, so we don't break anything in the working app.

Thanks for reviewing this Josue.

src/api/env.production
raygervais: Though I'd argue using /sed is the cleanest method, we can't assume that ALL are working in a bash environment. I'd opt for two files which removes technical requirements. It's simpler.

src/api/image/package.json
```diff
@@ -15,7 +15,7 @@
   },
   "homepage": "https://github.com/Seneca-CDOT/telescope#readme",
   "dependencies": {
-    "@senecacdot/satellite": "^1.3.0",
+    "@senecacdot/satellite": "^1.6.0",
```
raygervais, line 18: Given the rapid development of Satellite, should we remove the version pinning until it's stable / standardized and then pin?

src/api/readme.md
```diff
@@ -2,15 +2,49 @@
...
+
+```
+$ cp env.development .env
+```
+
+Or the equivalent command to copy the file on their operating system.
+
+## Running the Services via docker-compose
+
+Run the services via [docker-compose](https://docs.docker.com/compose/), which will use your `.env` to decide how
```
raygervais, line 30: I love this improvement, far cleaner than specifying files

raygervais (approved):

birtony:
It might be helpful to have a testing strategy in the PR description @humphd

raygervais:
Note, reviewed code but have not tested on machine itself. Will test today.

src/api/image/package.json
humphd: Good point, I guess using `^1.6.0` will mean any `1.x` gets used already.  I was being explicit, but it will still pick-up the new versions.  I'm not planning to do a major release update on satellite this term I don't think.

humphd: Updated to `^1.x`

humphd:
I've merged this so I can rebase the auth work.  If people want to test it and file new bugs, they can, and/or we'll fix things when we review auth.

src/api/env.production
manekenpix: This would only affect `env.staging` and `env.production`, which are used in production deployment (presumably, production deployments will always take place in bash environments), so it's safe for local development.

PR #1827, opened 2021-02-25
Update eslintignore to use glob patterns for node_modules deep in repo

humphd:
Reading a bit more into how [eslint's ignorefile](https://eslint.org/docs/user-guide/configuring/ignoring-code#the-eslintignore-file) works, I update ours to explicitly ignore more things, including `node_modules` deep in the tree.  For comparison, what [VSCode does](https://github.com/microsoft/vscode/blob/main/.eslintignore).

I also updated our scripts to not use `.` and make sure it's using the config file.  I'm not sure it was for all OSes by default (Ray's was really slow).

This runs fast now for me.


raygervais:
10000x faster on Fedora. 

:pray:  :fireworks:  :smile:

raygervais (approved):

manekenpix (approved):

PR #1830, opened 2021-02-26
Closes #1705: Add Let's Encrypt to Traefik Microservices router

### LINKED ISSUE #1705: Add Let's Encrypt to Traefik Microservices router
We want to have our microservices available on the web at the following URLs:

- https://api.telescope.cdot.systems/ (for production)
- https://dev.api.telescope.cdot.systems/ (for staging)

Doing so requires us to get SSL certificate management added to Traefik.  Luckily, this is already baked into Traefik, [we only need to configure it](https://doc.traefik.io/traefik/v1.7/user-guide/docker-and-lets-encrypt/).

Depends on:

- #1704
- #1636

manekenpix:
## Issue This PR Addresses

Closes #1705

## Type of Change

## Description

Currently, telescope has a microservice ([certbot](https://github.com/Seneca-CDOT/telescope/blob/9788288925ba547633c219441e8519a41c2a8585/docker-compose-production.yml#L77)) in its `docker-compose-production.yml` that takes care of our SSL certificates. With this PR, `traefik` will have access to those certificates so [`TLS`](https://www.networkworld.com/article/2303073/lan-wan-what-is-transport-layer-security-protocol.html) can be enabled for all the microservices.

**Note**
`Traefik` can generate and renew certificates automatically, but since there are pieces in telescope that are not part of `traefik` and need to have access to the certificates, I thought it'd be better to keep this task outside of the microservices workflow for now. This can be changed in the future.

Also, we'll have to make changes to our autodeployment server to set the certificates' path in a `env` variable. This way, `traefik` and anything outside of `traefik` that needs the certificates will know where to find them. This can be done in a follow-up PR.

## Checklist

### DIFF
--- a/src/api/config/certs.yml
+++ b/src/api/config/certs.yml
@@ -1,0 +1,13 @@
+tls:
+  certificates:
+    # The path to certs specified here is just a placeholder.
+    # During production deployment, it should be changed to `/certs/${PATH_TO_CERTS} (from .env)
+    - certFile: /certs/path/to/certs
+      keyFile: /certs/path/to/certs
+  options:
+    default:
+      sniStrict: true
+      minVersion: VersionTLS12
+
+    mintls13:
+      minVersion: VersionTLS13
--- a/src/api/production.yml
+++ b/src/api/production.yml
@@ -18,15 +18,20 @@ services:
       - '--api.insecure=false'
       - '--providers.docker=true'
       - '--providers.docker.exposedbydefault=false'
-      - '--entrypoints.web.address=:80'
-      # TODO: https/ssl setup...
+      - '--providers.file.directory=/config'
+      - '--providers.file.watch=true'
+      - '--entrypoints.web-secure.address=:443'
+      - '--entrypoints.web-secure.http.tls=true'
     ports:
-      - '80:80'
-      - '8080:8080'
+      - '443:443'
     labels:
       - 'traefik.enable=true'
       # GZip compression middleware: gzip_compress
       - 'traefik.http.middlewares.gzip_compress.compress=true'
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+      - ./config:/config
+      - ${PATH_TO_CERTS}:/certs
 
   # image service
   image:
--- a/src/api/env.production
+++ b/src/api/env.production
@@ -17,6 +17,9 @@ COMPOSE_FILE=docker-compose.yml;production.yml
 # NOTE: if you change this, change all other occurrences below too.
 API_HOST=api.telescope.cdot.systems
 
+# Path to certificates
+PATH_TO_CERTS=
+
 # The API Version, used as a prefix on all routes: /v1
 API_VERSION=v1
 
--- a/src/api/env.staging
+++ b/src/api/env.staging
@@ -17,6 +17,9 @@ COMPOSE_FILE=docker-compose.yml;production.yml
 # NOTE: if you change this, change all other occurrences below too.
 API_HOST=dev.api.telescope.cdot.systems
 
+# Path to certificates
+PATH_TO_CERTS=
+
 # The API Version, used as a prefix on all routes: /v1
 API_VERSION=v1
 


birtony:
This PR makes sense to me, but is there any way for me to test it @manekenpix?

src/api/config/certs.yml
```diff
@@ -0,0 +1,11 @@
+tls:
+  certificates:
+    - certFile: /certs/path/to/certs
```
humphd, line 3: Why does one line have a leading `-` and the next one not?

Also, are these paths real or meant to be place holders?  If the latter, can you leave comments explaining?

src/api/production.yml
```diff
@@ -18,15 +18,22 @@ services:
       - '--api.insecure=false'
       - '--providers.docker=true'
       - '--providers.docker.exposedbydefault=false'
+      - '--providers.file.directory=/config'
+      - '--providers.file.watch=true'
       - '--entrypoints.web.address=:80'
```
humphd, line 23: Should we disable 80 in staging/prod and force people to use HTTPS always?  A lot of sources online talk about this for services vs. HTML endpoints for users.

src/api/config/certs.yml
manekenpix: > Why does one line have a leading - and the next one not?
I was following the same format they use in [their documentation](https://doc.traefik.io/traefik/v2.0/https/tls/#user-defined).

> Also, are these paths real or meant to be place holders? If the latter, can you leave comments explaining?
:+1:  I'll leave a coment

src/api/production.yml
manekenpix: I think it's a good idea. I'll disable the entry point for port 80.

src/api/config/certs.yml
humphd: Can we do this in the docker-compose `production.yml` with labels instead of here in a separate file, so it can use the env vars in scope there?

manekenpix: @humphd That hasn't been implemented yet (but there's a lot of demand for it)

https://github.com/traefik/traefik/issues/5507

humphd: Interesting, I didn't realize you had to have an external file for this.  Seems odd.  Thanks for the link.

src/api/production.yml
```diff
@@ -18,15 +18,21 @@ services:
       - '--api.insecure=false'
       - '--providers.docker=true'
       - '--providers.docker.exposedbydefault=false'
-      - '--entrypoints.web.address=:80'
-      # TODO: https/ssl setup...
+      - '--providers.file.directory=/config'
+      - '--providers.file.watch=true'
+      - '--entrypoints.web-secure.address=:443'
     ports:
       - '80:80'
```
humphd, line 25: Let's remove this port binding

src/api/production.yml
```diff
@@ -18,15 +18,21 @@ services:
...
       - '8080:8080'
```
humphd, line 26: Do we care about the dashboard?  I'd remove it.

src/api/production.yml
```diff
@@ -42,6 +48,7 @@ services:
     labels:
       # Add gzip compression too
       - 'traefik.http.routers.image.middlewares=gzip_compress,strip_api_prefix'
+      - 'traefik.http.routers.image.tls=true'
```
humphd, line 51: So we have to explicitly enable tls on all these services?

humphd (changes requested):

src/api/production.yml
manekenpix: Yes, I thought it'd be a good idea to keep it flexible and only enabling it for specific services.
If we know that we want all the services to use tls, maybe I should enable it by default for all services?

humphd: I think that in the context of this file (btw, I'm refactoring it a bit in https://github.com/Seneca-CDOT/telescope/pull/1840 to separate our traefik+microservices from dev deps), anything we enable with traefik should be tls, no?  Especially if we disable :80, there's no point doing traefik for a container if we don't expose it via tls.

humphd (dismissed):

humphd (approved):

humphd (approved):

manekenpix:
Merged after rebasing

PR #1833, opened 2021-02-26
Added Service for Auto Discovery of Feeds for URLs

tonyvugithub:
## Issue This PR Addresses
Fixed #1645 

## Type of Change

- [ ] **Bugfix**: Change which fixes an issue
- [x] **New Feature**: Change which adds functionality
- [ ] **Documentation Update**: Change which improves documentation
- [ ] **UI**: Change which improves UI

## Description
Followed the layout of image microservice to add the feed discovery service. The service will return a list of potential feed urls if the user provided a valid blog url

## Checklist

image

Unit Tests
image
image

- [ ] **Quality**: This PR builds and passes our npm test and works locally
- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [x] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not (if applicable)
- [ ] **Documentation**: This PR includes updated/added documentation to user exposed functionality or configuration variables are added/changed or an explanation of why it does not(if applicable)

PR #1828, opened 2021-02-25
Post parsing/sync microservice

c3ho:
## Issue This PR Addresses
Fixes #1736 

## Type of Change

- [ ] **Bugfix**: Change which fixes an issue
- [x] **New Feature**: Change which adds functionality
- [ ] **Documentation Update**: Change which improves documentation
- [ ] **UI**: Change which improves UI

## Description
Still a WIP. Creates a post parsing microservice from our current code. There's a ton of files and I just want to get this up here for now. I opted to use the JS version of bull-queue, whether that is the correct choice or not I have no idea, but it shouldn't be too hard to make the switch to TS version in a future PR.

I started a small change in `processAllFeeds()` within the `index.js` file which will make a GET all user feeds request (that are not flagged) to the User ms.. I'll slowly improve this to do it in a pagination way.

I checked the Post api, there's no PUT routes yet to save all the Post data as per the most recent discussion. 

As satellite is waiting for Redis PR to be put in, this container will crash since I can't call `Redis()` yet. I'll also remove the test route I set up at `parserURL/`

Lists of tasks for this PR. I'll mark them off (maybe with a commit #) to track progress.
- [ ] Actually test this
- [ ] Write tests
- [x] Create readme
- [x] Create Dockerfile
- [ ] Update Feeds to store less info in Redis (need to doublecheck what we need and don't need)

## Checklist

- [ ] **Quality**: This PR builds and passes our npm test and works locally
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not (if applicable)
- [ ] **Documentation**: This PR includes updated/added documentation to user exposed functionality or configuration variables are added/changed or an explanation of why it does not(if applicable)

Issue #1735, opened 2021-02-13
Creating a Post microservice


HyperTHD:
**What would you like to be added**:

Telescope is going to undergo a huge overhaul with the backend as we're going to convert it to operate using microservices. As part of that effort, we should have a microservice to handle the posts. 

**Why would you like this to be added**:

A microservice to handle the posts would take away pressure from the feeds service we also plan on incorporating. The migration would be a little easier in that we have such operations currently running in our backend. It would also need to be updated to communicate with the user microservice work currently being done in #1642 

As per @humpd's vision of this microservice:

![Post Microservice](https://user-images.githubusercontent.com/16841702/107857674-d5706b00-6dfd-11eb-83af-55a0610bc718.png)

humphd:
I think the way to think about this service is to begin small.  Here is my initial suggestion:

- Look at the existing code in https://github.com/Seneca-CDOT/telescope/blob/master/src/backend/web/routes/posts.js and tests in https://github.com/Seneca-CDOT/telescope/blob/master/test/posts.test.js
- Port that code over to a stand-alone service in `src/api/posts`
- Add Redis to `src/api/docker-compose.yml` similar to how it works in our root `docker-compose.yml`

We will need to figure out how to connect this to our new User service for feed info.  We won't be storing feeds in Redis anymore, so Feed and User info will have to come from the User service.  I'm not sure if we should do that in the Posts service (i.e., have it amalgamate that info into the responses) or let the client do calls to both services and aggregate the data there.  cc @raygervais for some architectural advice.

Later on I think we'll want to expand this service to include triggering indexing.

manekenpix:
@HyperTHD I'm here if you need an extra hand. This sounds like a nice piece to work on.

Issue #1737, opened 2021-02-13
Create a search microservice


humphd:
We are moving toward microservices for our back-end, and I want to refactor our current search routes into a separate service.

We have some existing code for this at https://github.com/Seneca-CDOT/telescope/blob/master/src/backend/web/routes/query.js and a working implementation for searching Posts and Authors.

To begin, let's duplicate what we have, but provide a better REST API, maybe something like this:

- `GET /search/user/?q=...&page=..&per_page=..` - allows searching for user's by the given `q` (i.e., query) string.  This probably should search their names (first, last, display, github username).  It should return paged results (e.g., `page` results with `per_page` results per page).  See the [GitHub Search API](https://docs.github.com/en/rest/reference/search#search-commits) for examples of what this might look like.

- `GET /search/post/?q=...&page=..&per_page=..` - similar idea to the above, but let's you query with post data (title, body of post) and returns results.  We might want to expand this API a bit to support things that Elastic can do, since it is a lot more powerful with it's query API.  That could happen later, though.

There are likely other things we can add search endpoints for, but would be a good start.

raygervais:
I'd be interested in helping with this service, may not have time to fully implement but can def provide the foundation

humphd:
Cool @raygervais, do what you can, and we'll take over from there.

PR #1838, opened 2021-02-26
Adds Initial Service to Get Posts

HyperTHD:
## Issue This PR Addresses
This PR address #1735 

## Type of Change

## Description

This PR is a work-in-progress, meant to work on adding a Post Microservice. This microservice parses posts from user's feeds and returns them. There's a path that returns the 10 latest posts, and you can also retrieve information about a specific post using it's ID.

This PR includes files that are being used by other microservices. I recommend getting those PRs in first before this one so that I can update the docker-compose file to use those services. This would allow me to remove the files those services used and only keep what I need for my microservice. In particular, this would be the User Microservice, The Post parsing microservice (#1828), and possibly the Feed Parsing microservice (#1833).

UPDATE: Screenshot of successful loading of this route along with network tab open as proof
Posts-Microservice

## Checklist


HyperTHD:
This PR is not tested btw. I'm still working on the Docker related stuff. I will post screenshots of the output when I get it working

src/api/posts/server.js
```diff
@@ -0,0 +1,10 @@
+const { Satellite } = require('@senecacdot/satellite');
+const posts = require('./src/routes/posts');
+
+const service = new Satellite();
+
+service.router.use('/posts', posts);
```
humphd, line 6: You can do this at the root `/` vs. `/posts` since Traefik will add a `/v1/posts` prefix for you.

src/api/posts/src/bin/article-error.js
```diff
@@ -0,0 +1,13 @@
+class ArticleError extends Error {
```
humphd, line 1: Let's get rid of the `bin` directory completely, everything can live in src.

src/api/posts/src/routes/posts.js
```diff
@@ -0,0 +1,111 @@
...
+    }
+  } catch (err) {
+    logger.error({ err }, 'Unable to get posts from Redis');
+    res.status(503).json({
+      message: 'Unable to connect to database',
+    });
+  }
+});
+
+module.exports = posts;
```
humphd, line 111: Let's add a some more routes:

- `posts.post('/:id', ...)` for adding
- `posts.delete('/:id', ...)` for deleting.  Not sure if we need this one, or if we'd just delete at the feed level.  It's hard to remove a post, since it will just pop-up again the next time we crawl the feeds.
- `posts.put('/:id', ...)` for updating

All of these need to use the JWT protection I've added in Satellite (e.g., you have to be authenticated).  Furthermore, we'll need to deal with a user vs. admin:

- admin can do the operations above on any post (i.e., they don't need to own it)
- a user will only be able to do it for a post they own (i.e., belonging to a feed they own)

src/api/readme.md
```diff
@@ -38,16 +38,18 @@ docker-compose up --build
 This will (re)build all the services, pull any images that are necessary for dependent services, and start all the services
 together. You can then access the services via the api hostname (see `API_HOST` in your .env file). For example:
 
-http://api.telescope.localhost/v1/image/gallery
+<http://api.telescope.localhost/v1/image/gallery>
```
humphd, line 41: Why this change?

humphd (changes requested):
I did a first read through and left some feedback.

There's a lot of code in here that can be removed and done in #1828 by @c3ho (e.g., all the feed handling stuff).

Eventually I want to have the feed service use this service to get/set/update/delete posts.

src/api/readme.md
HyperTHD: Yikes, it seems when I updated the README and saved, it added those angle brackets together. That's something with the prettier settings iirc. I'll have to check on this

HyperTHD:
UPDATE: 

- I added the docker code to create the posts image. I may have done it wrong, but I needed to include redis and elasticsearch inside the docker-compose.yml file since it needs both to function.
- I added the redis and elasticsearch information to all 3 env example files for reason specified above.
- I modified the redis and elasticsearch host names so that they use the container names. Since we're running them off the container instead of localhost, it makes the most sense to do this.
- Lastly, I refactored the code to remove the bin folder completely.

Running the posts service atm loads the right page, however, NO POSTS ARE LOCATED AND YOU GET AN EMPTY ARRAY AS A RESULT. I feel this is because of the changed host names, since the redis data is locally served instead of via the cloud. I'm not too sure about this one. Otherwise, I feel good about this service and can move to add in the extra routes

src/api/posts/src/bin/article-error.js
HyperTHD: Done

src/api/posts/server.js
HyperTHD: Didn't know about that, nice to know, thanks!

src/api/docker-compose.yml
```diff
@@ -46,3 +46,48 @@ services:
       - 'traefik.http.middlewares.strip_api_prefix.stripprefix.prefixes=/${API_VERSION}/image'
       - 'traefik.http.middlewares.strip_api_prefix.stripprefix.forceSlash=true'
       - 'traefik.http.routers.image.middlewares=strip_api_prefix'
+
+  redis:
+    image: redis:latest
+    ports:
+      - ${REDIS_PORT}:${REDIS_PORT}
```
humphd, line 53: We don't want to expose Redis to the big-bad world.  We made this mistake once before, and got hacked, see https://docs.google.com/presentation/d/1xBYQyNEDJPS1JRlzw4JHAX1exuPTmTB-a950zHiU950/.  In Docker Compose, defining ports works like this:

```
ports:
  # Port definitions can take 1 or 2 values:
  # DOCKER_HOST_PORT: DOCKER_CONTAINER PORT
  6379: 6379
  # The above means: "use port 6379 on the HOST machine to expose port 6739 in the CONTAINER"
  # It has the effect of connecting the Redis server's port running in the container to the same port in
  # the host computer, effectively opening that port through to the internet, if the host is running on
  # the web (staging, production). 
```

This is what people tend to do when they work on development, and need to connect directly to some internal database or server.  By connecting the container's port to your local dev machine's port, you can access things running in the container.  It works great.  But in a production environment, you don't want to expose traffic via ports like this.  We want to route everything through Traefik, which can manage load-balancing, authentication requests, and container/port mappings.

For access to Redis and Elastic from the outside world, we want to use your service as our only direct means.  It shouldn't be possible to connect to https://telescope.cdot.systems:6739 for example.  Your service (and others that will talk to Redis/Elastic) will only be available on the outside via a set of well-defined HTTP endpoints, which can use JWT based authentication (see #1796).

The right way to do what you're doing here is to do something like you can see in these PRs:

1. https://github.com/Seneca-CDOT/telescope/pull/1840/files#diff-a3872c2e4d094f394e87eb51d5c934953e79319accea61231188218ba9b184b0  In this example I'm running a container for the Firebase emulator.  It uses all kinds of ports, but I only want to enable 2 of them: 4000 (Firestore) and 8088 (the Dashboard UI).  Notice that I say `- '4000'` (i.e., only 1 port).  I do this because I am defining the port that's being used in the container, but I am not exposing it to the outside world.  Instead, I do this to let Traefik map that internal port to an endpoint URL.  Going to http://ui.firebase.localhost will allow a developer to get to the container's port 4000, but I don't open port 4000 on the host machine:

```
      - 'traefik.http.routers.firebase_ui.rule=Host(`ui.firebase.localhost`)'
      - 'traefik.http.routers.firebase_ui.service=firebase_ui'
      - 'traefik.http.services.firebase_ui.loadbalancer.server.port=4000'
```

In the case of `8088` I do both a container port and a host port mapping.  I don't want to do this, but the Firebase UI expects to be able to use localhost:8088 in order to load JS/CSS and other static assets in the browser.  Because this is a development situation, I'm not that worried.  But FYI.

2. You can see what @raygervais is doing in https://github.com/Seneca-CDOT/telescope/pull/1802/files#diff-9b5638ed60397f4c9b63cc41ff257015946394e745dce431fda440513146f5de with the Search service and Elastic.  Here he defines the elastic port on the container, but doesn't open it to the host.

The docker-compose files are evolving quickly in various PRs that are in review right now, so don't worry too much about making this perfect.  I would advise that the easiest fix here is to drop the second port from each of these services so we don't expose anything.

I wanted to mention it to you so you'd know, since this isn't an obvious thing, and the cost of getting it wrong is much larger than you realize.

src/api/docker-compose.yml
```diff
@@ -46,3 +46,48 @@ services:
...
+    command: ['redis-server', '--appendonly', 'yes']
+    volumes:
+      - ../../../redis-data:/data
```
humphd, line 56: I wonder where we should put volumes related to the microservices.  The root is one option for sure, but maybe we should have a dedicated directory where we can find these things.  cc @manekenpix, @raygervais.

src/api/docker-compose.yml
```diff
@@ -46,3 +46,48 @@ services:
...
+
+  elasticsearch:
+    image: docker.elastic.co/elasticsearch/elasticsearch:7.9.3
+    environment:
+      - bootstrap.memory_lock=true
+      - 'ES_JAVA_OPTS=-Xms512m -Xmx512m'
+      - discovery.type=single-node
+    ports:
+      - ${ELASTIC_PORT}:${ELASTIC_PORT}
```
humphd, line 65: Same thing here.

src/api/posts/src/processHTML/dom.js
```diff
@@ -0,0 +1,18 @@
+const jsdom = require('jsdom');
```
humphd, line 1: We won't do any of the HTML parsing/processing of feeds in this service, you can remove this dir.

humphd (changes requested):

src/api/posts/src/processHTML/dom.js
HyperTHD: Removing it right now will prevent my service from running at all until this process is merged. I'd either need to wait until the service is merged or just merge it there and have the person writing that service remove the directory and modify this service to use theirs

src/api/docker-compose.yml
manekenpix: @humphd this is how we currently do it on prod and dev:

[Directory Structure](https://github.com/Seneca-CDOT/telescope/blob/master/docs/staging-production-deployment.md#directory-structure)

src/api/docker-compose.yml
HyperTHD: Interesting, I forgot that the second port being there exposes the service to the web via the host machine. Noted

src/api/posts/src/routes/posts.js
HyperTHD: Filed #1865 to handle these additional routes. This will likely involve talking with the User microservice or the Feed parsing microservice. Needs more discussion atm before we continue with this. Using the JWT protection in satellite should be easy enough

manekenpix:
I need some clarification for this PR. I thought this microservice was going to answer requests (serving posts), but I see a lot of code here related to processing feeds and posts. Shouldn't all that be part of a different microservice?

@HyperTHD @humphd

humphd:
@manekenpix I left a similar review comment, a lot of this code can get deleted.

HyperTHD:
@manekenpix @humphd The main problem with removing that code is that there's no way for me to really know if the service works or not. The files related to my service also require the other files needed for feeds as well. Since we don't have those other services up, I can't realistically remove those extra files without breaking CI

raygervais:
Why not test using mock services to pass in hypothetical data @HyperTHD ?

HyperTHD:
@raygervais I plan on doing that once I remove the unnecessary files here. The thing I'm having here is that I need to figure out what to remove. I see that I don't need elasticsearch so I'm gonna remove all relevant files for that. Since this service just makes the requests to redis and fetches the posts data, I only need files relevant to that, including Feeds possibly, I'm not sure on that.

HyperTHD:
@humphd Idk if Feeds or anything related to that is needed, but I kept that since the identity of a Post is based on that. Based on my testing, I realized that I'm only making requests to redis, so I removed all elasticsearch related files and other files that weren't needed. Unless I need to remove the feed stuff, this should be good and I can proceed with writing the tests

c3ho:
I thought we were using firebase to store feeds and elastic to store posts? In that case wouldn't redis only be used for bull-queue itself to manage queues?

humphd:
If/when we add a `POST` route to this, we'll need Elastic.  When a post gets added to the system, we do 2 things: cache it in Redis; index it in Elastic.

For the `GET` requests, we don't need Elastic (that's more for Search, see #1737).

The tension here around "how do I know if my code is working" is really best addressed through tests, which all services need.  The way you'd test this is to have the test harness put some posts into Redis (or our mock redis) and then try to get them again via `GET` requests. That's the best way to proceed.

src/api/docker-compose.yml
```diff
@@ -106,3 +106,52 @@ services:
       - 'traefik.http.routers.login.rule=Host(`login.localhost`)'
       # Specify the login port
       - 'traefik.http.services.login.loadbalancer.server.port=8080'
+      # Add middleware to this route to strip the /v1/posts prefix
```
humphd, line 109: This looks wrong.  Can you rebase?  I've updated how the docker-compose files are structured.

src/api/docker-compose.yml
```diff
@@ -106,3 +106,52 @@ services:
...
+      - 'traefik.http.routers.image.middlewares=strip_api_prefix'
+
+  redis:
+    image: redis:latest
+    container_name: api_redis
+    ports:
+      - ${REDIS_PORT}
+    command: ['redis-server', '--appendonly', 'yes']
+    volumes:
+      - ../../redis-data:/data
```
humphd, line 121: I wonder if we should move this volume into `src/api`?

src/api/docker-compose.yml
```diff
@@ -106,3 +106,52 @@ services:
...
+
+  elasticsearch:
```
humphd, line 123: If you want to remove this now, and let @raygervais add it in #1737, that's fine.

src/api/docker-compose.yml
```diff
@@ -106,3 +106,52 @@ services:
...
+      - redis
+    labels:
+      # Enable Traefik for the posts service container
+      - 'traefik.enable=true'
+      # Enables routing for the posts service at /v1/posts
+      - 'traefik.http.routers.posts.rule=Host(`${API_HOST}`) && PathPrefix(`/${API_VERSION}/posts`)'
+      # Specifiy the posts service port
+      - 'traefik.http.services.posts.loadbalancer.server.port=${POSTS_PORT}'
+      # Add middleware to this route to strip the /v1/posts prefix and use gzip compression
+      - 'traefik.http.middlewares.strip_api_prefix.stripprefix.prefixes=/${API_VERSION}/posts'
```
humphd, line 155: I need to research this more, but I don't think you can reuse this name `strip_api_prefix` again, since it's been used on the image service.  Probably it needs to be `strip_image_prefix` and `strip_posts_prefix` etc.

src/api/env.development
```diff
@@ -13,6 +13,24 @@ API_HOST=api.telescope.localhost
 # The API Version, used as a prefix on all routes: /v1
 API_VERSION=v1
 
+# Redis info
+REDIS_URL=http://redis:6379
+REDIS_PORT=6379
```
humphd, line 18: Let's simplify the code and just pass the URL in, removing port and password, since we don't mess with either in Docker.

humphd (dismissed):

src/api/docker-compose.yml
HyperTHD: I can move it there and remove it from the root folder.

humphd: Don't remove that one, it's our legacy one that's being used by the current setup.  I'm just thinking about isolating these two from each other while we get the services working.

src/api/docker-compose.yml
HyperTHD: This might have been the root cause of the traefik http routing issue I had whenever I would run both image and posts services together. I'll test and resolve if this works

src/api/docker-compose.yml
HyperTHD: Yeah, this was the result of the merge conflicts resolution thing I was fixing. Will update

src/api/posts/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@senecacdot/posts-service",
+  "version": "1.0.0",
+  "private": true,
+  "description": "A service for retrieving posts",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
humphd, line 7: We should add a couple of scripts so you can run this locally (start, dev).  See image service.

src/api/posts/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "url": "https://github.com/Seneca-CDOT/telescope/issues"
+  },
+  "homepage": "https://github.com/Seneca-CDOT/telescope#readme",
+  "engines": {
+    "node": ">=12.0.0"
+  },
+  "devDependencies": {
+    "del-cli": "^3.0.1",
+    "nodemon": "^2.0.7",
+    "pino-pretty": "^4.5.0"
```
humphd, line 21: Not necessary, you can remove.

src/api/posts/package.json
```diff
@@ -0,0 +1,35 @@
...
+  },
+  "dependencies": {
+    "@elastic/elasticsearch": "^7.11.0",
+    "@elastic/elasticsearch-mock": "^0.3.0",
+    "@senecacdot/satellite": "^1.6.1",
```
humphd, line 26: Change this to `"^1.x"`

src/api/posts/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "crypto": "^1.0.1",
```
humphd, line 27: Crypto is built-into node.js now, you shouldn't need this.  Also, if the only reason you're pulling this in is to do the hash function, you can use what @Metropass added to Satellite:

```js
const { hash } = require('@senecacdot/satellite');
```

src/api/posts/src/data/post.js
```diff
@@ -0,0 +1,99 @@
+const { getPost, addPost } = require('../storage');
+const textParser = require('../text-parser');
+const hash = require('../../../auth/src/hash');
```
humphd, line 3:
```js
const { hash } = require('@senecacdot/satellite');
```

src/api/posts/src/storage.js
```diff
@@ -0,0 +1,165 @@
...
+// "NirlSYranl" to "t:feed:NirlSYranl:invalid"
+const createInvalidFeedKey = (id) => createFeedKey(id).concat(invalidSuffix);
+// "NirlSYranl" to "t:feed:NirlSYranl:delayed"
+const createDelayedFeedKey = (id) => createFeedKey(id).concat(delayedSuffix);
+
+module.exports = {
+  /**
+   * Feeds
+   */
+  addFeed: async (feed) => {
```
humphd, line 28: In a follow-up PR, we can rip out the feed code from this service, and/or do it when we bring in the feed queue work (cc @c3ho).

src/api/posts/test/posts.test.js
```diff
@@ -0,0 +1,166 @@
...
+  beforeAll(() => posts.map((post) => addPost(post)));
+
+  test('default number of items should be returned', async (done) => {
+    const res = await request(app).get('/');
+
+    expect(res.status).toBe(200);
+    expect(res.get('Content-type')).toContain('application/json');
+    expect(res.get('X-Total-Count')).toBe(createdItems.toString());
+    expect(res.body.length).toBeGreaterThan(0);
+    expect(res.body instanceof Array).toBe(true);
```
humphd, line 37:
```
    expect(Array.isArray(res.body)).toBe(true);
```

src/api/posts/test/posts.test.js
```diff
@@ -0,0 +1,166 @@
...
+    done();
```
humphd, line 38: The way you're writing these tests, you probably don't need `done()` here, it's more when you have a callback function (e.g., if you used supertest's `end(function(...){...})` or something).  It's OK to leave it, or you can try removing too.

src/api/posts/test/posts.test.js
```diff
@@ -0,0 +1,166 @@
...
+  });
+
+  test('requests a specific number of items', async (done) => {
+    const res = await request(app).get('/?per_page=50');
+
+    expect(res.status).toEqual(200);
+    expect(res.get('Content-type')).toContain('application/json');
```
humphd, line 45: Instead of checking this in every test, write one test that does it and you can ignore it in all the others.

src/api/posts/test/posts.test.js
```diff
@@ -0,0 +1,166 @@
...
+    expect(res.get('X-Total-Count')).toBe(createdItems.toString());
```
humphd, line 46: Same thing here and below, you're duplicating tests.

src/api/posts/test/posts.test.js
```diff
@@ -0,0 +1,166 @@
...
+  });
+  test('request posts with non-integer per_page param', async (done) => {
+    const res = await request(app).get(`/?per_page=${nonInteger}`);
+    expect(res.status).toEqual(400);
+    done();
+  });
+});
+
+describe('test /posts/:id responses', () => {
+  // An array of keys used to test whether an id that doesn't exist cannot be found
```
humphd, line 97: This doesn't look like it matches the code it goes with.

src/api/posts/test/posts.test.js
```diff
@@ -0,0 +1,166 @@
...
+  const missingGuid = 'http://missing-guid';
+
+  test('A post with an id should be returned', async (done) => {
+    const res = await request(app).get('/');
+    const postIDRes = await request(app).get(`/${res.body[0].id}`);
```
humphd, line 102: Write another test like this, but use the `url` you get vs. the `id`, to make sure both work.

src/api/posts/test/posts.test.js
```diff
@@ -0,0 +1,166 @@
...
+    expect(postIDRes.get('Content-type')).toContain('application/json');
+    expect(postIDRes.body.id).toEqual(post.id);
+    done();
+  });
+
+  test("Pass an id that doesn't exist", async (done) => {
+    const res = await request(app).get(`/${hash(missingGuid)}`);
+    expect(res.status).toEqual(404);
+    expect(res.get('Content-type')).toContain('application/json');
+    expect(res.body instanceof Array).toBe(false);
```
humphd, line 114: These two checks aren't necessary (we just want a 404)

humphd (changes requested):

humphd:
I filed #1907 and #1908 for follow-up.

src/api/posts/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM node:lts:alphine as base
```
humphd, line 1: Typo here: `alpine` vs. `alphine`

humphd (changes requested):

src/api/posts/src/storage.js
HyperTHD: Sounds good

src/api/docker-compose.yml
```diff
@@ -82,3 +82,61 @@ services:
...
+      # Traefik routing for the image service at /v1/image
+      - 'traefik.http.routers.posts.rule=Host(`${API_HOST}`) && PathPrefix(`/${API_VERSION}/posts`)'
+      # Specify the posts service port
+      - 'traefik.http.services.posts.loadbalancer.server.port=${IMAGE_PORT}'
+      # Add middleware to this route to strip the /v1/posts prefix
+      - 'traefik.http.middlewares.strip_posts_prefix.stripprefix.prefixes=/${API_VERSION}/posts'
+      - 'traefik.http.middlewares.strip_posts_prefix.stripprefix.forceSlash=true'
+      - 'traefik.http.routers.posts.middlewares=strip_posts_prefix'
+
+  # SSO Identity Provider test service, https://simplesamlphp.org
```
humphd, line 120: You shouldn't add the login stuff in this PR, you can remove lines 120-142

src/api/env.development
```diff
@@ -77,3 +77,17 @@ IMAGE_PORT=4444
...
+# Posts Service
+################################################################################
+
+# Posts Service Port (default is 5555)
+POSTS_PORT=5555
+
+# Posts Service URL
+POSTS_URL=http://api.telescope.localhost/v1/posts
+
+
```
humphd, line 91: Remove this extra blank line

src/api/posts/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM node:lts:alpine as base
+
+RUN apk add dumb-init
+
+WORKDIR /app
+
+COPY --chown=node:node . .
+
+RUN npm install ci --only=production
```
humphd, line 9: Change this to:

```
RUN npm install --only=production --no-package-lock
```

src/api/posts/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "@elastic/elasticsearch": "^7.11.0",
+    "@elastic/elasticsearch-mock": "^0.3.0",
+    "@senecacdot/satellite": "^1.8.0",
+    "crypto": "^1.0.1",
+    "express-validator": "^6.10.0",
+    "ioredis": "^4.22.0",
+    "ioredis-mock": "^5.2.4",
+    "jsdom": "^16.4.0",
+    "normalize-url": "^5.3.0",
+    "set-interval-async": "^1.0.34"
```
humphd, line 28: This is unused, remove it.

src/api/posts/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "bugs": {
+    "url": "https://github.com/Seneca-CDOT/telescope/issues"
+  },
+  "homepage": "https://github.com/Seneca-CDOT/telescope#readme",
+  "engines": {
+    "node": ">=12.0.0"
+  },
+  "dependencies": {
+    "@elastic/elasticsearch": "^7.11.0",
+    "@elastic/elasticsearch-mock": "^0.3.0",
```
humphd, line 20: Both of these elastic deps are unused, remove them.

src/api/posts/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "@senecacdot/satellite": "^1.8.0",
```
humphd, line 21: Change this to `^1.x`

src/api/posts/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "crypto": "^1.0.1",
```
humphd, line 22: Remove this, it's unused.

src/api/posts/test/posts.test.js
```diff
@@ -0,0 +1,150 @@
...
+    // This will depend on the env value, so as long as we get back something.
+    expect(res.body.length).toBeGreaterThan(0);
+  });
+
+  test('request posts with non-integer page param', async (done) => {
+    const res = await request(app).get(`/?page=${nonInteger}`);
+    expect(res.status).toEqual(400);
+    done();
+  });
+  test('request posts with non-integer per_page param', async (done) => {
```
humphd, line 75: Newline above this

humphd (changes requested):

PR #1840, opened 2021-02-27
Add Firebase emulator via Docker

humphd:
In #1642, @chrispinkney is adding a service based on Firebase.  This is great for staging and production, but for local development and CI, we need a way to run an offline/mock Firebase.

Luckily, Google provides this via their [Emulator Suite](https://firebase.google.com/docs/emulator-suite).  There's a node version via [firebase-tools](https://www.npmjs.com/package/firebase-tools).

Chris was showing me today how it works on his machine, and it seemed like a perfect candidate for a Docker container, so we can run it like I do with the login container in #1796.

I found [this project](https://github.com/AndreySenov/firebase-tools-docker), which containerized the whole thing.  Based on what Chris learned and taught me, and what I could glean from [the docs](https://github.com/AndreySenov/firebase-tools-docker/blob/master/doc/guide/running_firebase_emulators.md), I was able to get it working without needing to login at all.

To run this, you use `npm run services:start firebase` and then you have two URLs you can access:

1. http://ui.firebase.localhost - the UI dashboard for the emulator tools
2. http://firestore.firebase.localhost - the Firestore service itself

I'll have to get @chrispinkney to test this with his current work on the User service, since I don't have any code that uses this.  But it all appears to work for me.


manekenpix (dismissed):
Tested. LGTM :+1:

humphd:
Had to rebase this and modify it a bit now that the auth service is in, and we're separating the development services from our core services in docker-compose.yml.

manekenpix (approved):

src/api/config/firebase.json
```diff
@@ -0,0 +1,13 @@
+{
+  "emulators": {
+    "ui": {
+      "enabled": true,
+      "host": "0.0.0.0",
+      "port": 4000
+    },
+    "firestore": {
+      "port": "8088",
```
tonyvugithub, line 9: is there a reason why this `port` property is a string while in `ui` it is a number?

tonyvugithub (changes requested):

src/api/config/firebase.json
humphd: Nice catch!  Hilariously, it's like this in the Firebase docs, so I must have copy/pasted and not noticed:

https://firebase.google.com/docs/emulator-suite/install_and_configure#port_configuration

```
{
  // Existing firebase configuration ...
  "database": {
    "rules": "database.rules.json"
  },
  "firestore" {
    "rules": "firestore.rules"
  },

  // ...

  // Optional emulator configuration. Default
  // values are used if absent.
  "emulators": {
    "firestore": {
      "port": "8080"
    },
    "ui": {
      "enabled": true,      // Default is `true`
      "port": 4000          // If unspecified, see CLI log for selected port
    },
    "auth": {
      "port": "9099"
    },
    "functions": {
      "port": "5001"
    },
    "database": {
      "port": "9000"
    },
    "pubsub": {
      "port": "8085"
    }
  }
}
```

I suspect it doesn't matter, but I'll switch to a number for all of them.

humphd:
@chrispinkney ping

birtony (approved):

chrispinkney (approved):
inserted data

Awesome. I was able to confirm that not only does this dockerized version run as expected, data can be inserted into it no problem.

Note: I used my microservice's backend for this and specified the project id [`telescope`](https://github.com/Seneca-CDOT/telescope/pull/1840/files#diff-a3872c2e4d094f394e87eb51d5c934953e79319accea61231188218ba9b184b0R21) in order to make this work. Hoping to have a draft of what I've been up to by tomorrow.

LGTM

humphd:
Awesome, thanks for testing @chrispinkney, really appreciate it, since I have no idea how to test this other than to know it "works."

PR #1843, opened 2021-02-27
React auth implementation using new JWT flow

humphd:
This implements the basics of what's needed in the front-end in order to use the new auth flow from #1796.  It's not possible to test this without also having the back-end auth service, so I might wait until it lands to get reviews, unless you want to merge that branch into this one locally.

This works by handling redirection on login/logout to the auth service and extracting the `?access_token=...` from the URL if it exists.


src/web/src/components/Login/LoggedOut.tsx
```diff
@@ -1,20 +1,11 @@
...
-      button={{
-        href: loginUrl,
-        title: 'Log In',
-        ariaLabel: 'log in',
-        Icon: ExitToAppIcon,
-      }}
-    />
-  );
+  const { login } = useAuth();
+  return <HeaderButton title="Log In" onClick={() => login()} Icon={ExitToAppIcon} />;
```
birtony, line 8: Can't we just do `onClick={login}`?

birtony (commented):
I really like the auth architecture, good job @humphd! Should the `components/MyFeeds.tsx` also be updated to use the new authentication hook? Otherwise, this PR would crash our staging environment, becasue Vercel preview is throwing this error:
image

src/web/src/components/Login/LoggedOut.tsx
humphd: Huh, this is weird that I didn't get this locally.  Thanks a lot for catching it, @birtony: review works!  I'll make a fix and push again.

birtony (approved):

manekenpix (dismissed):

humphd:
Rebased this and had to change quite a bit.  Can I get another review?

src/web/src/components/AuthProvider.tsx
```diff
@@ -0,0 +1,94 @@
...
+export interface AuthContextInterface {
+  login: (returnTo?: string) => void;
+  logout: (returnTo?: string) => void;
+  user?: User;
+  token?: string;
+}
+
+const AuthContext = createContext<AuthContextInterface>({
+  login() {
+    throw new Error('You need to wrap your component in <AuthProvider>');
```
birtony, line 19: We don't want to do login or logout on the server-side so we are throwing here?

humphd:
@birtony that code is sort of confusing.  With React Contexts, you have to provide a default object to use until you have your runtime values.  You'd never call this function directly.  Instead, you'd use the one we provide when we setup the AuthProvider component and use our own runtime implementation.

This is the equivalent of having a method in a base class that needs to be overridden by a sub-class that will provide its own implementation strategy.

humphd:
Can I get some reviews on this soon?  I'd like to land it now that the back-end is in.

birtony (dismissed):

humphd:
@DukeManh I'm not sure how to make eslint happy with that style.  If you know, I'll change it.

src/web/src/User.ts
```diff
@@ -0,0 +1,29 @@
...
+  // TODO: more user details...
+  id: string;
+
+  email: string;
+
+  name: string;
+
+  isAdmin: boolean;
+
+  constructor(id: string, email: string, name: string, isAdmin: boolean) {
```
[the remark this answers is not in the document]
DukeManh: > @DukeManh I'm not sure how to make eslint happy with that style. If you know, I'll change it.
I think these will do it. 
`
    /**
     *  Allow typescript empty constructor
     *  https://eslint.org/docs/rules/no-empty-function#options
     *  https://eslint.org/docs/rules/no-useless-constructor
     */
    'no-useless-constructor': 'off',
    'no-empty-function': ['error', { allow: ['constructors'] }],
`

humphd:
@DukeManh done, and rebased.  Can I get another review?

DukeManh (approved):

Metropass (approved):

PR #1864, opened 2021-03-03
Change MUI typography font to Spartan

### LINKED ISSUE #1856: UI 2.0 Change all fonts from Roboto to Spartan.
In our new UI, we are going to use the font `Spartan`. But our App still uses `Roboto`. 

tonyvugithub:
## Issue This PR Addresses
Fixed #1856 

## Type of Change

## Description
In this PR, I have updated the theme object so everything using MUI component will have Spartan font styling now. I have also removed `@fontsource/roboto` packages as i don't think we need it anymore after this merged. The rest of the changes are simply remove `fontFamily: 'Roboto'` on all components. 

 

## Checklist
image
image

### DIFF
--- a/src/web/package.json
+++ b/src/web/package.json
@@ -8,7 +8,6 @@
     "start": "next start -p 8000"
   },
   "dependencies": {
-    "@fontsource/roboto": "^4.2.1",
     "@fontsource/spartan": "^4.2.1",
     "@material-ui/core": "^4.11.2",
     "@material-ui/icons": "^4.11.2",
@@ -21,8 +20,8 @@
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
     "react-material-ui-form-validator": "^2.1.4",
-    "swr": "^0.4.2",
     "react-use": "^17.1.1",
+    "swr": "^0.4.2",
     "valid-url": "^1.0.9"
   },
   "devDependencies": {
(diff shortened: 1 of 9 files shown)


humphd (commented):
Why pull these fonts into our tree, where they won't get updates vs. leave them in npm, where they will be versioned and we can get updates?  What problem are you solving here?  I'm not saying you don't have a good reason, I just want to understand it from a maintenance point of view.

humphd (approved):
Works well, code looks good.

On a separate note, I don't love this font for body text.  I find it hard to read for a long time, which is what I do on Telescope.  Maybe it's good for titles, but I don't think it's right for content.  I'll probably file something later to revisit that aspect.

tonyvugithub:
@humphd : Initially I thought that MUI theme would not work with `@fontsource` but turns out it does so I made change to the PR to use `@fontsource/spartan` instead of importing the static font files manually.

tonyvugithub:
@PedroFonsecaDEV : on @humphd  comment above, maybe you can have another plan for post body font?

HyperTHD:
I think either solution works. Personally speaking, I would prefer to have the typography post font be something else instead of Spartan, but that's another discussion we need to have. As for doing it in here, I would also prefer it here, but if we need to get this PR in, then we can handle it in another issue

PedroFonsecaDEV (approved):

Issue #1855, opened 2021-02-28
Temporary Avatars for the new UI. 


PedroFonsecaDEV:
**What would you like to be added**:
We could use something like:
~https://evergreen.segment.com/components/avatar/~
https://material-ui.com/components/avatars/#avatar
To generate temporary avatars for our new UI while we wait for GitHub integration.

humphd:
MUI has this already, see https://material-ui.com/components/avatars/

yuanLeeMidori:
@humphd I tried both evergreen and Mui's avatar component, Mui's avatar doesn't seem to generate the initials for avatar, they just provide a circle, the style and the font size, initials, will be handled by ourselves.
But for evergreen, the avatar component also generate the initials for us.

Do we still use Mui avatar? Just curious.

humphd:
That's too bad about MUI not doing the letters.  Here's what Evergreen does to get the initials: https://github.com/segmentio/evergreen/blob/1d0927098597f7cb92461d4e8dfe529e34d1d80c/src/avatar/src/utils/getInitials.js and how they use them https://github.com/segmentio/evergreen/blob/1d0927098597f7cb92461d4e8dfe529e34d1d80c/src/avatar/src/Avatar.js.  Maybe you can hack something like that into our code to use MUI's component, but supply the initials if we don't have an image src?

yuanLeeMidori:
I'm using similar js to get initial with Mui avatar,
![Screenshot_2021-03-10 Telescope(1)](https://user-images.githubusercontent.com/50813726/110827566-31df7280-8264-11eb-8280-39ce01b09e46.png)

![Screen Shot 2021-03-10 at 20 27 16](https://user-images.githubusercontent.com/50813726/110827541-2be99180-8264-11eb-90ff-9fa3f5173a1a.png)

Should I make the component to check if we have img src first, and if we don't have, we apply initials? Or wait till we introduce the img src and only do initials for now? I'm not sure that we have img src for author in our current db.

humphd:
I would make our "Avatar" component take a name and image src URL.  If the latter is missing, fallback to the initials, yes.

This way it will work for every case.

yuanLeeMidori:
`<Avatar name={post.feed.author} img="src" />`

You mean something like this, and it display img src first if it exist; if it doesn't exist, show initials from name value?

humphd:
Yeah, it should be able to handle having an image prop or not (we don't have any data for this yet, but we will...)

[...]

PR #1911, opened 2021-03-12
Fix issue #1855: add temporary avatar

yuanLeeMidori:
## Issue This PR Addresses

Before GitHub integration or till we get the image from author, we need a temporary avatar to replace the current blank circle.
Fixes: #1855 

## Type of Change

## Description

This `PostAvatar` component takes name or img source (currently, I set img to optional because we don't have any image to pass). Ideally, if the img source is provided, it should display the image; if the img is not provided and the name is provided, it'll generate initials as an avatar.

#### Desktop
Screen Shot 2021-03-12 at 10 52 08
#### Mobile
Screenshot_2021-03-12 Telescope

## Checklist

### DIFF
--- a/src/web/src/components/Posts/PostAvatar.tsx
+++ b/src/web/src/components/Posts/PostAvatar.tsx
@@ -1,0 +1,50 @@
+import { createStyles, Theme, Avatar } from '@material-ui/core';
+import { makeStyles } from '@material-ui/core/styles';
+
+type AvatarProps = {
+  name: String;
+  img?: String;
+};
+
+const useStyles = makeStyles((theme: Theme) =>
+  createStyles({
+    circle: {
+      marginLeft: '0.5rem',
+      color: '#e5e5e5',
+      backgroundColor: '#121D59',
+      fontSize: '2.5rem',
+      width: '3em',
+      height: '3em',
+      [theme.breakpoints.down(1200)]: {
+        fontSize: '2em',
+        width: '2.5em',
+        height: '2.5em',
+      },
+    },
+  })
+);
+
+const PostAvatar = ({ name, img }: AvatarProps) => {
+  const classes = useStyles();
+
+  if (img) {
+    return <Avatar src="img" />;
+  }
+
+  if (name) {
+    const nameArr = name.split(' ');
+    let initials;
+    if (nameArr.length >= 2) {
+      const firstName = nameArr[0].substring(0, 1).toUpperCase();
+      const lastName = nameArr[nameArr.length - 1].substring(0, 1).toUpperCase();
+      initials = firstName + lastName;
+    } else if (nameArr.length === 1) {
+      initials = nameArr[0].substring(0, 1).toUpperCase();
+    }
+    return <Avatar className={classes.circle}>{initials}</Avatar>;
+  }
+
+  return <Avatar />;
+};
+
+export default PostAvatar;
--- a/src/web/src/components/Posts/PostAvatar.tsx
+++ b/src/web/src/components/Posts/PostAvatar.tsx
@@ -2,8 +2,8 @@ import { createStyles, Theme, Avatar } from '@material-ui/core';
 import { makeStyles } from '@material-ui/core/styles';
 
 type AvatarProps = {
-  name: String;
-  img?: String;
+  name: string;
+  img?: string;
 };
 
 const useStyles = makeStyles((theme: Theme) =>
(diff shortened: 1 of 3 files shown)


birtony (commented):
Is it expected to have the avatar slightly shifted towards the top of the avatar circle rather than Center it it? @yuanLeeMidori

yuanLeeMidori:
> Is it expected to have the avatar slightly shifted towards the top of the avatar circle rather than Center it it? @yuanLeeMidori

Hmm, no. I'm looking into it.

src/web/src/components/Posts/PostAvatar.tsx
```diff
@@ -0,0 +1,50 @@
...
+);
+
+const PostAvatar = ({ name, img }: AvatarProps) => {
+  const classes = useStyles();
+
+  if (img) {
+    return <Avatar className={classes.avatar} src="img" />;
+  }
+
+  if (name) {
```
humphd, line 34: Based on your TypeScript definition for the props, `name` is always defined.  You can drop this if-check, or maybe do you mean `name.length > 0`?

src/web/src/components/Posts/PostAvatar.tsx
```diff
@@ -0,0 +1,50 @@
...
+    const nameArr = name.split(' ');
+    let initials;
```
humphd, line 36: I wonder if we could simplify this a bit.  Something like:

```js
const initials = name.split(' ').map(word => word[0].toUpperCase()).join('').slice(0,2);
```

This should work for any name.  Question: do we want to split on space and/or dash?

humphd (changes requested):

src/web/src/components/Posts/PostAvatar.tsx
yuanLeeMidori: Currently there is no space between the initials, below are initials without anything, with dot, with dash, with space.

Screenshot_2021-03-12 Telescope(4)
Screenshot_2021-03-12 Telescope(3)
Screenshot_2021-03-12 Telescope(2)
Screenshot_2021-03-12 Telescope(1)

humphd: I don't mean we should include them in the final version, I more mean should we split on `/[ -]/`.  For example, what would you expect your name to be? YH or YL?  I agree that the final form should be 2 characters right beside each other (no space or dashes).

yuanLeeMidori: Ah I see, sorry for the misunderstanding.

I prefer split names by space, to my understanding, dash is more like linking? In my case, Yuan-Hsi is my first name, and everyone from Taiwan is using dash to link our 2-syllable first name. So YL makes much sense.

But I'm not sure which one is better.

humphd: I think you've answered it: we should just split on spaces, and assume a dash is a linked name.  Let's do that.

yuanLeeMidori: > I wonder if we could simplify this a bit. Something like:
> 
> ```js
> const initials = name.split(' ').map(word => word[0].toUpperCase()).join('').slice(0,2);
> ```
> 
> This should work for any name. Question: do we want to split on space and/or dash?

Also, another question, should the initials take the first 2 words' first letters, or the first and the last words' letters?

For example, name "Yuri Joe Chapman" should return YC or YJ? I believe it's YC, but the suggested code from your comment will return YJ instead. I'm not sure which one is preferred?

humphd: Ah, good point.  Yeah, modify my code to do first and last.

yuanLeeMidori: @humphd just pushed new commit, thank you!

src/web/src/components/Posts/PostAvatar.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  if (img) {
+    return <Avatar className={classes.avatar} src="img" />;
+  }
+
+  if (name.length > 0) {
+    const initials = name
+      .split(' ')
+      .map((splitName, i, arr) =>
+        i === 0 || i + 1 === arr.length ? splitName[0].toUpperCase() : null
```
humphd, line 38: Add a comment above this code explaining the logic, since this isn't an obvious way to do this (e.g., other devs might change it without understanding).

humphd (changes requested):

src/web/src/components/Posts/PostAvatar.tsx
```diff
@@ -0,0 +1,47 @@
...
+      },
+    },
+  })
+);
+
+const PostAvatar = ({ name, img }: AvatarProps) => {
+  const classes = useStyles();
+
+  if (img) {
+    return <Avatar className={classes.avatar} src="img" />;
```
HyperTHD, line 31: "img" should be {img} here no? Since you want the actual url

yuanLeeMidori: You're right, even though we don't have img now, but it is highly possible to be url. I'll change it to {img}. We can modify this when we have img value if needed.

humphd (approved):
Let's file a follow-up on the vertical alignment of the initials, and take this in 1.8.

birtony (approved):

yuanLeeMidori:
follow-up issue: #1919

PR #1935, opened 2021-03-13
Remove nginx from development.yml, update docs

humphd:
I initially thought I had to use `nginx` in development to get the Traefik ports to work in production without exposing 80.  I realize now I can do it without, and removing nginx will make things a little lighter in development.

I've also updated the docs to include recent changes that weren't written down.


manekenpix (approved):

humphd:
I need this for another PR, going to merge.

PR #1940, opened 2021-03-13
Use image service for dynamic image component

### LINKED ISSUE #1633: [next] Update DynamicImage to be fully dynamic

The current `DynamicImage` component in the next front-end is just a temporary placeholder.  It needs to be updated to work like this:

- use dynamic sizes (we need the image service being discussed in #1627)
- use different image types (jpg, webp, etc) via the `<picture>` element
- use random images from our Unsplash image set

We need to get the next front-end to have feature parity with gatsby, and this is one of the holes we still have.

### LINKED ISSUE #1939: Switch front-end to use image service for <DynamicImage />
The [DynamicImage](https://github.com/Seneca-CDOT/telescope/blob/2e6fe53b1c05fa9e5e0f505cc14fdd39fd696b92/src/web/src/components/DynamicImage.tsx) component currently uses a single image from our source tree.  Now that the image service is merged, we can switch to use that instead.

humphd:
Fixes #1633 
Fixes #1939

This switches the front-end to use our fancy new image service (hurray!).  It's touching a lot of files, because I also had to fix our docker and next build logic so that we can pass microservices URLs through to the front-end.  Specifically:

- Updated the `Dockerfile` for building next.js, so that it defines and uses `ARG`s, which are build-time variables we define in our `production.yml` file and pass through.  NOTE: you have to re-define `ARG` values in every stage of a docker build, so that's why you see me doing it over and over.  Once I get to the right stage, I shift the values over to being `ENV` variables named `NEXT_PUBLIC_*` so that next can see them in the front-end.  Anyone doing similar work to use a microservice URL in the front-end will have to do this same sort of thing, so this is a guide (cc @HyperTHD, who I expect will want to connect the posts service soon).
- The `next.config.js` now does a few steps to try and get the environment variables it will need.  First, we define them all in an array, and then we try loading them from `.env` (old Telescope 1.0 style) or `env.development` (new Telescope 2.0 style), or we just use them in the existing environment if they are there (e.g., docker build).  Then we copy them over to `NEXT_PUBLIC_*` style variables.
- I then get the `NEXT_PUBLIC_*` values in the front-end and expose them via `config.ts`.  I've reworked it a bit to make it nicer to import individual values, and also gotten rid of all the hold Gatsby style hook calls to `useSiteMetadata()` (we can use `config` everywhere).
- I've updated `src/web/src/components/DynamicImage.tsx` to use a [similar style to what next does with their Dynamic image component](https://image-component.nextjs.gallery/background): I define a bunch of images in a `srcSet` and let the browser figure out which one to load at runtime.  The image service will generate the appropriate image size based on this.
- @tonyvugithub, @huynguyez, @PedroFonsecaDEV: I need some help with the CSS on breakpoints for the image that include the proper sizing for the bottom nav-bar.  I did something, I but I don't think it's right (I left a comment).
- I removed the old static background image, which means that running the front-end will require the image service.  I need to figure out a clean way to let people override the URL to use, such that they can use the local docker version or staging (probably most people will want to do this).

NOTE: I'm going to need to update the variables manually in Vercel to have it pick up the right values.  I'll do that now so it works there.


src/web/src/components/DynamicImage.tsx
```diff
@@ -1,27 +1,41 @@
...
-    backgroundSize: 'cover',
-    transition: 'opacity 1s ease-in-out',
-    height: 'calc(100vh - 64px)',
-    [theme.breakpoints.between('xs', 'sm')]: {
-      height: 'calc(100vh - 56px)',
+    display: 'block',
+    height: '100%',
+    // TODO: this isn't right...
+    [theme.breakpoints.down(1024)]: {
+      height: 'calc(100vh - 64px)',
```
tonyvugithub, line 11: The way you do breakpoints here is legit, since MUI does not have any generic breakpoint at 1024px. Here is the chart of MUI's breakpoints
image
About having a more dynamic height for the bottom bar, I think we can use ref to refer to the `NavBar` component, something like this 
image
And then we can having a context to provide the height using `navBarRef.current.offsetHeight` to the `<Component />` in index page. Also, we could have a hook to use it and update a state in `<DynamicImage />` so it can rerender if there is a change in height of NavBar.
Reference: https://stackoverflow.com/questions/50747953/how-to-get-the-height-of-a-react-element

tonyvugithub (changes requested):
I think it works, just need to update to get the dynamic height of `<NavBar/>`

PedroFonsecaDEV (approved):
LGTM

HyperTHD (approved):
GTM, just needs a rebase. Tested on PC and Samsung Galaxy S7 Edge

PR #1933, opened 2021-03-13
Fix: expose ports in development

manekenpix:
## Type of Change

- [x] **Bugfix**: Change which fixes an issue
- [ ] **New Feature**: Change which adds functionality
- [ ] **Documentation Update**: Change which improves documentation
- [ ] **UI**: Change which improves UI

## Description

This PR exposes `Redis`' and `Elasticsearch`'s ports in development (we removed this in #1706)

For testing, simply start docker in development mode:
- `docker-compose --env-file config/env.development up --build`
or
- `npm run services:start`

and then run Telescope: `npm start`.

Telescope should be able to find `redis` and `elasticsearch`.

## Checklist

- [x] **Quality**: This PR builds and passes our npm test and works locally
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not (if applicable)
- [ ] **Documentation**: This PR includes updated/added documentation to user exposed functionality or configuration variables are added/changed or an explanation of why it does not(if applicable)

PR #1944, opened 2021-03-14
Fix #1943: Remove Docker localhost.* DNS for Windows development

### LINKED ISSUE #1942: /healthcheck for feed-discovery doesn't work through Traefik
Each microservice has a `/healthcheck` endpoint [automatically created by Satellite](https://github.com/Seneca-CDOT/satellite/blob/848aa89e32da50096e0c225abfe6f67f9223404b/src/satellite.js#L33-L40).

The new feed-discovery service has one too, but it only seems to work when run outside directly vs. through Traefik.

You can see what I mean by doing the following:

```
cd  src/api/feed-discovery
npm start
```

Now try to go to http://localhost:7777/healthcheck and you should get back JSON: `{"status":"ok"}`.

Next, try running this through Traefik.  You need Docker started on your machine before you do this:

```
# Go to the root of Telescope
cd ../../..
# Start the image and feed-discovery services
npm run services:start feed-discovery image
```

Compare these two URLs:

- http://api.telescope.localhost/v1/image/healthcheck works
- http://api.telescope.localhost/v1/feed-discovery/healthcheck fails

You can stop the services with `npm run services:stop`.

Looking at the code, I think the problem is in https://github.com/Seneca-CDOT/telescope/blob/master/docker/docker-compose.yml#L112-L118, specifically:

```docker
 - 'traefik.http.routers.feed_discovery.middlewares=strip_feed_discovery_prefix'
```

Notice the rest of that routing definition uses `feed-discovery` (dash) and this line uses `feed_discovery` (underscore).  Subtle bug, but it's breaking the middleware, so the `/v1/feed-discovery` path prefix isn't being stripped.

cc @tonyvugithub.

### LINKED ISSUE #1943: [Windows] api.telescope.localhost refused to connect
**What happened**:
When on Window and trying to access `api.telescope.localhost` domain, the connection cannot be established. 
image

We should be able to access it regardless of your OS
On a Window machine, start up the services using `npm run services:start` and try to curl any of service's URL 
- OS: Window
- Browser: Chrome

humphd:
Fixes #1942 
Fixes #1943 

It turns out that some versions of Docker on Windows can't do custom DNS resolution to things like `api.telescope.localhost`.  I was using this to make it easier to remember all the names of the services we run, but since we can't, I'm switching to use `localhost` and port numbers.

The only service I haven't changed is Elasticsearch which @manekenpix is doing in #1933.  We need to use 9200 in development for elasticsearch vs. `elasticsearch.localhost`.

### DIFF
--- a/config/env.development
+++ b/config/env.development
@@ -14,11 +14,11 @@ COMPOSE_FILE=docker/docker-compose.yml;docker/development.yml
 
 
 # The host where the Telescope 1.0 front-end and back-end are run.
-TELESCOPE_HOST=telescope.localhost
+TELESCOPE_HOST=http://localhost:3000
 
-# The host where all the microservices run (e.g., http://api.telescope.localhost)
+# The host where all the microservices run (e.g., http://localhost)
 # NOTE: if you change this, change all other occurrences below too.
-API_HOST=api.telescope.localhost
+API_HOST=http://localhost
 
 # The API Version, used as a prefix on all routes: /v1
 API_VERSION=v1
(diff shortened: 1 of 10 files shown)


humphd:
@tonyvugithub can you try this all on your local machine?  I'm mostly doing this so you can develop with us :)

humphd:
Rebased on #1933, which just landed.

tonyvugithub (approved):
Worked as expected 
image

humphd:
Awesome, thanks for testing @tonyvugithub.  Glad we got all this stuff sorted out today.

HyperTHD:
I never actually ran into this issue which was super weird. I'll test this tomorrow morning to make sure this works on my end too since I'm also using Windows to develop

HyperTHD (approved):
Tested using posts, image, and firestore emulator routing. feed-discovery health check route works fine as well. Just rebase this and it's GTM

PR #1952, opened 2021-03-15
Configure staging/Vercel to use new auth service

humphd:
I'm working on getting our authentication system moved from the 1.0 back-end to the new auth microservice.  Our SSO has been updated for staging by ITS, and now we need to make some adjustments to the config, build, and front-end code.

This fixes a few issues with ENV variables for staging, uses `AUTH_URL` in our Docker build for the front-end, and passes the value through to next.js, so we can hit the right end-point in staging/production.

I've also included all the various URLs we have for Vercel, so we can test things there against our backend.  NOTE: PR previews will not work, since we get a different sub-domain (and therefore different origin) on each build.  I've only been able to do this for stable URLs (e.g., whatever we merge to `master`).

cc @HyperTHD, this is another good example for you of how you pass the posts service URL through the Docker build to the frontend.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,7 @@
 ARG API_URL
 # API Service URLs, set via ENV in docker or next build
 ARG IMAGE_URL
+ARG AUTH_URL
 
 # Context: Build Context
 FROM node:lts-alpine as build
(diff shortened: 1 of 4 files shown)


manekenpix (approved):

humphd:
It's going to be almost impossible to test this, since we have to ship it to staging to test it.  Also, we only have a few auth accounts on staging (I have one, Josue has one).  On production, it will use regular Seneca accounts.

If you want to test this locally:

```
npm run services:start
npm run dev
```

Now browse to the next frontend and you should be able to click the "Login" link and complete the flow.  The instructions for login credentials is at https://github.com/Seneca-CDOT/telescope/blob/master/docs/login.md#running-an-sso-identity-provider

HyperTHD (approved):

PR #1954, opened 2021-03-16
Use proper AUTH_URL values in next.config.js and AuthProvider

humphd:
@manekenpix and @HyperTHD noticed a few bugs in the way the staging auth fix was being done.  This corrects the problems by properly defining the `AUTH_URL` for next's build, and using it in the `AuthProvider`.

### DIFF
--- a/config/env.development
+++ b/config/env.development
@@ -28,7 +28,7 @@ API_VERSION=v1
 # Auth Service
 ################################################################################
 
-# Auth Service Port (default is 4444)
+# Auth Service Port (default is 7777)
 AUTH_PORT=7777
 
 # Auth Service URL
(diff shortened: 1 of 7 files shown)


config/env.development
```diff
@@ -28,7 +28,7 @@ API_VERSION=v1
 # Auth Service
 ################################################################################
 
-# Auth Service Port (default is 4444)
+# Auth Service Port (default is 7777)
```
chrispinkney, line 31: Doesn't feed discovery use 7777?

humphd: It did, but I've now updated it.  We had two servers on the same port (auth was always on 7777 so it wasn't available when Tony chose it).

birtony (dismissed):
@humphd said we should test this on master

humphd:
Rebased to pickup the posts changes on master, I need two more reviews?  I'd love to get this in soon.

manekenpix (approved):

HyperTHD (approved):
We'll need to merge to see if this works, but I tested locally and it worked just fine

humphd:
Awesome, thanks!

Issue #1607, opened 2021-01-26
Improve SEO Tags and Create a Sitemap


birtony:
**What would you like to be added**:
It would be nice to add meta tags for social media, such as:
```javascript
<!-- Primary Meta Tags -->
<title>Meta Tags — Preview, Edit and Generate</title>
<meta name="title" content="Meta Tags — Preview, Edit and Generate">
<meta name="description" content="With Meta Tags you can edit and experiment with your content then preview how your webpage will look on Google, Facebook, Twitter and more!">

<!-- Open Graph / Facebook -->
<meta property="og:type" content="website">
<meta property="og:url" content="https://metatags.io/">
<meta property="og:title" content="Meta Tags — Preview, Edit and Generate">
<meta property="og:description" content="With Meta Tags you can edit and experiment with your content then preview how your webpage will look on Google, Facebook, Twitter and more!">
<meta property="og:image" content="https://metatags.io/assets/meta-tags-16a33a6a8531e519cc0936fbba0ad904e52d35f34a46c97a2c9f6f7dd7d336f2.png">

<!-- Twitter -->
<meta property="twitter:card" content="summary_large_image">
<meta property="twitter:url" content="https://metatags.io/">
<meta property="twitter:title" content="Meta Tags — Preview, Edit and Generate">
<meta property="twitter:description" content="With Meta Tags you can edit and experiment with your content then preview how your webpage will look on Google, Facebook, Twitter and more!">
<meta property="twitter:image" content="https://metatags.io/assets/meta-tags-16a33a6a8531e519cc0936fbba0ad904e52d35f34a46c97a2c9f6f7dd7d336f2.png">
```
We could probably use https://metatags.io/ to generate the meta tags for us, and also refer to https://css-tricks.com/essential-meta-tags-social-media/ for reference.
In terms of the sitemap, we could refer to https://medium.com/spemer/creating-a-sitemap-generator-for-next-js-3102fb5a297e or https://medium.com/frontend-digest/how-to-build-a-dynamic-sitemap-for-your-next-js-app-c69836c91f8a.
**Why would you like this to be added**:
This would improve our SEO and user experience when sharing links to Telescope on social media - they would have a logo, specified title and description to it, not just a blank url.

birtony:
I will work on the SEO Tags for this issue

HyperTHD:
Gonna also take a look at this as well

aserputov:
@HyperTHD can this one be closed too ?

PR #1962, opened 2021-03-17
Added new keywords, robots.txt, optimized meta tags 

birtony:
## Issue This PR Addresses
Related to #1607 

## Type of Change

## Description
In this PR, I am trying to improve our SEO. I have followed [Open Graph Docs](https://ogp.me/), [Facebook's Guide to Sharing for Webmasters](https://developers.facebook.com/docs/sharing/webmasters), [Twitter's Cards Markup Overview
](https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/markup) and [MetaTags.io](https://metatags.io/). In terms of the image used for the OG/Twitter tags, according to https://metatags.io/, the image's recommended size is 1200x628 pixels. We do not currently have such a logo size, so I've used the closest one and tested it with metatags.io. It looked pretty well to me, so I am not sure if there is a need to create and store another image.
image
image
image
Once merged, we will be able to use https://cards-dev.twitter.com/validator and https://developers.facebook.com/tools/debug/ for testing (both require registration).

I have also added `robots.txt` file to allow web crawlers to index our pages, as recommended by [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Glossary/Robots.txt). [According to Google](https://developers.google.com/search/docs/advanced/robots/robots_txt), it is almost a web standard nowadays and [Twitter also says they use it](https://developer.twitter.com/en/docs/twitter-for-websites/cards/guides/getting-started#crawling). I think I put this file in the right spot (have I?), but when I try to access it locally by running the front-end and going to http://localhost:8000/robots.txt I get a page not found. Take a look at https://github.com/robots.txt or https://twitter.com/robots.txt, for comparison.

Lastly, I have also modified/added some keywords to try to cover more things that users might be searching for that are related to Telescope. I also changed some page titles to try to make them more explicit. All ideas/opinions are welcome! Let's come to a team decision on these things together.

## Checklist

### DIFF
--- a/src/web/src/pages/about.mdx
+++ b/src/web/src/pages/about.mdx
@@ -1,9 +1,10 @@
+import SEO from '../components/SEO';
 import MDXPageBase from '../pages/layouts/MDXPageBase';
 import About from '../markdown-pages/about.md';
 const AboutPage = () => {
   return (
     <>
-      <MDXPageBase title={'About Us'}>
+      <MDXPageBase title={'About | Telescope'}>
         <About />
       </MDXPageBase>
     </>
--- a/src/web/src/pages/myfeeds.tsx
+++ b/src/web/src/pages/myfeeds.tsx
@@ -1,5 +1,4 @@
-import Head from 'next/head';
-
+import SEO from '../components/SEO';
 import BackToTopButton from '../components/BackToTopButton';
 import useAuth from '../hooks/use-auth';
 import MyFeeds from '../components/MyFeeds';
@@ -14,9 +13,7 @@ const MyFeedsPage = () => {
 
   return (
     <>
-      <Head>
-        <title>My Feeds</title>
-      </Head>
+      <SEO pageTitle="My Feeds | Telescope" />
       <BackToTopButton />
       <main className="main">
         <MyFeeds user={user} />
--- a/src/web/src/pages/search.tsx
+++ b/src/web/src/pages/search.tsx
@@ -4,7 +4,7 @@ import SearchPage from '../components/SearchPage';
 const Search = () => {
   return (
     <>
-      <SEO pageTitle="Search" />
+      <SEO pageTitle="Search | Telescope" />
       <SearchPage />
     </>
   );
--- a/src/web/public/robots.txt
+++ b/src/web/public/robots.txt
@@ -1,0 +1,2 @@
+# https://www.robotstxt.org/robotstxt.html
+User-agent: *
--- a/src/web/src/components/SEO.tsx
+++ b/src/web/src/components/SEO.tsx
@@ -9,6 +9,9 @@ type SEOProps = {
 const SEO = ({ pageTitle }: SEOProps) => {
   const { pathname } = useRouter();
   const currentUrl = `${telescopeUrl}${pathname}`;
+  // This variable explicitly points to our production API_URL.
+  // For reference, see https://developers.google.com/search/docs/advanced/crawling/consolidate-duplicate-urls#rel-canonical-header-method
+  const canonicalUrl = `https://telescope.cdot.systems${pathname}`;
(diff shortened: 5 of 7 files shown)


src/web/src/pages/myfeeds.tsx
```diff
@@ -14,9 +13,7 @@ const MyFeedsPage = () => {
 
   return (
     <>
-      <Head>
-        <title>My Feeds</title>
-      </Head>
+      <SEO pageTitle="My Feeds | Telescope" />
```
HyperTHD, line 16: Wouldn't this always result in My Feeds being printed instead of Telescope and thus, not worth keeping?

src/web/src/pages/search.tsx
```diff
@@ -4,7 +4,7 @@ import SearchPage from '../components/SearchPage';
 const Search = () => {
   return (
     <>
-      <SEO pageTitle="Search" />
+      <SEO pageTitle="Search | Telescope" />
```
HyperTHD, line 7: Same thing here too

src/web/src/pages/about.mdx
```diff
@@ -1,9 +1,10 @@
+import SEO from '../components/SEO';
 import MDXPageBase from '../pages/layouts/MDXPageBase';
 import About from '../markdown-pages/about.md';
 const AboutPage = () => {
   return (
     <>
-      <MDXPageBase title={'About Us'}>
+      <MDXPageBase title={'About | Telescope'}>
```
HyperTHD, line 7: Same thing here as well

HyperTHD (commented):
Nice to see this getting added. Have a few things I wanna know about

DukeManh (approved):

src/web/src/pages/myfeeds.tsx
birtony: @HyperTHD our current setup actually overwrites `Telescope` on each page, if you think about it in this way. This is page-specific, so for each page, we will have a page title (`My Feeds`, etc) + `| Telescope`. Otherwise when you share specific pages, for example, their titles will just say `About Us`, etc., which is not very explicit. A lot of websites use this format.

src/web/robots.txt
```diff
@@ -0,0 +1,2 @@
+# https://www.robotstxt.org/robotstxt.html
```
PedroFonsecaDEV, line 1: I think robots.txt should be inside `public` folder`
src/web/public/robots.txt`

https://nextjs.org/blog/next-9-1
```
Starting from Next.js 9.1 we are introducing a new directory called public.

Any files in the public directory will be mapped to the root of the domain.

For example, public/robots.txt will be mapped to /robots.txt.
```

PedroFonsecaDEV (changes requested):

src/web/robots.txt
birtony: Oh you're right! Thanks for the heads up 👍

birtony: I've moved it to the public folder and I can access the file under http://localhost:8000/robots.txt

PedroFonsecaDEV (dismissed):

humphd (changes requested):
This looks good to me.  One thing we should do is add a [canonical](https://yoast.com/rel-canonical/) link to the head so that identical pages on staging/production both point to production.

src/web/src/pages/myfeeds.tsx
HyperTHD: Ah, makes sense. Thanks for the clarifications!

HyperTHD (dismissed):
Rebase and you good to go

humphd (approved):

PedroFonsecaDEV (approved):

PR #1990, opened 2021-03-19
Changed site background in light theme to #FFFFFF

### LINKED ISSUE #1991: Change site background in light theme to White
A white background
Since white is the new grey

tonyvugithub:
## Issue This PR Addresses
Fixed #1991 

## Type of Change

## Description
Changed the background of site in light theme to #FFFFFF instead of #E5E5E5.

### DIFF
--- a/src/web/src/theme/index.ts
+++ b/src/web/src/theme/index.ts
@@ -21,7 +21,7 @@ export const lightTheme: Theme = createMuiTheme({
       main: red.A400,
     },
     background: {
-      default: '#E5E5E5',
+      default: '#FFFFFF',
       paper: '#d1d1d1',
     },
     text: {
@@ -29,7 +29,7 @@ export const lightTheme: Theme = createMuiTheme({
       secondary: '#121D59',
     },
     action: {
-      selected: '#ffffff',
+      selected: '#FFFFFF',
     },
   },
 });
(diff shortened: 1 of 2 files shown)


src/web/src/theme/index.ts
```diff
@@ -21,15 +21,15 @@ export const lightTheme: Theme = createMuiTheme({
...
+      default: '#FFFFFF',
       paper: '#d1d1d1',
     },
     text: {
       primary: '#000000',
       secondary: '#121D59',
     },
     action: {
-      selected: '#ffffff',
+      selected: '#FFFFFF',
```
humphd, line 32: What does `action.selected` mean?  When is it used, and should it be different from the regular background or the same?

tonyvugithub: this is due to @izhuravlev commit earlier for the search page. so when you click on a filter, it made the selected filter white box. I was surprised to but did not want to remove it for now.

izhuravlev: Action: selected should be different from the background. Search bar changes it's color on hover to theme.palette.action.selected. 
I added this so we can set up 2 different colours for both modes.

PedroFonsecaDEV (approved):

src/web/src/theme/index.ts
izhuravlev: To make sure everything looks good, click or hover on the search bar in the light mode to see how it looks.

birtony (commented):
@PedroFonsecaDEV do we use the upper- or the lowercase notation for hex values? I think we should standardize this and configure Prettier/Eslint accordingly.

PedroFonsecaDEV:
@birtony sounds good. Filed https://github.com/Seneca-CDOT/telescope/issues/1994 to fix this.

tonyvugithub:
@birtony : I think uppercase should be the way to go

birtony (approved):

PR #2008, opened 2021-03-22
Fixes #2000 redirect URL error

### LINKED ISSUE #2000: Logout error on /myfeeds
**What happened**:
Logging out from /myfeeds page returns an error.

Even though the logout is successful, it should redirect to the previous page. 

image

DukeManh:
## Issue This PR Addresses
Fixes #2000

## Type of Change

## Description
@humphd In this PR I have fixed the error when we log out from /myfeeds by allowing relative `redirect_uri` and add Telescope base URL to it.  However, there's another problem, when we call `logout()`, before the logout URL is reached, `myfeeds` component re-renders because `user` is now undefined, and it redirects us to login and cancels the logout request. I couldn't come up with an easy solution to prevent this. I was wondering if you can help me with this issue?  

## Checklist

### DIFF
--- a/src/web/src/pages/myfeeds.tsx
+++ b/src/web/src/pages/myfeeds.tsx
@@ -8,7 +8,7 @@ const MyFeedsPage = () => {
 
   // Redirect the user to login if they aren't authenticated already
   if (!user) {
-    return login('/myfeeds');
+    return login(`/myfeeds`);
   }
 
   return (
--- a/src/web/src/components/AuthProvider.tsx
+++ b/src/web/src/components/AuthProvider.tsx
@@ -4,11 +4,11 @@ import { useRouter } from 'next/router';
 import { nanoid } from 'nanoid';
 
 import User from '../User';
-import { loginUrl, logoutUrl } from '../config';
+import { loginUrl, logoutUrl, telescopeUrl } from '../config';
 
 export interface AuthContextInterface {
   login: (returnTo?: string) => void;
-  logout: (returnTo?: string) => void;
+  logout: () => void;
   user?: User;
   token?: string;
 }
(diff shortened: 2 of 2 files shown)


DukeManh:
@izhuravlev I missed the first part of the triage, I guess you were assigned to help me out? We might need to change the `authProvider` context to fix this.

humphd (changes requested):
The `redirect_uri` should never be relative.  If it is, it's a bug that should be fixed at the point where we send it to the auth server.

DukeManh:
> The `redirect_uri` should never be relative. If it is, it's a bug that should be fixed at the point where we send it to the auth server.

@humphd, thank you, I wasn't sure what way to do it, changing `redirect_uri` passed or making backend more relaxed. How can we go about fixing canceled logout request ?

humphd:
@DukeManh The fix for this is to always redirect back to the root of the application on logout.  [This code](https://github.com/Seneca-CDOT/telescope/blob/aef750f81fa39b91a46c8cca140289a82589b424/src/web/src/components/AuthProvider.tsx#L105-L113) needs to be changed:

```tsx
import { telescopeUrl, loginUrl, logoutUrl } from '../config';
...
const logout = () => {
    // Clear our existing token and state
    removeToken();
    removeAuthState();
    window.location.href = `${logoutUrl}?redirect_uri=${telescopeUrl}`;
};
```

You'll have to update the types for `logout` in this file as well.

src/web/src/pages/myfeeds.tsx
```diff
@@ -2,13 +2,14 @@ import SEO from '../components/SEO';
...
 import MyFeeds from '../components/MyFeeds';
+import { telescopeUrl } from '../config';
 
 const MyFeedsPage = () => {
   const { login, user } = useAuth();
 
   // Redirect the user to login if they aren't authenticated already
   if (!user) {
-    return login('/myfeeds');
+    return login(`${telescopeUrl}/myfeeds`);
```
humphd, line 12: I wonder if we should change the way that `login` works so that you can pass us a relative path, and we'll do the `${telescopeUrl}/` prefix for you.  That way people don't have to pull in `telescopeUrl` from config in order to accomplish this.

If so, `login` would need to be updated to something like this:

```tsx
    const url = new URL(returnTo || '', telescopeUrl)
    window.location.href = `${loginUrl}?redirect_uri=${url.href}&state=${loginState}`;
```

humphd (changes requested):
Looks good, one more thing.

DukeManh:
@humphd, I have updated the redirect_uri being passed to the `logout` function and it fixes the schema error. Though, the `logout` request is still canceled because `myfeeds` component rerenders and makes us `login` as in the picture.
image
It'll be the best to land this PR first and then we can file another issue for this.

src/web/src/pages/myfeeds.tsx
DukeManh: Will we need to call `encodeURIComponent` on `url` ?

humphd (dismissed):

HyperTHD (dismissed):

humphd (approved):

HyperTHD (approved):

HyperTHD:
@DukeManh Rebase this, and you gtm

birtony (approved):

Issue #2019, opened 2021-03-24
New environment setup instruction


yuanLeeMidori:
**What would you like to be added**:
A new document like `environment-setup.md` to tell developers how to run the application with microservices.
**Why would you like this to be added**:
Currently, we're moving to microservices. This brings big changes of how to run the application and the old docs might not be helpful. I want to add a new document for confused developers like me with some simple instructions.

yuanLeeMidori:
Close by #2079

PR #2020, opened 2021-03-24
Create temporary env-setup for microservices transition

yuanLeeMidori:
## Issue This PR Addresses
Fixes: #2019

## Type of Change

## Description
This PR adds a new document to help with the transition of microservice. The old `environment-setup.md` might not work. This new document should be able to help developers during the transition.

Please leave comments to help me improve this document! Thank you :grinning:

## Checklist

### DIFF
--- a/docs/tmp-env-setup.md
+++ b/docs/tmp-env-setup.md
@@ -1,0 +1,23 @@
+With the new microservices shipping, the old [environment-setup](environment-setup) document might not work. In the transition, this document may be helpful for developers. Here are some instructions for different scenarios, they are based on [next.config.js](../src/web/next.config.js) code logic. This `next.config.js` file configures which `env` file to use. It reads `.env` in the root first, if there is no `.env` in the root, it reads [env.development](../config/env.development).
+
+- Only want to work on front-end, don't need anything from back-end.
+
+  ```
+  # 1. copy config/env.staging to .env.
+  # If you have old .env, it is good to delete it before copy to prevent overwrite error
+  cp config/env.staging .env
+  # 2. start services
+  npm run services:start
+  # 3. run front-end application
+  npm run dev
+  ```
+
+- Want to run back-end services locally, and to have front-end using the services.
+
+  This is the default setting, you **don't need to copy or modify** any env file. Just run `npm run services:start` and `npm run dev`. However, if it doesn't work and you have old `.env` that was copied from old `.env.example`, delete it.
+
+- Want to mix and match services between local and staging.
+
+  This one depends on which part you're working on. For example, if you want to work with authorization, you need to specify the URL of AUTH in `.env`, which is, go to `.env` and find `AUTH_URL=...`. If you're testing auth locally, use `AUTH_URL=http://localhost/v1/auth`; otherwise, use staging, `AUTH_URL=http://dev.api.telescope.cdot.systems/v1/auth`
+
+  Then, do `npm run services:start`, and `npm run dev`.
--- a/docs/tmp-env-setup.md
+++ b/docs/tmp-env-setup.md
@@ -1,23 +1,29 @@
 With the new microservices shipping, the old [environment-setup](environment-setup) document might not work. In the transition, this document may be helpful for developers. Here are some instructions for different scenarios, they are based on [next.config.js](../src/web/next.config.js) code logic. This `next.config.js` file configures which `env` file to use. It reads `.env` in the root first, if there is no `.env` in the root, it reads [env.development](../config/env.development).
 
-- Only want to work on front-end, don't need anything from back-end.
+- Want to run back-end services locally, and to have front-end using the services.
+
+  This is the default setting, you **don't need to copy or modify** any env file. Just run `npm run services:start` and `npm run dev`. However, if it doesn't work and you have old `.env` that was copied from old `.env.example`, delete it.
+
+- Only want to work on front-end, and won't make any changes on the back-end.
 
   ```
   # 1. copy config/env.staging to .env.
   # If you have old .env, it is good to delete it before copy to prevent overwrite error
   cp config/env.staging .env
-  # 2. start services
-  npm run services:start
-  # 3. run front-end application
+  # 2. run front-end application
   npm run dev
   ```
-
-- Want to run back-end services locally, and to have front-end using the services.
-
-  This is the default setting, you **don't need to copy or modify** any env file. Just run `npm run services:start` and `npm run dev`. However, if it doesn't work and you have old `.env` that was copied from old `.env.example`, delete it.
 
 - Want to mix and match services between local and staging.
 
   This one depends on which part you're working on. For example, if you want to work with authorization, you need to specify the URL of AUTH in `.env`, which is, go to `.env` and find `AUTH_URL=...`. If you're testing auth locally, use `AUTH_URL=http://localhost/v1/auth`; otherwise, use staging, `AUTH_URL=http://dev.api.telescope.cdot.systems/v1/auth`
 
   Then, do `npm run services:start`, and `npm run dev`.
+
+- Want to run auth/image/posts service _solely_.
+
+  `npm run services:start auth` or `npm run services:start image` or `npm run services:start posts`
+
+- After making changes and would like to make Docker image reflects,
+
+  `npm run services:clean` and `npm run services:start` to rebuild the image.


docs/tmp-env-setup.md
```diff
@@ -0,0 +1,23 @@
+With the new microservices shipping, the old [environment-setup](environment-setup) document might not work. In the transition, this document may be helpful for developers. Here are some instructions for different scenarios, they are based on [next.config.js](../src/web/next.config.js) code logic. This `next.config.js` file configures which `env` file to use. It reads `.env` in the root first, if there is no `.env` in the root, it reads [env.development](../config/env.development).
+
+- Only want to work on front-end, don't need anything from back-end.
+
+  ```
+  # 1. copy config/env.staging to .env.
+  # If you have old .env, it is good to delete it before copy to prevent overwrite error
+  cp config/env.staging .env
+  # 2. start services
+  npm run services:start
```
DukeManh, line 10: We don't have to run microservices locally while only developing the frontend.

docs/tmp-env-setup.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Want to run back-end services locally, and to have front-end using the services.
+
+  This is the default setting, you **don't need to copy or modify** any env file. Just run `npm run services:start` and `npm run dev`. However, if it doesn't work and you have old `.env` that was copied from old `.env.example`, delete it.
+
+- Want to mix and match services between local and staging.
+
+  This one depends on which part you're working on. For example, if you want to work with authorization, you need to specify the URL of AUTH in `.env`, which is, go to `.env` and find `AUTH_URL=...`. If you're testing auth locally, use `AUTH_URL=http://localhost/v1/auth`; otherwise, use staging, `AUTH_URL=http://dev.api.telescope.cdot.systems/v1/auth`
+
+  Then, do `npm run services:start`, and `npm run dev`.
```
DukeManh, line 23: Here if we're only interested in Auth service, then we can run `services:start auth` to start up Auth service locally only.

docs/tmp-env-setup.md
```diff
@@ -0,0 +1,23 @@
...
+  npm run dev
+  ```
+
+- Want to run back-end services locally, and to have front-end using the services.
+
+  This is the default setting, you **don't need to copy or modify** any env file. Just run `npm run services:start` and `npm run dev`. However, if it doesn't work and you have old `.env` that was copied from old `.env.example`, delete it.
+
+- Want to mix and match services between local and staging.
+
+  This one depends on which part you're working on. For example, if you want to work with authorization, you need to specify the URL of AUTH in `.env`, which is, go to `.env` and find `AUTH_URL=...`. If you're testing auth locally, use `AUTH_URL=http://localhost/v1/auth`; otherwise, use staging, `AUTH_URL=http://dev.api.telescope.cdot.systems/v1/auth`
```
DukeManh, line 21: It's also worth mentioning that, to make a Docker image reflects the change we make to its code, we should rebuild the image by running `services:clean` and start it up again.

DukeManh (changes requested):

docs/tmp-env-setup.md
HyperTHD: I disagree. What you're saying only applies to developers working on UI changes in the front-end, in which case, the services aren't needed. But if they're making a change that requires a service to run, such as the search service in order to test the search page, then they should run the specific service needed

DukeManh: @yuanLeeMidori says this if for`Only want to work on front-end, don't need anything from back-end.`. I don't see why we need to run services locally here.

yuanLeeMidori: Sorry that I make you confused. Like @HyperTHD said, for most front-end changes, they still need the service running in order to get the posts, for example. In my opinion, it is rare that developers can work with just pure front-end without service running to provide functionalities.

How about, `Only want to work on front-end, and won't make any changes on the back-end.`?

docs/tmp-env-setup.md
yuanLeeMidori: Yes, that could be another scenario. I'll add one for starting different services solely.

Just to clarify, the example I made here is when you need auth to be locally, which means you might still need other services such as post. It's a mix situation (e.g., auth locally but perhaps post staging). I just want to specify that developer can change the url with different needs (local or staging).

docs/tmp-env-setup.md
yuanLeeMidori: Will do, thank you for the reminder.

docs/tmp-env-setup.md
DukeManh: I think that's how I understand it. When you `# 1. copy config/env.staging to .env.`, your frontend is already using services running on Vercel. Correct me if I'm wrong but on my machine, I can develop frontend and see posts and banner without opening running docker.

yuanLeeMidori: @DukeManh Just tried it, you're right. I'll update it!

yuanLeeMidori:
@DukeManh Just pushed a new commit, may you re-review it? Thanks!

rjayroso:
Hey Yuan, it looks good and was very helpful for me to get my posts to load. One hiccup while following the steps was that every time I ran the `npm run services:start` it would tell me that `docker-compose` was not recognized despite it being installed on my machine - I think you need to add a line noting that Docker must be running in order for the steps below to work since after I ran Docker everything was working properly.

docs/tmp-env-setup.md
```diff
@@ -0,0 +1,29 @@
+With the new microservices shipping, the old [environment-setup](environment-setup) document might not work. In the transition, this document may be helpful for developers. Here are some instructions for different scenarios, they are based on [next.config.js](../src/web/next.config.js) code logic. This `next.config.js` file configures which `env` file to use. It reads `.env` in the root first, if there is no `.env` in the root, it reads [env.development](../config/env.development).
+
+- Want to run back-end services locally, and to have front-end using the services.
+
+  This is the default setting, you **don't need to copy or modify** any env file. Just run `npm run services:start` and `npm run dev`. However, if it doesn't work and you have old `.env` that was copied from old `.env.example`, delete it.
+
+- Only want to work on front-end, and won't make any changes on the back-end.
+
```
rjayroso, line 8: You should note that Docker must be running before executing the steps below.

rjayroso (changes requested):

docs/tmp-env-setup.md
```diff
@@ -0,0 +1,29 @@
...
+  ```
+  # 1. copy config/env.staging to .env.
+  # If you have old .env, it is good to delete it before copy to prevent overwrite error
+  cp config/env.staging .env
+  # 2. run front-end application
+  npm run dev
+  ```
+
+- Want to mix and match services between local and staging.
```
c3ho, line 17: Before or after this section we should mention our original API docs like `You can find the API docs at <link to API docs>`

Then go into the more specific use cases either here or in the doc:

```Our documentation regarding our new APIs are available at (API doc link here).

To run all services: npm run services: start

The option to run only selected services is also available by using the command: npm run services: start <service1> <service2>

Prior to running the commands. Please make sure the .env file is configured appropriately.

For example (use the auth example from line 19 and on here)

**Review by c3ho — CHANGES_REQUESTED** [review]

**Comment by c3ho** [comment]
During the meeting we mentioned the old docs wouldn't cut it anymore with the new front-end and back-end we should work towards overhauling all the docs we have and not treat this as a temporary solution.

**Comment by yuanLeeMidori** [comment]
@c3ho Yes, but I'm not sure if we're able to start new changes in all related docs. To my understanding, microservice porting has not completed, therefore, isn't it hard to re-write/modify old docs now?

**Comment by c3ho** [comment]
True, we're not done porting, but a lot of the code is already in (we're just waiting for two more microservices to go in). What I mean is, don't treat this as a transition document and assume most of it will be used to update our `environment-setup.md` or will replace that document outright.

I'm mentioning this because if you follow the environment-setup.md steps in our master branch. The document will fail you at `After installing the prerequisites` section.

**Comment by yuanLeeMidori** [comment]
@c3ho I see! Yes, environment-setup is the main reason that I created this document. It's mostly out-dated and new comer cannot use it to run the application.

How about this, instead of keep updating this "tmp" document, I go through the old documents and update them step by step? For example, start with environment-setup, remove the outdated ones and try to update the instructions that work in current code.

File: docs/tmp-env-setup.md [hunk 6]
```diff
@@ -0,0 +1,29 @@
+With the new microservices shipping, the old [environment-setup](environment-setup) document might not work. In the transition, this document may be helpful for developers. Here are some instructions for different scenarios, they are based on [next.config.js](../src/web/next.config.js) code logic. This `next.config.js` file configures which `env` file to use. It reads `.env` in the root first, if there is no `.env` in the root, it reads [env.development](../config/env.development).
+
+- Want to run back-end services locally, and to have front-end using the services.
+
+  This is the default setting, you **don't need to copy or modify** any env file. Just run `npm run services:start` and `npm run dev`. However, if it doesn't work and you have old `.env` that was copied from old `.env.example`, delete it.
+
```
c3ho, line 6: We should add a blurb here, all commands should be run in the root of the telescope project. 

"If any of the commands below commands are failing, use the command `pwd` to find your current directory and navigate back to project root or provide an example of it here"

c3ho (changes requested):

docs/tmp-env-setup.md
c3ho: Yes, I personally think we should go with that.

birtony (commented):
I really like to see this effort @yuanLeeMidori! I think you should not hesitate and update the [`environment-setup.md`](https://github.com/Seneca-CDOT/telescope/blob/master/docs/environment-setup.md) directly to reflect on the changes we've made to the project. It might be helpful to let developers know that currently there are two ways to run things and then link them to two sections of the documentation: the old and the new one. We should make sure that services previously launched in an old-fashion way are shut down and do not interfere with the new ones when developer tries to runt he project.

yuanLeeMidori:
Thank you for all valuable comments! Really appreciate. This PR will be closed and I'll send another PR to modify `environment-setup.md` (see issue #2079)

PR #2022, opened 2021-03-25
Fixes issue-1929: Paginated get route

### LINKED ISSUE #1929: Add authorization middleware to Users service
The Users service exposes a `GET /` route, which lets you get all the users in Firebase.  We need to improve its current form in a few ways:

1. Only an admin should be able to do this (i.e., regular users or the general public can't do this request).  Satellite supports this via middleware:
```js
const { isAuthenticated, isAuthorized } = require('@senecacdot/satellite');
...
router.get('/', isAuthenticated(), isAuthorized({ roles: ["admin"] }), async (req, res, next) => {...}
```
2. We need to add paging.  Firestore [supports this](https://firebase.google.com/docs/firestore/query-data/query-cursors).  Let' follow the same pattern we have with [paging in the posts service](https://github.com/Seneca-CDOT/telescope/blob/master/src/api/posts/src/routes/posts.js#L8-L72), which uses `?page=...&per_page=...`

chrispinkney:
## Issue This PR Addresses
Closes #1929

## Type of Change

## Description
This PR paginates the get route in the users micro-service. The get route can now be used with a paginated front-end to request a specific number of users (defaulted to 5 at a time.)

I wrote a pretty basic test which demonstrates the functionality of this PR. I can always add more tests if required but I wasn't too happy with how hacky the code in tests file was getting. I'm up for suggestions.

## How to test
1. `npm run services:start firebase`
2. `npm run jest:e2e /src/api/users/test/e2e`

or

1. `npm run services:start firebase`
2. Add `process.env.FIRESTORE_EMULATOR_HOST = 'localhost:8088';` to `src\api\users\src\services\firestore.js`
3. `cd src/api/users && npm run dev`
3. `POST` a bunch of generic users with different `id`s using postman (see below)
4. `GET` `http://localhost:6666/?per_page=5&page=1`, this will get 5 users on page 1. You can change up both numbers at will to test more functionality.

####
Test User:
```
{
    "id": "1",
    "firstName": "Chris",
    "lastName": "Pinkney",
    "isAdmin": true,
    "displayName": "I am Not Chris",
    "isFlagged": true,
    "feeds": [
        "https://dev.to/feed/chrispinkney"
    ],
    "github": {
        "username": "chrispinkney",
        "avatarUrl": "https://avatars.githubusercontent.com/u/7242003?s=460&u=733c50a2f50ba297ed30f6b5921a511c2f43bfee&v=4"
    }
}
```

## Checklist

### DIFF
--- a/src/api/users/README.md
+++ b/src/api/users/README.md
@@ -62,7 +62,7 @@ By default the server is running on <http://localhost:6666/>.
 
 \- `GET /:id` - returns 200 with the user specified by the id, or 404 if a user does not exist.
 
-\- `GET /` - returns 200 with all Telescope users in an array, or 404 if the `users` collection is empty.
+\- `GET /?per_page=20&page=1` - returns 200 with the first 20 Telescope users (numerically sorted by `id`) in an array, or 404 if the `users` collection is empty.
 
 \- `POST /` - returns 201 if a Telescope user was successfully validated and added to the db, or 400 if the user already exists. (_An example of the JSON data to send as the POST body can be found in `api/user/test/user.test.js`_)
 
--- a/src/api/users/src/models/celebrateSchema.js
+++ b/src/api/users/src/models/celebrateSchema.js
@@ -1,5 +1,13 @@
 const { celebrate, Joi, Segments } = require('celebrate');
 
+const validatePagingParams = () =>
+  celebrate({
+    [Segments.QUERY]: {
+      per_page: Joi.number().integer().min(1).max(20).default(20),
+      page: Joi.number().integer().min(1).default(1),
+    },
+  });
+
 const validateId = () =>
   celebrate({
     [Segments.PARAMS]: {
@@ -10,7 +18,7 @@ const validateId = () =>
 const validateUser = () =>
   celebrate({
     [Segments.BODY]: Joi.object().keys({
-      id: Joi.number().required(),
+      id: Joi.number().integer().required(),
       firstName: Joi.string().required(),
       lastName: Joi.string().required(),
       displayName: Joi.string().default(''),
@@ -21,10 +29,9 @@ const validateUser = () =>
         username: Joi.string(),
         avatarUrl: Joi.string(),
       }).default({ username: '', avatarUrl: '' }),
-      created: Joi.string(),
-      updated: Joi.string(),
     }),
   });
 
 exports.validateUser = validateUser;
 exports.validateId = validateId;
+exports.validatePagingParams = validatePagingParams;
--- a/src/api/users/src/routes/users.js
+++ b/src/api/users/src/routes/users.js
@@ -1,6 +1,6 @@
 const { Router, logger } = require('@senecacdot/satellite');
 const { errors } = require('celebrate');
-const { validateId, validateUser } = require('../models/celebrateSchema');
+const { validatePagingParams, validateId, validateUser } = require('../models/celebrateSchema');
 const User = require('../models/users');
 const db = require('../services/firestore');
 
@@ -28,27 +28,35 @@ router.get('/:id', validateId(), async (req, res, next) => {
 });
 
 // get all users
-// rejected if the user collection is empty, otherwise users returned
-router.get('/', async (req, res, next) => {
+// rejected if the users collection is empty
+// otherwise n (specified via params) users are returned
+router.get('/', validatePagingParams(), async (req, res, next) => {
+  /*
+   *  celebrateSchema.js performs data validation via validatePagingParams() middleware
+   *  per_page is an integer validated to have a range between 1 and 20
+   *  page is an integer validated to have a min value of 1
+   */
+  const { per_page: perPage, page } = req.query;
+
+  // When requesting page 1 the user we start to build the query off of will be user 0
+  // when requesting page >= 2 the user we start at is determined by perPage * (page - 1)
+  const userToStartAt = page === 1 ? 0 : perPage * (page - 1);
+  const users = [];
+
   try {
-    const usersRef = db.collection('users');
-    const snapshot = await usersRef.get();
+    const query = await db
+      .collection('users')
+      .orderBy('id')
+      .startAt(userToStartAt)
+      .limit(perPage)
+      .get();
 
-    if (snapshot.empty) {
-      logger.debug(`All users were requested by ${req.ip} but could not be found.`);
-      res.status(404).json({
-        msg: `That collection could not be found, or contains no data.`,
-      });
-    } else {
-      logger.debug(`All users were requested by ${req.ip} and served successfully.`);
+    query.forEach((doc) => {
+      users.push(doc.data());
+    });
 
-      const usersArray = [];
-      snapshot.forEach((doc) => {
-        usersArray.push(doc.data());
-      });
-
-      res.status(200).json(usersArray);
-    }
+    logger.debug(`Users were requested by ${req.ip} and served successfully.`);
+    res.status(200).json(users);
   } catch (err) {
     next(err);
   }
--- a/src/api/users/test/e2e/users.test.js
+++ b/src/api/users/test/e2e/users.test.js
@@ -11,6 +11,8 @@ const getUser = (id) => request(app).get(`/${id}`);
 
 const getUsers = () => request(app).get('/');
 
+const getUsersParams = (perPage, page) => request(app).get(`/?per_page=${perPage}&page=${page}`);
+
 const createUser = async (editedUser = {}) => {
   const defaultUser = {
     // Use a unique id number for each user
@@ -94,6 +96,159 @@ describe('GET REQUESTS', () => {
         feeds: ['https://dev.to/feed/carlsagan'],
         github: {
           username: 'carlsagan',
+          avatarUrl:
+            'https://avatars.githubusercontent.com/u/7242003?s=460&u=733c50a2f50ba297ed30f6b5921a511c2f43bfee&v=4',
+        },
+      },
+    ]);
+  });
+
+  test('Accepted - Get all users (using params), only two users should be returned', async () => {
+    // create 3 users
+    const galileo1 = await createUser({ id: 10001 });
+    const galileo2 = await createUser({ id: 10002 });
+    const galileo3 = await createUser({ id: 10003 });
+
+    expect(galileo1.response.status).toBe(201);
+    expect(galileo2.response.status).toBe(201);
+    expect(galileo3.response.status).toBe(201);
+
+    // request 3 users per page on page 1, three users should be returned
+    const response = await getUsersParams(3, 1);
+    expect(response.status).toBe(200);
+    expect(response.body.length).toBe(3);
+    expect(response.body).toEqual([
+      {
+        id: 10001,
+        firstName: 'Galileo',
+        lastName: 'Galilei',
+        displayName: 'Galileo Galilei',
+        isAdmin: true,
+        isFlagged: true,
+        feeds: ['https://dev.to/feed/galileogalilei'],
+        github: {
+          username: 'galileogalilei',
+          avatarUrl:
+            'https://avatars.githubusercontent.com/u/7242003?s=460&u=733c50a2f50ba297ed30f6b5921a511c2f43bfee&v=4',
+        },
+      },
+      {
+        id: 10002,
+        firstName: 'Galileo',
+        lastName: 'Galilei',
+        displayName: 'Galileo Galilei',
+        isAdmin: true,
+        isFlagged: true,
+        feeds: ['https://dev.to/feed/galileogalilei'],
+        github: {
+          username: 'galileogalilei',
+          avatarUrl:
+            'https://avatars.githubusercontent.com/u/7242003?s=460&u=733c50a2f50ba297ed30f6b5921a511c2f43bfee&v=4',
+        },
+      },
+      {
+        id: 10003,
+        firstName: 'Galileo',
+        lastName: 'Galilei',
+        displayName: 'Galileo Galilei',
+        isAdmin: true,
+        isFlagged: true,
+        feeds: ['https://dev.to/feed/galileogalilei'],
+        github: {
+          username: 'galileogalilei',
+          avatarUrl:
+            'https://avatars.githubusercontent.com/u/7242003?s=460&u=733c50a2f50ba297ed30f6b5921a511c2f43bfee&v=4',
+        },
+      },
+    ]);
+  });
+
+  test('Accepted - Get 20 (uncreated) users using params, receive empty array', async () => {
+    // create no users, request 20 users
+    const response = await getUsersParams(20, 1);
+    expect(response.status).toBe(200);
+    expect(response.body.length).toBe(0);
+    expect(response.body).toEqual([]);
+  });
+
+  test('Accepted - Test negative perPage parameter, positive page parameter', async () => {
+    const galileo = await createUser({ id: 10001 });
+    expect(galileo.response.status).toBe(201);
+
+    const response = await getUsersParams(-20, 1);
+    expect(response.status).toBe(400);
+    expect(response.body.validation.query.message).toEqual(
+      '"per_page" must be greater than or equal to 1'
+    );
+  });
+
+  test('Accepted - Test negative perPage parameter, negative page parameter', async () => {
+    const galileo = await createUser({ id: 10001 });
+    expect(galileo.response.status).toBe(201);
+
+    const response = await getUsersParams(-20, -1);
+    expect(response.status).toBe(400);
+    expect(response.body.validation.query.message).toEqual(
+      '"per_page" must be greater than or equal to 1'
+    );
+  });
+
+  test('Accepted - Test positive perPage parameter, negative page parameter', async () => {
+    const galileo = await createUser({ id: 10001 });
+    expect(galileo.response.status).toBe(201);
+
+    const response = await getUsersParams(20, -1);
+    expect(response.status).toBe(400);
+    expect(response.body.validation.query.message).toEqual(
+      '"page" must be greater than or equal to 1'
+    );
+  });
+
+  test('Accepted - Test page and per_page parameters', async () => {
+    await createUser({ id: 0 });
+    await createUser({ id: 1 });
+
+    // create 2 users, request 1 user per page and page 2
+    const response = await getUsersParams(1, 2);
+    expect(response.body.length).toBe(1);
+    expect(response.body).toEqual([
+      {
+        id: 1,
+        firstName: 'Galileo',
+        lastName: 'Galilei',
+        displayName: 'Galileo Galilei',
+        isAdmin: true,
+        isFlagged: true,
+        feeds: ['https://dev.to/feed/galileogalilei'],
+        github: {
+          username: 'galileogalilei',
+          avatarUrl:
+            'https://avatars.githubusercontent.com/u/7242003?s=460&u=733c50a2f50ba297ed30f6b5921a511c2f43bfee&v=4',
+        },
+      },
+    ]);
+  });
+
+  test('Accepted - Test parameterized get, post one user, ask for five, receive one user', async () => {
+    const galileo = await createUser({ id: 10001 });
+
+    expect(galileo.response.status).toBe(201);
+
+    // request 5 users per page on page 2
+    const response = await getUsersParams(5, 2);
+    expect(response.status).toBe(200);
+    expect(response.body.length).toBe(1);
+    expect(response.body).toEqual([
+      {
+        id: 10001,
+        firstName: 'Galileo',
+        lastName: 'Galilei',
+        displayName: 'Galileo Galilei',
+        isAdmin: true,
+        isFlagged: true,
+        feeds: ['https://dev.to/feed/galileogalilei'],
+        github: {
+          username: 'galileogalilei',
           avatarUrl:
             'https://avatars.githubusercontent.com/u/7242003?s=460&u=733c50a2f50ba297ed30f6b5921a511c2f43bfee&v=4',
         },


humphd (dismissed):
Question: why do you not always use the `startAfter` cursor method vs. `limit` in many cases below?  Isn't it always possible to use a cursor to avoid the problems we discussed (i.e., can't you construct a query that uses `startAfter` or `startAt` for any position, even the page 1 case starting at 0)?

chrispinkney:
> (i.e., can't you construct a query that uses `startAfter` or `startAt` for any position, even the page 1 case starting at 0)?

Do you mean something like this?
```
if (page === 1) {
  const query = await db
    .collection('users')
    .orderBy('id')
    .startAt(1) <-- assuming perPage=5 = records 1-5
    .limit(perPage)
    .get();

  nextQuery.forEach((doc) => {
    usersArray.push(doc.data());
  });
if (page === 2) {
  const query = await db
    .collection('users')
    .orderBy('id')
    .startAfter(perPage) <-- assuming perPage=5 & page=2 = records 6-10
    .limit(perPage)
    .get();

  nextQuery.forEach((doc) => {
    usersArray.push(doc.data());
  });
} else {
  const query = await db
    .collection('users')
    .orderBy('id')
    .startAfter(perPage*(page - 1)) <-- assuming perPage=5 & page=3 = records 11-15
    .limit(perPage)
    .get();

  nextQuery.forEach((doc) => {
    usersArray.push(doc.data());
  });
}
```

Something like this could work to grab exactly how many users we want but there's still a few issues:
1. Won't the `id`s not be this simple (i.e. user 1, 2, 3, 4, etc.)? [Not sure what the hash returns](https://github.com/Seneca-CDOT/telescope/issues/1642#issuecomment-774355152).
2. [There's no way to know how many documents exist in a collection without grabbing the entire collection](https://stackoverflow.com/questions/46554091/cloud-firestore-collection-count) but as an alternative we could keep a counter and increment/decrement it in the `POST`/`DELETE` routes and use that number. We'd need to do this to know when to stop grabbing users (I do this [here](https://github.com/Seneca-CDOT/telescope/pull/2022/files#diff-f7f5bb5469b7259c5a222c50526058bcdfbec1a0f5cfeca13c613e83c6782b0fR59)) or else we'd make a query for undefined users.

So assuming my math is correct in the `startAfter`s from above, I'm not sure how this could work without redoing how we assign a users `id`. I also might be barking up the wrong tree entirely, either way LMK!

src/api/users/README.md
```diff
@@ -61,7 +61,9 @@ By default the server is running on <http://localhost:6666/>.
 
 \- `GET /:id` - returns 200 with the user specified by the id, or 404 if a user does not exist.
 
-\- `GET /` - returns 200 with all Telescope users in an array, or 404 if the `users` collection is empty.
+\- `GET /` - returns 200 with the first 5 Telescope users (numerically sorted by `id`) in an array, or 404 if the `users` collection is empty.
```
humphd, line 64: Consolidate this line and the next into one, since `/` is the same as `/?per_page=10&page=1`, it just defines default values.

Also, I think 5 is the wrong default.  Something like 10 or 20 seems better (i.e., if I was using this in a web page, I'd always want to show more than 5 at once, but probably not more than 20).

src/api/users/src/routes/users.js
```diff
@@ -28,29 +28,50 @@ router.get('/:id', validateId(), async (req, res, next) => {
...
 
 // get all users
-// rejected if the user collection is empty, otherwise users returned
+// rejected if the users collection is empty
+// otherwise n (specified via params) users are returned
 router.get('/', async (req, res, next) => {
-  try {
-    const usersRef = db.collection('users');
-    const snapshot = await usersRef.get();
+  const page = !parseInt(req.query.page, 10) ? 1 : parseInt(req.query.page, 10); // if req.query.page is 0 or unparsable set to 1, otherwise set to the req.query.page
```
humphd, line 34: Move your comments above vs. beside code, so the lines don't extend so far to the right.  Ideally we scroll down, not over when coding.

src/api/users/src/routes/users.js
```diff
@@ -28,29 +28,50 @@ router.get('/:id', validateId(), async (req, res, next) => {
...
+  const perPage = !parseInt(req.query.per_page, 10) ? 5 : parseInt(req.query.per_page, 10);
```
humphd, line 35: Let's use 10 or 20 as our default vs. 5.

src/api/users/src/routes/users.js
```diff
@@ -28,29 +28,50 @@ router.get('/:id', validateId(), async (req, res, next) => {
...
+  const usersArray = [];
```
humphd, line 36: Get rid of `Array` suffix.

src/api/users/src/routes/users.js
humphd: Also, use `Math.min()` here so that we only allow getting values between 10 and 100 (or some other upper-bound).  Don't let the user define some massive number.  For example, [GitHub's Users API](https://docs.github.com/en/rest/reference/users#list-users) sets an upper limit of 100.

src/api/users/src/routes/users.js
```diff
@@ -28,29 +28,50 @@ router.get('/:id', validateId(), async (req, res, next) => {
...
-      const usersArray = [];
-      snapshot.forEach((doc) => {
+    const query = await db
+      .collection('users')
+      .orderBy('id')
+      .startAt(userToStartAt)
+      .limit(perPage)
+      .get();
+
+    // if we're requesting too many users, return the last n (perPage) users in the db
```
humphd, line 55: What does this mean?

src/api/users/src/routes/users.js
```diff
@@ -28,29 +28,50 @@ router.get('/:id', validateId(), async (req, res, next) => {
...
-      res.status(200).json(usersArray);
+      logger.debug(`Users were requested by ${req.ip} and served successfully.`);
+      return res.status(200).json(usersArray.reverse()); // reversed to order numerically
     }
+
+    query.forEach((doc) => {
+      usersArray.push(doc.data());
+    });
+
+    logger.debug(`Users were requested by ${req.ip} and served successfully.`);
```
humphd, line 71: Let's not bother logging this, since we'll get default HTTP logging of the route info and success/failure automatically with satellite.

humphd (changes requested):

src/api/users/src/routes/users.js
chrispinkney: > Also, use Math.min() here

Love it!! Thanks!

src/api/users/src/routes/users.js
chrispinkney: Say a user requests something like `/?per_page=5&page=100` but only 1 user exists in the db, Firebase returns an empty array, hence `if (!query.size)`. I set it up to return the last `perPage` amount of users in this case.

I'm failing at succinctly saying that here.

humphd: If I request a page that doesn't exist, give back the empty array.  I think Firebase's behaviour is correct, and we should do the same.

chrispinkney: kk will remove then

HyperTHD (approved):
Tested, works like a charm. Rebase this and you good to go

src/api/users/src/routes/users.js
```diff
@@ -28,27 +28,41 @@ router.get('/:id', validateId(), async (req, res, next) => {
...
 // get all users
-// rejected if the user collection is empty, otherwise users returned
+// rejected if the users collection is empty
+// otherwise n (specified via params) users are returned
 router.get('/', async (req, res, next) => {
-  try {
-    const usersRef = db.collection('users');
-    const snapshot = await usersRef.get();
+  // if req.query.page is 0 or unparsable set to 1, otherwise set to the req.query.page
+  const page = !parseInt(req.query.page, 10) ? 1 : parseInt(req.query.page, 10);
```
humphd, line 35: Could we cache `parseInt(req.query.page, 10)` in a variable instead of doing it 4 times?

src/api/users/src/routes/users.js
```diff
@@ -28,27 +28,41 @@ router.get('/:id', validateId(), async (req, res, next) => {
 });
 
 // get all users
-// rejected if the user collection is empty, otherwise users returned
+// rejected if the users collection is empty
+// otherwise n (specified via params) users are returned
 router.get('/', async (req, res, next) => {
```
humphd, line 33: Have you added celebrate validation for these query params?

src/api/users/test/e2e/users.test.js
```diff
@@ -101,6 +103,51 @@ describe('GET REQUESTS', () => {
     ]);
   });
 
+  test('Accepted - Get all users (using params), only two users should be returned', async () => {
```
humphd, line 106: Let's add some tests for per_page and page, to make sure both work as expected.

humphd (dismissed):

src/api/users/src/models/celebrateSchema.js
```diff
@@ -1,5 +1,13 @@
 const { celebrate, Joi, Segments } = require('celebrate');
 
+const validateGetParams = () =>
+  celebrate({
+    [Segments.PARAMS]: {
+      per_page: Joi.string(),
```
humphd, line 6: These should both be numbers, not strings

src/api/users/src/models/celebrateSchema.js
```diff
@@ -28,3 +36,4 @@ const validateUser = () =>
 
 exports.validateUser = validateUser;
 exports.validateId = validateId;
+exports.validateGetParams = validateGetParams;
```
humphd, line 39: `validatePagingParams`

src/api/users/test/e2e/users.test.js
```diff
@@ -101,6 +103,87 @@ describe('GET REQUESTS', () => {
...
+        github: {
+          username: 'galileogalilei',
+          avatarUrl:
+            'https://avatars.githubusercontent.com/u/7242003?s=460&u=733c50a2f50ba297ed30f6b5921a511c2f43bfee&v=4',
+        },
+      },
+    ]);
+  });
+
+  test('Accepted - Get 20 uncreated users (using params), receive empty array', async () => {
```
humphd, line 152: Add a test with only `per_page` and only `page`, make sure defaults get applied as you expect.

src/api/users/src/routes/users.js
```diff
@@ -28,27 +28,41 @@ router.get('/:id', validateId(), async (req, res, next) => {
...
-    } else {
-      logger.debug(`All users were requested by ${req.ip} and served successfully.`);
+  const users = [];
+  let userToStartAt;
 
-      const usersArray = [];
-      snapshot.forEach((doc) => {
-        usersArray.push(doc.data());
-      });
+  if (page === 1) {
```
humphd, line 44: I'm unclear why you need to special-case 1 and 0, and further, why you're having them behave so differently, or why you'd not put the 0 case first in this list.

At the very least this needs a comment explaining your logic; however, I'm hesitant to have 3 separate code-paths through here like this for what should be doable with 1.  If we end up needing all 3, you need separate tests that exercise all 3 paths.

humphd (changes requested):

src/api/users/src/models/celebrateSchema.js
chrispinkney: Weird. I just tested this and it's working fine as `number`. I'm not sure if you remember but I had to make:
```
const validateId = () =>
  celebrate({
    [Segments.PARAMS]: {
      id: Joi.string().required(), <---- this a string
    },
  });
```
`id` into a string. Must be because I'm parsing `page` and `perPage` as ints whereas I'm not parsing `id` into an int, causing Firebase throws when attempting to get a doc using a number instead of a string (e.g. `Error: Value for argument "documentPath" is not a valid resource path. Path must be a non-empty string.`)

Cool though, thanks for the suggestion.

src/api/users/src/routes/users.js
```diff
@@ -28,27 +28,41 @@ router.get('/:id', validateId(), async (req, res, next) => {
...
-router.get('/', async (req, res, next) => {
-  try {
-    const usersRef = db.collection('users');
-    const snapshot = await usersRef.get();
+// rejected if the users collection is empty
+// otherwise n (specified via params) users are returned
+router.get('/', validateGetParams(), async (req, res, next) => {
+  // if req.query.page is 0 or unparsable set to 1, otherwise set to the req.query.page
+  const parseReqPage = parseInt(req.query.page, 10);
+  const page = !parseReqPage ? 1 : parseReqPage;
```
c3ho, line 36: What happens here if I enter a -1 for page? Should we add a  condition that checks for negative values or 0?

src/api/users/src/routes/users.js
```diff
@@ -28,27 +28,41 @@ router.get('/:id', validateId(), async (req, res, next) => {
...
+  // if req.query.per_page is 0 or unparsable, set to 20, otherwise set to the min of either req.query.page or 20
+  const parseReqPerPage = Math.min(parseInt(req.query.per_page, 10), 20);
+  const perPage = !parseReqPerPage ? 20 : parseReqPerPage;
```
c3ho, line 39: Same as above comment

c3ho (changes requested):

src/api/users/src/routes/users.js
chrispinkney: This was actually needed before I started parsing and using conditions to ensure only clean values can make it in. 

`const userToStartAt = page === 1 ? 0 : perPage * (page - 1)` this currently works as expected. Would this be okay?
If the page specified is the first page, the value to get users would naturally start at user 0, otherwise start at `perPage * (page - 1)`

src/api/users/src/routes/users.js
chrispinkney: Great question! Entering a negative number here returns the first page, and I think this behaviour is okay (I'm actually not sure why this works)

src/api/users/src/routes/users.js
chrispinkney: However entering a negative number here breaks stuff. What do you think about wrapping it in Math.abs to force the number to be positive?

`const parsePerPage = Math.abs(Math.min(parseInt(req.query.per_page, 10), 20));`

chrispinkney: Actually that might be a little hacky. What about this?

`const parsePerPage = req.query.per_page < 0 ? 20 : Math.min(parseInt(req.query.per_page, 10), 20);` (20 is what we're using as our default)

src/api/users/src/routes/users.js
c3ho: Let me understand this properly with the current code with `perPage = 20`

If I start at page 1, I'm getting the first 20 users
If I start at page 0, I'm getting 21 - 40 users?

src/api/users/src/routes/users.js
c3ho: I'm fine with that suggestion, although I'd break it up since `parsing` a value is different than checking perPage. I'd keep line 38 as-is. Then change line 39 to do the conditional check with an or statement.

Also, I think @manekenpix might hate magic numbers (for default value) though and prefers it being set in an env.

src/api/users/src/routes/users.js
c3ho: Should figure out if this behavior is cause of the db or something else. I feel like code shouldn't just "magically work" because of a idiot-proof mechanic that might only be specific to that tool. If this behavior is caused the result of a specific db, what happens if in the future we switch dbs to say mongo and this breaks?

src/api/users/src/routes/users.js
chrispinkney: @c3ho Requesting a page <= 1 always returns page 1 (for reasons that currently elude me, though I can always hardcode values <= 1 to always equal 1).

Requesting pages > 1 currently works as expected.

src/api/users/src/routes/users.js
chrispinkney: Okay awesome, I just hardcoded both page and perPage to disallow numbers less than 1.

I don't think setting this in an env is necessary, though I do appreciate it being a magic number. However, I can always change it if requested. It'll just get a bit hairy as I'll have to require in the env file in case we're running this outside of docker mode (i.e. `npm run dev`), which AFAIK we're trying to move away from.

src/api/users/src/routes/users.js
chrispinkney: This should be taken care of since I just hardcoded the values to not accept numbers < 1

src/api/users/src/routes/users.js
c3ho: Yes, the else case should def work cause 20 * (2 - 1) = 20,

But What I mean is do we need the (page === 0) in this case, isn't it the same as page === 1? Or am I missing something?

can we not do :

```
// Calculate how we're chunking the results
const startIndexAt = perPage * (page - 1)

// If page is <1 we start index at 0 otherwise we use the pagination
const userToStartAt = startIndexAt<= 0 ? 0 : startIndexAt
```

chrispinkney: The `page === 0` case is gone since the latest commit (🥳) and replaced by `const userToStartAt = page === 1 ? 0 : perPage * (page - 1)` which does exactly that. Currently with your conditional suggestion it works as expected when perPage or page are < 0 or 1

c3ho (changes requested):

src/api/users/test/e2e/users.test.js
```diff
@@ -101,6 +103,125 @@ describe('GET REQUESTS', () => {
...
+        feeds: ['https://dev.to/feed/galileogalilei'],
+        github: {
+          username: 'galileogalilei',
+          avatarUrl:
+            'https://avatars.githubusercontent.com/u/7242003?s=460&u=733c50a2f50ba297ed30f6b5921a511c2f43bfee&v=4',
+        },
+      },
+    ]);
+  });
+
```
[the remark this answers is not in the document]
chrispinkney: Awesome, great idea. I don't think it's extra I think it's practical, plus it should only take me a few minutes to ~~copy paste~~ write.

src/api/users/src/routes/users.js
```diff
@@ -28,27 +28,34 @@ router.get('/:id', validateId(), async (req, res, next) => {
...
-    if (snapshot.empty) {
-      logger.debug(`All users were requested by ${req.ip} but could not be found.`);
-      res.status(404).json({
-        msg: `That collection could not be found, or contains no data.`,
-      });
-    } else {
-      logger.debug(`All users were requested by ${req.ip} and served successfully.`);
+  const users = [];
+  // start at user 0 otherwise calculate what user to start at
+  const userToStartAt = page === 1 ? 0 : perPage * (page - 1);
```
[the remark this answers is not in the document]
chrispinkney: I just pushed a change for this, I think it's alright but LMK, I still have to rebase anyway.

src/api/users/src/models/celebrateSchema.js
```diff
@@ -1,5 +1,13 @@
 const { celebrate, Joi, Segments } = require('celebrate');
 
+const validatePagingParams = () =>
+  celebrate({
+    [Segments.PARAMS]: {
+      per_page: Joi.number(),
```
humphd, line 6: Make your validator do more work for you:

```js
per_page: Joi.number().integer().min(1).max(100),
page: Joi.number().integer().min(1),
```

src/api/users/src/routes/users.js
```diff
@@ -28,27 +28,35 @@ router.get('/:id', validateId(), async (req, res, next) => {
...
-// rejected if the user collection is empty, otherwise users returned
-router.get('/', async (req, res, next) => {
-  try {
-    const usersRef = db.collection('users');
-    const snapshot = await usersRef.get();
+// rejected if the users collection is empty
+// otherwise n (specified via params) users are returned
+router.get('/', validatePagingParams(), async (req, res, next) => {
+  // if req.query.per_page is 0 or unparsable, set to 20, otherwise set to the min of either req.query.page or 20
+  const parsePerPage = req.query.per_page < 0 ? 20 : Math.min(parseInt(req.query.per_page, 10), 20);
```
humphd, line 35: 100 per page is probably a more reasonable max.

src/api/users/test/e2e/users.test.js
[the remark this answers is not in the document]
humphd: Do the validation changes above, and then you can write a few tests that expect to get 400s (invalid params).

humphd (changes requested):

src/api/users/src/models/celebrateSchema.js
chrispinkney: Awesome, thanks so much for this.

birtony (commented):
These look more like unit tests, which would make sense for me to have to test the functionality being added in this PR. However, they are located under `e2e` folder, which makes me wonder if they actually need to be e2e tests? Why don't we test this with basic unit tests first and then add e2e iteratively?
cc @humphd @chrispinkney

humphd:
Re: e2e vs. unit, we have to use the Firebase service, and we need to change the tests to use the public service API.  We have another issue on that, so it's fine for now.

src/api/users/src/routes/users.js
```diff
@@ -28,27 +28,36 @@ router.get('/:id', validateId(), async (req, res, next) => {
...
-    const usersRef = db.collection('users');
-    const snapshot = await usersRef.get();
+// rejected if the users collection is empty
+// otherwise n (specified via params) users are returned
+router.get('/', validatePagingParams(), async (req, res, next) => {
+  // if req.query.per_page is < 1 or unparsable, set to 100, otherwise set to the min of either req.query.page or 100
+  const parsePerPage =
+    req.query.per_page < 1 ? 100 : Math.min(parseInt(req.query.per_page, 10), 100);
+  const perPage = !parsePerPage || parsePerPage < 1 ? 100 : parsePerPage;
+  // if req.query.page is 0 or unparsable set to 1, otherwise set to the req.query.page
```
humphd, line 38: It can't be 0 now, based on your validation rules, this comment should be updated.

src/api/users/src/routes/users.js
```diff
@@ -28,27 +28,35 @@ router.get('/:id', validateId(), async (req, res, next) => {
...
 // get all users
-// rejected if the user collection is empty, otherwise users returned
-router.get('/', async (req, res, next) => {
-  try {
-    const usersRef = db.collection('users');
-    const snapshot = await usersRef.get();
+// rejected if the users collection is empty
+// otherwise n (specified via params) users are returned
+router.get('/', validatePagingParams(), async (req, res, next) => {
+  // if req.query.per_page is < 1 or unparsable, set to 100, otherwise set to the min of either req.query.page or 100
```
humphd, line 34: This comment still doesn't reflect what the code is doing (100 vs. 20).  Either fix or remove it.

src/api/users/src/routes/users.js
```diff
@@ -28,27 +28,35 @@ router.get('/:id', validateId(), async (req, res, next) => {
...
+  const parsePerPage = req.query.per_page < 1 ? 20 : Math.min(parseInt(req.query.per_page, 10), 20);
```
humphd, line 35: Why are you assuming `req.query.per_page` is a number at the start of this (comparing to 1), but then you parse it as if it's a string later on?

src/api/users/src/routes/users.js
```diff
@@ -28,27 +28,35 @@ router.get('/:id', validateId(), async (req, res, next) => {
...
+  const perPage = !parsePerPage || parsePerPage < 1 ? 20 : parsePerPage;
```
humphd, line 36: No magic numbers: 20 should be defined above and used by name, `const defaultPerPage = 20;`

src/api/users/test/e2e/users.test.js
```diff
@@ -101,6 +103,158 @@ describe('GET REQUESTS', () => {
...
+    // create 3 users
+    const galileo1 = await createUser({ id: 10001 });
+    const galileo2 = await createUser({ id: 10002 });
+    const galileo3 = await createUser({ id: 10003 });
+
+    expect(galileo1.response.status).toBe(201);
+    expect(galileo2.response.status).toBe(201);
+    expect(galileo3.response.status).toBe(201);
+
+    // request 2 users per page on page 1, only two users should be returned
```
humphd, line 116: This comment doesn't match what you do in the next line

humphd (changes requested):

src/api/users/src/routes/users.js
chrispinkney: I added some default values to celebrate and removed all the validation logic we had in `users.js`, I figured since celebrate does the heavy lifting for us there's no reason to ensure that a number is an int, has a range, etc, since celebrate does all this for us and more.

What do you think?

humphd (approved):
So much better!

manekenpix (approved):

PR #2026, opened 2021-03-25
Fixes #1955: Replace normal hyphen with a non-breaking hyphen

### LINKED ISSUE #1955: Author Name Bug - Replace normal hyphen with a NON-BREAKING HYPHEN
**What happened**:
Right now are experiencing this: 
Screen Shot 2021-03-15 at 8 59 18 PM
Screen Shot 2021-03-15 at 8 59 02 PM

Royce's name should have the same behaviour as David's name:
```
Royce
Ayroso-ong
``` 
NON-BREAKING HYPHEN:
https://www.toptal.com/designers/htmlarrows/punctuation/non-breaking-hyphen/

**Environment**:

- OS:
- Browser:

rjayroso:
## Issue This PR Addresses
Fixes #1955 

## Type of Change

## Description
My name was being broken at the hyphen. I added a line of code to replace all instances of a normal hyphen within the author's name with a [non-breaking hyphen](https://www.toptal.com/designers/htmlarrows/punctuation/non-breaking-hyphen/). 

| Before |
| :---: |
| Screen Shot 2021-03-15 at 8 59 02 PM |
| After |
| image |

It was [mentioned](https://github.com/Seneca-CDOT/telescope/issues/1955#issuecomment-800256159) in the original issue that a possible fix was to expand the author section, especially since there are other authors with even longer names so I messed around with expanding the width/ giving the authors section a minimum width. This created unintended side effects when resizing the screen to be smaller as the names would begin to clip into the right side of the screen. That being said I'd like to know what you guys think - I can file an issue to better display(?) longer names.

## Checklist

## How to Test
View Vercel deployment and scroll down and see each name, when you see my name you will notice that my last name breaks as a whole instead of in the middle where the hyphen is.

### DIFF
--- a/src/web/src/components/Posts/PostInfo.tsx
+++ b/src/web/src/components/Posts/PostInfo.tsx
@@ -11,6 +11,9 @@ const useStyles = makeStyles((theme: Theme) =>
       flexDirection: 'column',
       justifyContent: 'center',
       borderLeft: '2.5px solid #707070',
+      [theme.breakpoints.up('lg')]: {
+        width: '22rem',
+      },
     },
     authorAvatarContainer: {
       shapeOutside: 'circle(50%) border-box',


yuanLeeMidori (dismissed):

HyperTHD:
Is there anyway we can test this with the list of names that David provided in the original issue? I know that currently, the vercel deployment for this will use the original backend which we no longer use so the posts won't load. 

@rjayroso Can you run this against the current Telescope backend using a few of those names from the issue this PR closes and add some screenshots regarding them?

HyperTHD (changes requested):
I'm gonna block this PR until you add some more screenshots using a few of the names David provided in the original issue. Once we know it works for more than just your name, we can merge this one in. No issue with the code, just hard to test with the other names atm since vercel deployment won't load in the posts until search service is in

PedroFonsecaDEV (approved):
Screen Shot 2021-03-26 at 1 46 33 AM

NON-BREAKING HYPHEN: 
HEX CODE:`&#x2011`

To test, you need to copy the char and paste it in some [code converter](https://r12a.github.io/app-conversion/) and check the HEX CODE. 

Nice work @rjayroso

humphd (changes requested):
I'm not sure this is a good solution.  First, the way you're doing your code, if you really want to make that replacement, you should use a unicode character vs. literal, since it's very hard to see that hyphen as being different in JS.

However, this breaks how hyphenation works on the web.  [Hyphens are supposed to provide wrap opportunities to the browser](https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/), and removing that seems wrong.

I would expand the width of this container.  We have more room on the page, why limit it to such a narrow region?

rjayroso:
> I'm not sure this is a good solution. First, the way you're doing your code, if you really want to make that replacement, you should use a unicode character vs. literal, since it's very hard to see that hyphen as being different in JS.
> 
> However, this breaks how hyphenation works on the web. [Hyphens are supposed to provide wrap opportunities to the browser](https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/), and removing that seems wrong.
> 
> I would expand the width of this container. We have more room on the page, why limit it to such a narrow region?

I've tried using CSS but it seems to me that, to make the hyphen non-breakable you will also subject the whole author name to also be unbreakable.

What I currently have up is: I've expanded the author info container and changed the string literal to a Unicode character and looks like so:
ezgif com-gif-maker

Though, I was thinking should we remove the Unicode stuff altogether? It won't solve the original hyphen-break issue but would still allow for the accommodation of larger names without messing with how the hyphen behaves.

src/web/src/components/Posts/PostInfo.tsx
```diff
@@ -67,6 +70,10 @@ type Props = {
 };
 const PostDesktopInfo = ({ authorName, postDate, blogUrl, postUrl }: Props) => {
   const classes = useStyles();
+
+  // Replace normal hyphen with non-breaking hyphen
```
humphd, line 74: agree, let's not do this.  I don't think it's a good idea.

humphd (changes requested):
Personally, I think the wrapping here when we hit a narrower size is fine.  If you think it's wrong, we should change our breakpoint logic so we hide the name, or switch to putting the name on the top vs. right-hand side sooner.

humphd (approved):

chrispinkney (dismissed):

humphd (approved):

HyperTHD (approved):

PR #2035, opened 2021-03-26
CSS changes to author info section

### LINKED ISSUE #1995: Tweaking author info section
I will change some padding/margin/font and its position.

Screen Shot 2021-03-19 at 1 21 41 PM

huynguyez:
UPDATED 

Closes #1995 
- only show it clear on mouse hover, less distraction on both sides while reading

Screen Shot 2021-03-26 at 1 09 34 PM

- add marginBottom to make 'sticky author info box' jump earlier (ugly to see both author close together on scrolling)
Screen Shot 2021-03-26 at 1 15 45 PM

Update:
with Left align in blog content, the info section placed on the right doesn't seem right. It's better on the left of the blog instead: 

Screen Shot 2021-04-03 at 1 59 55 PM

Screen Shot 2021-04-03 at 2 02 10 PM

The vertical separator bar is unnecessary and made it looks just unsightly in modern day design, or if you guys wanna keep it, leave it thin and use lighter color.

or remove it and add more margin between blog content (example for Medium)
Screen Shot 2021-04-03 at 1 58 22 PM

ALso about the Navigation menu, no matter what color you use, it will never pops out at first load + banner image in the current design. 
My suggestion: Return it back to horizontal top, hidden when scroll down and visible when scroll up.

### DIFF
--- a/src/web/src/components/Posts/PostInfo.tsx
+++ b/src/web/src/components/Posts/PostInfo.tsx
@@ -12,6 +12,9 @@ const useStyles = makeStyles((theme: Theme) =>
       flexDirection: 'column',
       justifyContent: 'center',
       width: '100%',
+      [theme.breakpoints.up('lg')]: {
+        width: '22rem',
+      },
     },
     authorAvatarContainer: {
       shapeOutside: 'circle(50%) border-box',
--- a/src/web/src/components/Posts/PostInfo.tsx
+++ b/src/web/src/components/Posts/PostInfo.tsx
@@ -10,10 +10,12 @@ const useStyles = makeStyles((theme: Theme) =>
       display: 'flex',
       flexDirection: 'column',
       justifyContent: 'center',
-      borderLeft: '2.5px solid #707070',
-      [theme.breakpoints.up('lg')]: {
-        width: '22rem',
-      },
+      marginBottom: '80vh',
+      opacity: '0.4',
+      transition: 'opacity .5s',
+      '&:hover': {
+        opacity: '1'
+      }
     },
     authorAvatarContainer: {
       shapeOutside: 'circle(50%) border-box',
@@ -30,12 +32,13 @@ const useStyles = makeStyles((theme: Theme) =>
     },
     authorNameContainer: {
       width: '100%',
+      height: '3rem'
     },
     author: {
       width: '100%',
       wordWrap: 'break-word',
-      fontSize: '2em',
-      lineHeight: '1.5em',
+      fontSize: '1.6em',
+      lineHeight: '1em',
       fontWeight: 'bold',
       marginLeft: '.5em',
       color: theme.palette.text.primary,
@@ -49,7 +52,7 @@ const useStyles = makeStyles((theme: Theme) =>
     },
     published: {
       marginLeft: '.5em',
-      fontSize: '1.5em',
+      fontSize: '1.2em',
       fontWeight: 'lighter',
       textDecoration: 'none',
       color: theme.palette.text.primary,
@@ -76,11 +79,11 @@ const PostDesktopInfo = ({ authorName, postDate, blogUrl, postUrl }: Props) => {
         <PostAvatar name={authorName} blog={blogUrl} />
       </div>
       <div className={classes.authorNameContainer}>
-        <h1 className={classes.author}>
+        <p className={classes.author}>
           <a className={classes.link} href={blogUrl}>
             {authorName}
           </a>
-        </h1>
+        </p>
       </div>
       <div>
         <a href={postUrl} rel="bookmark" className={classes.published}>
--- a/src/web/src/components/Posts/PostInfo.tsx
+++ b/src/web/src/components/Posts/PostInfo.tsx
@@ -8,20 +8,15 @@ const useStyles = makeStyles((theme: Theme) =>
       marginLeft: '2rem',
       padding: '0',
       display: 'flex',
-      borderLeft: '2.5px solid #CBCBCB',
+      borderLeft: '1.5px solid #EEEEEE',
       flexDirection: 'column',
       justifyContent: 'center',
-      opacity: '0.7',
-      transition: 'opacity .5s',
-      '&:hover': {
-        opacity: '1'
-      }
     },
     authorAvatarContainer: {
       shapeOutside: 'circle(50%) border-box',
       shapeMargin: '1rem',
       borderRadius: '50%',
-      marginLeft: '1.5rem'
+      marginLeft: '1.5rem',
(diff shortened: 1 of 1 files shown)


src/web/src/components/Posts/PostInfo.tsx
```diff
@@ -9,7 +9,12 @@ const useStyles = makeStyles((theme: Theme) =>
       display: 'flex',
       flexDirection: 'column',
       justifyContent: 'center',
-      borderLeft: '2.5px solid #707070',
```
PedroFonsecaDEV, line 12: Keep it.

src/web/src/components/Posts/PostInfo.tsx
```diff
@@ -9,7 +9,12 @@ const useStyles = makeStyles((theme: Theme) =>
...
+      marginBottom: '80vh',
+      opacity: '0.4',
```
PedroFonsecaDEV, line 13: I like the idea. But let's increase the opacity. 
`opacity: 0.6` or 0.7.

PedroFonsecaDEV (changes requested):

src/web/src/components/Posts/PostInfo.tsx
huynguyez: @PedroFonsecaDEV  looks weird with this line

huynguyez: @PedroFonsecaDEV or keep it but with a very light color

PedroFonsecaDEV: @huynguyez we can think about it in the future. For now let's keep ` borderLeft: '2.5px solid #707070'`

birtony (changes requested):
@huynguyez you have an interesting idea, but could you do some research on how similar websites handle this and share some results with us in the PR description, please? I think this would help us come to a team decision. Right now it looks to me like `opacity: 0.4` is too much, I would increase it a bit. In terms of the separator, I think we should maybe add some opacity to it and keep it, it helps identify the author info as a separate section from the content of the post. I wouldn't get rid of it. Let's take a look at how others handle such layouts and decide.

birtony (dismissed):
I personally like the latest update. It seems to be a good compromise for both of your points @PedroFonsecaDEV @huynguyez

PedroFonsecaDEV (changes requested):
I like the changes in the font size.
But I don't think the hover(opacity) is the way to go. 
Let's keep without the hover effect.

chrispinkney:
> 
> 
> I like the changes in the font size.
> But I don't think the hover(opacity) is the way to go.
> Let's keep without the hover effect.

I think I agree with this too. It's a great idea I just don't think it fits well here.

Also needs a rebase as the hyphen changes from Royce aren't included.

humphd:
Waiting on a rebase + more review, but looking close.

src/web/src/components/Posts/PostInfo.tsx
PedroFonsecaDEV: @huynguyez Keep this line.

PedroFonsecaDEV (changes requested):

PedroFonsecaDEV (approved):

humphd:
@rjayroso it looks like in the rebase for this, we lost the work that was done to increase the width of the container for the author's name, so that it doesn't always wrap on `-`, as with your name.  Can you review this and help us figure out what the right fix is before we take it?

humphd:
Since we're pushing this, let's also make the dark-mode version of the vertical line "darker" i.e., less visible to match what we do with the light theme.

rjayroso:
@humphd I've taken a look at the latest deployment (655e0c1b) and I see that the breakpoint for larger screens was removed. However, I think since Huy changed the font to be smaller, my name can now fully fit inside the container without breaking at the hyphen - which I guess is visually identical to what we had before. I've tested it with a longer name and it seems that it doesn't break at the `-` hypen like you mentioned. Personally, I don't mind since I always thought of my last name as one word (like "t-shirt"), but I still think we need to keep the increased container width because if I test out some of the longer names that you mentioned in the original issue (e.g. "Mohammad Abdullah Sajed Shadani", [see list here](https://github.com/Seneca-CDOT/telescope/issues/1955#issuecomment-800256159)) it starts to fall a part, see below:

image

I would say, keep the code for the extra width. Unless for whatever reason, it messes up your new proposed design.

PedroFonsecaDEV:
@huynguyez @rjayroso @humphd 

No regressions, please.

PedroFonsecaDEV:
@rjayroso Could you please review the latest commit (https://github.com/Seneca-CDOT/telescope/pull/2035/commits/adc81981febf6822f4c850a2bba42280bd9b5c1d) and check if there is some regression? 
The idea is to keep what you did and then add Huy's work to it.

PedroFonsecaDEV (changes requested):
@rjayroso

birtony:
@huynguyez, @rjayroso, @PedroFonsecaDEV let's review this one and merge if it's ready

src/web/src/components/Posts/PostInfo.tsx
```diff
@@ -5,20 +5,19 @@ import AdminButtons from '../AdminButtons';
...
+      marginLeft: '2rem',
       padding: '0',
       display: 'flex',
+      borderLeft: '1.5px solid #999999',
       flexDirection: 'column',
       justifyContent: 'center',
-      borderLeft: '2.5px solid #707070',
-      [theme.breakpoints.up('lg')]: {
-        width: '22rem',
-      },
```
rjayroso, line 16: Re-add the breakpoint to utilize the whitespace on larger screens. 

https://github.com/Seneca-CDOT/telescope/pull/2026#issuecomment-808944743

rjayroso (changes requested):

rjayroso (dismissed):

humphd:
@huynguyez is this ready for another review?

PedroFonsecaDEV (changes requested):
Screen Shot 2021-04-15 at 8 19 01 PM
Avatar needs to be left-aligned the way we have it on PROD or DEV now. 
and rebase

PedroFonsecaDEV (approved):

humphd:
@rjayroso an you please review?

huynguyez:
> @rjayroso an you please review?

@humphd I think I will merge it anyway because @rjayroso approved this yesterday, the latest commit just changed a bit of marginLeft.

humphd (approved):
Sounds good.

@rjayroso, file a follow-up if necessary.

Issue #2039, opened 2021-03-28
BlogSpot images are scaled down the same as inline images


rjayroso:
**What happened**:
Blogspot post images are scaled down to 32px along with inline images.
The latest iteration of our solution (see PR [here](https://github.com/Seneca-CDOT/telescope/pull/1910)) for issues: #1791 and #1412 works as intended (reduce the size of inline images without affecting the rest of the images in the post) except for Blogspot post images; *all* the images within a Blogspot post are being reduced to 32px.

**What should have happened**:
Blogspot post images should be unaffected, yet at the same time, inline images should be scaled down to 32px.

From my understanding after working on #1791, both Blogspot images and regular inline images both are structured like so:  
`div > a > img` - which means that if you apply CSS to scale down inline images, you would also affect Blogspot images and vice versa. 

**How to reproduce it (as precise as possible)**:
To see the *intended effect* of scaled-down inline images, navigate to the SearchPage and select the filter type `post` and search for "validation.js". Here you will see Tony Vu's GitHub avatar scaled down and inline with his GitHub profile link.

To see the bug, navigate to the SearchPage and select the filter type `author` and view Krystyna Lopez's BlogSpot blogs. Here you will see all the images within her post unintentionally scaled down to 32px.

**Anything else we need to know?**:
As Huy Nguyen mentions [here ](https://github.com/Seneca-CDOT/telescope/issues/1791#issuecomment-784429056) and [here](https://github.com/Seneca-CDOT/telescope/pull/1910#issuecomment-797540783), "Using `telescope-post-content div` will also affect those blogs from Blogspot."

**Environment**:
- OS: Windows
- Browser: Chrome

PR #2040, opened 2021-03-28
Fixes #1975: Decorate each post with a class that matches the blog host platform

### LINKED ISSUE #1975: Decorate each post with a class that matches the blog host platform
We should be able to distinguish each blog post by decorating the post with a class that matches their blogging host platform; this would allow us to style each blog post differently depending on which platform it is coming from. If the blogging platform cannot be identified we should just style it with our generic solution.

screen_shot_2021-03-18_at_10 04 03_am

Each post has a class which we can modify to look like so:
`<section class="telescope-post-content medium">` or `<section class="telescope-post-content wordpress">` etc

This would allow us to apply CSS to specific posts like so:
```CSS
img.medium {...}

img.wordpress {...}
```

This is especially useful since different blogging platforms execute their layout/CSS differently (which makes it hard to apply a one-size-fits-all solution). Furthermore, if we know what site the post is coming from, we can gather some hints from the original blogging platform to see how they do their images and use that in our own CSS.

In my latest PR ([here](https://github.com/Seneca-CDOT/telescope/pull/1910)), I ran into a problem where - if I apply CSS styling to handle inline images, it will also apply the same CSS styling to BlogSpot posts since they both share the same HTML structure (`div > a > img`). Thus, by fixing the styling of the inline images I had ruined the images from BlogSpot posts. By implementing the desired solution above, I would be able to apply the inline CSS stying to all posts *except* any BlogSpot posts.

rjayroso:
## Issue This PR Addresses
Fixes #1975 

## Type of Change

## Description
This essentially decorates `<section class="telescope-post-content">` with the blogging host platform name.  

**For example**, if the URL is `https://roycedev.medium.com/spending-the-weekend-thinking-575bcf535e58` it will extract **only** the `medium-com` and decorate the section class, like so:  `<section class="telescope-post-content medium-com">`  

image  

**Another example:**  
image

This will allow greater CSS customization with each post since each blogging platform styles their blogs/layout differently.
In reference to  the BlogSpot issue (#2039), we can fix it by styling **only** BlogSpot images to not be inline.

## How to Test
1. View the changed files for syntax/logic errors and possible changes/suggestions
2. View Vercel deployment OR pull this pr onto your local machine via `gh pr checkout` (see [here](https://cli.github.com/manual/gh_pr_checkout))
3. Scroll down to any blog post, and right-click the *content* (not the title/author name, aim for the paragraphs) and click inspect
image
4. View the `<section>` tag and see the blogging host platform name within the `className` 
image

## Checklist

### DIFF
--- a/src/web/src/components/Posts/Post.tsx
+++ b/src/web/src/components/Posts/Post.tsx
@@ -184,6 +184,20 @@ const formatPublishedDate = (dateString: string) => {
   }).format(date);
 };
 
+const extractBlogClassName = (url: string) => {
+  const blogClassName = new URL(url).hostname;
+  if (blogClassName.endsWith('medium.com')) {
+    return 'is-medium';
+  }
+  if (blogClassName.endsWith('dev.to')) {
+    return 'is-devto';
+  }
+  if (blogClassName.endsWith('blogspot.com')) {
+    return 'is-blogspot';
+  }
+  return 'is-generic';
+};
+
 const PostComponent = ({ postUrl }: Props) => {
   const classes = useStyles();
   const theme = useTheme();
@@ -284,7 +298,7 @@ const PostComponent = ({ postUrl }: Props) => {
       <div className={classes.content}>
         <section
           ref={sectionEl}
-          className="telescope-post-content"
+          className={`telescope-post-content ${extractBlogClassName(post.url)}`}
           dangerouslySetInnerHTML={{ __html: post.html }}
         />
       </div>


src/web/src/components/Posts/Post.tsx
```diff
@@ -193,6 +203,10 @@ const PostComponent = ({ postUrl }: Props) => {
   // Grab the post data from our backend so we can render it
   const { data: post, error } = useSWR<Post>(postUrl);
   const [expandHeader, setExpandHeader] = useState(false);
+  // If post exists, extract the blog platform name
```
humphd, line 206: Move this down to ~248 where you know post exists for sure.  Don't bother if it doesn't.  Also, I would suggest you should do it all in a function that you call vs. using a variable:

This needs a space between.  I'd do it like this:

```jsx
className={`telescope-post-content ${extractBlogClassName(post.url)}`}
```

src/web/src/components/Posts/Post.tsx
```diff
@@ -184,6 +184,16 @@ const formatPublishedDate = (dateString: string) => {
   }).format(date);
 };
 
+const parseHostName = (hostname: string) => {
+  // Check if there is a prefix to the hostname
```
humphd, line 188: I think you can simplify this a bit.  You are trying to deal with all possible cases, but we only have a handful that we care about (medium, dev.to, etc).  You can hard code them in here and check if the hostname ends with those strings:

```js
if(hostname.endsWith('medium.com')) {
  return "is-medium";
}
if(hostname.endsWith('dev.to')) {
  return "is-devto";
}
...
// Maybe add a "generic" version for blogs we don't know about specifically...
return 'is-generic';
```

const knownHosts = ['medium.com', 'dev.to', ...];
return 
```

File: src/web/src/components/Posts/Post.tsx [hunk 3]
```diff
@@ -284,7 +298,7 @@ const PostComponent = ({ postUrl }: Props) => {
       <div className={classes.content}>
         <section
           ref={sectionEl}
-          className="telescope-post-content"
+          className={'telescope-post-content' + blogClassName}
```
humphd, line 301: See my comment above.  I would call a function here that returns the string you want.

humphd (changes requested):

birtony (dismissed):
I tend to agree with @humphd's comments. It only really makes sense to add a platform-specific `className` if we know about the platform and apply appropriate styles. Otherwise, even with distinctive `className`s all of the posts will be identical for us, since we don't apply any specific styles. We can extend the list of the known platforms when needed and add corresponding styles.

rjayroso:
Implemented the changes suggested above to the best of my understanding.
1. Removed variable declaration/function call for `blogClassName` and combined it within the `className` like so `className={'telescope-post-content ${extractBlogClassName(new URL(post.url).hostname)}'}` 
2. `parseHostName` function changed to `extractBlogClassName` (we only care about the known hosts, otherwise label it as generic)

**Intended functionality is still preserved:**
Capture
Capture2

src/web/src/components/Posts/Post.tsx
```diff
@@ -284,7 +297,7 @@ const PostComponent = ({ postUrl }: Props) => {
       <div className={classes.content}>
         <section
           ref={sectionEl}
-          className="telescope-post-content"
+          className={`telescope-post-content ${extractBlogClassName(new URL(post.url).hostname)}`}
```
humphd, line 300: let's move the `new URL(post.url).hostname)` bit into `extractBlogClassName` and just pass the `post.url` here.

humphd (changes requested):
Nice.  One more change and I think this is good.

src/web/src/components/Posts/Post.tsx
rjayroso: Is something like this what you had in mind?
```js
const extractBlogClassName = (url: string) => {
  const blogClassName = new URL(url).hostname;
  if (blogClassName.endsWith('medium.com')) {
    return 'is-medium';
  } 
  // ... rest of function ...
}
```
This is what I pushed just now, I've checked and the class name decoration still works as intended

humphd (approved):

HyperTHD (commented):
Is this for specific blog posting websites? Cause wordpress is also used alot too.

EDIT: I read the comments above, this is GTM, just make sure you rebase first

HyperTHD (approved):

PR #2046, opened 2021-03-30
Fixes #1584: Create a SearchContext to pass props more easily to Search Components

### LINKED ISSUE #1584: [next] Create a context to Search components.
We are drilling props on our SeachPage components, let's create a context to solve this. 
I don't like to drill more than one level.

HyperTHD:
## Issue This PR Addresses
This PR closes #1584 

## Type of Change

## Description

This PR creates a SearchContext as well as a new hook that uses the context to make it more easier to access. This context includes the text parameter, the filter parameter, an onSubmit handler, an onTextChangeHandler, and an onFilterChangeHandler.
All Search related components now access these properties via the useSearchValue hook.

## Checklist

### DIFF
--- a/src/web/src/components/SearchResults.tsx
+++ b/src/web/src/components/SearchResults.tsx
@@ -2,9 +2,10 @@ import { makeStyles } from '@material-ui/core/styles';
 import { useSWRInfinite } from 'swr';
 import { Container, Box } from '@material-ui/core';
 
-import { postsServiceUrl } from '../config';
+import { telescopeUrl } from '../config';
 import Timeline from './Posts/Timeline';
 import Spinner from './Spinner';
+import useSearchValue from '../hooks/use-search-value';
 
 const NoResultsImg = '/noResults.svg';
 
@@ -48,15 +49,11 @@ const useStyles = makeStyles(() => ({
   },
 }));
 
-type SearchResultProps = {
-  text: string;
-  filter: 'post' | 'author';
-};
-
-const SearchResults = ({ text, filter }: SearchResultProps) => {
+const SearchResults = () => {
   const classes = useStyles();
+  const { text, filter } = useSearchValue();
   const prepareUrl = (index: number) =>
-    `${postsServiceUrl}/query?text=${encodeURIComponent(text)}&filter=${filter}&page=${index}`;
+    `${telescopeUrl}/query?text=${encodeURIComponent(text)}&filter=${filter}&page=${index}`;
 
   // We only bother doing the request if we have something to search for.
   const shouldFetch = () => text.length > 0;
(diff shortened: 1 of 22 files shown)


src/web/src/components/SearchResults.tsx
```diff
@@ -56,7 +56,7 @@ type SearchResultProps = {
 const SearchResults = ({ text, filter }: SearchResultProps) => {
   const classes = useStyles();
   const prepareUrl = (index: number) =>
-    `${telescopeUrl}/query?text=${encodeURIComponent(text)}&filter=${filter}&page=${index}`;
+    `${postsServiceUrl}/query?text=${encodeURIComponent(text)}&filter=${filter}&page=${index}`;
```
manekenpix, line 59: I don't understand this change. The `Post` MS doesn'thave a `/query` endpoint, how is this going to work?

HyperTHD: This was mentioned in that PR. I used the wrong url there, this is supposed to be the search service being used here instead. This change shouldn't have been picked up

src/web/src/pages/_app.tsx
```diff
@@ -44,8 +45,10 @@ const App = ({ Component, pageProps }: AppProps) => {
     <ThemeContext.Provider value={{ theme, themeName: theme.palette.type, toggleTheme }}>
       <ThemeProvider theme={theme}>
         <AuthProvider>
-          <NavBar />
-          <Component {...pageProps} />
+          <SearchProvider>
```
birtony, line 48: Do we want to wrap any component with this provider or only the search page component? I think it'd be best to only use the provider where we need it

birtony (commented):
Good job in cleaning this up @HyperTHD

src/web/src/pages/_app.tsx
HyperTHD: Good catch, didn't think I could only applied it to the search page. Moved it there, and works the same

src/web/src/pages/search.tsx
```diff
@@ -1,11 +1,14 @@
 import SEO from '../components/SEO';
 import SearchPage from '../components/SearchPage';
+import SearchProvider from '../components/SearchProvider';
 
 const Search = () => {
   return (
     <>
-      <SEO pageTitle="Search | Telescope" />
-      <SearchPage />
+      <SearchProvider>
```
birtony, line 8: I am thinking if it is still a good practice to keep things wrapped in react fragment here or should we just remove it since `SearchProvider` wraps everything up? 🤔

HyperTHD: It's a minor thing, works pretty much the same with or without it

birtony: Attention to details is key 😉

PR #2055, opened 2021-03-30
Expose JWT_ISSUER, JWT_AUDIENCE, and SECRET env vars to all services

humphd:
In Satellite we have `isAutenticated()` middleware that depends on a service knowing some info about our JWT setup.  Also, we're [adding the ability for a service token to be created in each service](https://github.com/Seneca-CDOT/satellite/pull/10), which means we need to share even more details with the services.  To do this, we expect environment variables to be set in Satellite.  Previously we weren't doing this, but this PR adds them in.

Future services will also require this (FYI @c3ho for Parser and @izhuravlev for Search).

We were also missing some `NODE_ENV=production` variables, which I've added.  They will always need to be `production` in production, so it's best to do it in the docker-compose file vs. the env.

Finally, I've stubbed out (with comments) the APM values we'll need later on when we add those, so it's clear what's missing.

### DIFF
--- a/docker/production.yml
+++ b/docker/production.yml
@@ -40,7 +40,7 @@ services:
       - ELASTIC_PORT=9200
       - ELASTIC_URL=http://elasticsearch
       # Environment values we expect from our legacy .env file
-      - NODE_ENV
+      - NODE_ENV=production
       - PORT
       - POSTS_URL
       - API_URL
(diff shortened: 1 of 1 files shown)


birtony (dismissed):

HyperTHD:
I'm assuming we plan on updating the posts service's routes as well as the any other future service to only be accessed by an authenticated user with this change

humphd:
Yeah, we need to add middleware for the JWT stuff in all routes that modify any data, or should be hidden behind admin.

humphd:
https://github.com/Seneca-CDOT/satellite/pull/10 just landed and needs this.

manekenpix (dismissed):

PR #2057, opened 2021-03-30
Fix path issus in production

manekenpix:
## Type of Change

## Description
This changes the path to the mapped directories used for the `portainer` and `users` services in `production.yml`. Instead of having everything in `config`, I think it's better to have separate directories.
Also, I added to `volumes` the name of the file where the settings for `firestore` are stored. It seems that mapping only the directory is not enough, it needs the name of the file.

## Checklist

### DIFF
--- a/docker/production.yml
+++ b/docker/production.yml
@@ -144,7 +144,7 @@ services:
       # - ELASTIC_APM_SERVER_URL=http://apm:8200
     volumes:
       # This will take care of copying the serviceAccountKey.json file needed by firestore.js
-      - ../../config/firebase:/app
+      - ../../firebase/serviceAccountKey.json:/app/serviceAccountKey.json
 
   ##############################################################################
   # Third-Party Dependencies and Support Services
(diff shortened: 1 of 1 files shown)


chrispinkney (approved):

docker/production.yml
```diff
@@ -238,8 +238,8 @@ services:
       - 9000
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock
-      - ../../config:/data
-      - ../../config/portainer:/data/portainer
+      - ../../portainer:/data
```
humphd, line 241: In the rest of this file you use `elastic-data` and `redis-data` so is this `portainer-data`?

Happy either way.

humphd (approved):

docker/production.yml
manekenpix: hmm, I didn't think of that. For `redis` and `ES` it makes sense since they're databases, but the `portainer` directory is not storing data, just config files, but I'm ok adding `data` for consistency.

Issue #2053, opened 2021-03-30
Staging and Production setup for Users Service


humphd:
Discussions today led us to the conclusion that we need a list of steps for everything we have to do in order to fully ship the Users service:

1. ~~Land #2027 for docker configs~~ (@chrispinkney)
2. ~~Modify Firebase schema to use email vs. id as the User identifier, in order to match Seneca's `nameID`.  This will involve changes to unit tests, Celebrate validation, and Users schema (@humphd https://github.com/Seneca-CDOT/telescope/pull/2060)~~
3. ~~Paginate data for Users / route #2022 (@chrispinkney)~~
4. ~~Modify https://github.com/Seneca-CDOT/telescope/blob/master/src/api/auth/src/authorization.js#L19-L27 to use nameID and have JWT token match what Users service expects, as well as include roles in JWTs (i.e., convert `isAdmin` to `["admin", ...]`) when we create the token (@humphd, https://github.com/Seneca-CDOT/telescope/pull/2058).~~
5. ~~Add `createServiceToken` to Satellite for inter-service communication, to allow for `role["service"]` @humphd in https://github.com/Seneca-CDOT/satellite/pull/10, https://github.com/Seneca-CDOT/telescope/pull/2055 to deal with missing env vars to make that possible.~~
6. ~~Add auth middleware to all routes of Users service based on JWT id (i.e., email) and/or role (e.g., admin or service account) (@humphd #2116)~~
7. Create production and staging Firebase projects (@humphd, @manekenpix) https://github.com/Seneca-CDOT/telescope/issues/1932
8. Share account access to Firebase projects with anyone who is involved with staging/production (@humphd)
9. Redis caching for Users service #2030 (@humphd, @chrispinkney, @Metropass, @manekenpix)
10. ~~Connect Users service to front-end~~
    i. ~~on login, we need to get user info from User service to populate the `user` object.  Instead of using the JWT payload for user info (see https://github.com/Seneca-CDOT/telescope/blob/master/src/web/src/components/AuthProvider.tsx#L81-L93), we need to do a GET request to the Users service, passing our JWT token, and get back the full user object (https://github.com/Seneca-CDOT/telescope/pull/2088 @humphd)~~
    ii. ~~signup flow in front-end #2148 (@PedroFonsecaDEV, @DukeManh, @Meneguini)~~
[...]

PR #2058, opened 2021-03-31
Add proper roles and User service to JWT authorization

humphd:
This does a number of things toward #2053, specifically:

- Updates the Auth service to know about the User service, so that we can look-up user info with the Users service during authentication.  This depends on https://github.com/Seneca-CDOT/satellite/pull/10 for the service token logic.  NOTE: this will mean our e2e tests are going to fail, since we'll need to put a user into Firebase before we can do the entire authentication flow.
- Adds `name` and `roles` claims to the JWT, so that we can include the user's display name (for convenience) and whether they are a user or admin, used by Satellite authorization middleware

I noticed while writing this that the schema we have for a user in the User service needs to be updated to include `email` and/or I need to change my code to use the `id` we get back from the Users service to set the email.  @chrispinkney we should figure this out ASAP, since a lot of things depend on the layout of these objects.


src/api/auth/src/token.js
```diff
@@ -2,19 +2,32 @@ const jwt = require('jsonwebtoken');
...
+    // User roles.  We have 4 currently:
+    //   1. seneca (authenticated with Seneca's SSO)
+    //   1. telescope (authenticated Telescope user)
+    //   2. admin (authenticated Telescope user with isAdmin=true in Firebase)
+    //   3. service (a Telescope microservice, see createServiceToken() in Satellite)
+    roles,
   };
 
-  const options = { expiresIn: JWT_EXPIRES_IN || '1h' };
+  const options = { expiresIn: JWT_EXPIRES_IN || '3h' };
```
birtony, line 30: Is there a way for us to extend this for PWA? I think most mobile apps let you stay logged in for a longer period of time than websites

src/api/auth/src/user.js
```diff
@@ -1,23 +1,98 @@
...
+  constructor(senecaProfile, telescopeProfile) {
+    // TODO: what to do about id?  Hash it like before or use email? hash(profile.nameID)
+    // this.id = id;
+
+    // All authenticated users are Seneca Users
+    this.seneca = {
+      'http://schemas.microsoft.com/identity/claims/displayname':
+        senecaProfile['http://schemas.microsoft.com/identity/claims/displayname'],
+      'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress':
+        senecaProfile['http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress'],
```
birtony, line 12: Are these links supposed to work when we access them? If not, how do they specify the data type?

src/api/auth/src/authentication.js
```diff
@@ -107,17 +107,43 @@ const strategy = new SamlStrategy(
      *   "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress": "first.last@email.com"
      *   "email": "first.last@email.com"
      * }
+     *
+     * We only really care about the emailaddress claim, which we also use in the Users service.
      */
+    const email =
+      senecaProfile['http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress'];
```
birtony, line 114: How does this link work? Should we add a comment to link to some docs about it?

src/api/auth/src/user.js
```diff
@@ -1,23 +1,98 @@
...
+
+    // But not all users are Telescope Users (yet) if they haven't signed up.
+    if (telescopeProfile) {
+      this.telescope = telescopeProfile;
+    }
+  }
+
+  // Normalize access to Standard Seneca user properties and optional Telescope properties
+  get email() {
+    return this.seneca['http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress'];
```
birtony, line 25: How does this link work? Should we add a comment to link to some docs about it?

src/api/auth/src/user.js
```diff
@@ -1,23 +1,98 @@
...
+  get github() {
+    return this.telescope?.github;
+  }
+
+  get isTelescopeUser() {
+    return !!this.telescope;
+  }
+
+  // Get a list of roles for this user
+  get roles() {
```
birtony, line 73: In https://github.com/Seneca-CDOT/telescope/pull/2058/files#diff-2c521e1745fe5dd3ad9df3a78ef80c8cd45f49c62e8367ab03026a5d7549df92R22 you say we have 4 roles, but this function only returns up to 3. Is there any chance we'd need the service role returned as well?

src/api/auth/test/user.test.js
```diff
@@ -0,0 +1,241 @@
...
+  firstName: 'First',
+  lastName: 'Last',
+  displayName: 'Telescope Display Name',
+  isAdmin: true,
+  isFlagged: true,
+  feeds: ['https://dev.to/feed/first-last'],
+  github: {
+    username: 'firstlast',
+    avatarUrl:
+      'https://avatars.githubusercontent.com/u/7242003?s=460&u=733c50a2f50ba297ed30f6b5921a511c2f43bfee&v=4',
```
birtony, line 34: Funny
cc @chrispinkney

src/api/auth/src/token.js
humphd: There's some risk in having tokens be long-lived, since they are effectively passwords that anyone can re-use without having to prove themselves.  We can extend this, for sure, but for how long?  With systems that do cookie based sessions on the server, you can have a login live forever, and not risk much.  With tokens that a browser holds, it's harder.  We could look into doing some kind of "refresh" token later, to let you renew your login.

Question: what would you do while logged in for longer?  The only time you need to log in is when you create or modify your account.

src/api/auth/src/user.js
humphd: They aren't links, they are URIs for specific schema property names.  It's a way for there to be less ambiguity about what "email" or "displayname" means: the URI version is very specific.  NOTE: I'm using what we get back from Microsoft, not choosing these names.

src/api/auth/src/authentication.js
humphd: You can read about the schema if you want more info http://docs.oasis-open.org/imi/ns/identity-200810

src/api/auth/src/user.js
humphd: No, we'd never create a `service` token from the user authentication flow.  See https://github.com/Seneca-CDOT/satellite/pull/10, and add a review, if you're willing.

src/api/auth/src/token.js
birtony: Understandable. I was just thinking if there is a safe way to let users stay logged in while using PWA. To answer your question, that's true, but only for now ;)

src/api/auth/src/authentication.js
birtony: 2010, wow...

birtony (approved):

src/api/auth/src/authentication.js
```diff
@@ -107,17 +107,43 @@ const strategy = new SamlStrategy(
...
+
+    async function lookupTelescopeUser(id) {
```
Metropass, line 116: I assume that when this function is invoked, we would have a check to see if the User creating the query has permissions to do so, right?

src/api/auth/src/authentication.js
```diff
@@ -107,17 +107,43 @@ const strategy = new SamlStrategy(
...
+      // We have the user's id (i.e. their email address) in a valid token. Now
+      // make a service-to-service request to the Users service in order to
+      // get this user's full profile information using the token's subject claim
+      // as the user's id (e.g., their email address).
+      try {
+        const res = await fetch(`${USERS_URL}/${id}`, {
+          headers: {
+            Authorization: `bearer ${createServiceToken()}`,
```
Metropass, line 124: We should probably add a Cache-Control header that contains `private` or `no-store` due to security risks on URI's/URL's

see:  https://tools.ietf.org/html/rfc6750#section-5

humphd:
Depends on #2060

humphd:
@chrispinkney @manekenpix @birtony @Metropass @HyperTHD this is finished (woo!) and ready for another review.  It's exceedingly hard to test locally, so the best thing to do is to read the e2e tests that I've written, and make sure they pass on CI.  Here's what I've done:

A user clicks Login in the front-end, it will go to the Auth service, and that will redirect to the SSO, where they enter credentials.  They then get redirected back to the Auth service.  Here, we use their Seneca email to check if a user exists in the Users service.  If one does, we grab that Telescope profile info and hold on to it for later; if not, we use their Seneca profile info instead.

We then create a JWT token for the user, and populate it with info for the front-end.  This includes their `name`, `sub` (subject in JWT--their email), `picture` (if we know it from GitHub), and `roles`.  The last one is an array of all authorized roles the user has: `seneca` (they are an authenticated Seneca SSO user); `telescope` (they are an authenticated Seneca SSO user who also exists in Telescope's Users service); `admin` (they are a Seneca + Telescope user but also have `isAdmin=true` in Firebase).

I've integrated all the pieces from various other PRs in Satellite and Telescope to do this, and also added a bit of convenience code for running things locally in development.

After this lands, we can update the front-end to use the `name` + `picture` from the token, and also start protecting all of our routes that need it.

Metropass (approved):
Took a while on my machine, but everything passed.

manekenpix (dismissed):

chrispinkney (dismissed):

DukeManh (approved):

PR #2060, opened 2021-03-31
Use hashed email as id for Users in Firestore

humphd:
In #2058 I'm dealing with connecting the Users and Auth services.  For this to work, we need to use a user's email address as the document path in Firestore.  Because the document path can't contain `/`, I'm hashing the email address so it's safe to use on URLs, using the Satellite `hash()` function.

To make this all work, I've done the following:

- Updated the 'user' model and renamed it to `user.js`.  I'm removing `created`/`upated` (I think we get this for free in Firestore, but we can add later if we want it), and including a generated `id` based on the email.  I've also added a `toObject()` to get rid of having to to do `JSON.parse(json.stringify(user))` before we pass to Firestore (we need an object with no custom prototype, which this accomplishes).
- Added unit tests for the `user` model
- Updated the Celebrate schema, and corrected some bugs it had
- use the hashed email `id` everywhere in routes
- added the `:id` to some routes that were missing the param
- added validation on all routes using the `:id` param
- gotten rid of extraneous debug logs
- switched to use proper errors and `next()` calls, which will give us logging
- removed error message checks in tests, since these are brittle (the error code is what we care about, error messages can change and break your code).
- got the e2e tests to work again


birtony (approved):

chrispinkney (changes requested):

src/api/users/src/models/celebrateSchema.js
```diff
@@ -1,28 +1,29 @@
...
+      email: Joi.string().email().required(),
+      displayName: Joi.string().default(generateDisplayName),
       isAdmin: Joi.boolean().default(false),
       isFlagged: Joi.boolean().default(false),
-      feeds: Joi.array().items(Joi.string()).required(),
+      feeds: Joi.array().items(Joi.string().uri()).required(),
       github: Joi.object({
         username: Joi.string(),
-        avatarUrl: Joi.string(),
-      }).default({ username: '', avatarUrl: '' }),
```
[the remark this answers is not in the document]
humphd: How will having `github.avatarUrl = ''` work better when you do `` or `github.username = ''` when you do `fetch('https://api.github.com/user/${github.username})`?  The empty value will belie the fact that we don't have any data.  I would prefer that we not even add `.github` to the user at all if we don't have the data, since it makes checking for it a lot easier: `if(user.github) { /* use the github info */ }`

humphd:
Thanks for the great reviews.  I've made the fixes.  I've also improved how we deal with GitHub data when validating and missing parts: it's now OK to not have any `.github` data for a user, but if you include it, it has to have both expected values in the form we want.  I've updated the tests too.

NOTE: I need to rebase this to pick-up the work from the paging PR landing, but I'm waiting on that til I get this approved, so the changes aren't so major.

src/api/users/src/models/celebrateSchema.js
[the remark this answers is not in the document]
chrispinkney: Ah gotcha, makes sense.

src/api/users/src/routes/users.js
```diff
@@ -16,7 +16,7 @@ router.get('/:id', validateId(), async (req, res, next) => {
     const doc = await userRef.get();
 
     if (!doc.exists) {
-      next(createError(404, `user ${id}) not be found.`));
+      next(createError(404, `user ${id}) not found.`));
```
chrispinkney, line 19: Is the extra parenthesis intended here? It's repeated a few times below too.

> `user ${id}) <---

chrispinkney:
> 
> 
> Thanks for the great reviews. I've made the fixes. I've also improved how we deal with GitHub data when validating and missing parts: it's now OK to not have any `.github` data for a user, but if you include it, it has to have both expected values in the form we want. I've updated the tests too.
> 
> NOTE: I need to rebase this to pick-up the work from the paging PR landing, but I'm waiting on that til I get this approved, so the changes aren't so major.

I landed that last night so you should be good to go!

src/api/users/src/routes/users.js
humphd: No, good grief, let me fix.  And NO, I DON'T USE COPY/PASTE, HOW COULD YOU SUGGEST SUCH A THING!!!???

Thanks for catching my terrible code.

humphd:
This is ready for another review.  I had to change quite a bit, but I think I have a workable solution.  I'll try to explain the changes.  I had to dig much deeper into Firebase, but I think I get it now.

We're moving from using a numeric id to a hex value (hashed email, like a git commit sha) for our document path (Objects in Firebase are stored with a path name, their id).  The way that Firebase does paging, you don't have any info about what's in your collection unless you grab the entire thing (like asking how many items are in a linked list: it's an O(n) problem, and you have to traverse everything to figure it out).

So what you can do is order your collection, then set a limit on how many items to return.  Let's say I have this set:

|id|data|
|-|-----|
|b| user b data...|
|a| user a data...|
|c| user c data...|
|...|...|

Let's say that I order them by `id`, so Firebase has `[a, b, c, ... ]`.  Now I ask for the documents in that ordered collection with a limit of 2.  I'm going to get back `[user a data, user b data]`.

I also have another useful piece of info, namely, the id of the last item in the results i got back: in this case, that would be `b`.

Now I can go to Firebase and ask for the same data again, but this time use `startAt('b')`, meaning, "give me this ordered data, but don't start at the beginning, start at id=b".

We can't convert our `per_page` and `page` concept into something Firebase can do.  But, we can approach it slightly differently.  Many REST APIs, will not only return results, but also a `Link` header with info about how to get the next/prev pages, for example, [GitHub does this](https://docs.github.com/en/rest/guides/traversing-with-pagination#basics-of-pagination).

We used to do this, but I think we removed the code at some point.  Regardless, we can use the idea here to have the Users service send back the current "page" of results, but also include a `Link` header to tell the browser/client the URL to use in order to get the `next` page of results.  You'd do this:

1. Request all results from Users
2. Get back the first 20, as well as a URL to use to get 41-59
3. Use that URL to get 60-80, and get URL for 81-100
4. ...repeat until you don't get back a `next` link in the headers

This lets the User service work with Firebase to keep track of things, and clients can page through data without ever worrying about keeping track of their place in the collection IDs.

I can't write all the tests I need until we switch to use proper URLs to the Users service vs. `request(app)`, since my `next` links are all absolute links to the service.  I commented those tests out for now, and we can fix things once that happens.

A few other things I changed:

- Firestore has a cool API for allowing you to write [custom data converters](https://firebase.google.com/docs/reference/js/firebase.firestore.FirestoreDataConverter) for your classes, so you don't have to manually serialize data all the time.  I added one to the User model, and use it in the routes
- I changed the `db` export for `firestore.js` so it does more work for us.  I notice a lot of people doing it this way vs. having callers do it manually all the time.
- I added logic to order things by the document id path, which isn't overly obvious how to do.  We could store the `id` in the user info, but since it's always generated from the email, it's not really necessary, and having both it and the email in the data could lead to synchronization issues
- I cleaned up the tests a bit more to remove duplication of data

Let me know what else isn't clear and I'll try to explain.

src/api/users/src/models/schema.js
```diff
@@ -0,0 +1,54 @@
...
+      }).and('username', 'avatarUrl'),
+    }),
+  });
+
+// Make sure that the :id param matches the hashed value of the user's email if present
+const validateEmailHash = () => (req, res, next) => {
+  const { email } = req.body;
+  const { id } = req.params;
+
+  if (id !== hash(email)) {
```
chrispinkney, line 44: This should have a comment specifying that the hash function in satellite only returns the first 10 characters of the generated hash. The id passed in the post request must be 10 characters long.

chrispinkney (changes requested):
This is just my first pass after reviewing some changes, I'm going to test `post` and `get` now.

src/api/users/src/routes/users.js
```diff
@@ -29,33 +34,29 @@ router.get('/:id', validateId(), async (req, res, next) => {
 
 // get all users
 // rejected if the users collection is empty
-// otherwise n (specified via params) users are returned
+// otherwise perPage (specified via params) users are returned, starting at
+// the beginning of the users collection or the provided document path id (if present)
 router.get('/', validatePagingParams(), async (req, res, next) => {
   /*
    *  celebrateSchema.js performs data validation via validatePagingParams() middleware
```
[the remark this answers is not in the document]
humphd: Fixed in https://github.com/Seneca-CDOT/telescope/pull/2060/commits/681c129eef9a5647893c52d152b43903186a5936.

src/api/users/src/models/schema.js
humphd: Fixed in https://github.com/Seneca-CDOT/telescope/pull/2060/commits/681c129eef9a5647893c52d152b43903186a5936.

src/api/users/src/util.js
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  // Get the id of the last user in this page of results
+  const lastUser = users[users.length - 1];
+  const lastId = lastUser.id;
+
+  // Construct the body of the header, giving the URI to use for the next page:
+  // '<https://api.telescope.cdot.systems/v1/users?startAfter=6Xoj0UXOW3&per_page=100>; rel="next"'
+  const link = new LinkHeader();
+  link.refs.push({ uri: `${USERS_URL}?startAfter=${lastId}&per_page=${perPage}`, rel: 'next' });
```
chrispinkney, line 24: Should `USERS_URL` have a default value of `localhost:6666/` here? After running the users microservice locally (via CLI) `USERS_URL` is undefined and therefore creates the following link in headers: `<undefined?startAfter=635ed9123d&per_page=100>; rel=next`

src/api/users/src/routes/users.js
```diff
@@ -29,112 +34,95 @@ router.get('/:id', validateId(), async (req, res, next) => {
...
-      const user = new User(req.body);
-      await db
-        .collection('users')
-        .doc(`${req.body.id}`)
-        .update(JSON.parse(JSON.stringify(user))); // the user object must be parsed and stringified to be persisted to firestore
-      res.status(200).json({ msg: `Updated user ${req.body.id}` });
-      logger.debug(`User added with id: ${req.body.id}:\n${JSON.stringify(req.body)} by ${req.ip}`);
+      const user = new User(body);
+      // NOTE: doc().update() doesn't use the converter, we have to make a plain object.
+      await db.doc(id).update(user.toJSON());
```
chrispinkney, line 103: I think something I forgot to add initially was to check the values of the user we're trying to create vs the values specified in the body. If they're the same let's not waste another hit to update and instead respond with.. idk, `204` and then not run `db.doc(id).update(user.toJSON())` ? If the values are different then `200`.

I just tried to spam the same `put` request over and over again and they all went through.

chrispinkney (changes requested):
Should be my last pass, everything else LGTM!

src/api/users/src/routes/users.js
humphd: This feels like a premature optimization that could potentially lead to us hiding some bug we don't intend.  Let's start out by doing the obvious thing, and adjust if it becomes a problem.

src/api/users/src/util.js
humphd: Good call.  Fixed in https://github.com/Seneca-CDOT/telescope/pull/2060/commits/6271c5af1b16d7b99cb84ef633ba25489d09a048, and doing so also revealed a bug in the docker file, so thanks for pointing this out.  Your reviewing skills are really improving.

humphd:
Review ping

src/api/users/src/util.js
chrispinkney, line 24: `"message": "\"starAfter\" is not allowed"`:

```suggestion
  link.refs.push({ uri: `${USERS_URL}?start_after=${lastId}&per_page=${perPage}`, rel: 'next' });
```

chrispinkney, line 24: `"message": "\"starAfter\" is not allowed"`:

```suggestion
  link.refs.push({ uri: `${USERS_URL}?start_after=${lastId}&per_page=${perPage}`, rel: 'next' });
```

The comment needs to reference the change as well.

src/api/users/src/routes/users.js
```diff
@@ -29,112 +34,95 @@ router.get('/:id', validateId(), async (req, res, next) => {
...
-    const userRef = db.collection('users').doc(`${req.body.id}`);
+    const userRef = db.doc(id);
     const doc = await userRef.get();
 
     if (doc.exists) {
-      logger.debug(`User (id: ${doc.id}) was posted by ${req.ip} but already exists.`);
-      res.status(400).json({
-        msg: `User with id ${doc.id} was requested to be added, but already exists in the db.`,
-      });
+      next(createError(400, `User with id ${id} already exists.`));
```
chrispinkney, line 77:
```suggestion
      next(createError(400, `user with id ${id} already exists.`));
```

Might seem pedantic but `user` and `User` are both used throughout the PR. We should be more consistent throughout this PR.

chrispinkney (changes requested):
Should be the last round 😄

src/api/users/src/util.js
humphd: Fixed in https://github.com/Seneca-CDOT/telescope/pull/2060/commits/dc9a3d07b38f2b048b9caa13d9c5459a4c300f2a

src/api/users/src/routes/users.js
humphd: Fixed in https://github.com/Seneca-CDOT/telescope/pull/2060/commits/dc9a3d07b38f2b048b9caa13d9c5459a4c300f2a

chrispinkney (approved):

manekenpix (approved):

chrispinkney (approved):

manekenpix (approved):

PR #2088, opened 2021-04-06
Rework User to use proper JWT payload data

humphd:
This updates our JWT parser to get the user's name, authorization roles, and avatar image URL (if present) from the access token.  It depends on #2058 for the changes to the JWT payload.

One thing this will also add is the ability to do this:

```js
const { user } = useAuth();

if(user?.isAdmin) {
  // do stuff for an admin
}

if(user?.isRegistered) {
  // do stuff for a Telescope user who isn't an admin
}

if(user) {
  // do stuff for a Seneca SSO user, who hasn't registered yet
}
```

### DIFF
--- a/src/web/src/components/BannerButtons.tsx
+++ b/src/web/src/components/BannerButtons.tsx
@@ -58,7 +58,7 @@ const LandingButtons = () => {
           >
             Sign out
           </Button>
-          <PostAvatar name={user.name} />
+          <PostAvatar name={user.name} img={user.avatarUrl} />
         </>
       ) : (
         <>
--- a/src/web/src/User.ts
+++ b/src/web/src/User.ts
@@ -1,20 +1,46 @@
 import jwtDecode, { JwtPayload } from 'jwt-decode';
 
+interface TelescopeJwtPayload extends JwtPayload {
+  name: string;
+  picture?: string;
+  roles: Array<string>;
+}
+
+// Check if the expiry time in seconds is missing or has passed.
+const isExpired = (exp?: number) => typeof exp === 'undefined' || new Date().getTime() / 1000 > exp;
+
 export default class User {
   constructor(
-    public id: string,
     public email: string,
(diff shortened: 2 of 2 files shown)


PedroFonsecaDEV (dismissed):
LGTM

humphd:
Depends on https://github.com/Seneca-CDOT/telescope/pull/2089

src/web/src/components/BannerButtons.tsx
```diff
@@ -58,7 +58,7 @@ const LandingButtons = () => {
           >
             Sign out
           </Button>
-          <PostAvatar name={user.name} />
+          <PostAvatar name={user.name} img={user.avatarUrl} />
```
HyperTHD, line 61: Won't the img give problems if the avatarUrl is not defined here? If decoded.picture is the user's avatar url, we might want to use that instead

PedroFonsecaDEV: The img will return first:
```
const PostAvatar = ({ name, img, blog = '' }: AvatarProps) => {
  const classes = useStyles();

  if (img) {
    return <Avatar className={classes.avatar} src={img} />;
  }

  if (name.length > 0) {
    const initials = name
   ```
   
   https://github.com/Seneca-CDOT/telescope/blob/baf9c91d25bcf96919951e10e4d9cb1caef81977/src/web/src/components/Posts/PostAvatar.tsx#L42

humphd: The `<PostAvatar>` only renders if `img` exists, otherwise we do the initials.  The `picture` property from the JWT payload is stored in the `User` type as `avatarUrl`.

humphd:
#2089 is in and I've rebased, which fixes the failing CI test.

PedroFonsecaDEV (approved):

manekenpix (approved):

PR #2089, opened 2021-04-06
Remove /myfeeds page, components, and deps

humphd:
We're moving to a Signup flow, with new UI/UX, and the current `/myfeeds` page can be retired.

This removes it, along with all the dependencies and child components I could find.  See if I've missed any.


src/web/package.json
```diff
@@ -17,22 +17,17 @@
...
-    "swr": "^0.5.5",
-    "valid-url": "^1.0.9"
+    "swr": "^0.5.5"
   },
   "devDependencies": {
     "@types/node": "^14.14.37",
     "@types/react": "^17.0.3",
-    "@types/react-material-ui-form-validator": "^2.1.0",
     "@types/react-test-renderer": "^17.0.1",
-    "@types/valid-url": "^1.0.3",
```
PedroFonsecaDEV, line 35: All of these dependencies being removed were being used only on`/myfeeds` ?

humphd: Yeah, I believe so.  If we need them in new code, add them back, or flag now and I can leave them in.

PedroFonsecaDEV: Okay. let's remove them.

PedroFonsecaDEV (approved):

HyperTHD (dismissed):

humphd:
Rebased to pick up changes on master, can I get another review?

HyperTHD (approved):

manekenpix (approved):

Issue #2041, opened 2021-03-29
[UI 2.0] [Landing Page] - The navBar should show up only when we reach the Posts.


PedroFonsecaDEV:
**What would you like to be added**:
In #1976 we are updating our lading page. 
For the new landing page, we will need that our NavBar only shows up when our first post reaches the top.

Just for clarification, if we were keeping the current landing page what this enhancement would look like:

No `NavBar` on our landing page. 
<img width="1668" alt="Screen Shot 2021-03-29 at 4 50 14 AM" src="https://user-images.githubusercontent.com/60857409/112811979-982d0900-904a-11eb-8667-a78af9bc8b3a.png">

`NavBar` Showing up when the first Post reachs the top of the page:
<img width="1015" alt="Screen Shot 2021-03-29 at 4 51 11 AM" src="https://user-images.githubusercontent.com/60857409/112812353-04a80800-904b-11eb-9325-581f00e061ef.png">

PR #2078, opened 2021-04-05
Landing page 2.0

PedroFonsecaDEV:
## Type of Change

- [ ] **Bugfix**: Change which fixes an issue
- [x] **New Feature**: Change which adds functionality
- [ ] **Documentation Update**: Change which improves documentation
- [x] **UI**: Change which improves UI

## Description
This is part of the solution for #1976.
The new landing page still needs #2041 and #2042 
To test, please check the Vercel Build

When all the landing page pieces landed, the NavBar will not show up when the user is on our landing page. 

Screen Shot 2021-04-05 at 3 02 41 AM
Screen Shot 2021-04-05 at 3 04 44 AM

Side effects: closes #2015 

## Checklist

- [ ] **Quality**: This PR builds and passes our npm test and works locally
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [x] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not (if applicable)
- [ ] **Documentation**: This PR includes updated/added documentation to user exposed functionality or configuration variables are added/changed or an explanation of why it does not(if applicable)

PR #2094, opened 2021-04-08
Only show navbar when scrolls to post

yuanLeeMidori:
## Issue This PR Addresses
Fixes: #2041

## Type of Change

## Description
This PR is to remove the navbar from the new banner (#2078); only shows the navbar when the user scrolls to where the posts (timeline) is on the top of the screen. In other word, navbar will only show when there is no banner on the screen.
desktop (1)
mobile (1)

## Checklist

### DIFF
--- a/src/web/src/components/Banner.tsx
+++ b/src/web/src/components/Banner.tsx
@@ -78,7 +78,6 @@ const useStyles = makeStyles((theme: Theme) =>
     },
     anchor: {
       position: 'relative',
-      top: '1rem',
     },
     container: {
       bottom: '0',
@@ -90,7 +89,11 @@ const useStyles = makeStyles((theme: Theme) =>
   })
 );
 
-export default function Banner() {
+type BannerProps = {
+  onVisibilityChange: (visible: boolean) => void;
+};
+
+export default function Banner({ onVisibilityChange }: BannerProps) {
   const classes = useStyles();
   const [gitInfo, setGitInfo] = useState({
     gitHubUrl: '',
@@ -100,7 +103,6 @@ export default function Banner() {
 
   const timelineAnchor = useRef<HTMLDivElement>(null);
   const bannerAnchor = useRef<HTMLDivElement>(null);
-
   const toTimelineTrigger = useScrollTrigger({
     threshold: 50,
     disableHysteresis: true,
@@ -147,6 +149,26 @@ export default function Banner() {
     }
     getGitData();
   }, []);
+
+  // Observer banner
+  useEffect(() => {
+    const options = {
+      root: null,
+      threshold: 0.9,
+    };
+
+    const observer = new IntersectionObserver(
+      (entries) => entries.forEach((entry) => onVisibilityChange(entry.isIntersecting)),
+      options
+    );
+    observer.observe(timelineAnchor.current!);
+
+    const timelineAnchorCopy = timelineAnchor.current;
+
+    return () => {
+      observer.unobserve(timelineAnchorCopy as HTMLDivElement);
+    };
+  }, [onVisibilityChange]);
 
   return (
     <>
--- a/src/web/src/pages/index.tsx
+++ b/src/web/src/pages/index.tsx
@@ -1,13 +1,17 @@
+import { useState } from 'react';
 import SEO from '../components/SEO';
 import Banner from '../components/Banner';
 import Posts from '../components/Posts';
+import NavBar from '../components/NavBar';
 
 const Home = () => {
+  const [bannerIsVisible, setBannerVisibility] = useState(true);
   return (
     <>
       <SEO pageTitle="Telescope" />
-      <Banner />
+      <Banner onVisibilityChange={(visible) => setBannerVisibility(visible)} />
       <main className="main">
+        <NavBar disabled={bannerIsVisible} />
         <Posts />
       </main>
     </>
--- a/src/web/src/components/NavBar/index.tsx
+++ b/src/web/src/components/NavBar/index.tsx
@@ -30,6 +30,7 @@ const useStyles = makeStyles((theme) => ({
     color: theme.palette.primary.main,
     backgroundColor: 'transparent',
     transition: 'width 100ms linear,top 300ms linear',
+    animation: `$navbar 400ms`,
     [theme.breakpoints.down(1200)]: {
       left: '-3vw',
     },
@@ -41,6 +42,25 @@ const useStyles = makeStyles((theme) => ({
       flexDirection: 'row',
       alignItems: 'center',
       background: theme.palette.background.default,
+      animation: `$mobileNavbar 400ms`,
+    },
+  },
+  '@keyframes navbar': {
+    '0%': {
+      opacity: 0,
+      transform: 'translateY(-200%)',
+    },
+    '100%': {
+      opacity: 1,
+      transform: 'translateY(0)',
+    },
+  },
+  '@keyframes mobileNavbar': {
+    '0%': {
+      opacity: 0,
+    },
+    '100%': {
+      opacity: 1,
     },
   },
   toolbar: {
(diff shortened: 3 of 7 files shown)


src/web/src/components/Banner.tsx
```diff
@@ -148,6 +150,37 @@ export default function Banner() {
...
+
+    const observer = new IntersectionObserver(
+      (entries) =>
+        entries.forEach((entry) => {
+          if (entry.isIntersecting) {
+            console.log('seen banner');
+            onVisibilityChange(true);
+          }
+          if (!entry.isIntersecting) {
+            console.log('no banner!~');
```
PedroFonsecaDEV, line 168: Remove the comments.

src/web/src/pages/index.tsx
```diff
@@ -1,13 +1,17 @@
...
 import Posts from '../components/Posts';
+import NavBar from '../components/NavBar';
 
 const Home = () => {
+  const [bannerIsVisible, setBannerVisibility] = useState<boolean>(true);
   return (
     <>
       <SEO pageTitle="Telescope" />
-      <Banner />
+      <Banner onVisibilityChange={(event) => setBannerVisibility(event)} />
```
PedroFonsecaDEV, line 12: `onVisibilityChange={()=> setBannerVisibility(!bannerIsVisible)}`

What do you think?

src/web/src/components/Banner.tsx
```diff
@@ -148,6 +150,37 @@ export default function Banner() {
...
+      root: null,
+      threshold: 0.9,
+    };
+
+    const observer = new IntersectionObserver(
+      (entries) =>
+        entries.forEach((entry) => {
+          if (entry.isIntersecting) {
+            console.log('seen banner');
+            onVisibilityChange(true);
```
PedroFonsecaDEV, line 165: Remove comments

PedroFonsecaDEV (changes requested):
The

chrispinkney:
Wow this is amazing.

A few small nits:
- I notice that you have to scroll a tiny bit down before the buttons actually appear (after scrolling once to get down to the posts), I imagine this is related to the scroll snapping feature we just landed. Any chance to have this ironed out a bit?
- Can we also have the **poof** animation appear, but in reverse when we hide the buttons? (i.e. scrolling back up to the banner)

src/web/src/pages/index.tsx
yuanLeeMidori: I tried this, it doesn't give the correct value somehow

yuanLeeMidori:
@chrispinkney Yes I think it's related to the scroll snap feature, I haven't figured it out yet, there seems to be a small piece between banner and timeline.

And yes that's good ides, let me try it.

PedroFonsecaDEV:
> Wow this is amazing.
> 
> A few small nits:
> 
> * I notice that you have to scroll a tiny bit down before the buttons actually appear (after scrolling once to get down to the posts), I imagine this is related to the scroll snapping feature we just landed. Any chance to have this ironed out a bit?

I don't think it's related to the snap-scroll because I'm tracking this space for some time. I will fix this in a follow-up PR like a little patch ASAP.

src/web/src/components/Banner.tsx
```diff
@@ -90,7 +89,11 @@ const useStyles = makeStyles((theme: Theme) =>
   })
 );
 
-export default function Banner() {
+type BannerProps = {
+  onVisibilityChange: (is: any) => void;
```
DukeManh, line 93: A boolean would be better here, we tend to avoid `any` in TS.
And why do you  name the `onVisibilityChange` argument `is` and `event` in two places.

DukeManh (changes requested):
Wow, great solution.

huynguyez (changes requested):
awesome, but the Nav isn't good on on mobile

src/web/src/components/Banner.tsx
yuanLeeMidori: @DukeManh thank you for the review!
Yes it used to be `=> boolean` but then it'll cause the typescript error when I do `onVisibilityChange={(e) => setBannerVisibility(e)` when I call `<Banner>`.  That's why I change it to `void` instead. 

Do you have suggestion how to solve this error?

yuanLeeMidori:
@huynguyez and the reason being? May you be more specific? Thanks!

src/web/src/components/Banner.tsx
DukeManh: @yuanLeeMidori It looks like the return type `void` is correct. I was referring to the argument `is`. It should be a boolean according to your usage on line 164 and 166.
```suggestion
  onVisibilityChange: (event: boolean) => void;
```

yuanLeeMidori: Got it! Thank you! I've pushed new commit with change :)

src/web/src/components/Banner.tsx
```diff
@@ -90,7 +89,11 @@ const useStyles = makeStyles((theme: Theme) =>
   })
 );
 
-export default function Banner() {
+type BannerProps = {
+  onVisibilityChange: (event: boolean) => void;
```
humphd, line 93: I would rename `event` to `visible`, since it isn't an event.

src/web/src/components/Banner.tsx
```diff
@@ -148,6 +150,35 @@ export default function Banner() {
...
+  // Observer banner
+  useEffect(() => {
+    const options = {
+      root: null,
+      threshold: 0.9,
+    };
+
+    const observer = new IntersectionObserver(
+      (entries) =>
+        entries.forEach((entry) => {
```
humphd, line 162: I think you can simplify these if blocks down to:

```js
entries.forEach((entry) => onVisibilityChange(entry.isIntersecting))
```

humphd: Confirm that I'm right, and there aren't cases I'm missing with this.

src/web/src/components/Banner.tsx
```diff
@@ -148,6 +150,35 @@ export default function Banner() {
...
+      options
+    );
+    observer.observe(timelineAnchor.current!);
+
+    const timelineAnchorCopy = timelineAnchor.current;
+
+    return () => {
+      observer.unobserve(timelineAnchorCopy as HTMLDivElement);
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
humphd, line 179: Is this necessary, or can you add `onVisibilityChange` to it?  Fine if you have to do it this way, just curious.

src/web/src/pages/index.tsx
```diff
@@ -1,13 +1,17 @@
+import { useState } from 'react';
 import SEO from '../components/SEO';
 import Banner from '../components/Banner';
 import Posts from '../components/Posts';
+import NavBar from '../components/NavBar';
 
 const Home = () => {
+  const [bannerIsVisible, setBannerVisibility] = useState<boolean>(true);
```
humphd, line 8: You can drop the `<boolean>` here, since it can infer that it's a boolean based on the initial value of `true`.

humphd (changes requested):
Really cool!  A few comments and questions.

src/web/src/components/Banner.tsx
```diff
@@ -148,6 +150,26 @@ export default function Banner() {
     getGitData();
   }, []);
 
+  // Observer banner
+  useEffect(() => {
+    const options = {
+      root: null,
```
humphd, line 156: We need the element to observer here, which should be the entire `<div>` for the banner I think.

humphd (changes requested):

src/web/src/components/Banner.tsx
```diff
@@ -148,6 +150,26 @@ export default function Banner() {
...
+      threshold: 0.9,
```
humphd, line 157: This should be `0` or something very small I think.

humphd (approved):

PedroFonsecaDEV (approved):

yuanLeeMidori:
follow-up #2112 
cc @DukeManh  @yuanLeeMidori @PedroFonsecaDEV

PR #2098, opened 2021-04-08
Fix login on localhost, add WEB_URL env variable support

humphd:
@PedroFonsecaDEV let me know about a bug locally with login/logout.  Currently, the auth redirect logic is using the Telescope 1.0 API backend URL (http://localhost:3000) as the front-end, but when you run the front-end on http://localhost:8000 it breaks.

This adds a new variable to our environment, `WEB_URL`, allowing us to explicitly set the URL where our front-end is hosted.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,8 @@ ARG API_URL
 ARG IMAGE_URL
 ARG AUTH_URL
 ARG POSTS_URL
+# Front-end Web URL, set via ENV in docker or next build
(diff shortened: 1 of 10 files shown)


HyperTHD (approved):
Was able to login and logout perfect. 
Found a CSS related issue with login though, gonna file an issue on that, but this is GTM

PedroFonsecaDEV (approved):

PR #2099, opened 2021-04-08
Allow Vercel preview PR URLs for auth redirects on staging

humphd:
I want to support testing authentication on staging, which requires that we allow wildcard URLs for Vercel.  This adds support for limited wildcard matching on origins that we'll allow the auth service to use when redirecting a user back with a token.

I split the middleware out into its own file, which is why it looks like I touched more than I did.

NOTE: for this to work fully in preview PRs, we'll also need #2098 to land for the `WEB_URL` fix.


humphd:
This needs https://github.com/Seneca-CDOT/telescope/pull/2104

HyperTHD (approved):

chrispinkney (approved):

PR #2101, opened 2021-04-08
Update access token with more user details for front-end

humphd:
I was talking with @PedroFonsecaDEV about the signup form, and realized that we'll need to include more user details from the Seneca profile in the initial access token for the signup flow.  This adds the following (NOTE: I'm using names [standardized by OpenID](https://openid.net/specs/openid-connect-basic-1_0.html#StandardClaims) for how you include this data in a token):

- `sub` - I switched this to the actual hashed email value, so we don't need to include the hashing function in the front-end
- `email` - I added this, since we wouldn't have the email otherwise (I was on `sub`)
-  `given_name` - the user's first/given name(s)
- `family_name` - the user's last/surname(s)

In the front-end, you can use `useAuth()` and get back a `user` instance that will have `.email`, `.firstName`, `.lastName`, etc.

### DIFF
--- a/config/simplesamlphp-users.php
+++ b/config/simplesamlphp-users.php
@@ -16,13 +16,13 @@ $config = array(
             'uid' => array('1'),
             'eduPersonAffiliation' => array('group1'),
             /*
-              NOTE: we need both `email` and `http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress`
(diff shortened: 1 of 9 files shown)


humphd:
This should get rebased onto https://github.com/Seneca-CDOT/telescope/pull/2103.

chrispinkney (approved):

manekenpix (approved):

humphd:
The failing Vercel build is fixed by https://github.com/Seneca-CDOT/telescope/pull/2104, and unrelated to this.  Merging.

PR #2113, opened 2021-04-09
Fix: add MS search info to env.staging and env.production

manekenpix:
Both the staging and production servers couldn't deploy telescope after adding the search microservice because `env.staging` and `env.production` don't have `port` and `url` information.

This PR adds that to those files.


HyperTHD (approved):

chrispinkney (approved):

humphd:
cc @izhuravlev, this got missed earlier.

PR #2116, opened 2021-04-10
Add auth middleware to Users service

humphd:
This adds *authentication* (makes sure the user has a valid JWT token) and *authorization* (makes sure their token matches what we expect per route) middleware from Satellite to the Users service.  It depends on https://github.com/Seneca-CDOT/satellite/pull/12 for some changes to the `isAuthorized()` middleware in Satellite.

I'm adding middleware to all of the routes, but it's different in each case:

- `GET /` this can be done only by an admin or another microservice (e.g., Parser service).  A regular user can't request this, nor can an anonymous, unauthenticated user.
- `GET /:id` this can be done by the user who owns the data (e.g., I can request my own data),  or by an admin or another microservice.  A user can't request someone else's data, nor can an anonymous, unauthenticated user.
- `POST /:id` this can be done by the user who owns the data (e.g., I can register my own data),  or by an admin or another microservice.  A user can't register someone else's data, nor can an anonymous, unauthenticated user.
- `PUT /:id` this can be done by the user who owns the data (e.g., I can update my own data),  or by an admin or another microservice.  A user can't update someone else's data, nor can an anonymous, unauthenticated user.
- `DELETE /:id` this can be done by the user who owns the data (e.g., I can delete my own data),  or by an admin or another microservice.  A user can't delete someone else's data, nor can an anonymous, unauthenticated user.

I haven't written pure e2e tests here, which would also test things like trying to hit the various routes as different users.  That would require the tests to be updated to full e2e tests first, so we'll have to come back to this.

A note about Satellite's middleware: a route **must** call `isAuthenticated()` **before** `isAuthorized()`, since the latter uses the parsed token created by the former.

The tests are going to fail until the Satellite fix is included.

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,6 +113,7 @@
     "eslint-config-airbnb": "18.2.1",
     "eslint-config-prettier": "8.1.0",
     "eslint-plugin-import": "2.22.1",
+    "eslint-plugin-jest-playwright": "0.2.1",
     "eslint-plugin-jsx-a11y": "6.4.1",
     "eslint-plugin-prettier": "3.3.1",
     "eslint-plugin-promise": "4.3.1",
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "pretest": "npm run lint",
     "test": "npm run jest",
     "jest": "cross-env NODE_ENV=test LOG_LEVEL=error MOCK_REDIS=1 FEED_URL_INTERVAL_MS=200 jest -c jest.config.js --",
-    "jest:e2e": "jest -c jest.config.e2e.js --",
+    "jest:e2e": "jest -c jest.config.e2e.js --runInBand --",
     "coverage": "cross-env NODE_ENV=test LOG_LEVEL=silent MOCK_REDIS=1 FEED_URL_INTERVAL_MS=200 jest -c jest.config.js --collectCoverage --",
     "jest-watch": "cross-env MOCK_REDIS=1 jest -c jest.config.js --watch --",
     "jest-update": "cross-env MOCK_REDIS=1 jest -c jest.config.js --updateSnapshot --",
--- a/src/api/users/test/e2e/users.test.js
+++ b/src/api/users/test/e2e/users.test.js
@@ -1,6 +1,6 @@
 const request = require('supertest');
 const firebaseTesting = require('@firebase/rules-unit-testing');
-const { hash } = require('@senecacdot/satellite');
+const { hash, createServiceToken } = require('@senecacdot/satellite');
 
 const { app } = require('../../src/index');
 const { User } = require('../../src/models/user');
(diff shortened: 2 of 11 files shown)


humphd:
Getting these tests to pass was more challenging than I'd expected, and has resulted in me having to fix a bunch of things in our docker setup and jest configs:

- I added the jest-playwright eslint plugin, so I could use the Playwright globals (page, browser, context)
- I shuffled around a bunch of environment variables in Docker, so things were defined properly in both development and production. We had some things only being set in production.  I need each microservice to be able to create a service token, which means they need JWT info.  I also simplified some things that don't need to be set multiple times (`NODE_ENV`, ports we don't change) or at all (APM stuff, until we do it).  @manekenpix I'm hoping you can help me review this part especially, so that I don't break production.
- I moved the details for how to switch Playwright from headless to having a UI to the `jest-playwright-config.js`, since I don't create the browsers in the tests anymore (you are supposed to let Playwright do it for you, apparently).
- I reduced our test timeout to 10s from 30s, since it made failed tests on CI take forever to finish.
- I changed the `e2e` tests to `runInBand` vs using parallelization, which should help things not get so bogged down.
- I refactored `src/api/auth/test/e2e/auth-flows.test.js` to use Playwright's global `page` and `context` instead of creating them manually.  This also results in all 3 browsers getting used vs. Chrome being used 3 times, like I was doing before.
- I added a service token to the user creation logic in `src/api/auth/test/e2e/auth-flows.test.js` so it would work with the new authenticated routes.

PedroFonsecaDEV (approved):

birtony (approved):

package.json
```diff
@@ -31,7 +31,7 @@
     "pretest": "npm run lint",
     "test": "npm run jest",
     "jest": "cross-env NODE_ENV=test LOG_LEVEL=error MOCK_REDIS=1 FEED_URL_INTERVAL_MS=200 jest -c jest.config.js --",
-    "jest:e2e": "jest -c jest.config.e2e.js --",
+    "jest:e2e": "jest -c jest.config.e2e.js --runInBand --",
```
HyperTHD, line 34: `-i` works here as well

src/api/users/test/e2e/users.test.js
```diff
@@ -367,15 +380,18 @@ describe('DELETE REQUESTS', () => {
...
 
   test('Rejected - Deleted a nonexistent user', async () => {
     const id = createUserHash('no-such-user@email.com');
-    const response = await request(app).delete(`/${id}`).set('Content-Type', 'application/json');
-
+    const response = await request(app)
+      .delete(`/${id}`)
+      .set('Content-Type', 'application/json')
+      .set('Authorization', `bearer ${createServiceToken()}`);
     expect(response.statusCode).toBe(404);
```
HyperTHD, line 395: Do we want to send a 404 status code here? Wouldn't we want to send a 303 and redirect so that when a user is deleted, they go back to the main telescope page instead?

HyperTHD (commented):
Looks good. A few stuff and some questions

package.json
humphd: I don't like using `-i` style flags in scripts, because you have no idea what they mean 6 months from now, and it's hard to search for on google, whereas doing a search for `jest runInBand` is easy.  When I'm typing commands over and over, the shorthand, single-letter version is great.  For scripts, it's almost always better to do full names.

src/api/users/test/e2e/users.test.js
humphd: The Users service has no concept of Telescope.  It's just processing a request to delete a user that isn't there, hence 404.  With microservices, you need to constrain your view to the request/response paradigm vs. thinking in terms of the larger app.  We might choose to do some redirect in the front-end based on this 404 later on, but that's a different issue.

manekenpix (approved):

humphd:
```
FAIL src/api/auth/test/token.test.js
  ● createToken() › should return a JWT with an issued at claim

    expect(received).toBeLessThan(expected)

    Expected: < 1618322319
    Received:   1618322319

      79 |     // The issued at time should be in the past
      80 |     const nowSeconds = Date.now() / 1000;
    > 81 |     expect(iat).toBeLessThan(nowSeconds);
         |                 ^
      82 |   });
      83 |
      84 |   it('should return a JWT without the picture claim, if picture not defined', () => {

      at Object.<anonymous> (src/api/auth/test/token.test.js:81:17)
```

I need to fix this in a follow-up.

PR #2130, opened 2021-04-12
Fixes #2001: Added bottom padding to improve visual break

### LINKED ISSUE #2001: Improve the visual break between posts in timeline
I'm reading @birtony's post and suddenly there is a title for `CoroVi App`, which seems like's it's part of his content vs. a new post:

Screen Shot 2021-03-20 at 12 07 45 PM

I think we need some kind of visual "break" between the end of one post and the start of the next.  Maybe we do this with whitespace or maybe we need something more visible?

rjayroso:
## Issue This PR Addresses
Fixes #2001

## Type of Change

## Description
I've thought about how to improve the visual break between posts. I've looked at how Instagram separates their infinite timeline of posts and it turns out that they don't (unlike tumblr). I think that the easiest solution would just to simply give it more space, see below:

https://user-images.githubusercontent.com/48869737/114341111-fe278f00-9b26-11eb-857c-7494dfe24f7f.mp4

From my reasoning, the next post's title looks like a subtitle for the current post is because the spacing is basically equal and the user struggles to identify who the title belongs to. By adding more space between, it'll be easier to see the title belongs to the post below rather than the previous post above. 

Let me know if we want a more drastic design change to improve the visual break or if the gap should be shortened or made larger.

## How to Test
1. View the file change, ask the following:
    - should I change the unit of measurement? (`px` instead of `rem`?)
2. View deployment, ask the following:
    - is the extra space enough? too little?
    - should this warrant a more drastic design change?

## Checklist

### DIFF
--- a/src/web/src/components/Posts/Post.tsx
+++ b/src/web/src/components/Posts/Post.tsx
@@ -24,6 +24,9 @@ const useStyles = makeStyles((theme: Theme) =>
   createStyles({
     root: {
       padding: '0',
+      paddingBottom: '2em',
+      marginBottom: '3em',
+      borderBottom: '1.5px solid #cccccc',
       fontSize: '1.5rem',
       width: '100%',
       backgroundColor: theme.palette.background.default,


PedroFonsecaDEV:
@rjayroso @humphd 

I think that increasing the space between the posts is not enough. We need to add something else. 
Follow my suggestions:
# 1
Screen Shot 2021-04-12 at 5 15 51 AM


# 2 (I'm using the wrong colors here).

**gif:**
<br/>


<br/>
<br/>

<br/>
<br/>

**img:**

Screen Shot 2021-04-12 at 5 30 52 AM

PedroFonsecaDEV (changes requested):
Please check the comment above.

rjayroso:
@PedroFonsecaDEV  I may have encountered a bug that was giving me trouble, though it may be intentional. That aside, do we still want to put the avatar at the bottom? I think when it is scaled down it will be hard to see and just be noise. I added the line and the "Written by: author" though and copied the colours and border width from the author container on the side, let me know what you think. 

image

https://user-images.githubusercontent.com/48869737/114985364-b96e6180-9e60-11eb-8c4d-6861d36d5889.mp4

---

**As for the aforementioned bug, there is an invisible image at the end of all medium posts.**

```css
.telescope-post-content img[src*='medium.com/_/stat'] {
  width: 1px;
  height: 1px;
}
```
image

If we put a line at the bottom, the gap will appear larger (specifically to medium posts).
I am unsure if I should remove it since I don't know if it serves a purpose elsewhere, thoughts?

src/web/src/components/Posts/Post.tsx
```diff
@@ -172,6 +172,12 @@ const useStyles = makeStyles((theme: Theme) =>
         color: theme.palette.action.selected,
       },
     },
+    contentFooter: {
+      borderBottom: '2.5px solid #707070',
```
PedroFonsecaDEV, line 176: I think we can go with `1.5px`

src/web/src/components/Posts/Post.tsx
```diff
@@ -302,6 +308,14 @@ const PostComponent = ({ postUrl }: Props) => {
           dangerouslySetInnerHTML={{ __html: post.html }}
         />
       </div>
+      <div className={classes.contentFooter}>
+        <p>
+          Written by:{' '}
```
PedroFonsecaDEV, line 313: Let's add the <TelescopeAvatar> as well.

PedroFonsecaDEV (changes requested):
I made some comments.
Regarding the medium ghost image, we can deal with it in a follow-up. Let's keep the scope of this PR.

src/web/src/components/Posts/Post.tsx
chrispinkney, line 313: I'm not sure how I feel about `Written by`, we already show the user on the right, why add it again at the bottom? I do think the line is nice though.

What do others think?

humphd:
What if we try only doing the bar with no extra info about the user or post for now, and iterate on that?  That seems to be the general feedback I'm hearing on our call today.

huynguyez:
@rjayroso @humphd I'll stick with the idea I mentioned earlier in the call:

Keep the header and add a line like the picture below (+ some margin to the upper post), this not only help 2 separate posts more visible but also we can remove the current author info on right side. That place, instead of author info, we can use that space (right) for the share feature in future, or a 'monthly/yearly timeline' (help user browsing old posts by that specific month/year when click on).

Screen Shot 2021-04-16 at 4 51 53 PM

rjayroso:
@PedroFonsecaDEV @huynguyez @chrispinkney, what do you think of the latest [deployment](https://telescope-67bgs5iao-humphd.vercel.app/)? I think for the whole removing the sidebar thing should be done in another issue, lets just get the line and spacing for the visual break. 

Anyways, should the line be close to the top post, in the middle, or as Huy suggested - closer to the bottom? Also: any comments on making a larger/smaller gap, darker/lighter, thinner/bolder line etc?

chrispinkney (changes requested):
image

Looks great! Just needs some more `margin-top` I think, since the gap between the next post and the end of the current post are noticeably different. If others are fine with it then I don't mind giving an approval!

rjayroso:
> image
> 
> Looks great! Just needs some more `margin-top` I think, since the gap between the next post and the end of the current post are noticeably different. If others are fine with it then I don't mind giving an approval!
> 
> I'm not sure what a darker/bolder line would look like but I'm curious to try it. Feel free to suggest some css changes I can apply to the deployment via the browser!

@chrispinkney done n done! rebased and squashed too

yuanLeeMidori (approved):

PedroFonsecaDEV (approved):
LGTM.

PR #2133, opened 2021-04-12
Use Redis for auth service session storage in production

humphd:
I was waiting for Redis to get added to Satellite before I could do this, and now that it's here, I'm switch our auth service to use Redis-backed session storage in production.  Similar to [what we do now in the back-end](https://github.com/Seneca-CDOT/telescope/blob/c385eaae08d20f806d4b1ef6363b81c8fc5f4a60/src/backend/web/app.js#L58).

### DIFF
--- a/src/api/auth/src/index.js
+++ b/src/api/auth/src/index.js
@@ -1,6 +1,7 @@
-const { Satellite } = require('@senecacdot/satellite');
+const { Satellite, Redis } = require('@senecacdot/satellite');
 const passport = require('passport');
 const session = require('express-session');
+const RedisStore = require('connect-redis')(session);
 
 // Setup SAML SSO-based Authentication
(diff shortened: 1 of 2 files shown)


src/api/auth/src/index.js
```diff
@@ -1,6 +1,7 @@
-const { Satellite } = require('@senecacdot/satellite');
+const { Satellite, Redis } = require('@senecacdot/satellite');
 const passport = require('passport');
 const session = require('express-session');
+const RedisStore = require('connect-redis')(session);
```
Metropass, line 4: Reading what `express-session` is, I assume we just pass in a new cookie with no options?

HyperTHD (approved):

birtony (approved):

PR #2142, opened 2021-04-14
Update auth tests to more closely match signup requirements

humphd:
Debugging the new signup flow with @PedroFonsecaDEV and @DukeManh last night made me realize I should expand the signup e2e test to better match what we do in the front-end.

I'm not 100% sure if this will work on all platforms, since I have to use http://host.docker.internal vs. http://localhost in the test.  I want to see how it performs on CI.

### DIFF
--- a/src/api/auth/test/e2e/signup-flow.test.js
+++ b/src/api/auth/test/e2e/signup-flow.test.js
@@ -19,7 +19,7 @@ const galileoGalilei = {
   displayName: 'Galileo Galilei',
   isAdmin: false,
   isFlagged: false,
-  feeds: ['https://hans.blog.com/feed'],
+  feeds: ['http://localhost:8888/feed.xml'],
   github: {
     username: 'hlippershey',
     avatarUrl:
@@ -36,12 +36,13 @@ describe('Signup Flow', () => {
   });
   afterAll(async () => {
     await browser.close();
+    await cleanupTelescopeUsers(users);
   });
 
   beforeEach(async () => {
     context = await browser.newContext();
     page = await browser.newPage();
-    await page.goto(`http://localhost:8888/`);
+    await page.goto(`http://localhost:8888/auth.html`);
   });
   afterEach(async () => {
     await context.close();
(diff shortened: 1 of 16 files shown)


humphd:
Writing these tests was much more involved than I was planning on originally.  To get this to work, I had to add an nginx container in development to serve static content to both the tests (http://localhost:8888) and the docker containers (http://test-web-content).  I also had to make the e2e tests always run in sequence vs. parallel, because the auth/signup tests depend on shared user accounts, and updating the data in parallel tests at the same time leads to random test failures.

cc @birtony, who is thinking about configurations for Jest, and @c3ho who has to write tests for the parser service.  The way I'm hosting the blog/feed in this PR could be used to test the parser service too, since it can connect to this web host running in docker to simulate download content.  You can easily add more and different types of static content for new tests.

PedroFonsecaDEV (approved):

src/api/auth/test/e2e/signup-flow.test.js
```diff
@@ -2,22 +2,24 @@
...
   lastName: 'Galilei',
   email: 'user2@example.com',
   displayName: 'Galileo Galilei',
   isAdmin: false,
   isFlagged: false,
-  feeds: ['https://imaginary.blog.com/feed/hans'],
+  feeds: ['http://localhost:8888/feed.xml'],
   github: {
     username: 'hlippershey',
     avatarUrl:
```
DukeManh, line 25: The form schema doesn't accept empty URLs as every GH account has a default avatar. 
`avatarUrl: Yup.string().url().required()`.
Should we allow it just for testing?

PedroFonsecaDEV: @humphd

humphd: It's on the next line down: `'https://avatars.githubusercontent.com/u/33902374?s=460&u=733c50a2f50ba297ed30f6b5921a511c2f43bfee&v=4',`

DukeManh (approved):
Pedro will file the signup page after this is landed.

PR #2148, opened 2021-04-14
SignUp Page

### LINKED ISSUE #1643: Improve Telescope sign-up flow

I'd like to improve how our sign-up flow works.  Here's my idea, see how you can improve this:

1. a Seneca student is either asked to join Telescope (part of a course), or decides on their own (self-motivated student looking to join the community).  I see a lot of both types of people right now.
2. The student visits Telescope and clicks some kind of "Sign Up" link
3. The begin a multi-step process to create their account(s)
    1. The are taken through the Seneca SAML SSO flow, and need to login
    2. We get their name, email, etc and display it in a form or something.  They have the chance to modify their name (e.g., display name).  Many students don't use the name that Seneca has for them.
    3. Next we ask them to either CREATE or PROVIDE their Blog URL.  If they don't have a blog, we provide instructions on where to go to do it, with links and a couple of suggestions.  If they already have a blog, they enter the URL and we try to guess the RSS Feed (we can do this for most blog providers, since the format is known).
    4. Next we ask them to either CREATE or PROVIDE their GitHub username.  If they don't have a GitHub account, we provide instructions on where to go to do it, with links.  If they already have a GitHub username, they enter it.  We use that to get info for the user, including their GitHub avatar picture URL
    5. We might give them a way to enter other URLs at this point (e.g., Twitter, website, etc)
4. We store the user's info and the various URLs collected above in our database (#1642)
5. The user is now registered in our system, and we take them to their `/profile` page, which shows the info just entered, and allows the user to edit and update their info, examine data being posted about them (e.g. posts) and modify that (e.g., delete a post from our system), delete their account, etc.

We have a very basic flow right now via the `/myfeeds` page, but it's quite limited.  Telescope is amazing, and I want to open it up to more of the Seneca tech community to use and get involved in.  The wiki was a reasonable way to work in the beginning, but we've moved beyond that now.

**Process**:

We'll need to split this work into various bugs, including:

- front-end UI designs
- front-end UI implementation
- back-end updates, tests
- data updates
- data migration from wiki

PedroFonsecaDEV:
## Issue This PR Addresses
Fixes #1643 
 Co-authored-by: @PedroFonsecaDEV 
 Co-authored-by: @Meneguini 
 Co-authored-by: @DukeManh 

## Type of Change

## Description

This PR adds the `signup` page. The style and the design still need to be polished, but we believe that we need to land this feature on our staging serv so we can test with our users (us). 
The styles of the page will be fixed in follow-ups. 

## Checklist

### DIFF
--- a/docker/production.yml
+++ b/docker/production.yml
@@ -48,6 +48,7 @@ services:
       - PORT
       - POSTS_URL
       - API_URL
+      - AUTH_URL
       - WEB_URL
       - SEARCH_URL
       - FEED_DISCOVERY_URL
--- a/docker/production.yml
+++ b/docker/production.yml
@@ -30,6 +30,7 @@ services:
         - AUTH_URL=${AUTH_URL}
         - POSTS_URL=${POSTS_URL}
         - SEARCH_URL=${SEARCH_URL}
+        - FEED_DISCOVERY_URL=${FEED_DISCOVERY_URL}
     container_name: 'telescope'
     restart: unless-stopped
     environment:
@@ -49,6 +50,7 @@ services:
       - API_URL
       - WEB_URL
       - SEARCH_URL
+      - FEED_DISCOVERY_URL
       - LOG_LEVEL
       - FEED_URL
       - FEED_URL_INTERVAL_MS
--- a/src/web/src/components/BannerButtons.tsx
+++ b/src/web/src/components/BannerButtons.tsx
@@ -37,12 +37,17 @@ const ButtonTooltip = withStyles({
 })(Tooltip);
 
 const LandingButtons = () => {
-  const classes = useStyles();
   const { login, logout, user } = useAuth();
+
+  const classes = useStyles();
+
   const router = useRouter();
 
   return (
-    <div className={classes.buttonsContainer}>
+    <div
+      className={classes.buttonsContainer}
+      style={user?.isRegistered ? { width: '155px' } : { width: '250px' }}
+    >
       {user && !user?.isRegistered && (
         <PopUp
           messageTitle="Telescope"
--- a/src/web/src/components/BannerButtons.tsx
+++ b/src/web/src/components/BannerButtons.tsx
@@ -1,8 +1,10 @@
 import Link from 'next/link';
+import { useRouter } from 'next/router';
 import { makeStyles, Tooltip, withStyles, Zoom } from '@material-ui/core';
 import Button from '@material-ui/core/Button';
 import useAuth from '../hooks/use-auth';
 import TelescopeAvatar from './TelescopeAvatar';
+import PopUp from './PopUp';
 
 const useStyles = makeStyles((theme) => ({
   buttonsContainer: {
@@ -37,9 +39,20 @@ const ButtonTooltip = withStyles({
 const LandingButtons = () => {
   const classes = useStyles();
   const { login, logout, user } = useAuth();
+  const router = useRouter();
 
   return (
     <div className={classes.buttonsContainer}>
+      {user && !user?.isRegistered && (
+        <PopUp
+          messageTitle="Telescope"
+          message={`Hello ${user?.name}, you need to create your Telescope account. Click SIGN UP to start.`}
+          agreeAction={() => router.push('/signup')}
+          agreeButtonText="SIGN UP"
+          disagreeAction={() => logout()}
+          disagreeButtonText="ABORT"
+        />
+      )}
       <Link href="/about" passHref>
         <Button
           style={{
@@ -52,20 +65,9 @@ const LandingButtons = () => {
           About us
         </Button>
       </Link>
-      {user ? (
+      {user?.isRegistered ? (
         <>
-          <Button
-            onClick={() => logout()}
-            style={{
-              border: 'none',
-              padding: 0,
-              fontSize: '1.2rem',
-            }}
-            variant="outlined"
-          >
-            Sign out
-          </Button>
-          <ButtonTooltip title="Logout" arrow placement="top" TransitionComponent={Zoom}>
+          <ButtonTooltip title="Sign out" arrow placement="top" TransitionComponent={Zoom}>
             <div>
               <TelescopeAvatar
                 action={() => logout()}
--- a/src/web/src/components/BannerButtons.tsx
+++ b/src/web/src/components/BannerButtons.tsx
@@ -51,7 +51,7 @@ const LandingButtons = () => {
       {user && !user?.isRegistered && (
         <PopUp
           messageTitle="Telescope"
-          message={`Hello ${user?.name}, you need to create your Telescope account. Click SIGN UP to start.`}
+          message={`Hello ${user?.name}, to sign in you need to create your Telescope account. Click SIGN UP to start.`}
           agreeAction={() => router.push('/signup')}
           agreeButtonText="SIGN UP"
           disagreeAction={() => logout()}
--- a/src/web/src/components/BannerButtons.tsx
+++ b/src/web/src/components/BannerButtons.tsx
@@ -1,5 +1,5 @@
 import Link from 'next/link';
-import { makeStyles } from '@material-ui/core';
+import { makeStyles, Tooltip, withStyles, Zoom } from '@material-ui/core';
 import Button from '@material-ui/core/Button';
 import useAuth from '../hooks/use-auth';
 import TelescopeAvatar from './TelescopeAvatar';
@@ -26,6 +26,13 @@ const useStyles = makeStyles((theme) => ({
     },
   },
 }));
+
+const ButtonTooltip = withStyles({
+  tooltip: {
+    fontSize: '1.5rem',
+    margin: 0,
+  },
+})(Tooltip);
 
 const LandingButtons = () => {
   const classes = useStyles();
@@ -58,7 +65,16 @@ const LandingButtons = () => {
           >
             Sign out
           </Button>
-          <TelescopeAvatar name={user.name} img={user.avatarUrl} size="40px" />
+          <ButtonTooltip title="Logout" arrow placement="top" TransitionComponent={Zoom}>
+            <div>
+              <TelescopeAvatar
+                action={() => logout()}
+                name={user.name}
+                img={user.avatarUrl}
+                size="40px"
+              />
+            </div>
+          </ButtonTooltip>
         </>
       ) : (
         <>
--- a/src/web/src/components/BannerButtons.tsx
+++ b/src/web/src/components/BannerButtons.tsx
@@ -55,7 +55,7 @@ const LandingButtons = () => {
           agreeAction={() => router.push('/signup')}
           agreeButtonText="SIGN UP"
           disagreeAction={() => logout()}
-          disagreeButtonText="ABORT"
+          disagreeButtonText="CANCEL"
         />
       )}
       <Link href="/about" passHref>
--- a/src/web/src/components/BannerButtons.tsx
+++ b/src/web/src/components/BannerButtons.tsx
@@ -73,14 +73,16 @@ const LandingButtons = () => {
           >
             Sign in
           </Button>
-          <Button
-            style={{
-              fontSize: '1.3rem',
-            }}
-            variant="outlined"
-          >
-            Sign up
-          </Button>
+          <Link href="/signup" passHref>
+            <Button
+              style={{
+                fontSize: '1.3rem',
+              }}
+              variant="outlined"
+            >
+              Sign up
+            </Button>
+          </Link>
         </>
       )}
     </div>
--- a/src/web/src/components/SignUp/Forms/GitHubAccount.tsx
+++ b/src/web/src/components/SignUp/Forms/GitHubAccount.tsx
@@ -109,38 +109,38 @@ const GitHubAccount = connect<{}, SignUpForm>((props) => {
   const controllerRef = useRef<AbortController | null>();
 
   const validateGit = async () => {
-    if (values.githubUsername) {
+    if (!values.githubUsername) {
+      setError('');
+      setFieldValue('github', {}, true);
+      return;
+    }
+    try {
       setValidating(true);
       if (controllerRef.current) {
         controllerRef.current.abort();
       }
       controllerRef.current = new AbortController();
-      try {
-        const response = await fetch(`${gitHubApiUrl}/${values.githubUsername}`, {
-          signal: controllerRef.current?.signal,
-        });
-        if (!response.ok) {
-          throw new Error(response.statusText);
-        }
-        const res = await response.json();
-
-        setFieldValue('github', {
-          username: res.login,
-          avatarUrl: res.avatar_url,
-        });
-        setError('');
-      } catch (err) {
-        console.error(err, 'Unable to get GitHub profile');
-
-        setError('Unable to get GitHub profile');
-        setFieldValue('github', {}, true);
-      } finally {
-        setValidating(false);
-        controllerRef.current = null;
+      const response = await fetch(`${gitHubApiUrl}/${values.githubUsername}`, {
+        signal: controllerRef.current?.signal,
+      });
+      if (!response.ok) {
+        throw new Error(response.statusText);
       }
-    } else {
+      const res = await response.json();
+
+      setFieldValue('github', {
+        username: res.login,
+        avatarUrl: res.avatar_url,
+      });
       setError('');
+    } catch (err) {
+      console.error(err, 'Unable to get GitHub profile');
+
+      setError('Unable to get GitHub profile');
       setFieldValue('github', {}, true);
+    } finally {
+      setValidating(false);
+      controllerRef.current = null;
     }
   };
 
--- a/src/web/src/components/SignUp/Forms/GitHubAccount.tsx
+++ b/src/web/src/components/SignUp/Forms/GitHubAccount.tsx
@@ -1,0 +1,172 @@
+/* eslint-disable camelcase */
+import { useEffect, useState } from 'react';
+import { createStyles, makeStyles, Theme } from '@material-ui/core';
+import { connect } from 'formik';
+import useSWR from 'swr';
+
+import { SignUpForm } from '../../../interfaces';
+import formModels from '../Schema/FormModel';
+import { TextInput, CheckBoxInput } from '../FormFields';
+import PostAvatar from '../../Posts/PostAvatar';
+
+const { githubUsername, github: githubModel, githubOwnership } = formModels;
+
+const useStyles = makeStyles((theme: Theme) =>
+  createStyles({
+    root: {
+      padding: '0',
+      margin: '0',
+      width: '100%',
+      position: 'relative',
+      minHeight: '100%',
+    },
+    container: {
+      display: 'grid',
+      gridTemplateAreas: '1fr',
+      textAlign: 'center',
+      justifyItems: 'center',
+      alignItems: 'start',
+      width: '100%',
+      position: 'absolute',
+      minHeight: '100%',
+      [theme.breakpoints.down(600)]: {
+        width: '90%',
+        marginLeft: '5%',
+      },
+    },
+    titlePage: {
+      fontSize: '1.5em',
+    },
+    subtitlePage: {
+      fontSize: '1.1em',
+      lineHeight: '1.8em',
+    },
+    infoContainer: {
+      display: 'grid',
+      gridTemplateColumns: '65% 35%',
+      gridGap: '1%',
+      textAlign: 'center',
+      justifyItems: 'center',
+      alignItems: 'center',
+      width: '90%',
+      [theme.breakpoints.down(600)]: {
+        gridTemplateColumns: '1fr',
+      },
+    },
+    inputsContainer: {
+      width: '100%',
+      display: 'grid',
+      gridTemplateColumns: '100%',
+      '& .MuiFormHelperText-root': {
+        fontSize: '0.9em',
+        color: 'black',
+      },
+      '& .MuiFormLabel-root': {
+        color: 'black',
+      },
+    },
+    avatarPreview: {
+      textAlign: 'center',
+      justifyItems: 'center',
+      alignItems: 'center',
+      justifySelf: 'end',
+      padding: '6%',
+      borderRadius: '5px',
+      [theme.breakpoints.down(600)]: {
+        justifySelf: 'center',
+        padding: '3%',
+        marginTop: '5%',
+      },
+    },
+    username: {
+      fontSize: '1.1em',
+    },
+  })
+);
+
+const gitHubApiUrl = 'https://api.github.com/users';
+
+const GitHubAccount = connect<{}, SignUpForm>((props) => {
+  const classes = useStyles();
+  const { values, setFieldValue } = props.formik;
+  const [username, setUsername] = useState(values.githubUsername);
+  const [inputTimeout, setInputTimeout] = useState(setTimeout(() => {}, 0));
+
+  const { data: github, error } = useSWR(
+    values.githubUsername ? `${gitHubApiUrl}/${values.githubUsername}` : null,
+    async (u) => {
+      try {
+        const response = await fetch(u);
+        if (!response.ok) {
+          throw new Error(response.statusText);
+        }
+        return response.json();
+      } catch (err) {
+        throw err;
+      }
+    }
+  );
+
+  const handleInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setUsername(e.target.value);
+    clearTimeout(inputTimeout);
+
+    // Update githubUsername 1000ms after input change
+    setInputTimeout(
+      setTimeout(() => {
+        setFieldValue('githubUsername', e.target.value);
+      }, 1000)
+    );
+  };
+
+  useEffect(() => {
+    if (error) {
+      setFieldValue('github', {}, true);
+    }
+
+    if (github) {
+      setFieldValue(
+        'github',
+        {
+          username: github.login,
+          avatarUrl: github.avatar_url,
+        },
+        true
+      );
+    }
+  }, [github, error, setFieldValue]);
+
+  return (
+    <div className={classes.root}>
+      <div className={classes.container}>
+        <h1 className={classes.titlePage}>GitHub Account</h1>
+        <h2 className={classes.subtitlePage}>Enter Github username and verify your profile</h2>
+        <div className={classes.infoContainer}>
+          <div className={classes.inputsContainer}>
+            <TextInput
+              label={githubUsername.label}
+              name={githubUsername.name}
+              error={!!error}
+              helperText={!!error && githubModel.invalidErrorMsg}
+              onChange={handleInputChange}
+              value={username}
+            />
+          </div>
+          {!error && github && (
+            <div className={classes.avatarPreview}>
+              <PostAvatar name={github.login} blog={github.avatar_url} img={github.avatar_url} />
+              <h2 className={classes.username}>{github.login}</h2>
+            </div>
+          )}
+        </div>
+        <CheckBoxInput
+          label={githubOwnership.label}
+          name={githubOwnership.name}
+          checked={values.githubOwnership}
+        />
+      </div>
+    </div>
+  );
+});
+
+export default GitHubAccount;
--- a/src/web/src/components/SignUp/Forms/GitHubAccount.tsx
+++ b/src/web/src/components/SignUp/Forms/GitHubAccount.tsx
@@ -1,15 +1,14 @@
-/* eslint-disable camelcase */
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
+import Button from '@material-ui/core/Button';
 import { createStyles, makeStyles, Theme } from '@material-ui/core';
 import { connect } from 'formik';
-import useSWR from 'swr';
 
 import { SignUpForm } from '../../../interfaces';
 import formModels from '../Schema/FormModel';
 import { TextInput, CheckBoxInput } from '../FormFields';
 import PostAvatar from '../../Posts/PostAvatar';
 
-const { githubUsername, github: githubModel, githubOwnership } = formModels;
+const { githubUsername, github, githubOwnership } = formModels;
 
 const useStyles = makeStyles((theme: Theme) =>
   createStyles({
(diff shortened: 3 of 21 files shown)


humphd:
Tagging some more reviewers.  I'd like to prioritize this PR so we can move toward getting the signup flow working in the front-end soon, and have time to fix bugs before 2.0.

huynguyez (changes requested):
is it a page but looks like a dialog? 
I think we should give user a chance to go back to homepage other than clicking on "back" button.

PedroFonsecaDEV:
> is it a page but looks like a dialog?
> I think we should give user a chance to go back to homepage other than clicking on "back" button.

@huynguyez 
Follow ups. We are landing the functionality.

yuanLeeMidori (commented):
image


I assume this is part of the sign-up and there will be more follow-up PRs, but should the "Login" button be "Next"? Or, should there be a "Next" button based on the instruction in this page? Or, it'll be in the upcoming PRs?

Great design as usual!

DukeManh:
@yuanLeeMidori, you are right, Pedro will update it. More PRs are coming but feel free to comment on anything.

HyperTHD (changes requested):
Don't forget to add the `USERS_URL` and the `FEED_DISCOVERY_URL` as args in both production.yml and the Dockerfile in the root directory, otherwise, you won't be able to test this in staging and production.

yuanLeeMidori (dismissed):

src/web/src/components/SignUp/Forms/GitHubAccount.tsx
```diff
@@ -0,0 +1,172 @@
...
+  })
+);
+
+const gitHubApiUrl = 'https://api.github.com/users';
+
+const GitHubAccount = connect<{}, SignUpForm>((props) => {
+  const classes = useStyles();
+  const { values, setFieldValue } = props.formik;
+  const [username, setUsername] = useState(values.githubUsername);
+  const [inputTimeout, setInputTimeout] = useState(setTimeout(() => {}, 0));
```
humphd, line 93: This is not a great pattern, because the GitHub API will rate-limit you if you make too many requests.  If a user types, deletes, types, etc over and over we may get locked out.  I'd suggest you add a button so they can indicate they are done, and do the request then.

This can happen now or in a follow-up, but make sure you file before you land this.

src/web/src/components/SignUp/Forms/GitHubAccount.tsx
```diff
@@ -0,0 +1,172 @@
...
+  const { data: github, error } = useSWR(
+    values.githubUsername ? `${gitHubApiUrl}/${values.githubUsername}` : null,
+    async (u) => {
+      try {
+        const response = await fetch(u);
+        if (!response.ok) {
+          throw new Error(response.statusText);
+        }
+        return response.json();
+      } catch (err) {
```
humphd, line 104: Don't bother with try/catch if all you do is throw the error.

src/web/src/components/SignUp/Forms/GitHubAccount.tsx
```diff
@@ -0,0 +1,172 @@
...
+      <div className={classes.container}>
+        <h1 className={classes.titlePage}>GitHub Account</h1>
+        <h2 className={classes.subtitlePage}>Enter Github username and verify your profile</h2>
+        <div className={classes.infoContainer}>
+          <div className={classes.inputsContainer}>
+            <TextInput
+              label={githubUsername.label}
+              name={githubUsername.name}
+              error={!!error}
+              helperText={!!error && githubModel.invalidErrorMsg}
```
humphd, line 150: It might be worth using your try/catch above to set some error text so this works, for example: "Unable to connect to GitHub".

src/web/src/components/SignUp/Forms/RSSFeeds.tsx
```diff
@@ -0,0 +1,264 @@
...
+  const classes = useStyles();
+  const { values, errors, setFieldValue } = props.formik;
+  const { token } = useAuth();
+
+  const [feedUrls, setFeedUrls] = useState<Array<string>>([]);
+  const [blogUrlError, setBlogUrlError] = useState('');
+  const [validating, setValidating] = useState(false);
+
+  // A controller to cancel undesired requests
+  const [controller, setController] = useState(new AbortController());
```
humphd, line 147: I think this makes more sense as a Ref vs. State, https://frontend-digest.com/cancelling-fetch-requests-in-react-applications-58a52a048e8e

src/web/src/components/SignUp/Forms/RSSFeeds.tsx
```diff
@@ -0,0 +1,264 @@
...
+
+  const validateBlog = async () => {
+    if (!errors.blogUrl && feedDiscoveryServiceUrl) {
```
humphd, line 150: When would `feedDiscoveryServieUrl` be null?

src/web/src/components/SignUp/Schema/FormSchema.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+// Each signup step has one validation schema
+export default [
+  // First step has no validation logic
+  Yup.object().shape({}),
+
+  Yup.object().shape({
+    [firstName.name]: Yup.string()
+      .required(`${firstName.requiredErrorMsg}`)
+      .test('len', firstName.invalidErrorMsg, validateLength(2, 16)),
```
humphd, line 30: Why not use the built-in string length assertions? https://github.com/jquense/yup#stringlengthlimit-number--ref-message-string--function-schema

humphd: Also, where did you come up with 2 and 16?  What if a user's name is longer?  I guarantee we'll have longer values.

humphd: I would suggest we require the name to exist and nothing about max length.

src/web/src/pages/signup.tsx
```diff
@@ -0,0 +1,271 @@
...
+      textAlign: 'center',
+      fontSize: '0.9em',
+      color: '#474747',
+    },
+  })
+);
+
+const SignUpPage = () => {
+  const classes = useStyles();
+  const [activeStep, setActiveStep] = useState<number>(0);
```
humphd, line 136: Let's use an `enum` to name these steps vs. 0, 1, 2... which are meaningless.

https://www.typescriptlang.org/docs/handbook/enums.html

src/web/src/pages/signup.tsx
```diff
@@ -0,0 +1,271 @@
...
+  const currentSchema = formSchema[activeStep];
+  const { user, token, login } = useAuth();
+  const [loggedIn, setLoggedIn] = useState(false);
```
humphd, line 139: Do we need this vs. using `!!user` below, or `const loggedIn = !!user;` here?

src/web/src/pages/signup.tsx
```diff
@@ -0,0 +1,271 @@
...
+          method: 'post',
+          headers: {
+            Authorization: `bearer ${token}`,
+            'Content-Type': 'application/json',
+          },
+          body: JSON.stringify(telescopeUser),
+        });
+
+        if (!response.ok) {
+          alert('Unable to create account.');
```
humphd, line 177: This is not ideal.  In a follow-up, let's do some proper UI for this.

humphd (changes requested):
This is so awesome to see, thank you to all of you for pushing on it.  I've tested it, and it's generally working.  There are things I'd like to improve about wording, UI, but they don't need to block this.  I left a few notes about issues, questions, and areas for follow-up.

src/web/src/components/SignUp/Forms/GitHubAccount.tsx
DukeManh: We should've logged the error there.

src/web/src/pages/signup.tsx
DukeManh: We used `!!user` but NextJS gave `Text content did not match` error and the buttons were not updated according to `user` value. This must have something to do with NextJS's server-side-rendering React.
See also https://aboutmonica.com/blog/server-side-rendering-react-hydration-best-practices.

image
`Next` button doesn't change color and is not clickable after login.

I wonder if we can fix this by changing how useAuth() passes values.

src/web/src/components/SignUp/Schema/FormSchema.tsx
DukeManh: > I would suggest we require the name to exist and nothing about max length.

Thanks, I wasn't sure what the right limits were.

src/web/src/components/SignUp/Forms/GitHubAccount.tsx
DukeManh: I imagine that would happen only when users type slowly, or they want to abuse the feature by feeding multiple values.
The profile is only fetched `1000ms` after users stop typing, which makes just 1 request when I test it.

DukeManh:
@humphd Thanks for the review. I left some comments.

src/web/src/components/SignUp/Forms/GitHubAccount.tsx
humphd: Regardless, let's use a button.  Network-requests-as-you-type are not a great idea

src/web/src/pages/signup.tsx
humphd: OK, let's leave it as is, and change your default value in 139 from `false` to `!!user`, to properly reflect the initial value.

humphd:
This needs a rebase.

src/web/src/pages/signup.tsx
DukeManh: Ok

src/web/src/components/SignUp/Forms/GitHubAccount.tsx
DukeManh: Ok, we'll fix this in this PR.

humphd:
When I land #2154 (either before or after this), we can change how we `POST` the form data, so it goes to the Auth service's `/register` endpoint, and then we can get back an updated token with the user's info, including the avatar URL.

src/web/src/pages/signup.tsx
```diff
@@ -0,0 +1,271 @@
...
+        const { firstName, lastName, email, displayName, github, feeds } = values;
+        const telescopeUser = {
+          firstName,
+          lastName,
+          email,
+          displayName,
+          github,
+          feeds,
+        };
+        const response = await fetch(`${usersServiceUrl}/${user?.id}`, {
```
PedroFonsecaDEV, line 167: @humphd @DukeManh 
After #2154 we need to change this fetch to the new route right?

DukeManh: Yeah, it would be /register so the token is updated.

PedroFonsecaDEV: OKay. Let's keep track of #2154.

src/web/src/components/BannerButtons.tsx
```diff
@@ -73,14 +73,16 @@ const LandingButtons = () => {
...
-            variant="outlined"
-          >
-            Sign up
-          </Button>
+          <Link href="/signup" passHref>
+            <Button
+              style={{
+                fontSize: '1.3rem',
+              }}
+              variant="outlined"
```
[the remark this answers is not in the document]
PedroFonsecaDEV: Okay. Investigating.

src/web/src/pages/signup.tsx
PedroFonsecaDEV: This problem persists after commit `Change request (#22)`

src/web/src/components/SignUp/Forms/GitHubAccount.tsx
PedroFonsecaDEV: Done.

src/web/src/components/SignUp/Schema/FormSchema.tsx
PedroFonsecaDEV: Done.

src/web/src/pages/signup.tsx
PedroFonsecaDEV: Done.

humphd:
@PedroFonsecaDEV @DukeManh these are the changes from our debugging session that I threw away, FYI:

```diff
diff --git a/src/web/src/pages/signup.tsx b/src/web/src/pages/signup.tsx
index a34a7bc8..9d7b341f 100644
--- a/src/web/src/pages/signup.tsx
+++ b/src/web/src/pages/signup.tsx
@@ -14,7 +14,7 @@ import DynamicImage from '../components/DynamicImage';
 import { SignUpForm } from '../interfaces';
 import formModels from '../components/SignUp/Schema/FormModel';
 import formSchema from '../components/SignUp/Schema/FormSchema';
-import { usersServiceUrl } from '../config';
+import { authServiceUrl } from '../config';
 
 const {
   firstName,
@@ -132,10 +132,15 @@ const useStyles = makeStyles((theme: Theme) =>
 
 const SignUpPage = () => {
   const classes = useStyles();
-  const [activeStep, setActiveStep] = useState<number>(0);
+  const [activeStep, setActiveStep] = useState(1);
   const currentSchema = formSchema[activeStep];
-  const { user, token, login } = useAuth();
-  const [loggedIn, setLoggedIn] = useState(!!user);
+  const { user, token, login, register } = useAuth();
+  // const [loggedIn, setLoggedIn] = useState(!!user);
+
+  // If the user hasn't logged in yet, present an overview and Login button
+  if (!user) {
+    return <Overview />;
+  }
 
   const handleNext = () => {
     setActiveStep(activeStep + 1);
@@ -145,13 +150,6 @@ const SignUpPage = () => {
     if (activeStep > 1) setActiveStep(activeStep - 1);
   };
 
-  useEffect(() => {
-    if (user) {
-      setLoggedIn(!!user);
-      handleNext();
-    }
-  }, []);
-
   const handleSubmit = async (values: SignUpForm, actions: FormikHelpers<SignUpForm>) => {
     if (activeStep === 4) {
       try {
@@ -164,8 +162,7 @@ const SignUpPage = () => {
           github,
           feeds,
         };
-        // TODO Update register URL
-        const response = await fetch(`${usersServiceUrl}/${user?.id}`, {
+        const response = await fetch(`${authServiceUrl}/register`, {
           method: 'post',
           headers: {
             Authorization: `bearer ${token}`,
@@ -177,7 +174,9 @@ const SignUpPage = () => {
         if (!response.ok) {
           throw new Error(response.statusText);
         }
-        login();
+
+        const result = await response.json();
+        register(result.token);
         return;
       } catch (err) {
         console.error(err, 'Unable to Post');
@@ -191,8 +190,6 @@ const SignUpPage = () => {
 
   const renderForm = () => {
     switch (activeStep) {
-      case 0:
-        return <Overview />;
       case 1:
         return <BasicInfo />;
       case 2:
@@ -238,23 +235,23 @@ const SignUpPage = () => {
             <>
               <Form autoComplete="off" className={classes.infoContainer}>
                 {renderForm()}
-                {!loggedIn && (
+                {!user && (
                   <div className={classes.text}>
                     <h3>Click LOGIN to start creating your Telescope Account.</h3>
                   </div>
                 )}
                 <div className={classes.buttonsWrapper}>
-                  {!loggedIn && (
+                  {!user && (
                     <Button className={classes.buttonLogin} onClick={() => login('/signup')}>
                       Login
                     </Button>
                   )}
-                  {activeStep > 0 && loggedIn && (
+                  {activeStep > 0 && user && (
                     <Button className={classes.button} onClick={handlePrevious}>
                       Previous
                     </Button>
                   )}
-                  {activeStep > 0 && loggedIn && (
+                  {activeStep > 0 && user && (
                     <Button className={classes.button} type="submit" disabled={isSubmitting}>
                       {activeStep === 4 ? 'Confirm' : 'Next'}
                     </Button>
```

humphd:
After you rebase, this is what I did to get the `AuthProvider` to work:

https://github.com/humphd/telescope/commit/8ff4ea863d40cc21de10e014bf4456239ddc3247

humphd:
1. Rebase this, get rid of all the `USERS_URL` stuff you added, since we really need only `AUTH_URL` in the front-end (next.config.js, config.ts, docker, etc)
2. `git fetch humphd` and `git cherry-pick 8ff4ea863d40cc21de10e014bf4456239ddc3247` to get the `AuthProvider` changes
3. Rework `signup.tsx` so the initial Login page logic is separate from your steps: if you have a user, render the steps, otherwise render the overview/login stuff.  Make sure you use bits from https://github.com/Seneca-CDOT/telescope/pull/2148#issuecomment-821314173 that are still correct (e.g., how we do the `fetch()` and get the new `token` to pass to `register()`

src/web/src/components/BannerButtons.tsx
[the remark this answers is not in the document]
humphd: Did you want to deal with this in a follow-up?  If so, let's file it please.

src/web/src/components/SignUp/Forms/GitHubAccount.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+const GitHubAccount = connect<{}, SignUpForm>((props) => {
+  const classes = useStyles();
+  const { values, setFieldValue } = props.formik;
+  const [validating, setValidating] = useState(false);
+  const [error, setError] = useState('');
+  const controllerRef = useRef<AbortController | null>();
+
+  const validateGit = async () => {
+    if (values.githubUsername) {
```
humphd, line 112: Deal with the inverse case first, exit early and left-align the bulk of the code.

```js
if(!values.githubUsername) {
   setError('');
   setFieldValue('github', {}, true);
   return;
}
// rest of code here, left aligned...

File: src/web/src/components/SignUp/Forms/RSSFeeds.tsx [hunk 13]
```diff
@@ -0,0 +1,263 @@
...
+
+  const validateBlog = async () => {
+    if (!errors.blogUrl) {
+      try {
+        setValidating(true);
+        if (controllerRef.current) {
+          controllerRef.current.abort();
+        }
+        controllerRef.current = new AbortController();
+        const response = await fetch(`${feedDiscoveryServiceUrl}`, {
```
humphd, line 155: `fetch(feedDiscoveryServiceUrl)` with out the `${}` since it's already a string.

src/web/src/components/SignUp/Forms/RSSFeeds.tsx
```diff
@@ -0,0 +1,263 @@
...
+  const { values, errors, setFieldValue } = props.formik;
+  const { token } = useAuth();
+
+  const [feedUrls, setFeedUrls] = useState<Array<string>>([]);
+  const [blogUrlError, setBlogUrlError] = useState('');
+  const [validating, setValidating] = useState(false);
+  const controllerRef = useRef<AbortController | null>();
+
+  const validateBlog = async () => {
+    if (!errors.blogUrl) {
```
humphd, line 148: Again, flip the logic and deal with the error case first, return, then do the rest of the code.

src/web/src/pages/signup.tsx
```diff
@@ -0,0 +1,276 @@
...
+  const handleNext = () => {
+    setActiveStep(activeStep + 1);
+  };
+
+  const handlePrevious = () => {
+    if (activeStep > 1) setActiveStep(activeStep - 1);
+  };
+
+  useEffect(() => {
+    console.log('useEffect-USER');
```
humphd, line 149: Remove

src/web/src/pages/signup.tsx
```diff
@@ -0,0 +1,276 @@
...
+    if (user) {
+      setLoggedIn(true);
+      console.log('USER:', user);
```
humphd, line 152: Remove

humphd (changes requested):
A few small things, this is close!

src/web/src/components/SignUp/Forms/GitHubAccount.tsx
```diff
@@ -0,0 +1,191 @@
...
+        controllerRef.current.abort();
+      }
+      controllerRef.current = new AbortController();
+      const response = await fetch(`${gitHubApiUrl}/${values.githubUsername}`, {
+        signal: controllerRef.current?.signal,
+      });
+      if (!response.ok) {
+        throw new Error(response.statusText);
+      }
+      const res = await response.json();
```
humphd, line 129: Nit: `res` isn't a great name here, since we often use it to mean response.  I think you might mean `result`, which would be a better choice.

src/web/src/components/SignUp/Forms/GitHubAccount.tsx
```diff
@@ -0,0 +1,191 @@
...
+            />
+            <Button className={classes.button} onClick={validateGit} disabled={validating}>
+              Get profile
+            </Button>
+          </div>
+          {!error && (
+            <div className={classes.avatarPreview}>
+              <PostAvatar
+                name={values.github.username || values.displayName}
+                blog={values.github.avatarUrl}
```
humphd, line 174: This is odd, since it's not a blog.  I filed https://github.com/Seneca-CDOT/telescope/issues/2169 to fix it later.

src/web/src/components/SignUp/Forms/RSSFeeds.tsx
```diff
@@ -0,0 +1,262 @@
...
+      setFieldValue('feeds', [], true);
+      return;
+    }
+    try {
+      setValidating(true);
+      if (controllerRef.current) {
+        controllerRef.current.abort();
+      }
+      controllerRef.current = new AbortController();
+      const response = await fetch(`${feedDiscoveryServiceUrl}`, {
```
humphd, line 157: This got missed.  Please remove the template literal around `feedDiscoveryServiceUrl` since it's already a string, and you're using the bare value without modification:

```js
const response = await fetch(feedDiscoveryServiceUrl, {
```

src/web/src/components/SignUp/Schema/FormModel.tsx
```diff
@@ -0,0 +1,55 @@
+export default {
+  displayName: {
+    name: 'displayName',
+    label: 'Display Name',
+    requiredErrorMsg: 'Display Name is required',
```
humphd, line 5: For consistency with the errors below, `Display name` (lower case `name`).

src/web/src/components/SignUp/Schema/FormModel.tsx
```diff
@@ -0,0 +1,55 @@
...
+  },
+  githubOwnership: {
+    name: 'githubOwnership',
+    label: 'I declare I’m the owner and the maintainer of this GitHub account',
+    invalidErrorMsg: 'You must be the owner of this account',
+  },
+  blogUrl: {
+    name: 'blogUrl',
+    label: 'Blog URl',
+    requiredErrorMsg: 'Blog Url is required',
```
humphd, line 39: `Url` -> `URL` for consistency with the message below.

src/web/src/components/SignUp/Schema/FormSchema.tsx
```diff
@@ -0,0 +1,57 @@
...
+  githubOwnership,
+  feeds,
+  allFeeds,
+  blogUrl,
+  blogOwnership,
+} = formModels;
+
+// Each signup step has one validation schema
+export default [
+  // First step has no validation logic
```
humphd, line 20: This array is confusing.  Can I suggest you name each of these or at the very least put a comment above each section and name it?

```js
// Comment explaining what step 1 is
const step1 = Yup.object().shape({}),

// Comment explaining what step 2 is
const step2 = ...
...

export default[
  step1,
  step2,
  ...
];
```

or

```js
export default[
  // Explain step 1
  Yup.object().shape({}),

  // Explain step 2
  ...

 // Explain step 3
 ...
];
```

humphd (changes requested):
I did another read through of the code, only a few things left to deal with, and I think this is good to land.  We can fix other bits in follow-up.

Testing this locally, the default case works well.  Some things I notice, which we can fix in follow-ups:

1. If I log in as a Seneca user (i.e., don't click signup), it's impossible to signup without logging out. We should probably include the Signup button if `!user.isRegistered`:
Screen Shot 2021-04-17 at 10 28 55 AM

2. After you click Confirm, there is an odd render where you get a Next and Previous button with no other content, maybe due to the value of token/user changing in the `AuthProvider`, and causing a re-render before the `router.push('/')` kicks in?  We might want to have some other state/check that shows a spinner or something vs. the form.

3. We need a way to exit the signup flow if you decide you don't want to do it, and just want to go back to the home page.  Maybe we should put an `X` in the top-right of the form/popup window and/or include a `Cancel` button in the steps?

4. After a user successfully signs up, it might be nice to render some kind of "Welcome!  Here's what you should do next..." and give some encouragement to go write a blog post introducing themselves, etc.  Right now we just drop them in the home page, and if you don't see the updated avatar in the top-right, you might not realize that it worked.

5. We need a Profile page, where a Telescope user can go and update their info, add/modify feeds, delete their account, etc.

6. Clicking the Avatar at the top-right of the screen does nothing.  But clicking the avatar at the left in the side-bar menu automatically logs you out.  I think they should do the same thing.  I wouldn't expect that clicking my avatar would log me out.

7. If I Signup and it works, logout, and click Signup again, it lets me go through the entire process, but then fails at the very last step with no UI to indicate what went wrong.  We shouldn't allow a user to do the signup flow if `user.isRegistered` is `true`.  We also need to add some error UI if the call to `/register` fails for some reason vs. letting the user sit there and not know what happened.  At the very least, we should give an error and a button to click to try again.

Screen Shot 2021-04-17 at 10 39 52 AM

src/web/src/components/BannerButtons.tsx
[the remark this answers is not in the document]
PedroFonsecaDEV: @humphd 
This bug seems to be solved after yesterday's interaction. 
Do you continue to experience this outline?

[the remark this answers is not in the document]
humphd: I couldn't reproduce, no.  I think it's fixed.

humphd:
@PedroFonsecaDEV I know you're active on this.  Let me know when to review it again, so that I don't waste your time reviewing stuff that's only in progress.

src/web/src/components/SignUp/Forms/GitHubAccount.tsx
PedroFonsecaDEV: Done

src/web/src/components/SignUp/Forms/RSSFeeds.tsx
PedroFonsecaDEV: I tried: `const response = await fetch(feedDiscoveryServiceUrl, {`

Then we have the following issue: 
```(alias) const feedDiscoveryServiceUrl: string | undefined
import feedDiscoveryServiceUrl
Argument of type 'string | undefined' is not assignable to parameter of type 'RequestInfo'.
  Type 'undefined' is not assignable to type 'RequestInfo'.ts(2345)
```


So I'm leaving this line in the same way:  
```
const response = await fetch(`${feedDiscoveryServiceUrl}`, {
```

src/web/src/components/SignUp/Schema/FormModel.tsx
PedroFonsecaDEV: Done

src/web/src/components/SignUp/Schema/FormModel.tsx
PedroFonsecaDEV: Done.

src/web/src/components/SignUp/Schema/FormSchema.tsx
PedroFonsecaDEV: Done

src/web/src/components/SignUp/Forms/RSSFeeds.tsx
PedroFonsecaDEV: @DukeManh 
I think this is done right?

humphd (dismissed):
I think this is ready to land, and anything else we do can happen in follow ups.  Amazing work.

docker/production.yml
```diff
@@ -30,6 +30,7 @@ services:
         - AUTH_URL=${AUTH_URL}
         - POSTS_URL=${POSTS_URL}
         - SEARCH_URL=${SEARCH_URL}
+        - FEED_DISCOVERY_URL=${FEED_DISCOVERY_URL}
```
PedroFonsecaDEV, line 33: @humphd 
Do we need `'USERS_URL'` here?

docker/production.yml
```diff
@@ -49,6 +50,7 @@ services:
       - API_URL
       - WEB_URL
       - SEARCH_URL
```
PedroFonsecaDEV, line 52: @humphd 
same here for: `  'AUTH_URL'` and `'USERS_URL'` ?
reference: https://github.com/Seneca-CDOT/telescope/pull/2148#discussion_r614091986

docker/production.yml
humphd: We should probably add it for future use.  But no, it's not needed now.

docker/production.yml
humphd: Yes, we need `AUTH_URL` in here, and if we're going to include `USERS_URL` above, we need it here too.

docker/production.yml
PedroFonsecaDEV: Ok. 
I won't add the `USER_URL` for now.

docker/production.yml
PedroFonsecaDEV: Okay.  
Tks. 

Done.

src/web/src/components/BannerButtons.tsx
```diff
@@ -27,12 +29,35 @@ const useStyles = makeStyles((theme) => ({
...
+      style={user?.isRegistered ? { width: '155px' } : { width: '250px' }}
+    >
+      {user && !user?.isRegistered && (
+        <PopUp
+          messageTitle="Telescope"
+          message={`Hello ${user?.name}, to sign in you need to create your Telescope account. Click SIGN UP to start.`}
+          agreeAction={() => router.push('/signup')}
+          agreeButtonText="SIGN UP"
+          disagreeAction={() => logout()}
+          disagreeButtonText="ABORT"
```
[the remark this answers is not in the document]
PedroFonsecaDEV: Done.

tonyvugithub (approved):

huynguyez:
kinda broken CSS in small screen.

PedroFonsecaDEV:
> kinda broken CSS in small screen.

@huynguyez 
I need this to land to work on the CSS. 
Is waiting for this PR to land #2171

humphd (approved):

PR #2154, opened 2021-04-15
Add /register route to auth service

humphd:
The work in #2148 revealed a few things about the signup flow that I wanted to fix.  First, instead of having the front-end `POST` the user's Telescope info to the Users service, I'm now proxying it through the Auth service via the `/register` route.  I do this in part because I want to make it possible to upgrade the user's access token to include the new info that's being added for them.

A user has to be authenticated as a Seneca user (i.e., have a token with the `'seneca'` role) before they can register--this part is the same.  However, after the user is done filling out the signup form, the front-end can `POST /register` on the auth service, sending the form data for the user.  If all goes well, a new user will be created in the Users service, and the Telescope profile data will be used to generate an upgraded access token, which gets returned to the front-end.  This will need to get set in the AuthProvider (we need code for this) and then the user sent to the home page.

We can land this before or after #2148, and then update the front-end to use this new route.

I've updated and expanded the tests to show how this all works, and to make sure we don't break it.  I also refactored the routes to make them a bit less overwhelming to read, putting all the various bits that are related into separate files.


PedroFonsecaDEV (approved):

DukeManh (approved):

PedroFonsecaDEV:
@humphd rebase and good to go ?

PR #2163, opened 2021-04-16
Fixes #1983: Improve SearchHelp logic

### LINKED ISSUE #1983: Improve SearchHelp logic
When there are no posts loaded in the SearchPage, the SearchHelp component should be visible. As it [currently is](https://github.com/Seneca-CDOT/telescope/pull/1909), the SearchHelp can only be viewed on the initial viewing of the SearchPage, and after the first search is submitted, can no longer be viewed again until the page is refreshed.

Although it is intended for the SearchHelp to disappear once the user searches, it would also be good for the user to see the SearchHelp anytime there are no posts loaded (i.e. if a user inputs a search that return no results OR if they clear the search bar, etc.)

rjayroso:
## Issue This PR Addresses
Fixes #1983 

## Type of Change

## Description
Previously, as soon as you hit the submit button the SearchHelp would disappear. The only way to get the SearchHelp to appear again would be to refresh the page.

It would be beneficial if the SearchHelp would appear when the User's search returned no results (case A) or if the search bar was empty (case B).

This is better logically speaking since in those two additional scenarios we can assume that the User either failed their search or are about to search for something, both of which would benefit from seeing the SearchHelp. That is what this PR accomplishes.

Read more [here](https://github.com/Seneca-CDOT/telescope/issues/1983)

## How to Test
1. Search `validation.js` (it should return a timeline of posts)
    - scroll all the way to the end to check if the `SearchHelp` component truly disappeared (it should be hidden)
2. (case A) Search `fasdfsadf` (it should return `No Results Found`)
    - the `SearchHelp` component should be visible
3. (case B) Clear your search bar and hit submit
    - the `SearchHelp` component should be visible

## Checklist

### DIFF
--- a/src/web/src/components/SearchPage.tsx
+++ b/src/web/src/components/SearchPage.tsx
@@ -21,14 +21,13 @@ const useStyles = makeStyles((theme: Theme) => ({
 
 const SearchPage = () => {
   const classes = useStyles();
-
   const { showHelp } = useSearchValue();
 
   return (
     <div className={classes.searchPage}>
       <SearchBar />
-      <SearchResults />
       {showHelp && <SearchHelp />}
+      <SearchResults />
     </div>
   );
 };
--- a/src/web/src/components/SearchProvider.tsx
+++ b/src/web/src/components/SearchProvider.tsx
@@ -10,6 +10,7 @@ export interface SearchContextInterface {
   textParam: string;
   filter: FilterProp['filter'];
   showHelp: boolean;
+  toggleHelp: (value: boolean) => void;
   onTextChange: (value: string) => void;
   onFilterChange: (value: FilterProp['filter']) => void;
   onSubmitHandler: (value: FormEvent) => void;
@@ -20,6 +21,9 @@ const SearchContext = createContext<SearchContextInterface>({
   textParam: '',
   filter: 'post',
   showHelp: true,
+  toggleHelp() {
+    throw new Error('This context must be wrapped inside SearchProvider');
+  },
   onTextChange() {
     throw new Error('This context must be wrapped inside SearchProvider');
   },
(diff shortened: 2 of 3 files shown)


chrispinkney (commented):
Found a minor bug (though I think this was pointed out during our meeting today) where the search help is not shown if a you search for nothing, after searching for something.

You can see it occur in this video:
https://user-images.githubusercontent.com/7242003/115086271-edd23400-9ed9-11eb-98ff-d38fb53dcc16.mp4

Not sure if this is related to your PR or not. Looks great otherwise though!

yuanLeeMidori (approved):
Except the one Chris mentioned, all look great! This search help logic makes more sense.

rjayroso:
> Found a minor bug (though I think this was pointed out during our meeting today) where the search help is not shown if a you search for nothing, after searching for something.
> 
> You can see it occur in this video:
> https://user-images.githubusercontent.com/7242003/115086271-edd23400-9ed9-11eb-98ff-d38fb53dcc16.mp4
> 
> Not sure if this is related to your PR or not. Looks great otherwise though!

```js
  // If there is no posts or if the search bar is empty, then show the search help, otherwise hide it
  if (!error && (isEmpty || textParam.length == 0)) {
    toggleHelp(true);
  } else {
    toggleHelp(false);
  }
```

So I noticed in your demo that it returned a server error; how I have it right now is that if there is an error we don't display the `SearchHelp`, but I'm thinking - maybe this shouldn't be the case?

chrispinkney:
> 
> 
> @chrispinkney
> 
> > Found a minor bug (though I think this was pointed out during our meeting today) where the search help is not shown if a you search for nothing, after searching for something.
> > You can see it occur in this video:
> > https://user-images.githubusercontent.com/7242003/115086271-edd23400-9ed9-11eb-98ff-d38fb53dcc16.mp4
> > Not sure if this is related to your PR or not. Looks great otherwise though!
> 
> ```js
>   // If there is no posts or if the search bar is empty, then show the search help, otherwise hide it
>   if (!error && (isEmpty || textParam.length == 0)) {
>     toggleHelp(true);
>   } else {
>     toggleHelp(false);
>   }
> ```
> 
> So I noticed in your demo that it returned a server error; how I have it right now is that if there is an error we don't display the `SearchHelp`, but I'm thinking - maybe this shouldn't be the case? There seems to be a weird search bug where it will return error if you search for nothing.

Weird. If the bug isn't coming from your PR then yeah I think the current behaviour is fine and it should not show the search help.

rjayroso:
@yuanLeeMidori @chrispinkney rebased

PR #2165, opened 2021-04-16
Fixes #1938: Updated healthCheck for Posts service

### LINKED ISSUE #1938: Update posts healthcheck with Redis check
Every [Satellite microservice includes a `/healthcheck`](https://github.com/Seneca-CDOT/satellite#satelliteoptions), which can be used to determine if the service is running.  For example, the current posts healthcheck is:

https://dev.api.telescope.cdot.systems/v1/posts/healthcheck

To make this really useful, we also need to know if any dependent service is also healthy.  In the case of posts, this means Redis needs to be up and accepting commands.

Redis has a [ping](https://redis.io/commands/PING) command that we can call and make sure it returns OK.

HyperTHD:
## Issue This PR Addresses
This PR closes #1938 

## Type of Change

## Description

With Redis added to Satellite, I'm able to update the posts service's health check to check if redis is functioning by using the ping function from Redis. This PR also updates the routes for the posts service to use the createError function and call that function using the `next` function as part of the catch.

## Checklist


src/api/posts/src/index.js
```diff
@@ -2,10 +2,10 @@ const { Satellite, Redis } = require('@senecacdot/satellite');
 
 const posts = require('./routes/posts');
 
+const redis = Redis();
+
 const service = new Satellite({
   healthCheck: async () => {
```
humphd, line 8: You should return the promise directly

```js
healthCheck: () => redis.ping(),
```

humphd (changes requested):
One small thing and this is ready.

Reading your code, I also realize that we need to add a way to tell the Redis connection to close just as we're shutting down.  Right now we don't expose a hook for this from Satellite, but we can add one.

src/api/posts/src/index.js
```diff
@@ -1,9 +1,17 @@
-const { Satellite } = require('@senecacdot/satellite');
+const { Satellite, Redis } = require('@senecacdot/satellite');
 
 const posts = require('./routes/posts');
 
-const service = new Satellite();
+const redis = Redis();
+
+const service = new Satellite({
+  healthCheck: () => {
```
humphd, line 8: You're not returning the promise here.  Should be: `healthCheck: () => redis.ping()`

humphd (changes requested):
I think to get the unit tests to pass, you're going to have to move your `redis` instance out of `index.js` and into its own file?  Not sure, but `redis` being undefined in the tests is odd.

chrispinkney (dismissed):
LGTM besides the nit pointed out by Dave.

HyperTHD:
@humphd. I might need to wait for the shutdown option to be merged into Satellite. Even in it's own file, redis seems to be undefined which is super weird

humphd:
I had an epiphany while on a walk today: did we add Redis since the last npm update for Satellite?  We probably need to ship an update that includes it.  I just triggered 1.12.0, which should hit npm within the next 30 mins.  Let's try that and see what happens here.

humphd (dismissed):
This looks good.  The only minor nit I have is that you're only exporting a single thing from `redis.js`, so I would `module.exports = redis` instead, so you can `const redis = require('./redis');` and not bother with destructuring.  But it's minor.

I agree, we should do a follow-up to add the shutdown logic for redis too, once that lands.

src/api/posts/src/index.js
```diff
@@ -1,8 +1,11 @@
 const { Satellite } = require('@senecacdot/satellite');
+const { redis } = require('./redis');
```
humphd, line 2:
```js
const = redis = require('./redis');
```

src/api/posts/src/storage.js
```diff
@@ -1,6 +1,7 @@
 const { logger } = require('@senecacdot/satellite');
-const { redis } = require('./libs/redis');
+const { redis } = require('./redis');
```
humphd, line 2: Same thing here, no `{...}` anymore.

humphd (changes requested):

src/api/posts/src/index.js
HyperTHD: You mean 
```js
const redis = require('./redis'); 
```
right?

humphd (dismissed):
Perfect.

chrispinkney (dismissed):
LGTM

HyperTHD:
Rebased, @humphd @chrispinkney, let's get this merged in

humphd (approved):

humphd (dismissed):

PR #2172, opened 2021-04-18
Fixes #1801 Code Highlighting

### LINKED ISSUE #1801: Code escaping/highlighting bug
Noticed reading [this post](https://dev.to/chrispinkney/il-meglio-e-l-inimico-del-bene-25f2) from @chrispinkney:

Screen Shot 2021-02-20 at 9 51 32 AM

Metropass:
## Issue This PR Addresses
Fixes #1801, Code Escaping Bug

## Type of Change

## Description

So, we've discussed about fixing this bug for quite a while.
Last week, I've audited the code, and described how the bug works, and went through it's call stacks to get an understanding of how the code works. 

@chrispinkney had the suggestion to use a fork of our `sanitize-html` tool specifically for React, however there's some really bad vulnerabilities that hasn't been patched, so we can't use it.

I then decided that the best solution was to use the DOMPurify tool instead:
https://www.npmjs.com/package/dompurify

There are quite a few problems though:
I needed to use the `RETURN_DOM` option in order for the sanitizer to NOT remove escape characters. Unfortunately, it wraps everything into a `<body>` tag. 

Since we're using DOMPurify, there's no option to whitelist `iframe` links, so we should find an npm package that will allow us to add our whitelist.

Screencaps of what I'm talking about:

Without Returning the String as a DOM (wrapping it with the `body` tag), the escape character is removed.
Screenshot_424

When we use the `RETURN_DOM` option to wrap everything, notice the Escape Characters not being removed, but the `body` tag remains.
Screenshot_425

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
     "cors": "2.8.5",
     "date-fns": "2.19.0",
     "docker-compose": "0.23.6",
+    "dompurify": "2.2.7",
     "dotenv": "8.2.0",
     "entities": "2.2.0",
     "express": "4.17.1",
@@ -91,7 +92,7 @@
     "pino": "6.11.2",
     "pino-elasticsearch": "5.4.0",
     "pino-pretty": "4.7.1",
-    "sanitize-html": "2.3.2",
+    "sanitize-html-react": "1.13.0",
     "set-interval-async": "1.0.34",
     "stoppable": "1.1.0",
     "valid-url": "1.0.9"
--- a/src/backend/utils/html/sanitize.js
+++ b/src/backend/utils/html/sanitize.js
@@ -1,12 +1,10 @@
 // Sanitize HTML and prevent XSS attacks
 // <img src=x onerror=alert(1)> becomes <img src="x">
 
-const sanitizeHtml = require('sanitize-html');
-
+const sanitizeHtml = require('dompurify');
 module.exports = function (dirty) {
-  return sanitizeHtml(dirty, {
-    // https://github.com/apostrophecms/sanitize-html#what-are-the-default-options
-    allowedTags: [
+  var e = sanitizeHtml.sanitize(dirty, {
+    ALLOWED_TAGS: [
       'a',
       'b',
       'blockquote',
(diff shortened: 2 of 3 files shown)


src/backend/utils/html/sanitize.js
```diff
@@ -40,19 +38,23 @@ module.exports = function (dirty) {
       'thread',
       'tr',
       'ul',
+      'MDXPageBase',
```
humphd, line 41: This is a losing battle.  There's no way that we can add every custom element that people will blog about, so no matter what we do, this will break when a new one pops up.

humphd (changes requested):
I don't think this moves us forward.  The only thing that's happened here is a new backend has been used, but you've had to manually add custom tag names to allow them.  I don't see that scaling.

package.json
```diff
@@ -91,7 +92,7 @@
     "pino": "6.11.2",
     "pino-elasticsearch": "5.4.0",
     "pino-pretty": "4.7.1",
-    "sanitize-html": "2.3.2",
+    "sanitize-html-react": "1.13.0",
```
humphd, line 95: This doesn't appear to be used anywhere.

humphd (changes requested):

src/backend/utils/html/sanitize.js
Metropass: The code actually works without adding `MDXPageBase` into the list, however we still have the issues of the new package that's still there.

I think we should change our plan of attack, and instead create a custom tag that the sanitizer will ignore. 

The past week I've been experimenting with the old `sanitize-html` package, and I found no probable way of fixing the About Us page without heavily changing the behaviour of how the sanitizer parsers our code (which I don't want to do since it's dangerous).

Should I bring this up on our meeting on Tuesday?

chrispinkney:
This is not currently fixable.

humphd:
We discussed this in triage the other day, and I don't think we can fix things this way.  We'll close for now and maybe revisit this issue later, or just leave it as a known limitation.

PR #2176, opened 2021-04-19
Fixes #2169: Rename blog prop to url in PostAvatar, deal with optional value

### LINKED ISSUE #2169: Rename blog prop to url in PostAvatar, deal with optional value
We use `blog` for the link's `href` in the `PostAvatar`:

```tsx
const PostAvatar = ({ name, img, blog = '' }: AvatarProps) => {
  const classes = useStyles();

  if (img) {
    return <Avatar className={classes.avatar} src={img} />;
  }

  if (name.length > 0) {
    const initials = name
      .split(' ')
      // splitName represents the current value, i represents its index, and arr represent the whole splitted name-word array
      // if the index is 0, means it's the first word in the name
      // if the index+1 equals to the array length, it means the current value is the last word in the name
      // anything rather than first word or last word will not be included in initials
      .map((splitName, i, arr) =>
        i === 0 || i + 1 === arr.length ? splitName[0].toUpperCase() : null
      )
      .join('');
    return (
      <a href={blog} className={classes.link}>
        <Avatar className={classes.avatar}>
          <p className={classes.text}>{initials}</p>
        </Avatar>
      </a>
    );
  }

  return <Avatar className={classes.avatar} />;
};
```

In the new signup code we're re-using this component to show the GitHub avatar in the signup form, and it's odd to name the URL we pass `blog` (it isn't a blog).

Let's rename `blog` to the more generic `url`, and update all callers to do the same.

Also, `blog` is an optional prop, but we always use it.  We should be returning an `<Avatar>` without a parent `<a>` if we don't get passed a `url`.

rjayroso:
## Issue This PR Addresses
Fixes #2169

## Type of Change

## Description
Changes the PostAvatar `blog` prop to be named `url`. 
Changes how the PostAvatar is rendered: if there is no url, then PostAvatar is rendered without an `<a>` element wrapped around it.

## How to Test
Pull this branch for testing, go to `components/Posts/Post.tsx` in VSCode (or your editor of choice) and scroll to the bottom - put a `<PostAvatar name="FirstName LastName" />` element somewhere that it can be rendered and seen. 

You should see an Avatar with the initials "FL" and it should also be unclickable (since their is no url prop)

Now add the prop `url="Test"` to the `PostAvatar` (e.g. `<PostAvatar name="FirstName LastName" url="Test"/>`

You should see that same Avatar now be clickable (if clicked it will give you Telescope's 404 because it is an invalid url).

## Checklist

### DIFF
--- a/src/web/src/components/Posts/PostAvatar.tsx
+++ b/src/web/src/components/Posts/PostAvatar.tsx
@@ -4,7 +4,7 @@ import { makeStyles } from '@material-ui/core/styles';
 type AvatarProps = {
   name: string;
   img?: string;
-  blog?: string;
+  url?: string;
 };
 
 const useStyles = makeStyles((theme: Theme) =>
@@ -39,7 +39,7 @@ const useStyles = makeStyles((theme: Theme) =>
   })
 );
 
-const PostAvatar = ({ name, img, blog = '' }: AvatarProps) => {
+const PostAvatar = ({ name, img, url }: AvatarProps) => {
   const classes = useStyles();
 
   if (img) {
@@ -58,11 +58,19 @@ const PostAvatar = ({ name, img, blog = '' }: AvatarProps) => {
       )
       .join('');
     return (
-      <a href={blog} className={classes.link}>
-        <Avatar className={classes.avatar}>
-          <p className={classes.text}>{initials}</p>
-        </Avatar>
-      </a>
+      <div>
+        {url?.length ? (
+          <a href={url} className={classes.link}>
+            <Avatar className={classes.avatar}>
+              <p className={classes.text}>{initials}</p>
+            </Avatar>
+          </a>
+        ) : (
+          <Avatar className={classes.avatar}>
+            <p className={classes.text}>{initials}</p>
+          </Avatar>
+        )}
+      </div>
     );
   }
 
--- a/src/web/src/components/Posts/Post.tsx
+++ b/src/web/src/components/Posts/Post.tsx
@@ -277,7 +277,7 @@ const PostComponent = ({ postUrl }: Props) => {
         {!desktop && (
           <>
             <div className={classes.authorAvatarContainer}>
-              <PostAvatar name={post.feed.author} blog={post.feed.link} />
+              <PostAvatar name={post.feed.author} url={post.feed.link} />
             </div>
             <div className={classes.authorNameContainer}>
               <h1 className={classes.author}>
--- a/src/web/src/components/Posts/PostInfo.tsx
+++ b/src/web/src/components/Posts/PostInfo.tsx
@@ -76,7 +76,7 @@ const PostDesktopInfo = ({ authorName, postDate, blogUrl, postUrl }: Props) => {
   return (
     <ListSubheader className={classes.root}>
       <div className={classes.authorAvatarContainer}>
-        <PostAvatar name={authorName} blog={blogUrl} />
+        <PostAvatar name={authorName} url={blogUrl} />
       </div>
       <div className={classes.authorNameContainer}>
         <p className={classes.author}>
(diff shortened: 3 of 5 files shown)


src/web/src/components/Posts/PostAvatar.tsx
```diff
@@ -58,11 +58,19 @@ const PostAvatar = ({ name, img, blog = '' }: AvatarProps) => {
       )
       .join('');
     return (
-      <a href={blog} className={classes.link}>
-        <Avatar className={classes.avatar}>
-          <p className={classes.text}>{initials}</p>
-        </Avatar>
-      </a>
+      <div>
+        {url.length > 0 ? (
```
HyperTHD, line 62: I'm not sure about this, but shouldn't we also be checking if url is not null as well, basically if it exists as well as if the length is greater than 0

rjayroso: So, I think from what I understand after reading the `PostAvatar.tsx` component is that, if no `url` is given in the props, then the `url` is assigned to a `""` empty string. Either way, I can add it quickly it's no problem

src/web/src/components/Posts/PostAvatar.tsx
```diff
@@ -59,7 +59,7 @@ const PostAvatar = ({ name, img, url = '' }: AvatarProps) => {
       .join('');
     return (
       <div>
-        {url.length > 0 ? (
+        {url.length > 0 && url != undefined ? (
```
HyperTHD, line 62: Think just `url.length > 0 && url` is fine enough

humphd:
Please update the title of this PR to give info about what it is, in addition to what it fixes.

src/web/src/components/Posts/PostAvatar.tsx
```diff
@@ -39,7 +39,7 @@ const useStyles = makeStyles((theme: Theme) =>
   })
 );
 
-const PostAvatar = ({ name, img, blog = '' }: AvatarProps) => {
+const PostAvatar = ({ name, img, url = '' }: AvatarProps) => {
```
humphd, line 42: Don't use `= ''` as a default.  If it's not there, we need to know that.

src/web/src/components/Posts/PostAvatar.tsx
humphd: `url?.length && (...)`

humphd (changes requested):

src/web/src/components/Posts/PostAvatar.tsx
```diff
@@ -58,11 +58,19 @@ const PostAvatar = ({ name, img, blog = '' }: AvatarProps) => {
       )
       .join('');
     return (
-      <a href={blog} className={classes.link}>
-        <Avatar className={classes.avatar}>
-          <p className={classes.text}>{initials}</p>
-        </Avatar>
-      </a>
+      <div>
+        {url != undefined && url?.length > 0 ? (
```
HyperTHD, line 62: This is really gonna sound nit-picky, but should be `!==` instead of `!=`

HyperTHD (changes requested):

src/web/src/components/Posts/PostAvatar.tsx
humphd: Doing `url?` is the same as checking if `url` exists, so you don't need this.  However, @HyperTHD is right: you never want to use != or == in JS, since they both allow implicit type casting to happen, whereas !== and === will also check the types.

humphd (changes requested):

yuanLeeMidori (dismissed):
Just do squash and merge, otherwise looks good for me :)

rjayroso:
Squashed and rebased

humphd (approved):

yuanLeeMidori (approved):

humphd:
@rjayroso you can rebase and merge this from GitHub, you don't have to keep rebasing manually.

humphd (approved):

yuanLeeMidori (approved):

PedroFonsecaDEV (approved):

humphd:
@rjayroso wait to merge this til we fix CI, which is failing atm

rjayroso:
> @rjayroso wait to merge this til we fix CI, which is failing atm

Yeah no worries, now that I know I can rebase and merge from GitHub I'm ready whenever 😅

PR #2213, opened 2021-04-23
Initial work on moving to pnpm from npm

humphd:
I have been experimenting with [pnpm](https://pnpm.io/) for improving our package installation speed/size requirements, and it's really very good.  I don't think we can merge this yet, since it doesn't work with Dependabot yet, though it is [likely to be added at some point](https://github.com/dependabot/dependabot-core/issues/1736).

The major benefit of pnpm is that it will install a dependency only once, and if multiple packages (e.g., sub-projects like the microservices) use the same package, you only have to install it once.  I've found that this is much, much faster, and also requires less disk space (~700M vs. 1.8G).

I wouldn't recommend running `pnpm install` in the same tree where you have already run `npm install`, but rather to do it in a clean fork if you want to test it.  To get rid of `node_modules` recursively see: https://css-tricks.com/how-to-delete-all-node_modules-directories-from-your-computer/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,6 +3,9 @@
 # Please see the documentation for all configuration options:
 # https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
 
+# TODO: update for pnpm when possible:
+# https://github.com/dependabot/dependabot-core/issues/1736
+
 version: 2
 updates:
   # Root package.json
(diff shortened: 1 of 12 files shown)


humphd:
I've filed a bug upstream on the pnpm GitHub Action, which is failing.

humphd:
I fixed one issue (most of CI passing now), but hit a new one, where passing args to `pnpm run` is currently broken.  There's a bug filed upstream, and I've switched to `npm` for that one case, which is fine for now.

humphd:
I think this is working now, but the e2e tests are failing because of a random installation failure in one microservice for the mock elasticsearch client.  I'm going to wait and try again tomorrow to see if it fixes itself.

PR #2239, opened 2021-05-14
Closes Issue #2153 -  Limit how a user becomes an admin #2153 

### LINKED ISSUE #2153: Limit how a user becomes an admin
We have the concept of an `admin` role in the User and Auth services, in the front-end, etc.  In Firebase it's a boolean, `isAdmin: true|false`.  We need to tighten how a user becomes an admin.

You shouldn't be able to POST that data to the Users service (we should ignore it).   The current system is too loose.

chrispinkney:
Co-authored-by: Josue manekenpix@fastmail.com

## Issue This PR Addresses
Closes #2153

## Type of Change

## Description
This PR limits how a user can become an admin:
- It does this by forcing the `PUT` and `POST` routes to set `user.isAdmin` to `false`.
- It also adds another route for admins to modify another user's `isAdmin`.

## Checklist


src/api/users/src/routes/users.js
```diff
@@ -111,6 +111,8 @@ router.post(
         next(createError(409, `user with id ${id} already exists.`));
       } else {
         const user = new User(body);
+        // no user can be created as an admin by default
+        user.isAdmin = false;
```
humphd, line 115: I would move this code into a new middleware function that happens just before the route's logic gets hit.  Maybe call it `validateUserRights()` and it can simply set `user.isAdmin` to `false`.

The logic in the route should be focused on talking to the db and returning HTTP results, nothing more.

src/api/users/src/routes/users.js
```diff
@@ -140,6 +142,43 @@ router.put(
...
 
+    try {
+      const userRef = db.doc(id);
+      const doc = await userRef.get();
+
+      if (!doc.exists) {
+        next(createError(404, `user ${id} not found.`));
+      } else {
+        const user = new User(body);
+        // no user can be created as an admin by default
```
humphd, line 153: Same, this can be done via middleware.

src/api/users/src/routes/users.js
```diff
@@ -140,6 +142,43 @@ router.put(
     const { id } = req.params;
     const { body } = req;
 
+    try {
+      const userRef = db.doc(id);
+      const doc = await userRef.get();
+
+      if (!doc.exists) {
```
humphd, line 149: This code seems to be duplicated between a few `put` calls now.  Can you extract it into middleware and share it?

src/api/users/src/routes/users.js
```diff
@@ -140,6 +142,43 @@ router.put(
...
+    }
+  }
+);
+
+// put (update) a user with a supplied id, validated by the schema
+// rejected if a user could not be found, otherwise user updated
+// this route is only accessible by administrators
+// it allows the modification of the isAdmin property
+router.put(
+  '/:id/admin',
```
humphd, line 170: This looks great, but one question: the naming implies that you are only setting the `admin` aspect of this record vs. all fields, yet you are allowing other fields to be updated here.  I would limit it to just the admin change.

src/api/users/src/routes/users.js
```diff
@@ -140,6 +142,43 @@ router.put(
...
+  isAuthenticated(),
+  validateId(),
+  validateUser(),
+  validateEmailHash(),
+  isAuthorized((req, user) => {
```
humphd, line 175: Arrow functions with body statements that immediately return a value should be rewritten as inline expressions:

```js
isAuthorized((req, user) => user.roles.includes('admin')),
```

humphd (changes requested):
This also lacks tests, which I'd like to see get added with these changes.

PedroFonsecaDEV:
Following.

src/api/users/src/routes/users.js
chrispinkney: `validateUserRights` created.

src/api/users/src/routes/users.js
chrispinkney: Done.

src/api/users/src/routes/users.js
chrispinkney: Duplication removed, code moved to `updateUser` in `schema.js`

src/api/users/src/routes/users.js
chrispinkney: Also done in `updateUser`

src/api/users/src/routes/users.js
chrispinkney: Done

chrispinkney:
Still have to add tests but I think this is heading in the right direction.

config/export/firebase-export-metadata.json
```diff
@@ -0,0 +1,8 @@
+{
```
humphd, line 1: Why did this file, and the other `config/export/*` files, get added?

docker/development.yml
```diff
@@ -45,7 +45,8 @@ services:
     volumes:
       # Copy firebase.config into the container so we get proper ip/port binding
       - ../config/firebase.json:/home/node/firebase.json
-    command: firebase emulators:start --project telescope --only firestore
+      - ../config/export:/export
+    command: firebase emulators:start --project telescope --only firestore --import=/export
```
humphd, line 49: Why this change?

src/api/users/src/models/schema.js
```diff
@@ -16,6 +16,34 @@ const validateId = () =>
...
+
+  try {
+    const userRef = db.doc(id);
+    const doc = await userRef.get();
+
+    if (!doc.exists) {
+      next(createError(404, `user ${id} not found.`));
+    } else {
+      // if makeAdmin is true turn the user into an admin, if not use body to update user.
+      const user = new User(makeAdmin ? { ...body, isAdmin: true } : body);
```
humphd, line 37: This won't allow downgrading a user from Admin to normal, which we should support.  We should accept a value `{isAdmin: true|false}` on the body and use that instead of hard-coding here.

humphd (changes requested):

config/export/firebase-export-metadata.json
chrispinkney: This and the change to `development.yml` add a single user to the emulator up its bootup. This user as the `isAdmin` flag set to true, as without this we'd have no way to make a user an admin during testing.

The problem stems from having no way of creating an admin without first having an admin to create an admin.

chrispinkney:
Given #2555, I wonder if this PR should be closed and migrated to a different issue at a later date (once the suprabase work has been merged)?

@manekenpix
@humphd

humphd:
Agree, this can get closed for now; though, I think it's a good starting point for what needs to happen.

PR #2242, opened 2021-05-17
Fixes #2240: Audit of microservices to use fetch from Satellite

### LINKED ISSUE #2240: Use Satellite's fetch() function and remove local deps for microservices
Now that #2135 is fixed, and v1.16.0 of Satellite is released, we can use `const { fetch }  = require('@senecacdot/satellite')` to get a `fetch()` function in our microservices.  Any of the services that are currently using `fetch()` or `got()` or the like to do client HTTP requests should switch to use this.

HyperTHD:
## Issue This PR Addresses
This PR closes #2240 

## Type of Change

## Description

Satellite version 1.16 has added node-fetch to it's api. This PR does an audit of all microservices that use `node-fetch` or `got` and switches to use fetch provided by Satellite.

NOTE: I'm not quite sure how to replicate `got.stream` with fetch, so I appreciate any feedback on that. Everything else should hopefully be good.

## Checklist

### DIFF
--- a/src/api/feed-discovery/src/middleware.js
+++ b/src/api/feed-discovery/src/middleware.js
@@ -1,7 +1,6 @@
-const got = require('got');
 const cheerio = require('cheerio');
 const validUrl = require('valid-url');
-const { logger, createError } = require('@senecacdot/satellite');
+const { logger, createError, fetch } = require('@senecacdot/satellite');
 
 // A middleware to check if Url provided is valid
 module.exports.checkValidUrl = function checkValidUrl() {
@@ -20,14 +19,14 @@ module.exports.checkValidUrl = function checkValidUrl() {
 module.exports.checkValidBlog = function checkValidBlog() {
   return async (req, res, next) => {
     try {
-      const { statusCode, headers, body } = await got(req.body.blogUrl);
-      const contentType = headers['content-type'];
+      const response = await fetch(req.body.blogUrl);
+      const contentType = response.headers;
       // If status code is not 200 or content-type is not text/html then send 400 error
-      if (!(statusCode === 200 && contentType.includes('text/html'))) {
+      if (!(response.status === 200 && contentType.includes('text/html'))) {
         next(createError(400, 'Invalid Blog'));
         return;
       }
-      res.locals.document = body;
+      res.locals.document = response.body;
       next();
     } catch (err) {
       // if there is err (eg: 404 status), return 400 error
--- a/src/api/feed-discovery/src/middleware.js
+++ b/src/api/feed-discovery/src/middleware.js
@@ -20,7 +20,7 @@ module.exports.checkValidBlog = function checkValidBlog() {
   return async (req, res, next) => {
     try {
       const response = await fetch(req.body.blogUrl);
-      const contentType = response.headers;
+      const contentType = response.headers.get('content-type');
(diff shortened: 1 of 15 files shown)


src/api/feed-discovery/src/middleware.js
```diff
@@ -20,14 +19,14 @@ module.exports.checkValidUrl = function checkValidUrl() {
 module.exports.checkValidBlog = function checkValidBlog() {
   return async (req, res, next) => {
     try {
-      const { statusCode, headers, body } = await got(req.body.blogUrl);
-      const contentType = headers['content-type'];
+      const response = await fetch(req.body.blogUrl);
+      const contentType = response.headers;
```
humphd, line 23:
```js
const contentType = response.headers.get('content-type);
```

src/api/image/bin/unsplash-download.js
```diff
@@ -18,7 +17,7 @@ const pipeline = promisify(stream.pipeline);
  * @param {string} filename - the local filename to use when writing
  */
 function downloadPhoto(url, filename) {
-  return pipeline(got.stream(url), fs.createWriteStream(filename)).then(() =>
+  return pipeline(fetch(url), fs.createWriteStream(filename)).then(() =>
```
humphd, line 20: I think this has to be the `body` of the response to get a stream, see https://github.com/node-fetch/node-fetch#streams

humphd (changes requested):
Nice work, a few things with the switch over from `got`.

src/api/image/bin/unsplash-download.js
HyperTHD: Thanks! Will do

src/api/feed-discovery/src/middleware.js
HyperTHD: Thanks!

src/api/image/bin/unsplash-download.js
```diff
@@ -16,8 +16,9 @@ const pipeline = promisify(stream.pipeline);
  * @param {string} url - the URL to the image
  * @param {string} filename - the local filename to use when writing
  */
-function downloadPhoto(url, filename) {
-  return pipeline(fetch(url), fs.createWriteStream(filename)).then(() =>
+async function downloadPhoto(url, filename) {
+  const response = await fetch(url);
```
humphd, line 20: I think `await` is incorrect here.  Node has a number of programming models: synchronous, asynchronous, and streams.  In this code, you are trying to create a pipeline out of two streams, where the `fetch` stream gets piped into the file write stream.  That way, as the bytes flow in from the `fetch`, they will get written to disk without having to accumulate in memory.

humphd (changes requested):
Looks like this is failing tests, so I suspect he switch from `got` needs more work.  Confirm we have the API usage correct with `node-fetch`.

manekenpix:
@HyperTHD what's the progress on this? can we include it in the 2.1 release?

HyperTHD:
> @HyperTHD what's the progress on this? can we include it in the 2.1 release?

ik 2.1 was just released but still gonna comment regardless

The main issue here is that tests need to be updated as well. Not to mention that `got` has uses here that aren't used with `node-fetch`. I'm gonna take another look through the files when I get some time and put up a commit that should hopefully get these tests to pass.

PR #2277, opened 2021-07-22
Investigating CI/CD frontend snapshots issue

PedroFonsecaDEV:
Just testing what is wrong with our CI/CD.


PedroFonsecaDEV:
The purpose of this PR was to investigate and test our CI/CD failures. Since the front-end snapshots pass on our local machines, I fired some codes against our CI/CD to check what was happening.  

As a result of the investigation, the following issue was open: 
#2278

PR #2308, opened 2021-09-17
Refactor autodeploy to use Satellite and a build queue

### LINKED ISSUE #2303: Dependabot pull requests cause development flow to crash
**What happened**:
`dev` crashed during deployment after merging a `dependabot` PR

`dev` should've deployed normally.

Merge a `dependabot` PR into master

Whenever a `dependabot` PR is merged, 2 webhook events are triggered, causing our deployment flow to crash.
image
Currently, our autodeployment server knows how to ignore certain type of events, but it doesn't know how to ignore 2 events in a row of the same type.
We should implement a way to ignore or queue webhook events.

humphd:
Fixes #2303 

This is an initial clean-up of the autodeployment code.  I've refactored it to use Satellite, so it matches the rest of our APIs.  I've also added some logic for dealing with multiple builds coming in one after the other.  I only keep track of the last build to finish, the current build, and the next one to be run.

I'm caching the GitHub data for the build/deploy so I can use it later in follow-up PRs.

I haven't fully tested this.  It might be good to try it on staging and see how it breaks, then I can fix.


tools/autodeployment/package.json
```diff
@@ -4,10 +4,9 @@
   "description": "A tool for automatic deployment",
   "main": "server.js",
   "dependencies": {
-    "@octokit/webhooks": "^9.1.0",
-    "date-fns": "^2.19.0",
+    "@octokit/webhooks": "^9.15.0",
```
manekenpix, line 7:
```suggestion
    "@octokit/webhooks": "^9.15.0",
    "@senecacdot/satellite": "^1.x",
```

humphd: Good catch

tools/autodeployment/builds.js
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  // Promote the next build to current and start it
+  const build = builds.pending;
+  builds.current = build;
+  builds.pending = null;
+
+  logger.debug('run() called, starting pending build');
+  build.proc = shell.exec(`./deploy.sh ${build.type}`, { silent: true }, (code) => {
+    build.finished(code);
```
manekenpix, line 75: This happens after a request was processed:
```sh
22:56:15 ✨ Server listening on port 4000
22:59:56 ✨ request completed POST 200 / 29ms
23:03:00 ✨ request completed POST 200 / 2ms
23:03:26 ✨ request completed GET 200 /log 6.4s
/usr/share/telescope_project/autodeployment/builds.js:75
    build.finished(code);
          ^

TypeError: build.finished is not a function
    at /usr/share/telescope_project/autodeployment/builds.js:75:11
    at /usr/share/telescope_project/autodeployment/node_modules/shelljs/src/exec.js:127:9
```

tools/autodeployment/credentials.js
```diff
@@ -0,0 +1,24 @@
+const { logger } = require('@senecacdot/satellite');
```
manekenpix, line 1:
```suggestion
const { logger } = require('@senecacdot/satellite');
const fs = require('fs');
const path = require('path');
```

manekenpix (changes requested):
I think it's very close to be ready.
I tested it on `dev` and I found a couple of issues.

tools/autodeployment/builds.js
humphd: Fixed

tools/autodeployment/credentials.js
humphd: Unbelievable...fixed!

manekenpix (approved):
@humphd I did some testing on staging using this PR and I think I found the problem with dependabot's PRs (actually, it's not dependabot's fault).
The issue is that dependabot executes 2 `push` actions whenever a PR is merged. One action pushes code to `master` and the other one to the branch that dependabot was using for the merged PR.
- Payloads of the webhook requests sent after merging a dependabot PR to master (notice `"ref"`)
image
image


Our autodeployment server doesn't filter webhook requests by branch, so pushing **to any branch** in the repo triggers a webhook `push` request, and our autodeployment server receives that request and initiates a deployment.
The fix added in this PR works, but I think we should also add a filter to only trigger a deployment when code is pushed to master. Maybe checking the value of `"ref"` in our [requestFilter](https://github.com/humphd/telescope/blob/0e536081cc56f387742db0a7d7f8d44c896229f0/tools/autodeployment/webhook.js#L36)?
We can file an issue to add the branch filter later or maybe it can be included in this PR. I'm ok with both.

humphd:
Based on your findings, I was about to add a simple check in this PR, but now I wonder if it's a bit more complicated.  Reading https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads, it looks like we have can have a few scenarios.  Depending on the `action` (which we don't seem to use?), we will get the following on the payload:

- `ref`: The git ref resource.  Could be a branch name (e.g., `refs/heads/master`) or a tag (`v1.0.2`)
- `ref_type`: The type of Git ref object created in the repository. Can be either branch or tag.
- `master_branch`: The name of the repository's default branch (usually `main` or `master`).

So I guess pseudo logic would be:

```js
if((ref_type === 'branch' && ref.endsWith(master_branch)) || ref_type === 'tag') {
  build()
}
```

But I suspect we should probably be using the `action` (created, published, etc) somehow too?

I'll file a follow-up.

humphd:
Filed https://github.com/Seneca-CDOT/telescope/issues/2312.

PR #2328, opened 2021-10-03
Run our front-end as a microservice

manekenpix:
## Type of Change

## Description

Currently, our front-end is built and served by our legacy back-end in development, and served by `nginx` in production.
This PR decouples the build and serving tasks from our legacy back-end, as another step in our transition to a full microservice architecture based web app. With this change, `nginx` is the one responsible for building and serving our front-end, in both `development` and `production`. The _manual_ way of building and serving the front-end will be kept for development and debugging purposes.

## Checklist


src/web/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+RUN apk add --no-cache tini util-linux
+ENTRYPOINT [ "/sbin/tini", "--"]
+# what is happening?
+WORKDIR /frontend
+
+COPY . .
+
+FROM base as dependencies
+
+COPY .git .git
```
humphd, line 26: I wanted to ask about this, but also see that we're failing CI due to this:

```
Step 14/32 : COPY .git .git
COPY failed: file not found in build context or excluded by .dockerignore: stat .git: file does not exist
```

@raygervais why did we include `.git`, was it for getting the version SHA in the frontend?  I wonder if we could be more clever about how we do this so we don't jam the entire `.git` into the build context.  NOTE: other than fixing this failure, we don't need to solve this in this PR, but reading it made me think.

humphd (changes requested):

src/web/Dockerfile
raygervais: That's correct, if memory serves me that was exactly the reason. Realistically, .git could be stored as a dedicated micro-service elsewhere which has it's own basic API for getting SHA, commit data etc. Would remove all the weight from frontend entirely (or, we write a script which produces a JSON that the frontend consumes.. but where's the fun in that?)

humphd: We should probably pass this as an arg to the build script, since we know it in the deployment server, and remove from docker.

manekenpix: > We should probably pass this as an arg to the build script, since we know it in the deployment server, and remove from docker.

I'm not really sure about how to do this. Would it make sense to extract the commit's sha from the webhook request's payload and pass it to docker via env var so it can be used while building telescope? something like:

```sh
GIT_COMMIT_SHA=98EEE45A830AA65F768EC194A091349A41113BD0 docker-compose -f config/production.yml...
```
Any ideas or suggestions are welcome.

humphd: We could just set it in the env early via command?

```sh
GIT_COMMIT=$(git rev-parse HEAD)
docker-compose -e GIT_COMMIT -f config/production.yml...
```

manekenpix: I think the front-end is fetching git info using the `/health` route (exactly [here](https://github.com/Seneca-CDOT/telescope/blob/65f148ad35727e3b43c74da70370e985e5727a1e/src/web/src/components/Banner.tsx#L149)), so I think not adding the `.git` folder to the build process should be enough.

src/web/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+# Front-end Web URL, set via ENV in docker or next build
+ARG WEB_URL
+
+FROM node:lts-alpine as base
+
+# Tini Entrypoint for Alpine
+# util-linux required to optimize builds using multiple cores
+RUN apk add --no-cache tini util-linux
+ENTRYPOINT [ "/sbin/tini", "--"]
+# what is happening?
```
humphd, line 19: Indeed...

humphd (approved):

src/web/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+ARG AUTH_URL
+ENV NEXT_PUBLIC_AUTH_URL ${AUTH_URL}
+
+ARG POSTS_URL
+ENV NEXT_PUBLIC_POSTS_URL ${POSTS_URL}
+
+ARG SEARCH_URL
+ENV NEXT_PUBLIC_SEARCH_URL ${SEARCH_URL}
+
+
```
humphd, line 42: Remove this extra blank line please.

src/web/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+ARG FEED_DISCOVERY_URL
+ENV NEXT_PUBLIC_FEED_DISCOVERY_URL ${FEED_DISCOVERY_URL}
+
+ARG WEB_URL
+ENV NEXT_PUBLIC_WEB_URL ${WEB_URL}
+
+RUN npm install --no-package-lock
```
humphd, line 49: Should we add `--production` to this?  I don't think we need or want dev dependencies in the container.

docker/docker-compose.yml
```diff
@@ -225,6 +225,35 @@ services:
...
+    container_name: 'nginx'
+    environment:
+      - TELESCOPE_HOST
+      - API_HOST
+    restart: unless-stopped
+    depends_on:
+      - traefik
+    ports:
+      # The HTTP port
+      - '80:80'
```
humphd, line 255: Will we run into any issues binding to port 80 on development machines?

humphd (changes requested):
This looks really good.  A few questions and nits.

PR #2344, opened 2021-10-08
Fix #1868: Use ADMINISTRATORS env var in auth service

### LINKED ISSUE #1868: [auth] Improve how we define admin accounts
In #1796, we are defining admin accounts via an env variable with `ADMINISTRATORS` being passed in.  This is probably OK for one super-admin, but the rest can be defined in Firebase with the #1642 (cc @chrispinkney).

humphd:
Fixes #1868.  This adds support for using the Telescope super user accounts defined in the `ADMINISTRATORS` env variable.  We did this previously with the old backend, but never ported it over to the auth service.  I haven't tested this fully yet, so I'll do it as a draft until I know tests are OK.

### DIFF
--- a/src/api/auth/src/roles.js
+++ b/src/api/auth/src/roles.js
@@ -2,3 +2,4 @@
 module.exports.seneca = () => ['seneca'];
 module.exports.telescope = () => ['seneca', 'telescope'];
 module.exports.admin = () => ['seneca', 'telescope', 'admin'];
+module.exports.superUser = () => ['seneca', 'admin'];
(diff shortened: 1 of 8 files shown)


humphd:
OK, this is working now.  Here's what I had to do:

- Create the concept of a new role for a super user, who is someone that has a Seneca SSO login and is defined in the `ADMINISTRATORS` env variable.  They may or may not have a Telescope account yet.
- When I set the `role` for a User, I deal with someone being a Telescope user/admin vs. a Seneca user/admin
- I've added tests for everything
- I refactored the old tests to move the check for users in the Users service to a util function, `ensureUsers`
- I made a few tests more clear with respect to the SSO SAML key Seneca uses vs. email

Tests are all passing.

humphd:
@Grommers00 I'm going to land this, and if you want to look at it later and file an issue if you have concerns, do that instead of me waiting on a review.

PR #2410, opened 2021-10-29
Closes Issue #2374: Include more details about issues and pull requests

### LINKED ISSUE #2374: Include more details about Issue and Pull Requests in title attribute
In #2367 we're adding GitHub info do a post.  We show Issues and Pull Requests discussed in a post as numbers only.  This is good for the space we have available, but when we hover over them, it would be good to add more detail.

For example, if a post talks about 2 or more repos, and then shows Issues and Pull Requests, which repo do they go with?  Instead of using `Issue #...` in the `title` attribute on hover, let's also include the repo details (user/org + repo name): `user/repo Issue#...`.  Same for Pull Requests and [Commits](https://github.com/Seneca-CDOT/telescope/issues/2372).

HyperTHD:
## Issue This PR Addresses:
This PR closes #2374 

## Type of Change

## Description

This PR adds additional information to indicate to the user where an issue or pull request comes from. Sentences are unoriginal, but the user will now know what repo an issue or PR comes from when hovering over the issue/PR number. 
This will also include the user as well

## Checklist

### DIFF
--- a/src/web/src/components/Posts/Commits.tsx
+++ b/src/web/src/components/Posts/Commits.tsx
@@ -41,6 +41,11 @@ const SHORT_SHA_LENGTH = 7;
 const getCommitNumber = (url: string, length?: number) =>
   url.replace(/.+\/(commit|commits)\/(\w{40}).*/, '$2').substr(0, length);
 
+const getCommitInfo = (commit: string) => {
+  const [, user, repo] = commit.split('/');
+  return `${user}/${repo}`;
+};
+
 type Props = {
   commitUrls: string[];
 };
@@ -61,7 +66,7 @@ const Commits = ({ commitUrls }: Props) => {
               href={`https://github.com${url}`}
               rel="bookmark"
               target="_blank"
-              title={'Commit ' + getCommitNumber(url)}
+              title={`${getCommitInfo(url)} Commit Hash ${getCommitNumber(url)}`}
               className={classes.link}
             >
               {getCommitNumber(url, SHORT_SHA_LENGTH)}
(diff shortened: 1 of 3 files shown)


humphd (dismissed):
@HyperTHD nice.  Do you want to merge this now, or add the same for Commits in this PR?  Happy either way, but let's file an issue for follow-up if we don't do it here.

HyperTHD:
> @HyperTHD nice. Do you want to merge this now, or add the same for Commits in this PR? Happy either way, but let's file an issue for follow-up if we don't do it here.

It's a simple add-on. I'll rebase and quickly add that in for the commits. Once that's done, you can merge this in

HyperTHD:
> That's an elegant way to extract the info, wonder if I can do the same for filtering urls.

I've made a hook with filtering search results before so it's definitely possible. Also, I find regex to be very confusing at times so if I can ever avoid using it, it's worth it. I only use it if absolutely necessary

Andrewnt219:
@HyperTHD if it works with all the cases in the comments, then let file an issue to change it. Probably have to write a test suite for that as well.

src/web/src/components/Posts/Commits.tsx
```diff
@@ -61,7 +66,7 @@ const Commits = ({ commitUrls }: Props) => {
               href={`https://github.com${url}`}
               rel="bookmark"
               target="_blank"
-              title={'Commit ' + getCommitNumber(url)}
+              title={`${getCommitInfo(url)} Commit Hash ${getCommitNumber(url)}`}
```
humphd, line 69: Let's drop the `Hash `, the rest is great.

humphd (changes requested):
Tested on staging + Vercel, works great!  One small fix and this is R+ from me.

src/web/src/components/Posts/Commits.tsx
HyperTHD: Ready to merge!

humphd (approved):

humphd:
@HyperTHD I'm fine for this to go in as is.  We're low on reviewers atm.

PedroFonsecaDEV (approved):
LGTM!

Issue #2345, opened 2021-10-09
Bull-board has been deprecated, update to use @bull-board/* packages in feed-parser microservice


manekenpix:
**What would you like to be added**:
Apply the changes added to our legacy back-end in #2341 to the `feed-parser` microservice.

**Why would you like this to be added**:
Telescope is in transition to become a fully microservice based web app, and this change is required since we'll eventually remove telescope's legacy back-end.

humphd:
I wonder if we should remove bull board completely.  We don't expose it.

PR #2441, opened 2021-11-03
Disable parser service until we fix #2111

### LINKED ISSUE #2111: Write tests and enable Parser service
We landed the Parser service today, but we have it not starting the actual parser, nor do we have any tests. We need to fix both things.  @c3ho added Jest configs for both unit and e2e tests, but there aren't any.

You can get some idea of how to work across services in tests by looking at my auth test [here](https://github.com/Seneca-CDOT/telescope/blob/master/src/api/auth/test/e2e/auth-flows.test.js).

humphd:
The parser service is crashing on staging due to #2345.  It's actually not really ready for use yet (#2111), so we're going to disable it until it's ready.  Once we finish the code and tests, we can revert this.

### DIFF
--- a/docker/development.yml
+++ b/docker/development.yml
@@ -93,11 +93,12 @@ services:
     depends_on:
       - test-web-content
 
-  parser:
-    environment:
-      # For development and testing, the Parser service needs to contact the users
-      # service directly via Docker vs through the http://localhost domain.
-      - USERS_URL=http://users:7000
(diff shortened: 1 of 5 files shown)


manekenpix (approved):

humphd:
Once this clears CI, I'll merge.

humphd:
Merging.

@DukeManh FYI, this touches the gitpod stuff, so you'll have to remove the `parser` service from your list, and make sure the lines I'm commenting out are also done in yours.

PR #2471, opened 2021-11-15
Fixes #2419: show feeds count on status dashboard

### LINKED ISSUE #2419: Show Posts count on status dashboard
In #2395 we are adding an expanded status service dashboard. One thing I want to include in it is the ability to see the current number of Posts and Feeds in the database.  We expose this data as a header when you query for `/posts` or `/feeds` from the Posts service:

https://github.com/Seneca-CDOT/telescope/blob/master/src/api/posts/src/routes/posts.js#L55

Let's add a widget to the dashboard that shows the current counts for each of these.

tuanthanh2067:
Fix #2419

## Type of Change

## Description

Add feeds count for status dashboard. I ran docker and used `localhost/v1/posts/feeds` to test it out. And I also have to add cors to expose headers in posts service
## Checklist

<a href="https://gitpod.io/#https://github.com/Seneca-CDOT/telescope/pull/2471"></a>

### DIFF
--- a/src/api/status/public/index.html
+++ b/src/api/status/public/index.html
@@ -16,6 +16,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
+    <base href="/v1/status/" />
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no" />
     <link rel="apple-touch-icon" sizes="76x76" href="assets/img/apple-icon.png" />
     <link rel="icon" type="image/png" href="assets/img/favicon.png" />
@@ -1277,7 +1278,6 @@
         </div>
       </div>
     </div>
-    <script src="./script.js"></script>
     <!--   Core JS Files   -->
     <script src="assets/js/core/popper.min.js"></script>
     <script src="assets/js/core/bootstrap.min.js"></script>
@@ -1288,7 +1288,12 @@
     <script src="assets/js/demo-sidenav.js"></script>
     <!-- Control Center for Material Dashboard: parallax effects, scripts for the example pages etc -->
     <script src="assets/js/material-dashboard.min.js?v=3.0.0"></script>
+    <<<<<<< HEAD
     <!-- Telescope JS Files -->
     <script src="js/github-stats.js"></script>
+    =======
+
+    <script src="./script.js"></script>
+    >>>>>>> 478f347 (made changes to multiple lines)
   </body>
 </html>
--- a/src/api/status/public/index.html
+++ b/src/api/status/public/index.html
@@ -16,7 +16,6 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
-    <base href="/v1/status/" />
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no" />
     <link rel="apple-touch-icon" sizes="76x76" href="assets/img/apple-icon.png" />
     <link rel="icon" type="image/png" href="assets/img/favicon.png" />
@@ -473,6 +472,22 @@
                 </div>
               </div>
             </div>
+          </div>
+        </div>
+        <div class="col-xl-3 col-sm-6 mb-xl-0 mb-4 mt-4">
+          <div class="card">
+            <div class="card-header p-3 pt-2">
+              <div
+                class="icon icon-lg icon-shape bg-gradient-success shadow-success text-center border-radius-xl mt-n4 position-absolute"
+              >
+                <i class="material-icons opacity-10">person</i>
+              </div>
+              <div class="text-end pt-1">
+                <p class="text-sm mb-0 text-capitalize">Total feeds</p>
+                <h4 class="mb-0" id="feeds-count"></h4>
+              </div>
+            </div>
+            <div class="card-footer"></div>
           </div>
         </div>
         <div class="row mt-4">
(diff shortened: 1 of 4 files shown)


src/api/status/env.local
```diff
@@ -1 +1,2 @@
 PLANET_PORT=1111
+API_HOST=dev.api.telescope.cdot.systems
```
humphd, line 2: Let's not do this change

src/api/status/public/index.html
```diff
@@ -356,8 +355,8 @@ <h4 class="mb-0">$53k</h4>
                   <i class="material-icons opacity-10">person</i>
                 </div>
                 <div class="text-end pt-1">
-                  <p class="text-sm mb-0 text-capitalize">Today's Users</p>
-                  <h4 class="mb-0">2,300</h4>
+                  <p class="text-sm mb-0 text-capitalize">Feeds Count</p>
+                  <h4 class="mb-0" id="feeds-count">2,300</h4>
```
humphd, line 359: Don't hard-code 2,300

humphd (changes requested):

src/api/status/public/index.html
```diff
@@ -356,7 +357,7 @@ <h4 class="mb-0">$53k</h4>
                 </div>
                 <div class="text-end pt-1">
                   <p class="text-sm mb-0 text-capitalize">Feeds Count</p>
-                  <h4 class="mb-0" id="feeds-count">2,300</h4>
+                  <h4 class="mb-0" id="feeds-count">0</h4>
```
humphd, line 360: Just leave it empty

src/api/status/public/index.html
```diff
@@ -1224,5 +1224,7 @@ <h6 class="mt-3">Thank you for sharing!</h6>
     <script src="assets/js/demo-sidenav.js"></script>
     <!-- Control Center for Material Dashboard: parallax effects, scripts for the example pages etc -->
     <script src="assets/js/material-dashboard.min.js?v=3.0.0"></script>
+
```
humphd, line 1227: Add a comment here `<!-- Telescope Dashboard -->`

src/api/status/public/index.html
```diff
@@ -1224,5 +1224,7 @@ <h6 class="mt-3">Thank you for sharing!</h6>
...
+    <script src="./script.js"></script>
```
humphd, line 1228: script.js is too generic.  Also, this should probably go in some kind of src dir.  @Andrewnt219, what's your suggestion, based on the build log work?

src/api/status/public/script.js
```diff
@@ -60,6 +64,5 @@ async function setCount() {
 
 window.onload = function () {
   checkStatus();
+  setCount();
```
humphd, line 67: Call `getFeedCount()` here, and have it call `setCount()`

humphd (changes requested):

src/api/status/public/index.html
Andrewnt219: it can go in `/public/js/feed/index.js`. I also suggest making an api.js file to export all the `fetch` functions partly because it's hard to remember all the different hosts.

```js
// /public/js/api.js 
// ...or /public/js/api/feed.js

const fetchFeeds = () => fetch(`https://${process.env.API_HOST}/v1/posts/feeds`);
const fetchStatus = () => fetch('/v1/status/status');

```

src/api/status/public/js/api.js
```diff
@@ -0,0 +1 @@
+export const fetchFeeds = () => fetch(`https://${process.env.API_HOST}/v1/posts/feeds`);
```
humphd, line 1: We don't have access to `process.env.API_HOST` in the Dashboard front-end

humphd (changes requested):

tuanthanh2067:
If everything seems ok, I can squash it and then commit.

src/api/status/public/index.html
humphd: This still needs to happen.  `script.js` isn't going to work.  Please do what @Andrewnt219 suggests.

src/api/status/public/js/feed/index.js
```diff
@@ -0,0 +1,12 @@
+const { fetchFeeds } = require('../api');
```
humphd, line 1: We can't use `require` on the web.  Use `import` and es6.

src/api/status/public/js/feed/index.js
```diff
@@ -0,0 +1,12 @@
...
+
+export async function getFeedCount() {
+  try {
+    const data = await fetchFeeds();
```
humphd, line 5: Because all you care about are the headers, use `fetch(url, {method: 'HEAD'})` to reduce how much gets sent back.

src/api/status/public/script.js
```diff
@@ -1,3 +1,5 @@
+const { getFeedCount } = require('./js/feed/index');
```
humphd, line 1: Same problem here, we don't have `require` in the browser.

humphd (changes requested):
It doesn't seem like you've tested this.  Can you make the fixes, get it running locally or in GitPod, and then post a screenshot?

tuanthanh2067:
This is what I have when running it locally

image

humphd:
Tested on GitPod (yeah @DukeManh!) and I think this is working:

Screen Shot 2021-11-18 at 8 59 41 PM

If we run it on the real server, there will be data, or should be.

humphd (approved):

DukeManh:
@humphd we have to change the status port from `1111` to something `>1024` to run the server on Gitpod.

I'm thinking if we should change all ports to be `>1024` so it works both locally and on Gitpod without subsisting.

DukeManh (approved):

humphd:
> @humphd we have to change the status port from `1111` to something `>1024` to run the server on Gitpod.

I'm confused, 1111 is > 1024 already.

DukeManh:
😀. I mean port 80 so people can test services with `npm run services:start <service>`.

humphd:
File an issue if you want to have some follow-up happen.

tuanthanh2067:
I think I have to move from using `window.onload` to `addEventListener` as mentioned [here](https://github.com/Seneca-CDOT/telescope/pull/2476#discussion_r753362604) @humphd @Andrewnt219

tuanthanh2067:
But still keep it in one file as I don't think bringing it to `script.js` is a good idea.

Issue #2373, opened 2021-10-20
Mobile UX/UI for GitHub post info


humphd:
In #2367 we are adding GitHub info to the sidebar for a post.  Currently, it only works in desktop, and there's nothing visible in mobile.  I'd like to see if we can add a design and UI for the mobile case.

One option might be to show a collapsed GitHub area that expands when you touch it.  If there's no GitHub details, we wouldn't show it.  But if a post does have any, we'd include a GitHub icon or something that you can use to drill down and see more.

Obviously we don't want to show this all the time, but it would be good if you could access it on demand.

irenejoeunpark:
Hi @humphd, If no one has started working on this, can I start contributing to the UI implementation for mobiles?

humphd:
@irenejoeunpark that would be great, please do.  I would encourage you to chat with people in the #telescope channel on Slack, as many of the devs/users will have opinions about this, so getting community input early will be key.  Also, we can do this in stages (e.g., you don't have to solve the whole thing in one PR).

Looking forward to seeing this happen!

sirinoks:
We have just completed changing a column view of isses/PRs to a row and fixed its wrapping. Await for the PR for that one. @TueeNguyen should make the PR.

There is another task to impove the view, which turned out to be more difficult than we first thought. I will create another issue describing it.

PR #2491, opened 2021-11-18
#2373 Update on github information on mobile view, need update on the titles

irenejoeunpark:
## Issue This PR Addresses

#2373 More change is needed after some feedbacks on styles

## Type of Change

## Description

This PR contains an update on the UI for mobile view, to include GitHub information on each post. 
When the user clicks on the GitHub icon, it expands the GitHub information of the post. 

Need feedback on the icon, design, and the idea of using Accordion.
Need to fix the title and the author name position on mobile view - need some advice on it. 

Screen Shot 2021-11-18 at 3 20 35 PM
Screen Shot 2021-11-18 at 3 20 43 PM

## Checklist

- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not -> Not a functional change, no need of functional tests to be added. 
- [ ] **Documentation**: This PR includes updated/added documentation to user exposed functionality or configuration variables are added/changed or an explanation of why it does not(if applicable) -> UI change, not applicable

<a href="https://gitpod.io/#https://github.com/Seneca-CDOT/telescope/pull/2491"></a>


src/web/src/components/Posts/Commits.tsx
```diff
@@ -14,6 +14,10 @@ const useStyles = makeStyles((theme: Theme) =>
       lineHeight: 'normal',
       fontSize: '1.2rem',
     },
+    GitHubMobile: {
```
DukeManh, line 17: This class is not used anywhere right?

humphd:
I think the idea is good.  The details need some work, but having it expand vertically to reveal the GitHub info is interesting.  We need to find a way to use the space more efficiently on mobile (not jammed to the right).  Also, I'm not sure where the GitHub icon should go, or what it should look like.  It might need some text to indicate there is `More...` somehow.

I would focus on figuring out how to arrange this same info in a smaller space efficiently on mobile (e.g., what's in the revealed area).

src/web/src/components/Posts/GitHubMobile.tsx
```diff
@@ -0,0 +1,156 @@
...
+    },
+    GitHubInfoContainer: {
+      margin: '2rem 0 0 1rem',
+    },
+    icon: {
+      fontSize: '4rem',
+      marginRight: '1rem',
+      verticalAlign: 'text-bottom',
+      paddingLeft: '6rem',
+      alignItems: 'right',
```
DukeManh, line 38: Did you mean `textAlign`? This property is invalid.

src/web/src/components/Posts/GitHubMobile.tsx
```diff
@@ -0,0 +1,156 @@
...
+      },
+      color: theme.palette.text.secondary,
+    },
+    GitHubInfoContainer: {
+      margin: '2rem 0 0 1rem',
+    },
+    icon: {
+      fontSize: '4rem',
+      marginRight: '1rem',
+      verticalAlign: 'text-bottom',
```
DukeManh, line 36: This property is invalid because `icon` is not an inline item

src/web/src/components/Posts/GitHubMobile.tsx
```diff
@@ -0,0 +1,156 @@
...
+      paddingLeft: '6rem',
+      alignItems: 'right',
+    },
+    accordionSummary: {
+      width: '10 rem',
```
DukeManh, line 41: You have extra space between '10' and 'rem' here. It's not valid.

src/web/src/components/Posts/GitHubMobile.tsx
```diff
@@ -0,0 +1,156 @@
...
+      paddingRight: '0',
+      justifyContent: 'flex-end',
+    },
+    accordion: {
+      backgroundColor: 'rgba(255, 255, 255, 0)',
```
DukeManh, line 46: You can have it inherit main bg color: `backgroundColor: inherit`.

src/web/src/components/Posts/GitHubMobile.tsx
```diff
@@ -0,0 +1,156 @@
...
+      border: 'none',
+      boxShadow: 'none',
+      justifyContent: 'flex-end',
```
DukeManh, line 49: This is not a flexbox so this won't apply.

src/web/src/components/Posts/GitHubMobile.tsx
```diff
@@ -0,0 +1,156 @@
...
+    },
+    '&:before': {
```
DukeManh, line 51: This doesn't seem to belong to any class. Also, it should be `&::before` with two colons.

src/web/src/components/Posts/GitHubMobile.tsx
```diff
@@ -0,0 +1,156 @@
...
+    },
+  })
+);
+
+type Props = {
+  ghUrls: string[];
+};
+
+const filterGitHubUrls = (urls: string[]) => {
+  const issues: Set<string> = new Set();
```
DukeManh, line 62: We should extract the logic for GitHub information and reuse them in both mobile and desktop components vs copying it.
Another way is to make `GithubInfo` component render the info differently using [useMediaQuery](https://mui.com/components/use-media-query/).

src/web/src/components/Posts/Issues.tsx
```diff
@@ -14,6 +14,10 @@ const useStyles = makeStyles((theme: Theme) =>
       lineHeight: 'normal',
       fontSize: '1.2rem',
     },
+    GitHubMobile: {
```
DukeManh, line 17: You might forget to remove this class too.

src/web/src/components/Posts/PullRequests.tsx
```diff
@@ -14,6 +14,10 @@ const useStyles = makeStyles((theme: Theme) =>
       lineHeight: 'normal',
       fontSize: '1.2rem',
     },
+    GitHubMobile: {
```
DukeManh, line 17: Same here

DukeManh (changes requested):
This is awesome.

I'm just thinking about whether we should make display the github info at the bottom of each post by default vs make users click a button and see a layout shift.

irenejoeunpark:
> I think the idea is good. The details need some work, but having it expand vertically to reveal the GitHub info is interesting. We need to find a way to use the space more efficiently on mobile (not jammed to the right). Also, I'm not sure where the GitHub icon should go, or what it should look like. It might need some text to indicate there is `More...` somehow.
> 
> I would focus on figuring out how to arrange this same info in a smaller space efficiently on mobile (e.g., what's in the revealed area).

I found another method for collapsable, called Colllapse in material-UI, here is an example of the usage.
https://codesandbox.io/s/77r47?file=/src/App.js:0-41
Should I use this to collapse the info horizontally? 
I am also thinking that no one would notice that the GitHub icon is for expanding information, and people would rarely click it. 
So, I am changing it to ```...``` which seems more reasonable and could lead more people to click it.

src/web/src/components/Posts/Issues.tsx
irenejoeunpark: Resolved!

src/web/src/components/Posts/GitHubMobile.tsx
irenejoeunpark: Fixed! Thanks :)

src/web/src/components/Posts/GitHubMobile.tsx
irenejoeunpark: Fixed. Thanks!

src/web/src/components/Posts/GitHubMobile.tsx
irenejoeunpark: Fixed. Thanks!

src/web/src/components/Posts/GitHubMobile.tsx
irenejoeunpark: Thanks for the recommendation, I learned a new thing. This is fixed, thanks!

src/web/src/components/Posts/GitHubMobile.tsx
irenejoeunpark: Removed &:before since this would affect nowhere. Thanks!

src/web/src/components/Posts/GitHubMobile.tsx
irenejoeunpark: Removed. Thanks

src/web/src/components/Posts/PullRequests.tsx
irenejoeunpark: Fixed as well!

src/web/src/components/Posts/Commits.tsx
irenejoeunpark: I removed it. Thanks for the review!

src/web/src/components/Posts/GitHubMobile.tsx
irenejoeunpark: I am working on this, when I get all the required information from GitHubInfo, would use ```import GitHubInfo from './GitHubInfo';``` be enough for it or any other additional step is needed?

irenejoeunpark: This has been updated. I removed those and reused them from GitHubInfo. Thanks

humphd (changes requested):
I think this needs more UX work.  Here's the comparison of what we have now (left) with what this does (right):

Screen Shot 2021-11-23 at 12 38 59 PM

I think adding another `...` clutters the view.  Here's my suggestion instead:

- Make the entire header (avatar logo, title, name, etc) a single clickable area
- If there is GitHub data to show, add a small GitHub icon to the right of the person's name, which would indicate that there is GitHub data to reveal
- Touching the header would open the header to a larger size

The current layout of the GitHub data isn't going to work on mobile:

Screen Shot 2021-11-23 at 12 40 59 PM

We need to use the entire width somehow.  cc @Meneguini, @PedroFonsecaDEV for ideas.

irenejoeunpark:
> I think this needs more UX work. Here's the comparison of what we have now (left) with what this does (right):
> 
> Screen Shot 2021-11-23 at 12 38 59 PM
> 
> I think adding another `...` clutters the view. Here's my suggestion instead:
> 
> * Make the entire header (avatar logo, title, name, etc) a single clickable area
> * If there is GitHub data to show, add a small GitHub icon to the right of the person's name, which would indicate that there is GitHub data to reveal
> * Touching the header would open the header to a larger size
> 
> The current layout of the GitHub data isn't going to work on mobile:
> 
> Screen Shot 2021-11-23 at 12 40 59 PM
> 
> We need to use the entire width somehow. cc @Meneguini, @PedroFonsecaDEV for ideas.

It seems like a good design. I will try to make the whole title container a clickable area and add a GitHub icon if there's any GitHub link attached.

irenejoeunpark:
> I think this needs more UX work. Here's the comparison of what we have now (left) with what this does (right):
> 
> Screen Shot 2021-11-23 at 12 38 59 PM
> 
> I think adding another `...` clutters the view. Here's my suggestion instead:
> 
> * Make the entire header (avatar logo, title, name, etc) a single clickable area
> * If there is GitHub data to show, add a small GitHub icon to the right of the person's name, which would indicate that there is GitHub data to reveal
> * Touching the header would open the header to a larger size
> 
> The current layout of the GitHub data isn't going to work on mobile:
> 
> Screen Shot 2021-11-23 at 12 40 59 PM
> 
> We need to use the entire width somehow. cc @Meneguini, @PedroFonsecaDEV for ideas.

@humphd I have revised the code to make the title container clickable, and show the GitHub icon beside the author name when there is any GitHub info available. 

Now I faced another challenge, which the title container, which has been pinned to the top before I make the change, is now scrolled down with the post content. I tried a few different things, but it never gets pinned on top even though I only updated it to put the whole thing in the AccordionSummary. <- might be related to the issue? 
I will ask for other reviews in the slack.

AmasiaNalbandian (changes requested):

AmasiaNalbandian:
Irene, when I click on the container to open the accordion, I believe right now that same title is a link. I can't open the containers without being navigated to the blog site. Maybe we should not use a link there?

src/web/src/components/Posts/Post.tsx
```diff
@@ -192,6 +196,31 @@ const useStyles = makeStyles((theme: Theme) =>
         width: 'auto',
       },
     },
+    accordionSummary: {
+      paddingRight: '0',
+      justifyContent: 'flex-end',
+      overflow: 'scroll',
+      textOverflow: 'ellipsis',
+    },
+    accordion: {
```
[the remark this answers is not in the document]
irenejoeunpark: Thanks for the suggestion! I just tried it and it works perfectly! New knowledge gained 😉

irenejoeunpark:
> Irene, when I click on the container to open the accordion, I believe right now that same title is a link. I can't open the containers without being navigated to the blog site. Maybe we should not use a link there?

Hi @AmasiaNalbandian  It should be only available when your screen size is below 1025 and below in width.

irenejoeunpark:
> I think this needs more UX work. Here's the comparison of what we have now (left) with what this does (right):
> 
> Screen Shot 2021-11-23 at 12 38 59 PM
> 
> I think adding another `...` clutters the view. Here's my suggestion instead:
> 
> * Make the entire header (avatar logo, title, name, etc) a single clickable area
> * If there is GitHub data to show, add a small GitHub icon to the right of the person's name, which would indicate that there is GitHub data to reveal
> * Touching the header would open the header to a larger size
> 
> The current layout of the GitHub data isn't going to work on mobile:
> 
> Screen Shot 2021-11-23 at 12 40 59 PM
> 
> We need to use the entire width somehow. cc @Meneguini, @PedroFonsecaDEV for ideas.

@humphd I fixed the title container to 

> > I think this needs more UX work. Here's the comparison of what we have now (left) with what this does (right):
> > Screen Shot 2021-11-23 at 12 38 59 PM
> > I think adding another `...` clutters the view. Here's my suggestion instead:
> > 
> > * Make the entire header (avatar logo, title, name, etc) a single clickable area
> > * If there is GitHub data to show, add a small GitHub icon to the right of the person's name, which would indicate that there is GitHub data to reveal
> > * Touching the header would open the header to a larger size
> > 
> > The current layout of the GitHub data isn't going to work on mobile:
> > Screen Shot 2021-11-23 at 12 40 59 PM
> > We need to use the entire width somehow. cc @Meneguini, @PedroFonsecaDEV for ideas.
> 
> @humphd I have revised the code to make the title container clickable, and show the GitHub icon beside the author name when there is any GitHub info available.
> 
> Now I faced another challenge, which the title container, which has been pinned to the top before I make the change, is now scrolled down with the post content. I tried a few different things, but it never gets pinned on top even though I only updated it to put the whole thing in the AccordionSummary. <- might be related to the issue? I will ask for other reviews in the slack.

Hi, @humphd I fixed the title container to be pinned to the top of the screen. And now it looks like below. 
Screen Shot 2021-11-23 at 11 09 57 PM

And below image is when the user click to expand. 
Screen Shot 2021-11-23 at 11 10 06 PM

I moved the information aligned left for a better look. 

Any feedback or suggestions are welcomed!

HyperTHD (changes requested):
Judging based on latest code, there's some stuff that I think we can fix

- There's no indication that the user has to click on the space just below the title to view the github info. You could probably try and see if you can fit like an arrow svg that can be used in the empty space allocated. Probably would want to make it work similar to a hamburger nav icon.

- You're gonna need some margin at the bottom, too much white-space at the end before the title. Also, there's some whitespace above the github info when it's opened, that's also needs to be fixed too.
Example post

I think the actual github info stuff looks fine. The github user icon is a bit small, but otherwise, the rest looks fine.

humphd (changes requested):
This is getting better.  A few things:

Let's reduce the amount of whitespace at the bottom when the accordion is collapsed (compare this to what's happening on staging/production):

Screen Shot 2021-11-26 at 9 19 21 AM

Let's remove the line **in the mobile view** on the left, which isn't needed, since it's purpose is to divide this from the content in desktop view.  Also, let's get rid of the excessive whitespace before the GitHub info starts (again, only on mobile).

Screen Shot 2021-11-26 at 9 17 40 AM

irenejoeunpark:
> This is getting better. A few things:
> 
> Let's reduce the amount of whitespace at the bottom when the accordion is collapsed (compare this to what's happening on staging/production):
> 
> Screen Shot 2021-11-26 at 9 19 21 AM
> 
> Let's remove the line **in the mobile view** on the left, which isn't needed, since it's purpose is to divide this from the content in desktop view. Also, let's get rid of the excessive whitespace before the GitHub info starts (again, only on mobile).
> 
> Screen Shot 2021-11-26 at 9 17 40 AM

Hi @humphd, I have updated the mentioned improvements. 
Below are the images after the update
Screen Shot 2021-11-26 at 4 39 16 PM
Screen Shot 2021-11-26 at 4 39 30 PM

HyperTHD:
@irenejoeunpark There's still a bit of whitespace at the bottom before the first sentence of the blog post that'll need to be fixed.
image

Also, we really need a way for users to be able to know that clicking at a certain spot brings over the github info in mobile view. 
Maybe an icon? Would like to hear from everyone else about this
image

humphd (changes requested):
This is really getting close, nice work.

One issue I'm hitting is that it's adding overflow and scrolling in the header:


https://user-images.githubusercontent.com/427398/143775243-0fb8371d-61cf-4710-b4fa-652fa56c96e4.mov

I also want to rework the layout of the components in the accordion, but we can do that in a followup.

humphd:
This seems to have stalled.  @irenejoeunpark can you update us and fix the merge conflicts and review issues so that we can move forward?

irenejoeunpark:
> This seems to have stalled. @irenejoeunpark can you update us and fix the merge conflicts and review issues so that we can move forward?

I am working on it! I am actually stuck on detecting which one causes that scrolling thing. I know it would be with height, margin, or padding, but even if I change every single one of them it still does that.

irenejoeunpark:
Hi @humphd, I have tested the mobile view in [https://dev.telescope.cdot.systems/], and the issue seems like it is not only happening in this branch. 
Filed a separate issue #2583 for the fixes of the bug.

humphd:
Excellent, thank you.  Can you rebase to pick up the changes to `src/web/src/components/Posts/*`?

humphd:
This isn't perfect, but it's better.  The mechanism of showing/hiding the info is working well.  I'm not sure about the font size change on the header text.  And the layout of the inner contents are still not right for mobile.

I'll be interested to see further improvements and hear more from other reviewers.

src/web/src/components/Posts/Users.tsx
```diff
@@ -15,6 +15,10 @@ const useStyles = makeStyles((theme: Theme) =>
       lineHeight: 'normal',
       fontSize: '1.2rem',
     },
+    GitHubMobile: {
```
DukeManh, line 18: This class isn't needed.

src/web/src/components/Posts/GItHubInfoMobile.tsx
```diff
@@ -0,0 +1,51 @@
...
+  })
+);
+
+const GitHubInfoMobile = ({ ghUrls }: Props) => {
+  const classes = useStyles();
+  const { repos, issues, pullRequests, commits, users } = filterGitHubUrls(ghUrls);
+
+  return (
+    <div>
+      <ListSubheader className={classes.root}>
```
DukeManh, line 38: Can you make this `<ListSubheader component="div">` and the same to every `ListSubheader` you add? We have a number of wrong uses of `ListSubheader` throughout the tree because we like its default style. By default it renders a `<li>`, `li` is meant for lists, we don't want to end up having nested `<li>`. I have fixed some `<ListSubheader component="div">` in [#2594](https://github.com/Seneca-CDOT/telescope/pull/2594) but to avoid this in the future I'll file an issue to replace all usage of `ListSubheader` with a different component while maintaining the same style.

src/web/src/components/Posts/Post.tsx
```diff
@@ -311,62 +353,90 @@ const PostComponent = ({ postUrl, currentPost, totalPosts }: Props) => {
...
-                </a>
-              </h1>
-            </div>
-            <div className={classes.publishedDateContainer}>
-              <h1 className={classes.published}>
-                <a href={post.url} rel="bookmark" className={classes.link}>
-                  <time dateTime={post.updated}>{`${formatPublishedDate(post.updated)}`}</time>
-                </a>
+      {!desktop && (
+        <>
```
DukeManh, line 357: This fragment isn't needed.

src/web/src/components/Posts/Post.tsx
```diff
@@ -311,62 +353,90 @@ const PostComponent = ({ postUrl, currentPost, totalPosts }: Props) => {
...
+                </div>
+              </ListSubheader>
+            </AccordionSummary>
+            <AccordionDetails>
+              {!!extractedGitHubUrls.length && <GitHubInfoMobile ghUrls={extractedGitHubUrls} />}
+            </AccordionDetails>
+          </Accordion>
+        </>
+      )}
       {desktop && (
```
DukeManh, line 409: It'll be nice if y can combine these two expression into something like `desktop ? <A/> : <B/>`

src/web/src/components/Posts/Repos.tsx
```diff
@@ -15,6 +15,11 @@ const useStyles = makeStyles((theme: Theme) =>
       wordWrap: 'break-word',
       lineHeight: 'normal',
     },
+    GitHubMobile: {
```
DukeManh, line 18: This class isn't needed.

DukeManh (changes requested):
This is getting closer. I left some reviews.

src/web/src/components/Posts/GItHubInfoMobile.tsx
```diff
@@ -0,0 +1,51 @@
...
+    },
+    GitHubInfoContainer: {
+      margin: '0 0 0 1rem',
+    },
+  })
+);
+
+const GitHubInfoMobile = ({ ghUrls }: Props) => {
+  const classes = useStyles();
+  const { repos, issues, pullRequests, commits, users } = filterGitHubUrls(ghUrls);
```
DukeManh, line 34: let's cache the result of this function call with useMemo so that it doesn't get called every time the `Accordion` is open. 
```suggestion
  const { repos, issues, pullRequests, commits, users } = useMemo(
    () => filterGitHubUrls(ghUrls),
    [ghUrls]
  );
```
Can you do the same for `GithubInfo`?

src/web/src/components/Posts/Post.tsx
```diff
@@ -203,6 +209,42 @@ const useStyles = makeStyles((theme: Theme) =>
...
+      border: 'none',
+      boxShadow: 'none',
+      top: '0',
+      zIndex: 1,
+      position: 'sticky',
+      '& .MuiAccordionDetails-root': {
+        padding: '0 0 0 1rem',
+      },
+    },
+    gitHubIcon: {
```
DukeManh, line 234: You don't use a githubIcon anymore right.

DukeManh (changes requested):

DukeManh:
One area of the inner layout to be improved is the repos list to take up some of the horizontal space. If it looks clustered to you, feel free to design around and add visual separators between them :).
Column | Row
:-------------------------:|:-------------------------:
Screen Shot 2021-12-16 at 11 33 09 PM | Screen Shot 2021-12-16 at 11 33 55 PM

HyperTHD:
> One area of the inner layout to be improved is the repos list to take up some of the horizontal space. If it looks clustered to you, feel free to design around and add visual separators between them :).
> 
> Column	Row
> Screen Shot 2021-12-16 at 11 33 09 PM	Screen Shot 2021-12-16 at 11 33 55 PM

I'm definitely leaning more towards the row style, given a suitable max-width ofc. Column view makes it easier to view the list of repos, at the expense of a much longer height and unused width space being used.

humphd:
Using row vs. column `flex` with this breakpoint could be the answer, I agree.  That looks great.

sirinoks:
We are closing this PR and making our own branch to revert some changes and work on the issue. I new PR will be opened once it is ready.

PR #2496, opened 2021-11-19
Fix startup hang in backend by updating pino

humphd:
We're getting hangs/crashes in the backend.  Bisecting led me to this change: https://github.com/Seneca-CDOT/telescope/commit/c8326da5547fd1b4664138de494128c94829207d, where we updated [pino-pretty](https://github.com/pinojs/pino-pretty) but not pino itself.

This bumps all the loggers used in the backend.  We should do a proper audit and update of the logger infrastructure, since pino has changed how you are supposed to do pretty printing of logs, see https://getpino.io/#/docs/pretty:

```
(node:20671) [PINODEP008] PinoWarning: prettyPrint is deprecated, look at https://github.com/pinojs/pino-pretty for alternatives.
(Use `node --trace-warnings ...` to show where the warning was created)
```

With this change, I'm able to run `node src/backend` and have it start.  I'll file an issue on the prettyPrint warning.


humphd:
Filed #2497 to get rid of the warning.

manekenpix (approved):

PR #2510, opened 2021-11-22
issue-2423: Added state to store student quote

### LINKED ISSUE #2423: Student Quote changes on scroll
In #2422 we added a way for the student quote to be randomly changed on the front page.  However, it's not stable per load, and changes when I scroll the page (i.e., new random number is generated on re-render).

Steps to Reproduce:

1. Go to https://dev.telescope.cdot.systems/
2. Scroll down a few inches, then back up, then down
3. Notice the quote changes over and over again

We need to deal with this function and return value differently:

```js
const getQuote = () => {
  return quotes[Math.floor(Math.random() * quotes.length)];
};
```

We need that value to be stable across renders. cc @AmasiaNalbandian

AmasiaNalbandian:
## Issue This PR Addresses
fixes #2423

## Type of Change
This PR adds the sets a state for the student quote, so that when you scroll on the page, it doesn't grab a new quote to display. 

## Checklist

### DIFF
--- a/src/web/src/components/Banner.tsx
+++ b/src/web/src/components/Banner.tsx
@@ -106,12 +106,10 @@ const useStyles = makeStyles((theme: Theme) =>
   })
 );
 
+const studentQuote = quotes[Math.floor(Math.random() * quotes.length)];
+
 type BannerProps = {
   onVisibilityChange: (visible: boolean) => void;
-};
-
-const getQuote = () => {
-  return quotes[Math.floor(Math.random() * quotes.length)];
 };
 
 export default function Banner({ onVisibilityChange }: BannerProps) {
@@ -121,7 +119,6 @@ export default function Banner({ onVisibilityChange }: BannerProps) {
     sha: '',
     version: '',
   });
-  const studentQuote = getQuote();
 
   const timelineAnchor = useRef<HTMLDivElement>(null);
   const bannerAnchor = useRef<HTMLDivElement>(null);


src/web/src/components/Banner.tsx
```diff
@@ -110,18 +110,14 @@ type BannerProps = {
...
-
 export default function Banner({ onVisibilityChange }: BannerProps) {
   const classes = useStyles();
   const [gitInfo, setGitInfo] = useState({
     gitHubUrl: '',
     sha: '',
     version: '',
   });
-  const studentQuote = getQuote();
+  const [studentQuote] = useState(() => quotes[Math.floor(Math.random() * quotes.length)]);
```
humphd, line 120: Let's consider doing what @DukeManh suggested and move this out of the component.  Put it on line 113 above and do:

```ts
const studentQuote = quotes[Math.floor(Math.random() * quotes.length)]);
```

Then you can use it in the component, but it shouldn't change its value on re-render.

humphd (changes requested):

src/web/src/components/Banner.tsx
AmasiaNalbandian: Perfect. I tried the solution but didn't quite understand we had to move it out, and it had not worked. This works now! Thank you.

humphd (approved):
👍

DukeManh (approved):

PR #2520, opened 2021-11-24
Issue 2412  GET /feeds/invalid

### LINKED ISSUE #2412: Add /feeds/invalid route to Posts service
In #2395 we are adding an expanded status service dashboard.  One thing I want to include in it is the ability to see a list of all feeds that have been flagged as `invalid` in our system.  @joshuali7536 did some work toward making this possible in #2388, and @manekenpix has done some more in #2396 to include the `/feeds` endpoint in the Posts service.

I'd like to see a new route get added to Posts:

```
GET /feeds/invalid
```

This should return a JSON array of Objects with the feed info as well as the reason why it was invalidated.

Doing so will require an update to how the Redis database code works in https://github.com/Seneca-CDOT/telescope/blob/master/src/backend/utils/storage.js.  Currently when we invalidate a feed, we add the `:invalid` suffix to the feed key, and store the reason as its value.

Redis supports doing pattern matching on keys using the [`SCAN` command with `match`](https://redis.io/commands/scan).  You can see an example in JS of what this looks like at https://github.com/luin/ioredis/blob/88b8a3125cd039afdac928c529a1b9958009979c/test/functional/scan_stream.ts#L50-L78.   What we'd do is look for all keys like this:

```
t:feed:*:invalid
```

We can then use these keys to lookup the actual reason for each invalid feed, which is stored in the value for those keys.  We'd use this to return the feed id (and probably URL) along with the reason why this feed was marked invalid.

Later we can use this route to create a widget on the status dashboard with a table of feeds and reasons for being marked invalid.

hphan9:
## Issue This PR Addresses
Fixes #2412 

## Type of Change

## Description

This is PR for issue- #2412: adding GET /feeds/invalid.  I have added the route in posts/src/routes/feeds.js file and new methods to get invalids keys from Redis in posts/src/storage.js. I have not modified the Redis database code in https://github.com/Seneca-CDOT/telescope/blob/master/src/backend/utils/storage.js as prof David mentioned. 
I have not written the unit tests for this but I tested it in local and the GET /feeds/invalid returned the expected result. 
Thank you for reviewing and please let me know if I am on the right track.

## Checklist

### DIFF
--- a/src/api/posts/src/routes/feeds.js
+++ b/src/api/posts/src/routes/feeds.js
@@ -1,6 +1,6 @@
 const { Router, logger, isAuthenticated, isAuthorized } = require('@senecacdot/satellite');
 const Feed = require('../data/feed');
-const { getFeeds } = require('../storage');
+const { getFeeds, getInvalidFeeds } = require('../storage');
 const { validateNewFeed, validateFeedsIdParam } = require('../validation');
 
 const feeds = Router();
@@ -25,6 +25,23 @@ feeds.get('/', async (req, res, next) => {
         id,
         url: `${feedURL}/${id}`,
       }))
+  );
+});
+
+feeds.get('/invalid', async (req, res, next) => {
+  let invalidFeeds;
+  try {
+    invalidFeeds = await getInvalidFeeds();
+  } catch (error) {
+    logger.error({ error }, 'Unable to get invalid feeds from Redis');
+    return next(error);
+  }
+  res.set('X-Total-Count', invalidFeeds.length);
+  return res.json(
+    invalidFeeds.map((element) => ({
+      ...element,
+      url: `${feedURL}/${element.id}`,
+    }))
   );
 });
 
--- a/src/api/posts/src/storage.js
+++ b/src/api/posts/src/storage.js
@@ -22,6 +22,25 @@ const createInvalidFeedKey = (id) => createFeedKey(id).concat(invalidSuffix);
 // "NirlSYranl" to "t:feed:NirlSYranl:delayed"
 const createDelayedFeedKey = (id) => createFeedKey(id).concat(delayedSuffix);
 
+const getFeedKeysUsingScanStream = (matchPattern) => {
+  const keys = new Set();
+  const stream = redis.scanStream({
+    match: matchPattern,
+  });
+  return new Promise((resolve, reject) => {
+    stream.on('data', (data = []) => {
+      data.forEach((k) => keys.add(k));
+    });
+    stream.on('error', (err) => {
+      logger.error({ err }, 'Error while scanning redis keys');
+      reject(new Error('Error while scanning redis keys'));
+    });
+    stream.on('end', () => {
+      resolve([...keys]);
+    });
+  });
+};
+
 module.exports = {
   /**
    * Feeds
@@ -56,6 +75,20 @@ module.exports = {
   },
 
   getFeeds: () => redis.smembers(feedsKey),
+
+  getInvalidFeeds: async () => {
+    const invalidKeys = await getFeedKeysUsingScanStream(`${feedNamespace}*${invalidSuffix}`);
+    return Promise.all(
+      invalidKeys.map(async (key) => {
+        const reason = await redis.get(key);
+        const id = key.replace(feedNamespace, '').replace(invalidSuffix, '');
+        return {
+          id,
+          reason,
+        };
+      })
+    );
+  },
 
   getFlaggedFeeds: () => redis.smembers(flaggedFeedsKey),
 
--- a/src/api/posts/test/feeds.test.js
+++ b/src/api/posts/test/feeds.test.js
@@ -1,0 +1,39 @@
+const request = require('supertest');
+const { app } = require('../src');
+const Feed = require('../src/data/feed');
+
+describe('Test GET /feeds/invalid endpoint', () => {
+  const createdItems = 150;
+  // Array of feeds
+  const feeds = [...Array(createdItems).keys()].map((item) => {
+    return new Feed('foo', `http://telescope${item}.cdot.systems`);
+  });
+  beforeAll(() => Promise.all(feeds.map((feed) => feed.save())));
+
+  it('Should return 200 and empty body when there are no invalid feeds', async () => {
+    const res = await request(app).get('/feeds/invalid');
+
+    expect(res.status).toEqual(200);
+    expect(res.get('Content-type')).toContain('application/json');
+    expect(res.get('X-Total-Count')).toBe('0');
+    expect(res.body.length).toBe(0);
+    expect(res.body instanceof Array).toBe(true);
+  });
+
+  it('Should return 200 and invalid feeds object', async () => {
+    // arrange invalid feeds
+    const createdInvalidFeeds = 20;
+    for (let i = 0; i < createdInvalidFeeds; i += 1) {
+      feeds[i].setInvalid('test');
+    }
+    const res = await request(app).get('/feeds/invalid');
+
+    expect(res.status).toEqual(200);
+    expect(res.get('Content-type')).toContain('application/json');
+    expect(res.get('X-Total-Count')).toBe(createdInvalidFeeds.toString());
+    expect(res.body.length).toBe(createdInvalidFeeds);
+    expect(res.body instanceof Array).toBe(true);
+    expect(res.body[0].reason).toBe('test');
+    expect(res.body[0].id).toBe(feeds[0].id);
+  });
+});


src/api/posts/src/routes/feeds.js
```diff
@@ -34,6 +34,32 @@ feeds.get('/', async (req, res, next) => {
...
 });
 
+//This should return a JSON array of Objects with the feed info as well as the reason why it was invalidated.
+feeds.get('/invalid', async (req, res, next) => {
+  let keys = [];
+  try {
+    keys = await getInvalidFeeds();
+  } catch (error) {
+    logger.error({ error }, 'Unable to get invalid feeds from Redis');
+    next(error);
```
humphd, line 44: You need a `return` here so that it doesn't keep going.

src/api/posts/src/routes/feeds.js
```diff
@@ -34,6 +34,32 @@ feeds.get('/', async (req, res, next) => {
   );
 });
 
+//This should return a JSON array of Objects with the feed info as well as the reason why it was invalidated.
+feeds.get('/invalid', async (req, res, next) => {
+  let keys = [];
```
humphd, line 39: I would rename this `invalidFeeds` vs. `keys`. Also, no need for initialization here, just `let keys;` is enough.

src/api/posts/src/routes/feeds.js
```diff
@@ -34,6 +34,32 @@ feeds.get('/', async (req, res, next) => {
...
+  try {
+    keys = await getInvalidFeeds();
+  } catch (error) {
+    logger.error({ error }, 'Unable to get invalid feeds from Redis');
+    next(error);
+  }
+  const invalidsFeeds = keys.map(async (element) => {
+    const feedInfo = await Feed.byId(element.hashId);
```
humphd, line 47: Do we need the feed details?  If you already have the `id` and `reason`, that's probably enough.  I would skip getting all the feeds by id.

src/api/posts/src/storage.js
```diff
@@ -22,6 +22,26 @@ const createInvalidFeedKey = (id) => createFeedKey(id).concat(invalidSuffix);
 // "NirlSYranl" to "t:feed:NirlSYranl:delayed"
 const createDelayedFeedKey = (id) => createFeedKey(id).concat(delayedSuffix);
 
+const getInvalidKeys = async () => {
+  let keysArray = [];
+  const stream = redis.scanStream({
+    match: 't:feed:*:invalid',
+  });
+  return new Promise((res, rej) => {
+    stream.on('data', async (keys) => {
```
humphd, line 31: You don't need `async` here, since you aren't `await`ing any promise.

humphd: I would also set a default for `keys` so you don't have to check its value: `(keys = []) => {...}`

src/api/posts/src/storage.js
```diff
@@ -22,6 +22,26 @@ const createInvalidFeedKey = (id) => createFeedKey(id).concat(invalidSuffix);
...
+      keysArray = keysArray.concat(keys);
+    });
+    stream.on('end', () => {
+      //remove duplicate keys
```
humphd, line 35: There won't be any, since keys are unique.  You can skip this bit and just `resolve` the array.

src/api/posts/src/storage.js
```diff
@@ -22,6 +22,26 @@ const createInvalidFeedKey = (id) => createFeedKey(id).concat(invalidSuffix);
 // "NirlSYranl" to "t:feed:NirlSYranl:delayed"
 const createDelayedFeedKey = (id) => createFeedKey(id).concat(delayedSuffix);
 
+const getInvalidKeys = async () => {
+  let keysArray = [];
+  const stream = redis.scanStream({
+    match: 't:feed:*:invalid',
+  });
+  return new Promise((res, rej) => {
```
humphd, line 30: Prefer `resolve` and `reject` vs. `res`, since it also could mean `response`.  It's better to be clear.

src/api/posts/src/storage.js
```diff
@@ -22,6 +22,26 @@ const createInvalidFeedKey = (id) => createFeedKey(id).concat(invalidSuffix);
...
+    stream.on('data', async (keys) => {
+      keysArray = keysArray.concat(keys);
+    });
+    stream.on('end', () => {
+      //remove duplicate keys
+      let filterArray = [...new Set(keysArray)];
+      res(filterArray);
+    });
```
humphd, line 38: Let's add something deal with the error case:

```js
stream.on('error', (err) => {
  logger.error({ err }, 'Error while scanning redis keys');
  reject([]);
});
```

src/api/posts/src/storage.js
```diff
@@ -22,6 +22,26 @@ const createInvalidFeedKey = (id) => createFeedKey(id).concat(invalidSuffix);
...
+  }).catch((err) => {
```
humphd, line 39: You can get rid of this, in favour of the stream error handling I suggested.

src/api/posts/src/storage.js
```diff
@@ -57,6 +77,22 @@ module.exports = {
...
+  getInvalidFeeds: async () => {
+    const invalidKeys = await getInvalidKeys();
+    const feeds = invalidKeys.map(async (key) => {
+      //do we need to check the response of redis.get
+      const reason = await redis.get(key);
+      const id = key.replace(invalidSuffix, '');
+      const hashId = id.replace(feedNamespace, '');
+      return {
+        id: id,
+        hashId: hashId,
```
humphd, line 89: I think we only care about the actual feed id.  I would send that as `id`.

src/api/posts/src/storage.js
```diff
@@ -57,6 +77,22 @@ module.exports = {
...
+        reason: reason,
+      };
+    });
+    return Promise.all(feeds);
```
humphd, line 93: Move this up to line 82 and get rid of your variables:

```js
const invalidKeys = await getInvalidKeys();
return Promise.all(invalidKeys.map(...));

**Review by humphd — CHANGES_REQUESTED** [review]
This looks like its heading in the right direction, nice work.  I left some feedback.

File: src/api/posts/src/routes/feeds.js (hunk 1 above)
Comment (hphan9) [reply src/api/posts/src/routes/feeds.js:44]: Thank you Prof, I will fix it

File: src/api/posts/src/routes/feeds.js (hunk 3 above)
Comment (hphan9) [reply src/api/posts/src/routes/feeds.js:47]: that is great, it will be cleaner

File: src/api/posts/src/storage.js (hunk 7 above)
Comment (hphan9) [reply src/api/posts/src/storage.js:38]: This is great recommendation, thank you

File: src/api/posts/src/storage.js (hunk 5 above)
Comment (hphan9) [reply src/api/posts/src/storage.js]: Hi prof, about this , in the ioredis documentation , they mention that keys might contain duplicate values due to Scan's implementation in redis.
example: 
```
const redis = new Redis();
// Create a readable stream (object mode)
const stream = redis.scanStream();
stream.on("data", (resultKeys) => {
  // `resultKeys` is an array of strings representing key names.
  // Note that resultKeys may contain 0 keys, and that it will sometimes
  // contain duplicates due to SCAN's implementation in Redis.
  for (let i = 0; i < resultKeys.length; i++) {
    console.log(resultKeys[i]);
  }
});
stream.on("end", () => {
  console.log("all keys have been visited");
});
``` 
I am new to Redis and ioredis so  I am not sure about this issue ?

Comment (humphd) [reply src/api/posts/src/storage.js]: OK, if that's the case, I would use a `Set()` vs. an `Array` to store these, and then you can [`.add()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/add) as many times as you want and we won't get duplicates.  When it's done (i.e., in `"end"`), you can give back `[...set]` or `Array.from(set)` or something to convert it to an Array.

Comment (hphan9) [reply src/api/posts/src/storage.js]: Thank you, I have made a change according to your suggestion.

File: src/api/posts/src/storage.js [hunk 11]
```diff
@@ -22,6 +22,25 @@ const createInvalidFeedKey = (id) => createFeedKey(id).concat(invalidSuffix);
 // "NirlSYranl" to "t:feed:NirlSYranl:delayed"
 const createDelayedFeedKey = (id) => createFeedKey(id).concat(delayedSuffix);
 
+const getFeedKeysUsingScanStream = (matchPattern) => {
+  const keysArray = new Set();
```
humphd, line 26: Because this isn't an `Array`, let's not name it that.  `keys` is fine.

humphd (changes requested):
Really nice, one fix.

src/api/posts/src/storage.js
hphan9: Hi prof, thank you, I have made the change. Please review :)

src/api/posts/src/storage.js
```diff
@@ -57,6 +76,20 @@ module.exports = {
 
   getFeeds: () => redis.smembers(feedsKey),
 
+  getInvalidFeeds: async () => {
+    const invalidKeys = await getFeedKeysUsingScanStream('t:feed:*:invalid');
```
humphd, line 80: Let's not hard-code the key values, use the variables we have:

```js
`${feedNamespace}*${invalidSuffix}`
```

humphd (changes requested):
This looks great.  One minor change and I think it's ready.  Please make that fix, squash and rebase.  Then we can get one more reviewer to take a look.  Well done.

src/api/posts/src/storage.js
hphan9: Thank you, it is better than hardcode. I have rebased and updated the Pr.

humphd (approved):
Really nice work!

manekenpix (approved):

PR #2522, opened 2021-11-24
Fixes #2489: Add SCSS compile step to dashboard

### LINKED ISSUE #2489: Add SCSS compiler step to status dashboard
Use SCSS for status dashboard

The primary color and many other presets/components are defined in the Material Dashboard 2 SCSS files. It's inconvenient to keep overriding the same presets/components.

Andrewnt219:
## Issue This PR Addresses
Closes #2489

## Type of Change

## Description

Add docs for setting up the status dashboard.

Safe to ignore scss files. All the files are copied from the original Material Dashboard 2, with one change to the $primary-color.

The CSS folder in assets is deleted because it is now generated by scss. Scss is not inside the public folder because it should not be included in the production build.

~~Node engine is limited to 14 because `gulp-sass` depends on `node-sass` 4 which depends on node 12-14. This can be solve by upgrading to `gulp` 5 and use `sass` instead of `node-sass`. However, there is a bit of migration work needed to fix the scss between `sass` and `node-sass`, so I won't do that in this PR.~~ 
Move to sass, should be good now. The migration includes converting [deprecated division](https://sass-lang.com/documentation/breaking-changes/slash-div) to a multiplier (e.g. /2 -> *0.5). Trust my math skills.

I don't know if I do the right step for docker. Need someone to check on that.

~~EDIT: Tests using Node 16 are failing, guess I have to fix the node-sass problem in this PR after all.~~

## Test plan
```bash
# Start the server
npm run dev

# Compile scss to css
npm run compile:scss
```

Make a change in scss
```diff
scss/material-dashboard/_variable.scss
- $primary-gradient: #182677 !default;
+ $primary-gradient: #000 !default;
```

```bash
# Compile scss to CSS again
npm run compile:scss
```

See the active navlink on side bar change to black.

## Checklist


src/api/status/Dockerfile
```diff
@@ -9,6 +9,8 @@ COPY --chown=node:node . .
 
 RUN npm install --only=production --no-package-lock
 
+RUN npm run compile:scss
```
manekenpix, line 12: This is not going to work since we're using `node:lts-alpine` as our base image (`lts` is 16).
I'd suggest changing the first line of this file to:

```yml
FROM node:14.18-alpine as base
```
I tested this and docker can build and launch it.

src/api/status/package.json
```diff
@@ -18,10 +20,17 @@
     "express": "^4.17.1"
   },
   "engines": {
-    "node": ">=12.0.0"
+    "node": ">=12.0.0 <=14"
```
humphd, line 23: Yeah, we can't use <16 with other parts of how our stuff works.

src/api/status/Dockerfile
humphd, line 12: This looks fine

src/api/status/package.json
humphd: Actually, one thing we could do is split the Dockerfile into two stages, and have the SCSS compile step use node 14 or whatever, then copy the built artifacts out to the next stage and use node 16.  That won't solve the local development case, though, if that matters?

humphd (changes requested):

src/api/status/package.json
```diff
@@ -18,10 +20,17 @@
...
     "env-cmd": "^10.1.0",
-    "nodemon": "^2.0.15"
+    "nodemon": "^2.0.15",
+    "gulp": "^4.0.2",
+    "gulp-autoprefixer": "^7.0.1",
+    "gulp-clean": "^0.4.0",
+    "gulp-install": "^1.1.0",
+    "gulp-sass": "^4.1.0",
+    "gulp-sourcemaps": "^2.6.5",
+    "gulp-open": "^3.0.1"
```
manekenpix, line 34: These can't be `devDependencies` since we [don't install them in production](https://github.com/Seneca-CDOT/telescope/blob/17d5c1522e7b799d4d9b826ec0223c4bb3714ea9/src/api/status/Dockerfile#L10) (this is probably why the `e2e` tests are failing).

src/api/status/package.json
manekenpix: That should do it, something like this:

```yml
FROM node:14.18-alpine as base

WORKDIR /app

COPY . .

RUN npm install --only=production --no-package-lock

RUN npm run compile:scss

FROM node:lts-alpine as deployment

WORKDIR /app

# https://snyk.io/blog/10-best-practices-to-containerize-nodejs-web-applications-with-docker/
RUN apk add dumb-init

COPY --chown=node --from=base /app /app
USER node

CMD ["dumb-init", "node", "src/server.js"]
```

manekenpix: But I'm not sure if it's a good idea to install dependencies and build using one version of node and then deploy with a different version?
I think we should start having `.nvmrc` files for scenarios like this one.

Andrewnt219: @manekenpix So should I go with the split or your first suggestion?

src/api/status/package.json
Andrewnt219: That makes sense, I'll move them.

src/api/status/package.json
manekenpix: I'm not sure, Having a `.nvmrc` file would help with local development, but I'm not sure if we should do that in this PR. As for the `Dockerfile`, I'd use only one version of `node` in the container, but maybe @humphd has a better approach.

Andrewnt219: We still have another option that is using `sass` instead of `node-sass`. It only needs a small migration work, which is replacing deprecated API. Last time I checked there were 20 lines needed replacing, not much.

humphd: If you can do this by using standard node packages on 16+, I'd do that vs. complex Docker or dev setups.

humphd:
@Andrewnt219 can you update me a bit on what I'm reviewing here?  Some of what @manekenpix hasn't been addressed, but maybe that's OK now?  This needs clarification, as the PR is massive to read without knowing what I'm looking for.

Andrewnt219:
I updated it in the PR description. @humphd

Andrewnt219:
The migration can be reviewed in commit [migrate to sass](https://github.com/Seneca-CDOT/telescope/commit/589a97462d4d084244d009244801fcdbb3426075)

src/api/status/Dockerfile
humphd: @Andrewnt219 does this still apply with your latest change to sass vs. node-sass?  If it's not a problem any more, please resolve this.

Andrewnt219: @manekenpix do we still need this line when the node engine is no longer restricted to 14 and below?

manekenpix: I don't think so. Your latest changes seem to work with node v16, so I think we're ok.

src/api/status/gulpfile.js
```diff
@@ -0,0 +1,28 @@
+var gulp = require('gulp');
+var path = require('path');
+var sass = require('gulp-sass')(require('sass'));
+var autoprefixer = require('gulp-autoprefixer');
+var sourcemaps = require('gulp-sourcemaps');
+var open = require('gulp-open');
+
+var Paths = {
```
manekenpix, line 8: Can we use `const` here?

Andrewnt219: good catch, I just blindly copied that from the template.

manekenpix (dismissed):
LGTM :+1:

src/api/status/package.json
```diff
@@ -15,7 +17,15 @@
   "homepage": "https://github.com/Seneca-CDOT/telescope#readme",
   "dependencies": {
     "@senecacdot/satellite": "^1.x",
-    "express": "^4.17.1"
+    "express": "^4.17.1",
+    "gulp": "^4.0.2",
```
humphd, line 21: These are all `devDependencies`, no?

humphd (changes requested):

src/api/status/package.json
manekenpix: https://github.com/Seneca-CDOT/telescope/pull/2522#discussion_r756479096

Andrewnt219: @humphd e2e doesn't install dev dependencies so it crashes the tests. 

https://github.com/Seneca-CDOT/telescope/pull/2522#discussion_r756479096

Andrewnt219: Another way is manually run the sass compiler
```
- npm run compile:scss
+ npm run sass .\scss\material-dashboard.scss:.\public\assets\css\material-dashboard.css
```

humphd: It's not a huge deal if we have to do it this way.

Andrewnt219:
If gulp is not used in production, may as well remove it from development. [33e6aaf remove gulp](https://github.com/Seneca-CDOT/telescope/pull/2522/commits/33e6aafb96472914a1ad1992bb8edf174a812d1f)

manekenpix (approved):
:+1:

humphd (approved):

PR #2551, opened 2021-11-30
Add API status to the new dashboard

### LINKED ISSUE #2418: Rewrite the Microservices API status check in the new status dashboard
In #2395 we are adding an expanded status service dashboard. One thing I want to see done is to port the existing Microservices API status check code over to the new dashboard.  The current status check UI is at:

- staging https://dev.api.telescope.cdot.systems/v1/status
- production https://api.telescope.cdot.systems/v1/status

It is a simple UI that queries all the health routes for each service to see if they are responding or not.  At the moment it seem to be done (it crashes sometimes, which would be good to investigate and fix).

If we added `CORS` to the [healthchecks that Satellite creates](https://github.com/Seneca-CDOT/satellite/blob/d42a831a8ca6666dd5759eeccdbe932fe34239ca/src/satellite.js#L20), we could probably get rid of the backend code and do it all in the browser.

Since we're going to be replacing [all this current code](https://github.com/Seneca-CDOT/telescope/tree/master/src/api/status), I want to make sure we don't lose it, and the functionality of being able to see if services are OK should be duplicated in the new dashboard.

Kevan-Y:
## Issue This PR Addresses
Fixes #2418 

## Type of Change

## Description
- Add API route to get all health information
- Update dashboard to display the status information
- Add function to fetch and display to dashboard 

https://user-images.githubusercontent.com/58233223/144141115-6b10b87a-919b-4b55-8a86-d70fd3f8dae5.mp4

## Checklist

### DIFF
--- a/src/api/status/public/js/serviceStatus.js
+++ b/src/api/status/public/js/serviceStatus.js
@@ -1,0 +1,71 @@
+const reportTbody = () => document.querySelector('#report');
+const requestMessageReport = () => document.querySelector('#request-message');
+
+// Create td for service name
+const createTdText = (content) => {
+  const td = document.createElement('td');
+  const div = document.createElement('div');
+  const h6 = document.createElement('h6');
+
+  h6.className = 'text-capitalize mb-0 text-sm';
+  h6.innerHTML = content;
+  div.className = 'text-start px-2 py-1';
+  div.appendChild(h6);
+  td.appendChild(div);
+
+  return td;
+};
+
+// Create td for icon
+const createTdIcon = (status) => {
+  const td = document.createElement('td');
+  const icon = document.createElement('i');
+
+  td.className = 'align-middle text-center';
+  icon.title = status;
+  icon.className =
+    status >= 200 && status <= 299
+      ? 'text-lg text-success fas fa-check-circle'
+      : 'text-lg text-danger fas fa-exclamation-circle';
+
+  td.appendChild(icon);
+
+  return td;
+};
+
+// Create row of the table
+const serviceRowFormat = (result) => {
+  const {
+    name,
+    status: { staging, production },
+  } = result;
+  const tr = document.createElement('tr');
+
+  tr.appendChild(createTdText(name));
+  tr.appendChild(createTdIcon(staging));
+  tr.appendChild(createTdIcon(production));
+  reportTbody().appendChild(tr);
+};
+
+export default function getAllServicesStatus() {
+  fetch('/v1/status/status')
+    .then((res) => {
+      if (res.ok) return res.json();
+
+      const messageReport = requestMessageReport();
+      if (messageReport) {
+        messageReport.innerHTML = '';
+        const icon = document.createElement('i');
+        icon.className = 'fas fa-server px-2';
+        messageReport.appendChild(icon);
+        messageReport.innerHTML += `${res.status} Unable to get API status.`;
+      }
+      throw new Error('unable to get API status');
+    })
+    .then((results) => {
+      reportTbody().innerHTML = '';
+      results.forEach((result) => serviceRowFormat(result));
+      return setTimeout(getAllServicesStatus, 5000);
+    })
+    .catch((err) => console.error(err));
+}
--- a/src/api/status/src/server.js
+++ b/src/api/status/src/server.js
@@ -1,6 +1,7 @@
 const { Satellite } = require('@senecacdot/satellite');
 const { static: serveStatic } = require('express');
 const path = require('path');
+const { check } = require('./services');
 
 const host = process.env.API_HOST || 'localhost';
 
@@ -42,6 +43,16 @@ if (process.env.PATH_PREFIX) {
 } else {
   service.router.use('/', serveStatic(path.join(__dirname, '../public')));
 }
+
+service.router.get('/v1/status/status', (req, res) => {
+  check()
+    .then((status) => {
+      // This status response shouldn't be cached (we need current status info)
+      res.set('Cache-Control', 'no-store, max-age=0');
+      return res.json(status);
+    })
+    .catch((err) => res.status(500).json({ error: err.message }));
+});
 
 const port = parseInt(process.env.STATUS_PORT || 1111, 10);
 service.start(port);
--- a/src/api/status/public/index.html
+++ b/src/api/status/public/index.html
@@ -519,42 +519,11 @@
             <div class="card">
               <div class="card-header pb-0">
                 <div class="row">
-                  <div class="col-lg-6 col-7">
-                    <h6>Projects</h6>
+                  <div class="col">
+                    <h6>Telescope API Status</h6>
                     <p class="text-sm mb-0">
-                      <i class="fa fa-check text-info" aria-hidden="true"></i>
-                      <span class="font-weight-bold ms-1">30 done</span> this month
+                      Status of the Telescope Web and Microservices API servers.
                     </p>
-                  </div>
-                  <div class="col-lg-6 col-5 my-auto text-end">
-                    <div class="dropdown float-lg-end pe-4">
-                      <a
-                        class="cursor-pointer"
-                        id="dropdownTable"
-                        data-bs-toggle="dropdown"
-                        aria-expanded="false"
-                      >
-                        <i class="fa fa-ellipsis-v text-secondary"></i>
-                      </a>
-                      <ul
-                        class="dropdown-menu px-2 py-3 ms-sm-n4 ms-n5"
-                        aria-labelledby="dropdownTable"
-                      >
-                        <li>
-                          <a class="dropdown-item border-radius-md" href="javascript:;">Action</a>
-                        </li>
-                        <li>
-                          <a class="dropdown-item border-radius-md" href="javascript:;"
-                            >Another action</a
-                          >
-                        </li>
-                        <li>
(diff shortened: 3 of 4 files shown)


humphd (changes requested):
On GitPod I get:

Screen Shot 2021-11-30 at 7 40 56 PM

It's doing:

Screen Shot 2021-11-30 at 7 42 49 PM

The problem is in:

```js
export default function getAllServicesStatus() {
  fetch('/status')
```

This may be a problem specific to GitPod's URL/port layout, but I suspect that this isn't quite right yet.

Kevan-Y:
> On GitPod I get:
> 
> Screen Shot 2021-11-30 at 7 40 56 PM
> 
> It's doing:
> 
> Screen Shot 2021-11-30 at 7 42 49 PM
> 
> The problem is in:
> 
> ```js
> export default function getAllServicesStatus() {
>   fetch('/status')
> ```
> 
> This may be a problem specific to GitPod's URL/port layout, but I suspect that this isn't quite right yet.


Notice in GitPods is not running the server which allows the API to fetch all status to work.
In  `src/api/status/src/server.js`
 ```JS
 service.router.get('/status', (req, res) => {
  check()
    .then((status) => {
      // This status response shouldn't be cached (we need current status info)
      res.set('Cache-Control', 'no-store, max-age=0');
      return res.json(status);
    })
    .catch((err) => res.status(500).json({ error: err.message }));
});
```
Could it also be this issue?

**Comment by humphd** [comment]
How is the status page showing up if it's not running?  It gets hosted by that server.

**Comment by humphd** [comment]
cc @DukeManh for gitpod wisdom

File: src/api/status/public/js/serviceStatus.js [hunk 1]
```diff
@@ -0,0 +1,65 @@
...
+  const tr = document.createElement('tr');
+
+  tr.appendChild(createTdText(result.name));
+  tr.appendChild(createTdIcon(result.status.staging));
+  tr.appendChild(createTdIcon(result.status.production));
+  reportTbody().appendChild(tr);
+};
+
+export default function getAllServicesStatus() {
+  fetch('/status')
```
manekenpix, line 47:
```suggestion
  fetch('/v1/status')
```

src/api/status/public/js/serviceStatus.js
```diff
@@ -0,0 +1,65 @@
...
+    .then((res) => {
+      if (!res.ok) {
+        requestMessageReport().innerHTML = '';
+        const icon = document.createElement('i');
+        icon.className = 'fas fa-server px-2';
+        requestMessageReport().appendChild(icon);
+        requestMessageReport().innerHTML += `${res.status} Unable to get API status.`;
+        throw new Error('unable to get API status');
+      }
+      return res.json();
```
manekenpix, line 57:
```suggestion
      return res;
```

manekenpix (changes requested):

src/api/status/public/js/serviceStatus.js
Kevan-Y: Changed 🙏

src/api/status/public/js/serviceStatus.js
```diff
@@ -0,0 +1,65 @@
...
+    })
```
DukeManh, line 58: You should exit early and reuse variable.
```suggestion
      if (res.ok) {
        return res.json();
      }
        const messageReport = requestMessageReport();
        if (messageReport) {
          messageReport.innerHTML = '';
          const icon = document.createElement('i');
          icon.className = 'fas fa-server px-2';
          messageReport.appendChild(icon);
          messageReport.innerHTML += `${res.status} Unable to get API status.`;
          throw new Error('unable to get API status');
        }
    })
```

src/api/status/public/js/serviceStatus.js
```diff
@@ -0,0 +1,65 @@
...
+  return td;
+};
+
+// Create row of the table
+const serviceRowFormat = (result) => {
+  const tr = document.createElement('tr');
+
+  tr.appendChild(createTdText(result.name));
+  tr.appendChild(createTdIcon(result.status.staging));
+  tr.appendChild(createTdIcon(result.status.production));
```
DukeManh, line 42:
```suggestion
  const {
    name,
    status: { staging, production },
  } = result;
```

DukeManh (changes requested):
It works on Gitpod when I run `npm run dev` from `src/api/status`.

I think the variable naming can be better

src/api/status/src/server.js
```diff
@@ -43,5 +44,15 @@ if (process.env.PATH_PREFIX) {
   service.router.use('/', serveStatic(path.join(__dirname, '../public')));
 }
 
+service.router.get('/v1/status/status', (req, res) => {
```
HyperTHD, line 47: `/v1/status` instead of `/v1/status/status`

HyperTHD (dismissed):

HyperTHD:
Ignore the approval, misclick. Need to fix what I mentioned above

src/api/status/public/js/serviceStatus.js
DukeManh: @manekenpix, the status server is serving `/status` endpoint and hosting this public folder, I think it's ok to fetch from `/status`.

src/api/status/src/server.js
manekenpix: @HyperTHD the endpoint defined to get the status is `/status`, which makes the whole route `/v1/status/status`

src/api/status/public/js/serviceStatus.js
manekenpix: Yes, but the `status` endpoint (`/v1/status/status`) is the one providing the `json` with all the services' status for both staging and production, you need to hit `/v1/status/status` to get that data.

src/api/status/src/server.js
Kevan-Y: Yup I think API wouldn't work if we changed to `/v1/status` since `/v1/status` fetch the static site

src/api/status/public/js/serviceStatus.js
Kevan-Y: Thanks, changed.
I don't think we need a check for messageReport.

src/api/status/public/js/serviceStatus.js
Kevan-Y: Fixed, I always forgot about destructuring assignment if I got the term right. Thanks, I will remind myself to do this in the future.

src/api/status/public/js/serviceStatus.js
DukeManh: It's a safe error handling because `querySelector` can return Element | null. I got used to Typescript strictness.

Kevan-Y: If so, I have to add a empty `return;` after the if, since then() required either an return or throw

DukeManh: nit:
```suggestion
if (messageReport) {
          messageReport.innerHTML = '';
          const icon = document.createElement('i');
          icon.className = 'fas fa-server px-2';
          messageReport.appendChild(icon);
          messageReport.innerHTML += `${res.status} Unable to get API status.`;
        }
 throw new Error('unable to get API status');
```
@Kevan-Y

Kevan-Y: Fixed

Andrewnt219 (approved):
Nice enhancement 🙌.

humphd:
@Kevan-Y can you rebase to pick up the change to `src/api/status/public/index.html`?

humphd (changes requested):
I love this!

Screen Shot 2021-12-04 at 2 46 41 PM

Can we only show the services for the current deployment (i.e., staging or production, but not both at once)?  I think if you are looking at the status service on staging, we should show you staging only; same for prod.

Kevan-Y:
> I love this!
> 
> Screen Shot 2021-12-04 at 2 46 41 PM
> 
> Can we only show the services for the current deployment (i.e., staging or production, but not both at once)? I think if you are looking at the status service on staging, we should show you staging only; same for prod.

That possible, I just looked at the `env.local` what `API_HOST` for? 
Can we use this to check if it's production URL, or staging URL?

humphd:
You should be able to use the URL the page is loaded on to figure it out, i.e., https://dev.api.telescope.cdot.systems/v1/status (staging) vs https://api.telescope.cdot.systems/v1/status (production).  We can't access env vars in the dashboard.

Kevan-Y:
> You should be able to use the URL the page is loaded on to figure it out, i.e., https://dev.api.telescope.cdot.systems/v1/status (staging) vs https://api.telescope.cdot.systems/v1/status (production). We can't access env vars in the dashboard.

Using URL will work but not on Gitpot since they are using another URL.

humphd:
When we fix #2553 we can do this server side.

Kevan-Y:
Ok sure, I will leave it like that for now and when working on  #2553 I will do it to show only staging/production

humphd:
This is failing prettier

HyperTHD:
How can I test this PR to make sure it works? The services seem to not be able to appear despite running. Tried on both gitpod and locally by pulling this PR @Kevan-Y

Kevan-Y:
@HyperTHD , to run it, you will have to go to `src\api\status` and do a `npm run dev`

HyperTHD (approved):

Andrewnt219 (approved):
Well done👍.

Issue #2519, opened 2021-11-24
Extend build terminal in dashboard to full height of viewport


humphd:
The build terminal (see https://dev.api.telescope.cdot.systems/v1/status/pages/build.html) is only occupying the top portion of the page.  Let's have it extend to fill the available space vertically.

<img width="1254" alt="Screen Shot 2021-11-23 at 6 58 11 PM" src="https://user-images.githubusercontent.com/427398/143147402-718cb809-0d3e-448a-8fd0-6438331dda98.png">

Yoda-Canada:
@humphd I would like to work on this for release 0.4

humphd:
@Yoda-Canada sounds good.  NOTE: this isn't enough for 0.4, so look for some other things to pair it with.

Yoda-Canada:
@humphd where is the code? I find the code in the webpage, but can't find it on GitHub

humphd:
https://github.com/Seneca-CDOT/telescope/blob/master/src/api/status/public/pages/build.html and https://github.com/Seneca-CDOT/telescope/blob/master/src/api/status/public/js/pages/build.js

Yoda-Canada:
I only can find the id="terminal" and I try to "go to definition", but can't find anything. 
![image](https://user-images.githubusercontent.com/55161917/143319573-76551d25-c534-4636-b325-590aecfbbac6.png)

humphd:
Yes, it gets created dynamically by xterm.js at runtime.  You need to stretch the `#terminal` parent element to fill the vertical space, and the terminal should fill it automatically.

Yoda-Canada:
@humphd I try to use the following code to solve this issue, but it doesn't change.  Then I check the container size and the main class size doesn't fit the whole screen. I changed the max-height, but it not useful. Could you give some clues?
```
<div id="terminal">
      <script>
        var term = new Terminal();
        var fitAddon = new FitAddon();
        term.loadAddon(fitAddon);
        term.open(document.getElementById('terminal'));
        fitAddon.fit();
      </script>
</div>
```
![image](https://user-images.githubusercontent.com/55161917/143374410-ca7ff93e-14e6-4988-b440-6ad394535da8.png)

humphd:
Short of solving this myself, I'm not sure.  You need to do some research, work with your classmates to try and dig into things, etc.  You can do this.

HyperTHD:
[...]

PR #2556, opened 2021-12-01
edit terminal.js to full height of view port

Yoda-Canada:
issue #2519:
The terminal port can't auto fit the height because the Browser can't get the size of the container.  So I edit the terminal.js file and get the window's width and height, then calculate the row height and column width based on these data.
If the user has more than one screen, just refresh the screen, the terminal window will fit the new screen.
**Before:**
e01f38d53a69df584f6bb85c87a20d2
**After:**
6e3615f1006fa113d90c12ff899bea3

## Type of Change

## Checklist

### DIFF
--- a/src/api/status/public/css/terminal.css
+++ b/src/api/status/public/css/terminal.css
@@ -1,3 +1,4 @@
 .xterm {
   padding: 0.5rem;
+  height: 89vh;
 }


src/api/status/public/js/build-log/terminal.js
```diff
@@ -20,9 +20,12 @@ const theme = {
   white: '#CCCCC6',
   yellow: '#FD971F',
 };
+
+const width = window.innerWidth;
+const height = window.innerHeight;
```
humphd, line 25: This will work, but it duplicates what we already have with https://www.npmjs.com/package/xterm-addon-fit.  The right way to do this is to have the parent element that holds the terminal fill the available width and height, and then it will automatically use that.  Can you fix our CSS for the page so that it does this?

humphd (changes requested):

src/api/status/public/js/build-log/terminal.js
Yoda-Canada: Okay, I will try to fix CSS. Actually I have checked some information about why the xterm-addon-fit command invalid. One explanation is that the container does not open until the browser has checked the size, so the xterm-addon-fit command is invalid. That is why I check windows' size first.

Yoda-Canada:
@humphd I have edited the CSS ,and the terminal full height of view port. Please review it, thx.
image

Andrewnt219 (changes requested):
This is not working. Reproduce

```js
// terminal.js
terminal.write(`
  lorem
  ipsum
`);
```

The terminal should display the text on two lines, but it has glitchy text instead. 

Maybe we can use [`vh`](https://developer.mozilla.org/en-US/docs/Web/CSS/length) instead of percentage?

Yoda-Canada:
@Andrewnt219 You are right. I have changed it and please review.

src/api/status/public/css/terminal.css
```diff
@@ -1,3 +1,6 @@
 .xterm {
   padding: 0.5rem;
+  position: relative;
+  width: 100vh;
+  height: 89vh;
```
Andrewnt219, line 5: I think `vh` is the correct way, but seem like we need to apply it on a different class or in combination with another class.

In the picture, the text should go all the way to the bottom, but it turns to scroll at half the height.

image

src/api/status/public/css/terminal.css
```diff
@@ -1,3 +1,6 @@
 .xterm {
   padding: 0.5rem;
+  position: relative;
+  width: 100vh;
```
Andrewnt219, line 4: These two doesn't seem to be necessary. `vh` should not be used for width as well.

Andrewnt219 (changes requested):

src/api/status/public/css/terminal.css
Yoda-Canada: I deleted the position and width properties. I know why the text turns to scroll at half the height. I think you have two screen and you changed the screen from one to another. When we change the screen, we need to refresh it.

Andrewnt219: Yeah you're right.

Andrewnt219 (approved):

humphd (approved):

PR #2565, opened 2021-12-04
#2554 Search page re-renders on entering input 

### LINKED ISSUE #2554: Search page re-renders on entering input
**What happened**:
When entering any text inside the search bar, the search page re-renders and focus is removed from the search bar.

When entering any text, it should keep focus so that you can continue entering your search text. This is a render issue, most likely a variable/function introduced in one of our recent PRs that caused this issue.

The issue is caused by the removal of the `autoFocus` property in [this recent commit](https://github.com/Seneca-CDOT/telescope/commit/d8a7f3330d670ec89ca499d69d94055133f46feb). Without that property, upon entering a character, our `onChange` event handler will cause a useEffect re-render from our SearchProvider, thus, the search page itself will be rendered again and the search input text field loses focus.

drew5494:
This PR closes #2554 

## Type of Change

## Description

When entering any text into search, it should keep focus so that you can continue entering your search text. The <code>autofocus</code> attribute is not recommended by ESLint, unless the page is mainly used for searching, as it is in this case. See [The accessibility of HTML 5 autofocus](https://brucelawson.co.uk/2009/the-accessibility-of-html-5-autofocus/).

## Checklist

### DIFF
--- a/src/web/src/components/SearchInput/AuthorSearchInput.tsx
+++ b/src/web/src/components/SearchInput/AuthorSearchInput.tsx
@@ -35,13 +35,10 @@ const AuthorSearchInput = () => {
       <input
         autoFocus
         className={classes.input}
-        list="search-suggestions"
         placeholder="How to contribute to Open Source"
         value={text}
         onChange={(event) => onTextChange(event.target.value)}
       />
-
-      <datalist aria-label="search telescope" id="search-suggestions" />
     </>
   );
 };
--- a/src/web/src/components/SearchInput/SearchInput.tsx
+++ b/src/web/src/components/SearchInput/SearchInput.tsx
@@ -5,7 +5,7 @@ import useSearchValue from '../../hooks/use-search-value';
 const SearchInput = () => {
   const { filter } = useSearchValue();
 
-  return filter === 'author' ? <AuthorSearchInput /> : <PostSearchInput />;
+  return <>{filter === 'author' ? <AuthorSearchInput /> : <PostSearchInput />}</>;
 };
 
 export default SearchInput;


src/api/status/env.local
```diff
@@ -1,3 +1,7 @@
 PLANET_PORT=1111
```
HyperTHD, line 1: This file is unrelated to PR's intended fix, remove this.

src/api/status/src/server.js
```diff
@@ -44,4 +44,4 @@ if (process.env.PATH_PREFIX) {
 }
 
 const port = parseInt(process.env.STATUS_PORT || 1111, 10);
```
HyperTHD, line 46: Same with this, remove this file

src/web/src/components/SearchInput/AuthorSearchInput.tsx
```diff
@@ -33,6 +33,8 @@ const AuthorSearchInput = () => {
   return (
     <>
       <input
+        // eslint-disable-next-line jsx-a11y/no-autofocus
```
HyperTHD, line 36: We removed the `autoFocus` property specifically because of the eslint fix that came in. The solution for this will have to come from a different source rather than adding the autoFocus prop back in. Remove this comment and the autoFocus prop from here.

src/web/src/components/SearchInput/PostSearchInput.tsx
```diff
@@ -35,6 +35,8 @@ const PostSearchInput = () => {
   return (
     <>
       <input
+        // eslint-disable-next-line jsx-a11y/no-autofocus
```
HyperTHD, line 38: Same thing here, remove the comment and the autoFocus prop

HyperTHD (changes requested):
Alright, lots of work gonna be needed to fix this. Aside from the files being added that aren't supposed to be here, the fix you added is one we've decided to not add back anymore. You can read [this link](https://stackoverflow.com/questions/2180645/is-automatically-assigning-focus-bad-for-accessibility) to see why it's not a good idea.

drew5494:
Alright next time close the issue first. 

> Alright, lots of work gonna be needed to fix this. Aside from the files being added that aren't supposed to be here, the fix you added is one we've decided to not add back anymore. You can read [this link](https://stackoverflow.com/questions/2180645/is-automatically-assigning-focus-bad-for-accessibility) to see why it's not a good idea.

As stated it may be acceptable for pages used for searching: "It's clear why your focus winds up in the Google search box so that doesn't bother me"
Are you still looking for an alternative to autofocus ?

humphd:
@drew5494 we recently had to move away from using `autoFocus`, so it would be good to find a way to accomplish this that doesn't rely on it.  However, I'd have to dig up the PR that did it (it might have been `eslint` related)?

It's interesting to see that Google uses it:

```html
<input class="gLFyf gsfi" jsaction="paste:puy29d;" maxlength="2048" name="q" type="text" aria-autocomplete="both" aria-haspopup="false" autocapitalize="none" autocomplete="off" autocorrect="off" autofocus="" role="combobox" spellcheck="false" title="Search" value="" aria-label="Search" data-ved="0ahUKEwjz1dCF7sr0AhXbG80KHc4SAIAQ39UDCAY">
```

So maybe we can make eslint happy *and* do this?  Not sure.

DukeManh:
This is introduced in my eslint PR that removed `autofocus` from input because of [Eslint's accessibility recommendations](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-autofocus.md). 

I have investigated a bit, the problem is here in `searchInput`:
https://github.com/Seneca-CDOT/telescope/blob/78f3324cf79f253db17a33f832e99a6e5072a4c7/src/web/src/components/SearchInput/SearchInput.tsx#L8

We have a `useSearchValue` hook to get the search `filter` and `value`, when the input changes, all components that subscribe to the hook re-renders, hence`searchInput` rerenders and returns a new component every time, either `AuthorInput` or `PostsInput`, which makes the input lose focus.

If we move the conditional into a JSX expression to make sure it returns the same component, the problem should go away.

```js
  return <>{filter === 'author' ? <AuthorSearchInput /> : <PostSearchInput />}</>;
```

src/web/src/components/SearchInput/AuthorSearchInput.tsx
```diff
@@ -33,6 +33,8 @@ const AuthorSearchInput = () => {
...
+        autoFocus
         className={classes.input}
         list="search-suggestions"
```
DukeManh, line 39: While you are, can you remove this `list` attribute and the `data-list` field below? They're used for input suggestions but we don't have any.

DukeManh (changes requested):

src/web/src/components/SearchInput/AuthorSearchInput.tsx
drew5494: sure @DukeManh

src/api/status/env.local
```diff
@@ -1,3 +1,7 @@
...
 PATH_PREFIX=/v1/status
+<<<<<<< HEAD
```
humphd, line 3:
```
git checkout master src/api/status/env.local
```

src/api/status/src/server.js
humphd:
```
git checkout master src/api/status/src/server.js
```

src/web/src/components/SearchInput/SearchInput.tsx
```diff
@@ -5,7 +5,7 @@ import useSearchValue from '../../hooks/use-search-value';
 const SearchInput = () => {
   const { filter } = useSearchValue();
 
-  return filter === 'author' ? <AuthorSearchInput /> : <PostSearchInput />;
+  return <>{filter === 'author' ? <AuthorSearchInput /> : <PostSearchInput />}</>;
```
humphd, line 8: Can you help me understand more about this change?  What does wrapping in an element do for us in this case that returning the component won't?

humphd (changes requested):

src/web/src/components/SearchInput/SearchInput.tsx
DukeManh: @humphd, `filter === 'author' ? <AuthorSearchInput /> : <PostSearchInput />;` is like an function that's called in every re-render, and its return result in unknown. When either `AuthorSearchInput` or `PostSearchInput` is returned, the current input is unmounted and re-rendered so the input loses focus without `autofoucs` attribute.


`<>{filter === 'author' ? <AuthorSearchInput /> : <PostSearchInput />}</>` is a React Component that's has conditional rendering, `AuthorSearchInput` and `PostSearchInput` won't re-mount if it doesn't change.

I believe that's how React works, let me tag Dan Abramov to confirm :).

DukeManh:
@drew5494 , can you rebase to pickup the new changes?

src/web/src/components/SearchInput/SearchInput.tsx
humphd: In both cases the React component tree will be different, whether at the root or child level. "filter === 'author' ? <AuthorSearchInput /> : <PostSearchInput />; is like an function" I don't think so.  It's the same expression you're evaluating in `<>{filter === 'author' ? <AuthorSearchInput /> : <PostSearchInput />}</>`

```js
function ComponentA() { return <div></div>; }
function ComponentB() { return <div></div>; }

const x = 1;

const element = <>{x === 1 ? <ComponentA /> : <ComponentB />}</>;
// var element = /*#__PURE__*/React.createElement(React.Fragment, null, x === 1 ? /*#__PURE__*/React.createElement(ComponentA, null) : /*#__PURE__*/React.createElement(ComponentB, null));

const element2 = x === 1 ? <ComponentA /> : <ComponentB />;
// var element2 = x === 1 ? /*#__PURE__*/React.createElement(ComponentA, null) : /*#__PURE__*/React.createElement(ComponentB, null);
```

If there's a real difference when wrapping in a fragment, related to re-rendering, I'd like to understand what it is. Both are going to evaluate to, and produce the same DOM.

humphd: Are you saying that the fragment provides stability between re-renders in order to maintain focus?

DukeManh: > Are you saying that the fragment provides stability between re-renders in order to maintain focus?

Yes, I mean that.
The difference is that the React fragment prevents unmounting, both will produce the same DOM, but the first one doesn't re-mount and the second one does on every re-render.

humphd: OK, that makes sense.

drew5494:
> @drew5494 , can you rebase to pickup the new changes?

Sure. Once I have a chance to figure this out.

DukeManh:
@drew5494  
```
git fetch upstream master
git rebase upstream/master
```

TueNguyen2911:
Looks like you still have 2 commits, rebase is not correct

DukeManh:
@TueNguyen2911 it's ok, don't have to squash during rebase, we can squash when this is merged.

TueNguyen2911:
> @TueNguyen2911 it's ok, don't have to squash during rebase, we can squash when this is merged.

Aah, that's good then

humphd (dismissed):
This is a nice fix.  Works great on Vercel.

DukeManh (dismissed):

humphd:
@drew5494 can you squash and rebase this for me so I can merge?

DukeManh:
@drew5494, your branch is still out-of-date, I wonder if you have the right remote URL.
```
git remote -v
git remote set-url upstream https://github.com/Seneca-CDOT/telescope.git
git fetch upstream master
git rebase upstream/master
```

humphd:
Rebased via GitHub.

PR #2573, opened 2021-12-06
Fix prettier-check issues on master

humphd:
We have a few prettier issues on `master`:

1. we shouldn't be running lint on pre commit, it's messing with prettier.  I think this is what is messing up people's prettier commits, since eslint and prettier are fighting each other
2. we are using deprecated prettier config
3. we aren't ignore generated workbox scripts

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -9,3 +9,5 @@ out
 
 # We don't maintain these files
 src/api/status/public/assets
+src/web/public/sw*
+src/web/public/workbox*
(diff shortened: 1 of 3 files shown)


humphd:
It is an interesting side-issue that eslint and prettier are not agreeing on how this should get formatted. I wonder if we have a config issue somewhere else, too.

DukeManh (approved):

Kevan-Y (approved):

PR #2575, opened 2021-12-07
Adding /feeds/delayed route to Posts service, adding test in posts/te…

### LINKED ISSUE #2413: Add /feeds/delayed route to Posts service
In #2395 we are adding an expanded status service dashboard.  One thing I want to include in it is the ability to see a list of all feeds that have been flagged as `delayed` in our system, which happens when we get an HTTP 429 back from the server. @manekenpix has done some more in #2396 to include the `/feeds` endpoint in the Posts service.

I'd like to see a new route get added to Posts:

```
GET /feeds/delayed
```

This should return a JSON array of Objects with the feed info as well as how long until it will stopped being delayed (e.g., number of seconds that we put into the db).

Doing so will require an update to how the Redis database code works in https://github.com/Seneca-CDOT/telescope/blob/master/src/backend/utils/storage.js.  Currently when we delay a feed, we add the `:delayed` suffix to the feed key and set a time for expiry in seconds.

Redis supports doing pattern matching on keys using the [`SCAN` command with `match`](https://redis.io/commands/scan).  You can see an example in JS of what this looks like at https://github.com/luin/ioredis/blob/88b8a3125cd039afdac928c529a1b9958009979c/test/functional/scan_stream.ts#L50-L78.   What we'd do is look for all keys like this:

```
t:feed:*:delayed
```

We can then use these keys to lookup the time to live (TTL) each delayed feed, see https://redis.io/commands/TTL.  We'd use this to return the feed id (and probably URL) along with the time left in the delay, either in seconds or ms or pretty formatted to a human readable time.

Later we can use this route to create a widget on the status dashboard with a table of feeds and times for being marked delayed.

hphan9:
…st/feeds.test.js, fixing the setDelayedFeeds function in posts/src/storage.js and backend/utils/storage.js

## Issue This PR Addresses
Fixes #2413 

## Type of Change

## Description

I added a new route /feeds/delayed to Posts service. I also fixed the small bug in the setDelayedFeed function to set a time for expiry in seconds. Thank you and please review my pull request.
## Checklist

### DIFF
--- a/src/api/posts/src/storage.js
+++ b/src/api/posts/src/storage.js
@@ -90,6 +90,16 @@ module.exports = {
     );
   },
 
+  getDelayedFeeds: async () => {
+    const delayedKeys = await getFeedKeysUsingScanStream(`${feedNamespace}*${delayedSuffix}`);
+    return delayedKeys.map((key) => {
+      const id = key.replace(feedNamespace, '').replace(delayedSuffix, '');
+      return {
+        id,
+      };
+    });
+  },
+
   getFlaggedFeeds: () => redis.smembers(flaggedFeedsKey),
 
   getFeed: (id) => redis.hgetall(feedNamespace.concat(id)),
@@ -127,7 +137,7 @@ module.exports = {
 
   isInvalid: (id) => redis.exists(createInvalidFeedKey(id)),
 
-  setDelayedFeed: (id, seconds) => redis.set(createDelayedFeedKey(id), seconds, 1),
+  setDelayedFeed: (id, seconds) => redis.set(createDelayedFeedKey(id), seconds, 'EX', seconds),
 
   isDelayed: (id) => redis.exists(createDelayedFeedKey(id)),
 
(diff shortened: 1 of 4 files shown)


src/api/posts/src/storage.js
```diff
@@ -90,6 +90,20 @@ module.exports = {
     );
   },
 
+  getDelayedFeeds: async () => {
+    const delayedKeys = await getFeedKeysUsingScanStream(`${feedNamespace}*${delayedSuffix}`);
+    return Promise.all(
+      delayedKeys.map(async (key) => {
+        const timeLeft = await redis.ttl(key);
```
humphd, line 97: I can't decide if we need this info or not.  The fact that it will require so many extra `async` calls to Redis makes me want to say that we don't.  Just knowing that it's delayed is probably enough?  What's your thinking?

humphd (commented):
Looks great, one question.

src/api/posts/src/storage.js
hphan9: Yeah, I think we should not have this timeLeft infor. we don't use it for anything. I will remove it and fix the test. Thank you :)

humphd (dismissed):

AmasiaNalbandian (dismissed):

humphd:
@hphan9 please squash and rebase this.

Issue #2110, opened 2021-04-09
Advanced search dialog in SearchBar 


huyxgit:
Show advanced search dialog from clicking 'gear' button in searchBar 

![Screen Shot 2021-04-09 at 2 42 32 PM](https://user-images.githubusercontent.com/10421864/114226719-ef33b780-9941-11eb-8c24-d68c8f2e7f3a.jpg)

AmasiaNalbandian:
@humphd I can take a look into this again now, I saw the last two PRs that you linked and it seems like it's been left open. Should I continue from the work there or are there new suggestions for the search?

humphd:
@AmasiaNalbandian Sure, if you like.  There are a few PRs already open on this that you could start with.

PR #2584, opened 2021-12-08
Issue-2110: Search by Date

AmasiaNalbandian:
## Issue This PR Addresses
This PR helps us move forward for [issue-2110](https://github.com/Seneca-CDOT/telescope/issues/2110)

## Type of Change

## Description
We are slowly trying to enhance the search bar. To begin, I have added a date search.
The UI for this will eventually be included in a separate modal, which can be built off [this existing PR](https://github.com/Seneca-CDOT/telescope/pull/2199). My plan is to have the [TextField with input as date](https://v4.mui.com/components/pickers/#datepickers), so this will eventually be much more user friendly. Therefore, this PR only addresses the addition of the logic for this.

chrome_H75P2I5IlX

## Checklist

### DIFF
--- a/src/api/search/src/search.js
+++ b/src/api/search/src/search.js
@@ -13,6 +13,8 @@ const createFieldsFromFilter = (filter) => {
   switch (filter) {
     case 'author':
       return ['author'];
+    case 'date':
+      return ['published'];
     case 'post':
     default:
       return ['text', 'title'];
--- a/src/web/src/components/SearchBar.tsx
+++ b/src/web/src/components/SearchBar.tsx
@@ -95,7 +95,7 @@ const SearchBar = () => {
 
   const { filter, onFilterChange, onSubmitHandler } = useSearchValue();
 
-  const searchOptions = ['post', 'author'];
+  const searchOptions = ['post', 'author', 'date'];
 
   return (
     <Box className={classes.root}>
--- a/src/web/src/components/SearchInput/DateSearchInput.tsx
+++ b/src/web/src/components/SearchInput/DateSearchInput.tsx
@@ -1,0 +1,45 @@
+import { makeStyles, createStyles, Theme } from '@material-ui/core/styles';
+import useSearchValue from '../../hooks/use-search-value';
+
+const useStyles = makeStyles((theme: Theme) =>
+  createStyles({
+    input: {
+      fontSize: '1.6rem',
+      '&:hover': {
+        borderColor: theme.palette.primary.main,
+      },
+      '&:focus': {
+        borderColor: theme.palette.primary.main,
+      },
+      '& > *': {
+        fontSize: '1.6rem !important',
+        color: theme.palette.text.primary,
+      },
+      height: '55px',
+      backgroundColor: 'transparent',
+      paddingLeft: '10px',
+      paddingRight: '60px',
+      border: 'none',
+      outline: 'none',
+      color: theme.palette.text.primary,
+    },
+  })
+);
+
+const DateSearchInput = () => {
+  const classes = useStyles();
+  const { onTextChange } = useSearchValue();
+  const today = new Date().toISOString().split('T')[0];
+
+  return (
+    <>
+      <input
+        className={classes.input}
+        placeholder={today}
+        onChange={(event) => onTextChange(event.target.value)}
+      />
+    </>
+  );
+};
+
+export default DateSearchInput;
--- a/src/web/src/components/SearchInput/SearchInput.tsx
+++ b/src/web/src/components/SearchInput/SearchInput.tsx
@@ -1,11 +1,23 @@
 import PostSearchInput from './PostSearchInput';
 import AuthorSearchInput from './AuthorSearchInput';
+import DateSearchInput from './DateSearchInput';
 import useSearchValue from '../../hooks/use-search-value';
 
 const SearchInput = () => {
   const { filter } = useSearchValue();
 
-  return filter === 'author' ? <AuthorSearchInput /> : <PostSearchInput />;
+  function getComponent(f: string) {
+    switch (f) {
+      case 'author':
+        return <AuthorSearchInput />;
+      case 'date':
+        return <DateSearchInput />;
+      default:
+        return <PostSearchInput />;
+    }
+  }
+
+  return getComponent(filter);
 };
 
 export default SearchInput;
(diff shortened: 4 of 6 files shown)


AmasiaNalbandian:
There is a known [bug](https://github.com/Seneca-CDOT/telescope/issues/2554) with the input field that is causing this branch difficulty to test when hosted.

src/web/src/components/SearchInput/DateSearchInput.tsx
```diff
@@ -0,0 +1,45 @@
...
+const DateSearchInput = () => {
+  const classes = useStyles();
+  const { onTextChange } = useSearchValue();
+  const today = new Date().toISOString().split('T')[0];
+
+  return (
+    <>
+      <input
+        className={classes.input}
+        placeholder={today}
```
HyperTHD, line 38: Probably should make the placeholder, "Enter date", instead of putting today's date

src/web/src/components/SearchProvider.tsx
```diff
@@ -35,6 +35,19 @@ const SearchContext = createContext<SearchContextInterface>({
...
 });
 
+// Returns the string in the filter accordingly after loading results.
+function getFilter(filter: any) {
+  switch (filter) {
+    case 'author':
+      return 'author';
+    case 'date':
+      return 'date';
+    case 'post':
```
HyperTHD, line 45: Since we default the search option to `'post'`, this is probably not needed and can be removed.

HyperTHD (changes requested):
I like what I'm seeing so far. It sucks that the search bug is stalling this PR, otherwise, we really got something going here! 

Aside from some stuff that I noticed you change, I'm a bit curious as to how you enter the date you want to search specifically. I assume it's in the format you mentioned in the placeholder such that the search service can check it.

src/web/src/components/SearchInput/DateSearchInput.tsx
```diff
@@ -0,0 +1,45 @@
...
+);
+
+const DateSearchInput = () => {
+  const classes = useStyles();
+  const { onTextChange } = useSearchValue();
+  const today = new Date().toISOString().split('T')[0];
+
+  return (
+    <>
+      <input
```
HyperTHD, line 36: Also, I'd like some feedback on this. What if we made the type of input here `date`?

AmasiaNalbandian: Hey thanks for the review - so for this PR I am only putting in the logic, but as mentioned I am going to be using the datepicker once we put in the advanced search dialogue:

https://v4.mui.com/components/pickers/#datepickers

This will make sure the format is not skewed up. Unfortunately if I do this in the searchbar, which is also an input I'll just end up trying to add a bunch of front end to make it look nice. The idea is that this date input will be seen in the advanced search modal that I am hoping to work on in addition to solving issue-2110.

src/web/src/components/SearchInput/DateSearchInput.tsx
AmasiaNalbandian: This is not a bad idea! I had it to the date, so that users can see what kind of format to follow.

src/web/src/components/SearchProvider.tsx
AmasiaNalbandian: Let me ping you about this, because I had some difficulty understanding the login on line 50 of the original.

src/api/search/src/search.js
```diff
@@ -13,6 +13,8 @@ const createFieldsFromFilter = (filter) => {
   switch (filter) {
     case 'author':
       return ['author'];
+    case 'date':
+      return ['published'];
```
humphd, line 17: The date seems like something you'd want to stack with other filters.  For example, I might want to search for posts by an author in a given date range.  I'm not sure that returning just `[published]` alone is all we want here.

src/web/src/components/SearchBar.tsx
```diff
@@ -95,7 +95,7 @@ const SearchBar = () => {
 
   const { filter, onFilterChange, onSubmitHandler } = useSearchValue();
 
-  const searchOptions = ['post', 'author'];
+  const searchOptions = ['post', 'author', 'date'];
```
humphd, line 98: We have 2 dates in a post, so maybe we should keep the wording "published" somehow?

src/web/src/components/SearchInput/DateSearchInput.tsx
```diff
@@ -0,0 +1,45 @@
...
+      outline: 'none',
+      color: theme.palette.text.primary,
+    },
+  })
+);
+
+const DateSearchInput = () => {
+  const classes = useStyles();
+  const { onTextChange } = useSearchValue();
+  const today = new Date().toISOString().split('T')[0];
```
humphd, line 32: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat

src/web/src/components/SearchInput/SearchInput.tsx
```diff
@@ -1,11 +1,23 @@
 import PostSearchInput from './PostSearchInput';
 import AuthorSearchInput from './AuthorSearchInput';
+import DateSearchInput from './DateSearchInput';
 import useSearchValue from '../../hooks/use-search-value';
 
 const SearchInput = () => {
   const { filter } = useSearchValue();
 
-  return filter === 'author' ? <AuthorSearchInput /> : <PostSearchInput />;
```
humphd, line 8: Let's rebase to pick up the change that @drew5494 just added to fix the search autofocus problem.  I don't want to re-introduce it.

As an aside (@DukeManh), the fix for that feels really brittle, and easy to break again.  We probably need a comment to help future devs not regress.

src/web/src/components/SearchInput/SearchInput.tsx
```diff
@@ -1,11 +1,23 @@
...
+  function getComponent(f: string) {
```
humphd, line 9: Why bother with a function you immediately call?  Just inline the function body.

src/web/src/components/SearchInput/SearchInput.tsx
```diff
@@ -1,11 +1,23 @@
...
+    switch (f) {
```
humphd, line 10: Let's use full words vs. `f`.

humphd (changes requested):
This needs a test.  I'm not clear how to use it, and a test would help us prove how it works.

Question: I think it's unlikely that I'll know the exact date that I want to search for; I'm more likely to have a range.  I know that Elastic supports range queries, and has some cool ways to do offset searches.  Can we explore that, either here or in a follow up?

AmasiaNalbandian:
@humphd I think we can close this PR. The changes are going to be incoming on the front end in the advanced search only. and I will keep the feedback for that one.

humphd:
Perfect.

Issue #2553, opened 2021-12-01
Add a template engine for status dashboard


Andrewnt219:
**What would you like to be added**:
A template engine somewhat like this

```handlebars
<!-- side-bar.html -->

<nav>
  Sidebar content
</nav>
```

```handlebars
<!-- index.html -->

{{> sidebar }}

<main>
  Main content 

  <ul>
  {{#each items}}
    <li>{{this.title}}</li>
  {{/each}}
 </ul>
</main>
```

At the moment, I'm looking at the good old [handlebars](https://handlebarsjs.com) from WEB322. It's the closest to HTML and has a [simple compiler setup](https://handlebarsjs.com/installation/precompilation.html#getting-started) like SCSS.

**Why would you like this to be added**:
Two main reasons:
- The duplicate of HTML: With only two active pages, we already have a problem with different sidebar on both pages in #2505 
- The verbose code to render HTML from JS in #2551

humphd:
Sounds like a good idea to me.

Kevan-Y:
Maybe we can separate into several issues maybe, one to setup handlebars, other to add components

DukeManh:
Let's eventually use React to write it 😃.

Kevan-Y:
> Let's eventually use React to write it 😃.

That what I was thinking too using Next.Js.

humphd:
@DukeManh I'm hoping to use a different set of technologies so students not as interested in React have a place to work as well.  We have all the Next.js we can handle already :)

Kevan-Y:
I can try and work on it.

PR #2585, opened 2021-12-08
Added jobs count to dashboard

TueNguyen2911:
## Issue This PR Addresses

This fixes issue #2414 which is also a follow up of PR #2541 

## Type of Change

- [ ] **Bugfix**: Change which fixes an issue
- [x] **New Feature**: Change which adds functionality
- [ ] **Documentation Update**: Change which improves documentation
- [ ] **UI**: Change which improves UI

## Description

- Added `jobCountCard.hbs` to display jobs in feeds queue in the backend 
- Added to `src/api/status/src/js` a new file `queue-stats.js` to fetch queue's information using posts API `posts/feeds/info`

### To test this feature: 

- Added `*` or `http://localhost/v1/posts/feeds/info` to `connectSrc` in [server.js](https://github.com/Seneca-CDOT/telescope/blob/master/src/api/status/src/server.js)
- Change the fetch url in `queue/index.js` to `http://localhost/v1/posts/feeds/info`
- Run `pnpm run services:start posts` in root folder and `npm run dev` in `api/status`, you'll get `feeds/info` initial response 
- (Optional) Run all services `pnpm run services:start` and `pnpm start` to start the backend feeds queue. Reload to see changes

https://user-images.githubusercontent.com/58532267/145457103-3a3643b1-d981-4949-8f42-388254193405.mp4

## Checklist

- [ ] **Quality**: This PR builds and passes our npm test and works locally
- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [x] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not (if applicable)
- [ ] **Documentation**: This PR includes updated/added documentation to user exposed functionality or configuration variables are added/changed or an explanation of why it does not(if applicable)

PR #2590, opened 2021-12-10
Fix fetch url between microservices + Fix sidebar root path links + cache github info

### LINKED ISSUE #2592: Broken url when trying to go back to dashboard from build page on staging
**What happened**:
When on the build log page on the dev view for the status page and trying to go back to the dashboard, it leads to an empty page that doesn't exist.

image
image

Kevan-Y:
## Issue This PR Addresses
Related #2553 
Fixes #2592

## Type of Change

## Description
- Add `STATUS_URL`, `POSTS_URL`, `POSTS_PORT` environment for status microservice
- Using `STATUS_URL` to check if production or staging.
- Change fetch URL for the total post, total feed.
- Change sidebar root link from `/` to `/v1/status`

### Option 1
1. `pnpm run services:start`
2. Go to the link `http://localhost/v1/status`

### Option 2
1. `pnpm run services:start posts`
2. `cd .\src\api\status\`
3. `pnpm run dev`
4. Go to the link `http://localhost:1111/v1/status/`
5. Add cache for github info (cache for 10 min)

## Checklist

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -19,6 +19,7 @@ services:
       - NODE_ENV
       - API_HOST
       - STATUS_PORT
+      - STATUS_URL
       # Satellite authentication/authorization support
       - JWT_ISSUER
       - JWT_AUDIENCE
--- a/src/api/status/env.local
+++ b/src/api/status/env.local
@@ -1,4 +1,5 @@
 PLANET_PORT=1111
 PATH_PREFIX=/v1/status
 API_HOST=localhost
-API_URL= http://localhost:3000
+POSTS_URL=http://localhost/v1/posts
+MOCK_REDIS=1
--- a/src/api/status/src/js/feed-stats.js
+++ b/src/api/status/src/js/feed-stats.js
@@ -1,14 +1,11 @@
 const { fetch } = require('@senecacdot/satellite');
 
 module.exports = async function getFeedCount() {
-  let feedCount = 0;
   try {
-    const data = await fetch(`${process.env.API_URL}/v1/posts/feeds`, {
-      method: 'HEAD',
-    });
-    feedCount = data.headers.get('x-total-count');
+    const data = await fetch(`${process.env.POSTS_URL}/feeds`, { method: 'HEAD' });
+    return data.headers.get('x-total-count');
   } catch (err) {
     console.error(err);
   }
-  return feedCount;
+  return 0;
 };
--- a/src/api/status/src/js/github-stats.js
+++ b/src/api/status/src/js/github-stats.js
@@ -1,74 +1,108 @@
 const { fetch } = require('@senecacdot/satellite');
 
+const redis = require('../redis');
+
+const cacheTime = 60 * 10; // 10 mins of cache time
+
 const fetchGitHubApi = (owner, repo, path) =>
   fetch(`https://api.github.com/repos/${owner}/${repo}/${path}`, {
     headers: { Accept: 'application/vnd.github.v3+json' },
   });
 
-module.exports = async function getGitHubData(owner, repo) {
-  const returnedValue = {};
-
+const getStatsParticipation = async (owner, repo) => {
   // get weekly commits for last year: https://docs.github.com/en/rest/reference/repos#get-the-weekly-commit-count
-  try {
-    const res = await fetchGitHubApi(owner, repo, 'stats/participation');
-    const participationResponse = await res.json();
+  const res = await fetchGitHubApi(owner, repo, 'stats/participation');
+  const participationResponse = await res.json();
+  if (!res.ok) {
+    throw new Error(`[Code ${res.status}] - ${participationResponse.message}`);
+  }
 
-    if (!res.ok) throw new Error(`[Error ${res.status}] - ${participationResponse.message}`);
-
-    returnedValue.weeklyCommits = {
+  return {
+    weeklyCommits: {
       commits: participationResponse.all[participationResponse.all.length - 1],
-    };
-    returnedValue.yearlyCommits = {
+    },
+    yearlyCommits: {
       commits: participationResponse.all.reduce((a, b) => a + b),
-    };
-  } catch (error) {
-    console.error(error);
-  }
+    },
+  };
+};
 
+const getStatsCodeFrequency = async (owner, repo) => {
   // get weekly commits activity: https://docs.github.com/en/rest/reference/repos#get-the-weekly-commit-activity
-  try {
-    const res = await fetchGitHubApi(owner, repo, 'stats/code_frequency');
-    const codeFrequencyResponse = await res.json();
+  const res = await fetchGitHubApi(owner, repo, 'stats/code_frequency');
+  const codeFrequencyResponse = await res.json();
 
-    if (!res.ok) throw new Error(`[Error ${res.status}] - ${codeFrequencyResponse.message}`);
-
-    const [, linesAdded, linesRemoved] = codeFrequencyResponse[codeFrequencyResponse.length - 1];
-    returnedValue.weeklyCommits = { linesAdded, linesRemoved, ...returnedValue.weeklyCommits };
-  } catch (error) {
-    console.error(error);
+  if (!res.ok) {
+    throw new Error(`[Code ${res.status}] - ${codeFrequencyResponse.message}`);
   }
 
+  const [, linesAdded, linesRemoved] = codeFrequencyResponse[codeFrequencyResponse.length - 1];
+  return { weeklyCommits: { linesAdded, linesRemoved } };
+};
+
+const getCommitsInfo = async (owner, repo) => {
   // get latest author from commits list: https://docs.github.com/en/rest/reference/repos#list-commits
-  try {
-    const res = await fetchGitHubApi(owner, repo, 'commits');
-    const commitResponse = await res.json();
+  const res = await fetchGitHubApi(owner, repo, 'commits');
+  const commitResponse = await res.json();
 
-    if (!res.ok) throw new Error(`[Error ${res.status}] - ${commitResponse.message}`);
-
-    const lastCommitResponse = commitResponse[0];
-    returnedValue.avatar = lastCommitResponse.author.avatar_url;
-    returnedValue.username = lastCommitResponse.author.login;
-    returnedValue.fullName = lastCommitResponse.commit.author.name;
-    returnedValue.commitURL = lastCommitResponse.html_url;
-    returnedValue.shortSha = lastCommitResponse.sha.substr(0, 7);
-    returnedValue.usernameURL = `https://github.com/${lastCommitResponse.author.login}`;
-  } catch (error) {
-    console.error(error);
+  if (!res.ok) {
+    throw new Error(`[Code ${res.status}] - ${commitResponse.message}`);
   }
 
+  const lastCommitResponse = commitResponse[0];
+
+  return {
+    avatar: lastCommitResponse.author.avatar_url || '',
+    username: lastCommitResponse.author.login || '',
+    fullName: lastCommitResponse.commit.author.name,
+    commitURL: lastCommitResponse.html_url,
+    shortSha: lastCommitResponse.sha.substr(0, 7),
+    usernameURL: `https://github.com/${lastCommitResponse.author.login || ''}`,
+  };
+};
+
+const getContributorsInfo = async (owner, repo) => {
   // get total contributors: https://docs.github.com/en/rest/reference/repos#list-repository-contributors
-  try {
-    const res = await fetchGitHubApi(owner, repo, 'contributors?per_page=1');
+  const res = await fetchGitHubApi(owner, repo, 'contributors?per_page=1');
 
-    if (res.status >= 400) throw new Error(`[Error ${res.status}] - ${(await res.json()).message}`);
-
-    const contributorsResponse = await res.headers.get('link');
-    [, returnedValue.totalContributors] = contributorsResponse.match(
-      /.*"next".*&page=([0-9]*).*"last".*/
-    );
-  } catch (error) {
-    console.error(error);
+  if (res.status >= 400) {
+    throw new Error(`[Code ${res.status}] - ${(await res.json()).message}`);
   }
 
-  return returnedValue;
+  const contributorsResponse = await res.headers.get('link');
+  const [, totalContributors] = contributorsResponse.match(/.*"next".*&page=([0-9]*).*"last".*/);
+
+  return {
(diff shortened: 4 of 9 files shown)


Kevan-Y:
Waiting #2585

docker/docker-compose.yml
```diff
@@ -19,6 +19,9 @@ services:
       - NODE_ENV
       - API_HOST
       - STATUS_PORT
+      - STATUS_URL
+      - POSTS_URL=http://posts
+      - POSTS_PORT=5555
```
humphd, line 24: You shouldn't need the port.  That's only for internal use.  You should access Posts via our microservice router (Traefik).  Every other service that uses the Posts service just works with the public `POSTS_URL`, for example: https://github.com/Seneca-CDOT/telescope/blob/master/config/env.development#L131 and https://github.com/Seneca-CDOT/telescope/blob/master/config/env.production#L121.

docker/docker-compose.yml
```diff
@@ -19,6 +19,9 @@ services:
       - NODE_ENV
       - API_HOST
       - STATUS_PORT
+      - STATUS_URL
+      - POSTS_URL=http://posts
```
humphd, line 23: This shouldn't need to be defined, just forward what is already in the environment.

src/api/status/src/js/feed-stats.js
```diff
@@ -1,11 +1,15 @@
...
+
 module.exports = async function getFeedCount() {
   let feedCount = 0;
   try {
-    const data = await fetch(`${process.env.API_URL}/v1/posts/feeds`, {
-      method: 'HEAD',
-    });
+    const API_URL = POSTS_URL
+      ? `${POSTS_URL}:${POSTS_PORT}/feeds`
+      : `http://${process.env.API_HOST}/v1/posts/feeds`;
```
humphd, line 10: You should be able to use `POSTS_URL` in the backend without having to play with it.  For local dev, change what's defined in `env.local`.

src/api/status/src/js/feed-stats.js
```diff
@@ -1,11 +1,15 @@
...
 
+const { POSTS_URL, POSTS_PORT } = process.env;
+
 module.exports = async function getFeedCount() {
   let feedCount = 0;
   try {
-    const data = await fetch(`${process.env.API_URL}/v1/posts/feeds`, {
-      method: 'HEAD',
-    });
+    const API_URL = POSTS_URL
```
humphd, line 8: I don't think you should redefine what `API_URL` means, since we use it everywhere to mean something different than this.

src/api/status/src/server.js
```diff
@@ -77,17 +77,17 @@ service.router.get(process.env.PATH_PREFIX || '/', async (req, res) => {
   let totalFeeds;
 
   try {
-    [telescope, satellite, totalPost, totalFeeds] = await Promise.all([
-      getGitHubData('Seneca-CDOT', 'telescope'),
-      getGitHubData('Seneca-CDOT', 'satellite'),
+    [totalPost, totalFeeds, telescope, satellite] = await Promise.all([
```
humphd, line 80: Catching up on your discussion on Slack last night about rate limiting, what I'd do is separate this into a few files.  You want the HTTP route to work with data, but not worry about where it comes from.  That data should either be cached (in Redis) or live from GitHub, and you can combine it all into 1 object.

```js
// a single method returns an object with all the data we need (totalPosts, totalFeeds, etc.)
const telescopeStats = await getTelescopeData();
```

The `getTelescopeData()` function can look like this:

```js
const { redis } = require('@senecacdot/satellite');

const totalPosts = async () => ({
  // Step 1: see if we have this value stored in Redis already
  const cached = await redis.get('github:totalPosts');
  if(cached) {
    return cached;
  }

  // Step 2. get the value from GitHub, cache, and return
  const res = await fetch(...);
  // ...
  const totalPosts = ...;
  // cache this value for 10 minutes before we need to get it again (arbitrary, could be more or less)
  const tenMinutesInSeconds = 10 * 60
  redis.set('github:totalPosts', totalPosts, 'EX', tenMinutesInSeconds);

  return totalPosts;
});

// Caller needs to deal with errors (wrap in try/catch)
const getTelescopeData = async () => ({
  totalPosts: await getTotalPosts(),
  totalFeeds: await totalFeeds(), 
  /* and so on... */
});
```

You can streamline this, but that's the idea.

humphd (changes requested):

docker/docker-compose.yml
Kevan-Y: But in the environment `POSTS_URL` is `http://localhost/v1/posts` but I need to call `http://posts` to be able to talk with post microservices.

src/api/status/src/js/feed-stats.js
Kevan-Y: Noted, will change to another variable names

docker/docker-compose.yml
humphd: OK, please do it like this then https://github.com/Seneca-CDOT/telescope/blob/master/docker/development.yml#L107

docker/docker-compose.yml
humphd: Understood.  See my comment above.  Let's do what planet does.

src/api/status/src/server.js
Kevan-Y: Let me experiment with it. This would be my first time using Redis.

docker/docker-compose.yml
```diff
@@ -19,6 +19,8 @@ services:
       - NODE_ENV
       - API_HOST
       - STATUS_PORT
+      - STATUS_URL
+      - POSTS_URL=http://posts:5555
```
humphd, line 23: This belongs in the development docker-compose, not in the base one.  The only place we have the localhost issue is running locally.  Same idea as here https://github.com/Seneca-CDOT/telescope/blob/master/docker/development.yml#L107.

src/api/status/src/server.js
Kevan-Y: Added cache with redis but I'm only able to run it with  `pnpm run services:start`.
If I do `pnpm run dev` I'm getting 
```
[ioredis] Unhandled error event: Error: getaddrinfo ENOTFOUND redis
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:71:26)
 ```

File: src/api/status/src/js/feed-stats.js [hunk 7]
```diff
@@ -3,9 +3,8 @@ const { fetch } = require('@senecacdot/satellite');
 module.exports = async function getFeedCount() {
   let feedCount = 0;
   try {
-    const data = await fetch(`${process.env.API_URL}/v1/posts/feeds`, {
-      method: 'HEAD',
-    });
+    const postsFeedURL = `${process.env.POSTS_URL || `${process.env.API_URL}/posts`}/feeds`;
```
humphd, line 6: Let's not do this.  We should be able to rely on `POSTS_URL` in all cases.  If you need to pass it in for `npm` based startup, put it in `env.local`

src/api/status/src/js/github-stats.js
```diff
@@ -1,74 +1,124 @@
...
   fetch(`https://api.github.com/repos/${owner}/${repo}/${path}`, {
     headers: { Accept: 'application/vnd.github.v3+json' },
   });
 
-module.exports = async function getGitHubData(owner, repo) {
-  const returnedValue = {};
+const statsParticipation = async (owner, repo) => {
+  // Get data from cached data
+  const cached = await redis.get(`github:statsParticipation-${repo}`);
+  if (cached) return JSON.parse(cached);
```
humphd, line 13: Always brace `if` statement bodies.

src/api/status/src/js/github-stats.js
```diff
@@ -1,74 +1,124 @@
...
-    };
-    returnedValue.yearlyCommits = {
-      commits: participationResponse.all.reduce((a, b) => a + b),
-    };
-  } catch (error) {
-    console.error(error);
-  }
+  const res = await fetchGitHubApi(owner, repo, 'stats/participation');
+  const participationResponse = await res.json();
+  const returnedValue = {};
```
humphd, line 18: Don't bother with this variable, do this at the end:

```js
return {
  weeklyCommits: {
    commits: participationResponse.all[participationResponse.all.length - 1],
  },
  yearlCommits: {
    commits: participationResponse.all.reduce((a, b) => a + b),
  },
};

File: src/api/status/src/js/github-stats.js [hunk 10]
```diff
@@ -1,74 +1,124 @@
...
+  if (!res.ok) throw new Error(`[Code ${res.status}] - ${participationResponse.message}`);
```
humphd, line 19: Brace this please.

src/api/status/src/js/github-stats.js
```diff
@@ -1,74 +1,124 @@
...
+  };
+  returnedValue.yearlyCommits = {
+    commits: participationResponse.all.reduce((a, b) => a + b),
+  };
+
+  await redis.set(
+    `github:statsParticipation-${repo}`,
+    JSON.stringify(returnedValue),
+    'EX',
+    10 * 60
```
humphd, line 32: No magic numbers.  Please create a variable for this that specifies what it is, and the units (seconds).

src/api/status/src/js/github-stats.js
```diff
@@ -1,74 +1,124 @@
...
+  );
+
+  return returnedValue;
+};
+
+const statsCodeFrequency = async (owner, repo) => {
+  // Get data from cached data
+  const cached = await redis.get(`github:statsCodeFrequency-${repo}`);
```
humphd, line 40: Quite a bit of duplication here with the previous function.  Can you refactor out the common bits and only do what's different in each?

src/api/status/src/js/github-stats.js
```diff
@@ -1,74 +1,124 @@
...
+  );
+
+  await redis.set(`github:contributorsInfo-${repo}`, JSON.stringify(returnedValue), 'EX', 10 * 60);
 
   return returnedValue;
 };
+
+module.exports = async function getGitHubData(owner, repo) {
+  const [statsParticipation_, statsCodeFrequency_, commitsInfo_, contributorsInfo_] =
+    await Promise.all([
```
humphd, line 111: `try/catch` or you'll crash the server when this fails

humphd (changes requested):
Nice changes.  Some more things to fix.

src/api/status/src/server.js
humphd: Correct, since you have a dependency on redis now.

In your `env.local` add `MOCK_REDIS=1` to mock it in development, see https://github.com/Seneca-CDOT/satellite/blob/d42a831a8ca6666dd5759eeccdbe932fe34239ca/src/redis.js#L9

src/api/status/src/js/github-stats.js
Kevan-Y: But I still need this variable to pass to `redis.set`

src/api/status/src/js/github-stats.js
Kevan-Y: Noted

src/api/status/src/js/feed-stats.js
Kevan-Y: Fixed

docker/docker-compose.yml
Kevan-Y: Moved to development.yml.

src/api/status/src/js/github-stats.js
humphd: ah,, gotcha.  I'm reviewing too quickly, apologies.

src/api/status/src/js/github-stats.js
Kevan-Y: Refactored now it's one caching.

src/api/status/src/js/github-stats.js
Kevan-Y: Fixed

src/api/status/env.local
```diff
@@ -1,4 +1,5 @@
 PLANET_PORT=1111
 PATH_PREFIX=/v1/status
 API_HOST=localhost
-API_URL= http://localhost:3000
+POSTS_URL=http://localhost/v1/posts
+MOCK_REDIS=1
```
humphd, line 5: This is probably fine.  The only risk of doing this is that we won't be able to connect to a live Redis instance when running this way, but I think that's acceptable, since you can do it via docker.

src/api/status/src/js/feed-stats.js
```diff
@@ -3,9 +3,7 @@ const { fetch } = require('@senecacdot/satellite');
 module.exports = async function getFeedCount() {
   let feedCount = 0;
```
humphd, line 4: Get rid of this and return on line 7 and again in the catch.

humphd (approved):
One nit, but otherwise this looks great.  I haven't tested the code, only read it.

src/api/status/src/js/feed-stats.js
Kevan-Y: I think if I do that I'm getting this 
image
Same thing for getFeedCount()

humphd:
```js
  try {
    const data = await fetch(`${process.env.API_URL}/v1/posts/feeds`, {
      method: 'HEAD',
    });
    const data = await fetch(`${process.env.POSTS_URL}/feeds`, { method: 'HEAD' });
    return data.headers.get('x-total-count');
  } catch (err) {
    console.error(err);
    return 0;
  }
```

humphd (dismissed):

src/api/status/src/js/github-stats.js
```diff
@@ -1,74 +1,108 @@
...
+      ]);
 
-    const contributorsResponse = await res.headers.get('link');
-    [, returnedValue.totalContributors] = contributorsResponse.match(
-      /.*"next".*&page=([0-9]*).*"last".*/
-    );
-  } catch (error) {
-    console.error(error);
-  }
+    returnGithubInfo = {
```
humphd, line 96: You call this function, `getGitHubData` so let's call this `githubData` or just `data`.  An object shouldn't have a verb as part of its name (e.g., no `return`).  This is `data` and we'll `return` it later.

humphd (changes requested):
Looks good, one small nit.

src/api/status/src/js/github-stats.js
Kevan-Y: Changed

humphd (approved):
👍

manekenpix (approved):

PR #2581, opened 2021-12-08
Add backend and API functionality for uploaded videos

dbelokon:
## Issue This PR Addresses

#1026, since this issue has a lot of features involved, this PR will not cover all of them.

## Type of Change

- [ ] **Bugfix**: Change which fixes an issue
- [X] **New Feature**: Change which adds functionality
- [ ] **Documentation Update**: Change which improves documentation
- [ ] **UI**: Change which improves UI

## Description

This PR focuses on the backend for tracking and processing feeds of video channels.

The frontend is addressed in PR #2596.

## Checklist

- [ ] **Quality**: This PR builds and passes our npm test and works locally
- [x] **Tests**: A test has been added to make sure that the post is of type 'video'.
- [ ] **Screenshots**: There is no PR changes.
- [ ] **Documentation**: This PR includes updated/added documentation to user exposed functionality or configuration variables are added/changed or an explanation of why it does not(if applicable)

Issue #1026, opened 2020-04-18
Feature Request: integrate the Telescope Podcast into the app


humphd:
I [wrote in my blog](https://blog.humphd.org/online-week-3/#the-telescope-podcast) about the so-called "Telescope Podcast,"

> Their calls were mostly voice-only (I was surprised to observe that they almost never used video), with either screensharing or Visual Studio Code Live Share (collaborative editing, syncing between editors)...
>
> The "Telescope Podcast" is an interesting phenomenon to me, because it's not possible to realize it using only one tool or platform.  This is a "Worse is Better" Google Wave without a single product or product owner and with loosely connected, swappable layers:
> 
> * GitHub for code sharing and review
> * Zeit Now for quick deployment and testing without local setup
> * Microsoft Teams for voice chat and screensharing
> * Visual Studio Code Live Share for remote pair/group programming
> * Slack for text chat
> * Blog posts for documenting knowledge outcomes and shared learning 

I'd like to propose that we explore the potential for making this workflow a defining aspect of Telescope itself.

## Proposal

Make it possible for Telescope to host real-time, online, collaborative development so that people can engage with both past content (blogs) and current content (voice/video streams, chat, screensharing).

I don't exactly what I mean by this, but here are the pieces that I imagine we could use and explore:

- integrate streaming into the timeline.  Maybe it's a separate page, or maybe live streams show-up at the top, sort of like what happens with Live Events in Twitter (they are pinned to the top while they happen).  Maybe we can use Twitch or YouTube Live to host this?
- integrate chat (if it's useful) so that people can ask questions of the live stream.  This might be something that Twitch/YouTube APIs enable, or maybe we can connect to Slack or Teams, or host another app to make it happen.
- use our SSO to only allow authenticated Seneca users to engage with the stream.  Maybe this means that only Seneca users can ask questions.
- is there a way to integrate a VSCode Live Share session in the browser in view-only mode?
- we we automatically record these to YouTube, and have them get hosted (as a feed) in Telescope when it's all done?

## Goals

There's an increased awareness of real-time chat, video, audio, and screensharing due to us all being at home for COVID-19.  I think some people will continue to want to create and consume edutainment this way.  Seneca is full of people who are trying hard to learn how all this technology works, and we could use Telescope as a way to show them how.

[...]

PR #2596, opened 2021-12-11
Show videos in the main timeline

dbelokon:
## Issue This PR Addresses
#1026: This will address the front end implementation for viewing videos in Telescope

## Type of Change

## Description

Display videos on the main timeline, processed by feed of a YouTube channel. This depends on PR #2581 to display any videos at all.

## Checklist

- [ ] **Tests**: This PR does not need tests because it changes the UI.
- [X] **Screenshots**: image
- [ ] **Documentation**: This PR does not documentation, as it adds UI changes.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -347,6 +347,7 @@ importers:
       react: 17.0.2
       react-dom: 17.0.2
       react-icons: 4.3.1
+      react-lite-youtube-embed: 2.1.3
       react-transition-group: 4.4.2
       react-use: 17.3.2
       smoothscroll-polyfill: 0.4.4
@@ -374,6 +375,7 @@ importers:
       react: 17.0.2
       react-dom: 17.0.2_react@17.0.2
       react-icons: 4.3.1_react@17.0.2
+      react-lite-youtube-embed: 2.1.3_react-dom@17.0.2+react@17.0.2
       react-transition-group: 4.4.2_react-dom@17.0.2+react@17.0.2
       react-use: 17.3.2_react-dom@17.0.2+react@17.0.2
       smoothscroll-polyfill: 0.4.4
@@ -2858,6 +2860,7 @@ packages:
         integrity: sha512-sdpgI/PL56QVsEJldwEe4FFaFTLUqN+rd7sSZiRCdx2E/C7z5yK0y/khAWVBH24tXwto7I1hCzNWfJGZIYJKnw==,
       }
     engines: { node: '>=8.0.0' }
+    deprecated: 'You can now upgrade to @mui/material. See the guide: https://mui.com/guides/migration-v4/'
     peerDependencies:
       '@types/react': ^16.8.6 || ^17.0.0
       react: ^16.8.0 || ^17.0.0
@@ -3738,7 +3741,7 @@ packages:
       chalk: 4.1.2
       dom-accessibility-api: 0.5.10
       lz-string: 1.4.4
-      pretty-format: 27.4.2
+      pretty-format: 27.4.6
     dev: true
 
   /@testing-library/react/12.1.2_react-dom@17.0.2+react@17.0.2:
@@ -12325,17 +12328,6 @@ packages:
       supports-color: 8.1.1
     dev: false
 
-  /jest-worker/27.4.2:
-    resolution:
-      {
-        integrity: sha512-0QMy/zPovLfUPyHuOuuU4E+kGACXXE84nRnq6lBVI9GJg5DCBiA97SATi+ZP8CpiJwEQy1oCPjRBf8AnLjN+Ag==,
-      }
-    engines: { node: '>= 10.13.0' }
-    dependencies:
-      '@types/node': 16.11.18
-      merge-stream: 2.0.0
-      supports-color: 8.1.1
-
   /jest-worker/27.4.6:
     resolution:
       {
@@ -16320,6 +16312,19 @@ packages:
         integrity: sha512-w2GsyukL62IJnlaff/nRegPQR94C/XXamvMWmSHRJ4y7Ts/4ocGRmTHvOs8PSE6pB3dWOrD/nueuU5sduBsQ4w==,
       }
 
+  /react-lite-youtube-embed/2.1.3_react-dom@17.0.2+react@17.0.2:
+    resolution:
+      {
+        integrity: sha512-pTWl7bcRnnxj1Tz0zHEn1SG0sW9+ucumF9m8zrm2qx6lYDuSPsLN+GCTuAFHn0KuKV9N3nUIvVwCfC0Zx9tNWA==,
+      }
+    peerDependencies:
+      react: '>=16.0.8'
+      react-dom: '>=16.0.8'
+    dependencies:
+      react: 17.0.2
+      react-dom: 17.0.2_react@17.0.2
+    dev: false
+
   /react-refresh/0.8.3:
     resolution:
       {
@@ -17730,6 +17735,7 @@ packages:
       {
         integrity: sha512-cPiFOTLUKvJFIg4SKVScy4ilPPW6rFgMgfuZJPNoDuMs3nC1HbMUycBoJw77xFIp6z1UJQJOfx6C9GMH80DiTw==,
       }
+    deprecated: See https://github.com/lydell/source-map-url#deprecated
     dev: false
 
   /source-map/0.5.6:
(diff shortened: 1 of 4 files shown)


src/web/src/components/Posts/Post.tsx
```diff
@@ -61,6 +62,12 @@ const useStyles = makeStyles((theme: Theme) =>
         padding: '1em 0 1em 0',
       },
     },
+    video: {
+      display: 'block',
+      margin: 'auto',
+      width: '85%',
```
humphd, line 68: Why 85% and not fill the available space in the parent container?

src/web/src/components/Posts/Post.tsx
```diff
@@ -374,6 +382,11 @@ const PostComponent = ({ postUrl, currentPost, totalPosts }: Props) => {
         </ListSubheader>
       )}
       <div className={classes.content}>
+        {isMedia && (
+          // a 'guid' from a YouTube video is usually written as 'yt:video:id'
+          <YouTube videoId={post.guid.split(':')[2]} className={classes.video} />
```
humphd, line 387: Question: what's the difference between the `guid` and `url` for  a YouTube video?

humphd: Also, let's make a function that does the id extraction, so you can name it `extractId` or something so this is self documenting.

humphd (changes requested):
This is very cool.  One other thing I'd like to see, either here or in follow-up, is to make links into HTML `<a>` elements, so they are clickable.  I can't remember the subset of HTML that YouTube lets you put in the description, but it's pretty small, and we could likely write a simple parser for it.

pnpm-lock.yaml
```diff
@@ -348,6 +348,7 @@ importers:
       react-icons: ^4.3.1
       react-transition-group: ^4.4.2
       react-use: ^17.2.1
+      react-youtube: ^7.13.1
```
DukeManh, line 351: I've seen [lite-youtube-embed](https://github.com/paulirish/lite-youtube-embed) which claimed to render 224x faster while giving the same functionalities. Let's try it!

DukeManh (changes requested):

src/web/src/components/Posts/Post.tsx
dbelokon: I thought that the 85% looked better that covering the whole container. It seemed that the video was too close to the summary information on the right.

src/web/src/components/Posts/Post.tsx
dbelokon: The `url` already has the link to the youtube video, so there is no need to extract the video `id` from the `guid`.

pnpm-lock.yaml
dbelokon: How would I load this into the front-end project? I aready `pnpm install`ed it, but I am not sure how to load those files. The NPM package description says to simply treat it like a regular JS file, but I am not sure if I can do that here. @DukeManh

dbelokon: I figured it out! I had to add a few `import` lines in the `_app.tsx` file. 😌

dbelokon: Actually, nevermind, it seemed to work once and when I restarted the program it stopped working...

humphd: You use `pnpm add package-name` to add a package, and `pnpm install` to install dependencies and update the lock file.  With `npm` you use `install` for both meanings.

dbelokon: So, I was searching for a way to make this work, and I found this issue: vercel/next.js#11619. It seems that we are supposed to directly load the file with `import`, so that the `HTMLElement is not defined` error does not happen. I am not sure if there is a better way to make this work.

humphd: What if you disable SSR and load it dynamically https://nextjs.org/docs/advanced-features/dynamic-import#with-no-ssr.  I've had to do this in the past with other similar components.

DukeManh: @dbelokon they have a React ports too, have you tried that?

DukeManh: https://github.com/ibrahimcesar/react-lite-youtube-embed

dbelokon: I'm going to try using the React port. I didn't know they had one

dbelokon: The React port project worked great!

dbelokon:
@humphd, would it be possible to modify the [feed list in the CDOT wiki](https://wiki.cdot.senecacollege.ca/wiki/Planet_CDOT_Feed_List) to include your YouTube channel? I would like to see whether the back-end with the front-end would display the video.

humphd:
> @humphd, would it be possible to modify the [feed list in the CDOT wiki](https://wiki.cdot.senecacollege.ca/wiki/Planet_CDOT_Feed_List) to include your YouTube channel? I would like to see whether the back-end with the front-end would display the video.

Done.  I added this:

```
[http://blog.humphd.org/tag/seneca/rss/]
name=David Humphrey

[https://www.youtube.com/feeds/videos.xml?channel_id=UCqaMbMDf01BLttof1lHAo2A]
name=David Humphrey (YouTube)
```

dbelokon:
> One other thing I'd like to see, either here or in follow-up, is to make links into HTML <a> elements, so they are clickable.

I will submit an issue to address this.

dbelokon:
I checked the preview deployment after @humphd added his YouTube feed to the CDOT wiki feed list. Here are the results:

## PC screens:

image

## Phone screens: 

image

This is on the screen resizing of the browser, so on an actual phone it might be different. It can be improved, but I am not sure how we would like to show it for mobile user (maybe with a redirect to the YouTube mobile app if they have it?).

Let me know of any other small changes I may do.

cc @humphd, @DukeManh, @cindyledev

humphd:
Awesome!  In a follow-up, we should add a side-bar that includes YouTube links and the YouTube icon, like we do with GitHub info.  @dbelokon can you file that for me please?

src/web/package.json
```diff
@@ -27,6 +27,7 @@
     "react": "17.0.2",
     "react-dom": "17.0.2",
     "react-icons": "4.3.1",
+    "react-lite-youtube-embed": "^2.1.3",
```
humphd, line 30: We're using pinned versions, so this should be `2.1.3` without the `^` prefix.

src/web/src/pages/_app.tsx
```diff
@@ -13,6 +13,8 @@ import '../styles/globals.css';
 import '@fontsource/spartan';
 import '@fontsource/pt-serif';
 
+import 'react-lite-youtube-embed/dist/LiteYouTubeEmbed.css';
```
humphd, line 16: Should this be in the post component where we use it vs. here?  If we don't load that component, it's not needed.

humphd (commented):
Looks good, a couple small things.

src/web/src/pages/_app.tsx
dbelokon: Yeah, it should be in the `Post` component. I saw an example leaving it in the `_app.tsx`, so I didn't know it could be in another place.

src/web/src/components/Posts/Post.tsx
```diff
@@ -251,6 +260,11 @@ const extractGitHubUrlsFromPost = (htmlString: string): string[] => {
   );
 };
 
+const extractVideoId = (post: Post): string => {
```
humphd, line 263: When a function returns the result of an expression, and you use an arrow function, you don't need a function body:

```ts
const extractVideoId = (post: Post): string => post.guid.split(':')[2];
```

src/web/src/interfaces/index.ts
```diff
@@ -8,11 +8,13 @@ export type Feed = {
 export type Post = {
   feed: Feed;
   id: string;
+  guid: string;
   post: string;
   title: string;
   updated: string;
   url: string;
   html: string;
+  type: string;
```
humphd, line 17:
```ts
type: 'blogpost' | 'video';
```

humphd (changes requested):

humphd (approved):
R+ with CI/CD passing.

DukeManh:
Great work on implementing both backend and frontend.

nit: I think the video could expand to full width on Mobile as the screen is smaller.

>  redirect to the YouTube mobile app if they have it?

I tested this on my phone, clicking on the video title does redirect to the youtube app. We can still watch in the browser directly.

humphd (approved):

DukeManh (dismissed):

dbelokon:
@TDDR, since the reviews approved the changes and the PR is done, I'll be rebasing and merging once the CI checks have completed.

PR #2600, opened 2021-12-13
Handling invalid feed

### LINKED ISSUE #2569: Improve invalid feed handing
Now that @hphan9's work on #2520 is merged, we can finally see what's happening with our feed parser and invalid feeds: https://dev.api.telescope.cdot.systems/v1/posts/feeds/invalid.

Looking through these I think we need to fix some things:

- remove `\n` from the `reason` so that we can display it as a single line.  We could replace them with `' '` (space)
- figure out why 304s are showing up as invalid.  This means that content hasn't been modified (and we don't need to re-download) so this is odd to see in here
- we should set a `ttl` for these `invalid` keys in Redis, so they expire and we can retry.  A blog might be invalid today, and OK in a few days or a week.  Maybe we should set a ttl of 7 days or something?

I'm excited to see us get this data into the dashboard so we can act on it.  Do we have an issue on that yet?  If not, we should file one.

tpmai22:
## Issue This PR Addresses
Fixed #2569 

I got the 2/3 of the issue that we are having, handling invalid feed. In this PR I have replaced all the `\n` with whitespace `' '`. Adding a `ttl` so as the `invalid` blog post will be expire in 7 days. I have test through `redis-cli` after I `flushall` the old invalid key and using `ttl (invalid key)` it return the associate expired time which is 7 days.  

image

## Type of Change

## Checklist

### DIFF
--- a/src/api/posts/src/storage.js
+++ b/src/api/posts/src/storage.js
@@ -84,7 +84,7 @@ module.exports = {
         const id = key.replace(feedNamespace, '').replace(invalidSuffix, '');
         return {
           id,
-          reason,
+          reason: reason.replace(/\n/g, ' '),
         };
       })
     );
@@ -108,7 +108,8 @@ module.exports = {
 
   setInvalidFeed: (id, reason) => {
     const key = createInvalidFeedKey(id);
-    return redis.set(key, reason);
+    const sevenDaysInSeconds = 60 * 60 * 24 * 7; // Expire after 7 days
+    return redis.set(key, reason, 'EX', sevenDaysInSeconds);
   },
 
   /**
--- a/src/backend/utils/storage.js
+++ b/src/backend/utils/storage.js
@@ -64,7 +64,8 @@ module.exports = {
 
   setInvalidFeed: (id, reason) => {
     const key = createInvalidFeedKey(id);
-    return redis.set(key, reason);
+    const sevenDaysInSeconds = 60 * 60 * 24 * 7; // Expire after 7 days
+    return redis.set(key, reason, 'EX', sevenDaysInSeconds);
   },
 
   /**


humphd:
So it sounds like you need to file a new issue to deal with the 1/3rd you aren't fixing here, so it doesn't get lost.

src/api/posts/src/storage.js
```diff
@@ -80,8 +80,9 @@ module.exports = {
     const invalidKeys = await getFeedKeysUsingScanStream(`${feedNamespace}*${invalidSuffix}`);
     return Promise.all(
       invalidKeys.map(async (key) => {
-        const reason = await redis.get(key);
+        let reason = await redis.get(key);
         const id = key.replace(feedNamespace, '').replace(invalidSuffix, '');
+        reason = reason.replace(/\n/g, ' ');
```
humphd, line 85: Instead of making `reason` let vs. const, do this:

```
return {
  id,
  reason: reason.replace(/\n/g, ' '),
}
```

Don't modify or create variables when you can avoid it.

src/api/posts/src/storage.js
```diff
@@ -108,7 +109,8 @@ module.exports = {
 
   setInvalidFeed: (id, reason) => {
     const key = createInvalidFeedKey(id);
-    return redis.set(key, reason);
+    const expireAfter = 1000 * 60 * 60 * 24 * 7; // Expire after 7 days
```
humphd, line 112: Let's name this what it is: `sevenDaysInSeconds`

src/backend/utils/storage.js
```diff
@@ -64,7 +64,8 @@ module.exports = {
 
   setInvalidFeed: (id, reason) => {
     const key = createInvalidFeedKey(id);
-    return redis.set(key, reason);
+    const expireAfter = 1000 * 60 * 60 * 24 * 7; // Expire after 7 days
```
humphd, line 67: Same

humphd (changes requested):

humphd (approved):
Looks good, but can't take this until that other issue is filed.

hphan9 (approved):
It looks great

PR #2617, opened 2021-12-15
added advanced search query for elasticsearch

### LINKED ISSUE #2110: Advanced search dialog in SearchBar

AmasiaNalbandian:
## Issue This PR Addresses
This PR closes #2110 

## Type of Change

## Description
This PR Is aimed to help make search queries more flexible. What I have done here is I looked into Elasticsearch so that we can further query different fields we have indexed in Posts. I looked at the data presented back by Elasticsearch and realized we have more fields we can use to search: 
- post
- title
- author
- date published

This PR allows for us to query all of this on the backend, with search and Elasticsearch. 

Additionally, this query seems to be more promising than the original query for the post/author search. We should eventually use the query provided here, as the `zero_terms_query: 'all',` means that if that field is not provided, to search without it. This means we can use this query to search for one, two, three... or all fields, acting the same way as the previous implementation but doing more. The query will also work if we want to search for all posts published on X date, or from x to y dates.

## Steps to test: 
1. delete redis-data file if it exists
2. `cp config/env.development .env`
3. `docker-compose --env-file ./config/env.development up --build search elasticsearch redis posts`
4. to rebuild search: `docker-compose --env-file ./config/env.development up --build -d search posts` (NOT Necessary - Skip! just FYI)
5. in another terminal run `pnpm start` to start the feed parser. Wait a bit for some posts to populate.
6. As the changes are only reflected in the API, you can test queries using such:
```http://localhost/v1/search/advanced/?post=elasticsearch&author=amasia```
feel free to add more filters such as author, title, from, to and post

After you are done ensure your containers are down by running `docker-compose --env-file ./config/env.development down `
## Checklist

Example of how the search works:

The screenshot is taken from elasticsearch on localhost:9200 for reference of the post: 
chrome_eujzknH9TX

### DIFF
--- a/src/api/search/src/bin/validation.js
+++ b/src/api/search/src/bin/validation.js
@@ -1,4 +1,4 @@
-const { check, validationResult } = require('express-validator');
+const { oneOf, check, validationResult } = require('express-validator');
 
 const queryValidationRules = [
   // text must be between 1 and 256 and not empty
@@ -31,8 +31,59 @@ const queryValidationRules = [
     .bail(),
 ];
 
-const validateQuery = (rules) => {
+/**
+ * Advanced search is more flexible, only needs at least ONE field, but can run without any too.
+ * Date formats must be YYYY-MM-DD
+ */
+const advancedQueryValidationRules = [
+  oneOf([
+    check('post')
+      .exists({ checkFalsy: true })
+      .withMessage('post should not be empty')
+      .bail()
+      .isLength({ max: 256, min: 1 })
+      .withMessage('post should be between 1 to 256 characters')
+      .bail(),
+    check('author')
+      .exists({ checkFalsy: true })
+      .withMessage('author should exist')
+      .bail()
+      .isLength({ max: 100, min: 2 })
+      .withMessage('invalid author value')
+      .bail(),
+    check('title')
+      .exists({ checkFalsy: true })
+      .withMessage('title should exist')
+      .bail()
+      .isLength({ max: 100, min: 2 })
+      .withMessage('invalid title value')
+      .bail(),
+  ]),
+  check('to').optional().isISO8601().withMessage('invalid date format').bail(),
+
+  check('from').optional().isISO8601().withMessage('invalid date format').bail(),
+  check('perPage')
+    .optional()
+    .isInt({ min: 1, max: 10 })
+    .withMessage('perPage should be empty or a number between 1 to 10')
+    .bail(),
+
+  check('page')
+    .optional()
+    .isInt({ min: 0, max: 999 })
+    .withMessage('page should be empty or a number between 0 to 999')
+    .bail(),
+];
+
+/**
+ * Validates query by passing rules. The rules are different based on the pathname
+ * of the request. If the pathname is '/' it is the basic route.
+ * Otherwise, if '/advanced/' it is the advanced search
+ */
+const validateQuery = () => {
   return async (req, res, next) => {
+    const rules = req.baseUrl === '/' ? queryValidationRules : advancedQueryValidationRules;
+
     await Promise.all(rules.map((rule) => rule.run(req)));
 
     const result = validationResult(req);
@@ -45,4 +96,4 @@ const validateQuery = (rules) => {
   };
 };
 
-module.exports.validateQuery = validateQuery(queryValidationRules);
+module.exports.validateQuery = validateQuery();
--- a/src/api/search/src/routes/query.js
+++ b/src/api/search/src/routes/query.js
@@ -1,6 +1,6 @@
 const { Router, createError } = require('@senecacdot/satellite');
 const { validateQuery } = require('../bin/validation');
-const search = require('../search');
+const { search, advancedSearch } = require('../search');
 
 const router = Router();
 
@@ -8,6 +8,15 @@ router.get('/', validateQuery, async (req, res, next) => {
   try {
     const { text, filter, page, perPage } = req.query;
     res.send(await search(text, filter, page, perPage));
+  } catch (error) {
+    next(createError(503, error));
+  }
+});
+
+// route for advanced
+router.get('/advanced', validateQuery, async (req, res, next) => {
+  try {
+    res.send(await advancedSearch(req.query));
   } catch (error) {
     next(createError(503, error));
   }
--- a/src/api/search/src/search.js
+++ b/src/api/search/src/search.js
@@ -76,4 +76,94 @@ const search = async (
   };
 };
 
-module.exports = search;
+/**
+ * Advanced search allows you to look up multiple or single fields based on the input provided
+ * @param options.post    - text to search in post field
+ * @param options.author  - text to search in author field
+ * @param options.title   - text to search in title field
+ * @param options.from    - published after this date
+ * @param options.to      - published before this date
+ * @return all the results matching the fields text
+ * Range queries: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#_ranges
+ * Match field queries: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-ppmatch-query.html#query-dsl-match-query-zero
+ */
+const advancedSearch = async (options) => {
+  const results = {
+    query: {
+      bool: {
+        must: [],
+      },
+    },
+  };
+
+  const { must } = results.query.bool;
+
+  if (options.author) {
+    must.push({
+      match: {
+        author: {
+          query: options.author,
+          zero_terms_query: 'all',
+        },
+      },
+    });
+  }
+
+  if (options.post) {
+    must.push({
+      match: {
+        text: {
+          query: options.post,
+          zero_terms_query: 'all',
+        },
+      },
+    });
+  }
+
+  if (options.title) {
+    must.push({
+      match: {
+        title: {
+          query: options.title,
+          zero_terms_query: 'all',
+        },
+      },
+    });
+  }
+
+  if (options.from || options.to) {
+    must.push({
+      range: {
+        published: {
+          gte: options.from || '2000-01-01',
+          lte: options.to || new Date().toISOString().split('T')[0],
+        },
+      },
+    });
+  }
+
+  if (!options.perPage) {
+    options.perPage = ELASTIC_MAX_RESULTS_PER_PAGE;
+  }
+
+  if (!options.page) {
+    options.page = 0;
+  }
+
+  const {
+    body: { hits },
+  } = await client.search({
+    from: calculateFrom(options.page, options.perPage),
+    size: options.perPage,
+    _source: ['id'],
+    index,
+    type,
+    body: results,
+  });
+
+  return {
+    results: hits.total.value,
+    values: hits.hits.map(({ _id }) => ({ id: _id, url: `${POSTS_URL}/${_id}` })),
+  };
+};
+module.exports = { search, advancedSearch };

### CODE REVIEW

**Comment by humphd** [comment]
This needs a proper description of what it is, what it changes, how to test it, etc.

**Comment by AmasiaNalbandian** [comment]
> This needs a proper description of what it is, what it changes, how to test it, etc.

This was not meant to be compared to Seneca Master... I am still cherry picking commits - sorry about this.

File: src/api/search/src/bin/validation.js [hunk 1]
```diff
@@ -31,8 +31,73 @@ const queryValidationRules = [
     .bail(),
 ];
 
-const validateQuery = (rules) => {
+/**
+ * Advanced search is more flexible, only needs at least ONE field, but can run without any too.
+ * Date formats must be YYYY-MM-DD
+ */
+const advancedQueryValidationRules = [
+  oneOf([
```
humphd, line 39: This needs to be fixed for eslint.

src/api/search/src/bin/validation.js
```diff
@@ -31,8 +31,73 @@ const queryValidationRules = [
...
+      .withMessage('dateEnd should exist')
+      .bail()
+      .isLength({ max: 100, min: 2 })
+      .withMessage('invalid author value')
+      .bail()
+      .isISO8601()
+      .withMessage('invalid date format')
+      .bail(),
+
+    check('perPage')
```
humphd, line 77: The values above seem separate from these next two, where you are providing a modifier of the results vs. a filter used to produce those results.  I would break them out of your `oneOf` logic.

src/api/search/src/routes/query.js
```diff
@@ -13,4 +13,16 @@ router.get('/', validateQuery, async (req, res, next) => {
   }
 });
 
+// route for advanced
+router.get('/advanced', validateQuery, async (req, res, next) => {
```
humphd, line 17: Per your discussion in the description of this PR, I don't think that there's any point maintaining the existing search endpoint separate from the new advanced one, since the latter is a always special case of the former.  Let's have only one search endpoint, drop the `advanced` bit, and either pass the `postText` or `author` fields as an indication of which we are doing.  Later the front-end can be more precise.  Please file a follow-up issue on this, detailing what you think needs to happen for that to work, so we don't lose track of it here.

src/api/search/src/search.js
```diff
@@ -76,4 +76,81 @@ const search = async (
...
+ * @param authorText  - text to search in author field
+ * @param titleText   - text to search in title field
+ * @param dateStart   - published after this date
+ * @param dateEnd     - published before this date
+ * @return all the results matching the fields text
+ * Range queries: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#_ranges
+ * Match field queries: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-match-query.html#query-dsl-match-query-zero
+ */
+const advancedSearch = async (
+  postText = '',
```
humphd, line 91: Question: would it make more sense to pass in an Object with these options on it?  This would allow us to omit some of them.  Then, you could build your query Object by inspecting the values of the options passed in:

```js
const search = async (options) => {
  const query = {
    bool: {
      must: [],
    },
  };
  const must = query.bool.must;

  if(options.author) {
    must.push({
      match: {...},
    })
  }
  ...
  return query;
}
```

humphd (changes requested):
This needs tests.

src/api/search/src/bin/validation.js
AmasiaNalbandian: fixed.

src/api/search/src/routes/query.js
AmasiaNalbandian: We can't remove the existing one in this PR, but the one I was working on with the front end does work on removing it - we have to go through and refactor the react context for this change.

src/api/search/src/search.js
AmasiaNalbandian: The `zero_terms_query: 'all',` will omit if the value is not provided. 

The docs are explained here: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-match-query.html#query-dsl-match-query-zero

humphd: I don't think it makes sense to send values that we don't actually want, and to set defaults for things that were not passed to us by the query (e.g., setting things to `''`).  If a value isn't defined, don't add it.

src/api/search/src/routes/query.js
humphd: Then we need an issue filed to harmonize these into a single route.  Can you file that?

src/api/search/src/bin/validation.js
AmasiaNalbandian: I fixed this, but I want to know is the search going to be useful for just date range? should we leverage that... my test right now I fixed so that you need some context (author,title or post) AND date. Let me know we can play around with this.

src/api/search/src/routes/query.js
AmasiaNalbandian: filed: https://github.com/Seneca-CDOT/telescope/issues/2618

src/api/search/src/bin/validation.js
humphd: I think requiring at least one of `author`, `title`, `post`, or the dates is useful.  The `perPage` stuff is separate, since it governs the results of any of those.

src/api/search/src/bin/validation.js
```diff
@@ -31,8 +31,60 @@ const queryValidationRules = [
...
+
+/**
+ * Validates query by passing rules. The rules are different based on the pathname
+ * of the request. If the pathname is '/' it is the basic route.
+ * Otherwise, if '/advanced/' it is the advanced search
+ */
+const validateQuery = () => {
   return async (req, res, next) => {
+    const rules =
+      req._parsedUrl.pathname === '/' ? queryValidationRules : advancedQueryValidationRules;
```
humphd, line 86: This is failing eslint

humphd (changes requested):

src/api/search/src/search.js
AmasiaNalbandian: I think I should read about how the zero_terms_query works... but eventually this will ignore the field completely in the search... both ways will work. If its a matter of code looking clean I can change this. I'm not sure if the must.push will work though, in theory it looks logical but I will try.

src/api/search/src/bin/validation.js
AmasiaNalbandian: Can we change the eslint rules? or should I try another method?

AmasiaNalbandian:
> This needs tests.

Where can I find the tests to write for this? I know you've mentioned it a bunch of times... do we want to just test the routes?

humphd:
They should be in https://github.com/Seneca-CDOT/telescope/tree/master/src/api/search, but apparently we have no tests for this service?  Please file an issue for us to add them, since they don't exist, and we'll do it all at once.

src/api/search/src/bin/validation.js
humphd: Whenever possible, we try to listen to eslint's advice.  Here it's advising you not to rely on an internal API, which could change.  I would use the public API, `originalUrl`, see https://expressjs.com/en/api.html#req.originalUrl

AmasiaNalbandian:
> They should be in https://github.com/Seneca-CDOT/telescope/tree/master/src/api/search, but apparently we have no tests for this service? Please file an issue for us to add them, since they don't exist, and we'll do it all at once.

I've filed this: https://github.com/Seneca-CDOT/telescope/issues/2621

humphd:
OK, so let's get my other comments addressed in updates and try to land this ASAP.

src/api/search/src/search.js
```diff
@@ -76,4 +76,93 @@ const search = async (
...
+        },
+      },
+    });
+  }
+
+  if (options.from || options.to) {
+    must.push({
+      range: {
+        published: {
+          gte: options.from || '2000-01-01',
```
AmasiaNalbandian, line 138: @humphd is this valid in react.... It was working when I was querying but I am not too certain.

AmasiaNalbandian:
chrome_RWzZN27Wkp

Queries after requested changes

src/api/search/src/search.js
```diff
@@ -76,4 +76,93 @@ const search = async (
...
+const advancedSearch = async (options) => {
+  const query = {
+    query: {
+      bool: {
+        must: [],
+      },
+    },
+  };
+
+  const must = query.query.bool.must;
```
humphd, line 99:
```js
const { must } = query.query.bool;

File: src/api/search/src/search.js [hunk 8]
```diff
@@ -76,4 +76,93 @@ const search = async (
...
+ * @param authorText  - text to search in author field
+ * @param titleText   - text to search in title field
+ * @param dateStart   - published after this date
+ * @param dateEnd     - published before this date
+ * @return all the results matching the fields text
+ * Range queries: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#_ranges
+ * Match field queries: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-ppmatch-query.html#query-dsl-match-query-zero
+ */
+const advancedSearch = async (options) => {
+  const query = {
```
humphd, line 91: Lets' name this `result` so we don't have `query.query`

src/api/search/src/search.js
```diff
@@ -76,4 +76,93 @@ const search = async (
...
+    },
+  };
+
+  const must = query.query.bool.must;
+
+  if (options.author) {
+    must.push({
+      match: {
+        author: {
+          query: options.author,
```
humphd, line 105: Above you say `authorText` and `postText` but in this code you use `options.author` and `options.post`.  Please confirm which one is right, and correct the other.

humphd (changes requested):
Getting close, a few more things

src/api/search/src/search.js
AmasiaNalbandian: I updated to the simpler one, but I need to update the comments you are correct.

src/api/search/src/search.js
humphd: React doesn't matter here, I'm not sure I understand the question.

AmasiaNalbandian:
I keep rebasing but its saying i'm merging master... I also think I detached and I cant squash my commits all into 1.

humphd (dismissed):
Without tests, this is hard to verify at the runtime level, but the code looks OK to me.

humphd:
Please don't merge `master` *into* your branch, we merge your branch into `master`.  Use `rebase` vs. `merge` or `pull`.

humphd (approved):

JerryHue (changes requested):
I cannot find any other thing to comment about. Overall, the code is very concise!

src/api/search/src/search.js
```diff
@@ -76,4 +76,94 @@ const search = async (
...
+      },
+    });
+  }
+
+  if (options.from || options.to) {
+    must.push({
+      range: {
+        published: {
+          gte: options.from || '2000-01-01',
+          lte: options.to || Date.now,
```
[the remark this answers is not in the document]
JerryHue: @AmasiaNalbandian, I think this line is still the same in your latest update, right?

src/api/search/src/bin/validation.js
```diff
@@ -31,8 +31,59 @@ const queryValidationRules = [
...
+      .exists({ checkFalsy: true })
+      .withMessage('title should exist')
+      .bail()
+      .isLength({ max: 100, min: 2 })
+      .withMessage('invalid title value')
+      .bail(),
+  ]),
+  check('to').optional().isISO8601().withMessage('invalid date format').bail(),
+
+  check('from').optional().isISO8601().withMessage('invalid date format').bail(),
```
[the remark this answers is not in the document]
JerryHue: What have we decided on this? @humphd, @AmasiaNalbandian

src/api/search/src/search.js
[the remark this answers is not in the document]
AmasiaNalbandian: Yes I only rebased it.... I wasn't able to find time to sit down and build it yet.

src/api/search/src/bin/validation.js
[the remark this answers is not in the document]
AmasiaNalbandian: I basically did it so the more information you provide the more restrictive the results should be. So if we only do date, we can get all posts for that date/date range. My idea was that if someone wants to search the posts for a certain semester, they will be able to by providing the dates. So yes, while there is no context beyond date, it still provides a useful criteria.

humphd:
Let us know when this is ready for review.

AmasiaNalbandian:
it is ready now! Thank you @JerryHue!
chrome_CEwLHrVK1V

AmasiaNalbandian:
I updated all the instructions if you want to build and try it out. :) But also for my own memory...

JerryHue (commented):
Where can I find the build instructions?

HyperTHD (approved):
Tested, works good. My only concern is the lack of tests but the search service doesn't have any tests at all. Might want to file a follow-up to get those added in unless that's been filed already

AmasiaNalbandian:
> Tested, works good. My only concern is the lack of tests but the search service doesn't have any tests at all. Might want to file a follow-up to get those added in unless that's been filed already

It's been filed and assigned! :)

HyperTHD:
Just noticed #2621 so once you get another approval, you can rebase and merge this in @AmasiaNalbandian

humphd (approved):

JerryHue (approved):

PR #2623, opened 2021-12-16
Added .dockerignore file to search service

HyperTHD:
## Issue This PR Addresses
No issue was made for this PR

## Type of Change

## Description

While debugging an issue with my search service not starting due to a file access, I realized that the `Dockerfile` in our search service was checking the `node_modules` folder, something that shouldn't be happening. This PR adds a `.dockerignore` file to the search service so that when building the search service, the files/folders that should be ignored are ignored.

## Checklist

### DIFF
--- a/src/api/search/.dockerignore
+++ b/src/api/search/.dockerignore
@@ -1,0 +1,6 @@
+.dockerignore
+node_modules
+npm-debug.log
+Dockerfile
+.git
+.gitignore


AmasiaNalbandian:
should we also add *.md ?

HyperTHD:
> should we also add *.md ?

md files don't affect the build at all. They're mainly text based files after all. The files included are mostly log files, files that contain secrets that can't be exposed and the node_modules folder which makes building any node application long and can cause problems when messing with access permissions like we do in each of our services dockerfiles

AmasiaNalbandian:
Can I ask why we include .dockerignore?

humphd (approved):

HyperTHD:
> Can I ask why we include .dockerignore?

When we build a docker container via a dockerfile, we might have some variables or files that we want to have ignored when the container is being built. Think of it like a `.gitignore`. That file allows git to ignore whatever folder/file we put in it when using git to commit and push changes. 

In the case of a `.dockerignore` file, we're telling docker to ignore certain folders/files when building the application either via `docker build` or with docker-compose. In this case, we want to ignore the node_modules folder we copy over as well as any logs file since they aren't needed and end up making the build longer than it should be.

humphd:
@AmasiaNalbandian we use a `.dockerignore` file to help reduce the size of the build context we use when creating a Docker image.  For example, we don't want the `node_modules` folder to be copied into the image, since we need to set it up for Linux vs. Windows etc.

src/api/search/.dockerignore
```diff
@@ -0,0 +1,6 @@
+.dockerignore
```
AmasiaNalbandian, line 1: I'm really curious about this .dockerignore

AmasiaNalbandian (approved):

src/api/search/.dockerignore
AmasiaNalbandian: I'm understanding the structure of this project more now. Thank you! :)

humphd: Asking great questions is how you learn, keep it up! Code review is about fixing issues, but it's also about communicating how things work, and it's a perfect time to ask questions so that the community understands why a change is being made, and what it does.

HyperTHD: > do we have two? oh I see now.

Just for future reference, we don't want the dockerignore file to be included as well so we add it here too 👍🏾

PR #2629, opened 2021-12-18
Handle invalid blog url

### LINKED ISSUE #2628: Search page crash | Unhandled invalid URL
**What happened**:
Go to [search](https://telescope.cdot.systems/search/) and search for `humphrey`, the page will crash because of an unhandled exception when calling `new URL`. 
**Environment**:

- OS: OSX
- Browser: Firefox

DukeManh:
## Issue This PR Addresses
Fixes #2628. This PR  handles the exception of `new URL` with try-catch. Some blog URLs are relative paths and therefore not valid.
The search page won't crash anymore but clicking on the relative URLs directs us to the 404 not found. I'd like hear if we should handle this somehow or is it ok.

Example blog having invalid URL: https://telescope-5ab9u0wbz-humphd.vercel.app/search/?text=alex%20wang&filter=author

## Type of Change

## Checklist

### DIFF
--- a/src/api/parser/src/data/post.js
+++ b/src/api/parser/src/data/post.js
@@ -43,7 +43,8 @@ class Post {
     this.html = html;
     this.published = ensureDate(datePublished);
     this.updated = ensureDate(dateUpdated, datePublished);
-    this.url = postUrl;
+    // create an absolute url if postURL is relative
+    this.url = new URL(postUrl, feed.url).href;
     this.guid = guid;
 
     if (!(feed instanceof Feed)) {
--- a/src/api/posts/src/data/post.js
+++ b/src/api/posts/src/data/post.js
@@ -43,7 +43,8 @@ class Post {
     this.html = html;
     this.published = ensureDate(datePublished);
     this.updated = ensureDate(dateUpdated, datePublished);
-    this.url = postUrl;
+    // create an absolute url if postURL is relative
+    this.url = new URL(postUrl, feed.url).href;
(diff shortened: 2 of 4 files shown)


manekenpix (approved):

humphd (dismissed):
Really nice, great find!  Question: have you also found a bug in the RSS parser we're using?  It's odd to me that it would produce a URL that can't be parsed by `URL()`.

DukeManh:
@humphd, I think it's a problem with the [users' feed](https://codebeautify.org/rssviewer). The RSS parser returns whatever it reads from the feed. 

We get a list of posts [here](https://github.com/Seneca-CDOT/telescope/blob/ee0234a776a15cd213970d392ec619e1169afb83/src/backend/feed/processor.js#L193) and only check if the URL [is missing](https://github.com/Seneca-CDOT/telescope/blob/ee0234a776a15cd213970d392ec619e1169afb83/src/backend/data/post.js#L104) and then create a Post object from the data.

humphd:
Interesting.  So when the URL is relative like this, `/2018/01/22/first-blog-post.html`, we could pass the feed URL as the `base` argument to `URL()` so it resolves: https://developer.mozilla.org/en-US/docs/Web/API/URL/URL#parameters

DukeManh:
>we could pass the feed URL as the base argument to `URL()`

Yeah, we can do that.

humphd (approved):

manekenpix (approved):

PR #2653, opened 2022-01-13
Cache and display build logs

### LINKED ISSUE #2535: Cache build log for current/previous build
When you go to the build log in the dashboard, it shows you what's happening with the current build:

Screen Shot 2021-11-28 at 10 49 36 AM

That's useful, but not as useful as it could be.  What I'd really like is to see is the entire log for the current build get sent on load, and then have it continue when there is more data.

Ideally we would cache the build log and send that first, then stream the rest of the log in real-time.

If that's not doable, we can also consider switching to Web Sockets.

humphd:
Fixes #2535.

This adds a cache to the build log stream, so that when new clients connect, we start by sending the beginning of the log that they missed (if any).  It also allows us to stream the previous build's log, so we can always display the last build that completed.

The basic idea is that when we create the stream for the build process, we also create a `cache` stream buffer, which we fill with the data from the build output.  Then, whenever we need to send that data back again, we can get its contents and return.  This also allows us to remember the previous build's log and send it to whoever wants it later on.

I have tested nothing in this PR (yolo!) and I'm not sure how to test this without running it on staging.  This is just a draft to get things started.


src/api/status/public/js/build-log/check-for-build.js
```diff
@@ -1,48 +1,47 @@
...
 
-async function finish() {
-  try {
-    await reader.cancel();
-  } finally {
-    build = null;
-  }
-}
+  // Prefer the current build, but fallback to the previous one
+  const build = status.current || status.previous;
```
Andrewnt219, line 9: Nit
```suggestion
  const build = status.current ?? status.previous;
```

Andrewnt219 (approved):
Can't say it works for sure because I cannot run the auto-deployment server, but the logic looks straightforward to me. As I understand, the server now additionally writes to the `build.cache` while streaming the new data. And on response, it writes the cache first before appending new data.

Some logic changes in the client but it should work just fine. The code refactoring is a nice addition 👍. Maybe we will need a date library down the road or just wait for [Temporal](https://tc39.es/proposal-temporal/docs/) lol.

src/api/status/public/js/build-log/check-for-build.js
```diff
@@ -1,48 +1,47 @@
...
+    terminal.write(value);
+
+    // Recursively invoke processLog until `done === true`
+    return reader.read().then(processLog).catch(finish);
+  };
+
+  terminal.clear();
+
+  // Start processing the log output messages
+  reader.read().then(processLog).catch(finish);
```
DukeManh, line 46: Would this be a bit cleaner. 
```suggestion js
const processLog = () => {
    reader
      .read()
      .then(({ done, value }) => {
        if (done) {
          return finish();
        }
        // Write the data to the terminal
        terminal.write(value);
        // Recursively invoke processLog until `done === true`
        return processLog();
      })
      .catch(finish);
  };

  // Start processing the log output messages
  processLog();
```

DukeManh (changes requested):
Great work 👍 coding in the dark.

Andrewnt219 (dismissed):

Andrewnt219 (dismissed):

src/api/status/public/js/build-log/check-for-build.js
```diff
@@ -1,48 +1,48 @@
...
+          return finish();
+        }
+        // Write the data to the terminal
+        terminal.write(value);
+        // Recursively invoke processLog until `done === true`
+        return processLog();
+      })
+      .catch(finish);
+  };
+
```
DukeManh, line 45: I removed the `terminal.clear()` in the suggestion. Sorry.

tpmai22 (dismissed):
I think it's good, I learned a lot.

Andrewnt219 (approved):
LGTM🤞.

DukeManh (approved):

humphd:
I'm not sure what else to do on this without actually merging it and testing live.  @manekenpix I assume you would have to manually set this up if I merged?  Is there a time that's good for us to try that, so I don't break staging?

manekenpix:
@humphd I'll set up the autodeployment server later today (at 6ish) and launch telescope using this branch for testing.

humphd:
Fixed a bunch of crashes @manekenpix found on staging, rebased and integrated the work that landed recently on the build log header.  I think this is ready (again).

humphd:
I missed a lock file fix that went in with @dbelokon's video changes.  Rebased and updated.

manekenpix (approved):

HyperTHD (approved):

PR #2686, opened 2022-01-20
Fix CSP error and build log crashes in dashboard

humphd:
This is a follow-up to the build log cache PR I just merged.  I have a few bugs still in it:

- we need to connect back to the `/deploy` route of the web host vs. api host, so that CSP won't block it.
- we need to deal with the build being null, and show the error

I also cleaned up a path in the build server logic.

### DIFF
--- a/src/api/status/env.local
+++ b/src/api/status/env.local
@@ -1,5 +1,6 @@
 PLANET_PORT=1111
 PATH_PREFIX=/v1/status
 API_HOST=localhost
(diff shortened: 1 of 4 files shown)


LuigiZaccagnini (approved):

manekenpix (approved):

HyperTHD (approved):

humphd:
Thanks for the quick reviews, everyone.

Issue #2650, opened 2022-01-07
Clicking on Version v2.4.0 leads to an undefined commit on https://telescope.cdot.systems/


raygervais:
**What happened**:

When you select the version, it's redirects to an undefined (404) commit

![image](https://user-images.githubusercontent.com/14265337/148476120-05ca0431-1edd-4a10-b649-02f1bda5a522.png)

**What should have happened**:

When you select the version, it's supposed to take you to the GitHub commit relating to that release version

manekenpix:
Some context:

We process the payload sent by GitHub [here](https://github.com/Seneca-CDOT/telescope/blob/cfc8f8f30f67869715f3dceed836cd76c4c76275/tools/autodeployment/webhook.js#L66), so we can  pass the sha to our deployment script [here](https://github.com/Seneca-CDOT/telescope/blob/cfc8f8f30f67869715f3dceed836cd76c4c76275/tools/autodeployment/builds.js#L75), and this sha is added to the `env` file (deployment, staging, production) used in the deployment.

The payload sent when a push event takes place includes an `after` prop that contains the commit sha used in this process (see [here](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#push)). This prop doesn't exist in the payload sent for release events (see [here](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#release)).

We need to find a way to get the commit sha from a payload sent during a release event.

manekenpix:
Checking the payloads sent by github, it seems their documentation might not be up to date.
The docs about release payloads say that there's a prop called `target_commitish` that's supposed to look like this:

```json
"tag_name": "0.0.1",
"target_commitish": "master",
"name": null,
```
but checking our logs, we get this:

```json
"tag_name": "2.5.0",
"target_commitish": "321f0afe7b78bf7e0fb774719b1cb3a231ca5d8b",
"name": "Release 2.5.0",
```
I think it'd be a good idea to try `target_commitish` as a way to get the release's commit sha.

manekenpix:
Closed via #2756

PR #2695, opened 2022-01-20
Switch autodeployment server to use streaming cache for build logs

humphd:
Fixes #2695
I noticed a bug with the way I was caching the build logs: the cached contents get erased the first time they are requested.  So it was working the first time, but not on the second, third, ... request.

This switches the log cache to use an LRU stream cache, see https://github.com/LaurentZuijdwijk/streaming-cache.

It also lets us hold on to more build logs from the past based on git sha.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -9,6 +9,7 @@ overrides:
   node-fetch@<2.6.7: '>=2.6.7'
 
 importers:
+
   .:
     specifiers:
       '@babel/core': 7.16.7
@@ -409,8 +410,8 @@ importers:
       dotenv: ^10.0.0
       merge-stream: ^2.0.0
       pm2: ^5.1.1
+      rereadable-stream: ^1.4.13
       shelljs: ^0.8.4
-      stream-buffers: ^3.0.2
     dependencies:
       '@octokit/webhooks': 9.22.0
       '@senecacdot/satellite': 1.17.0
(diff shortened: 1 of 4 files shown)


AmasiaNalbandian (dismissed):
the code looks good! I'm just not sure how we can confirm it's working.... Also can you please resolve conflicts for the pnpm package?

humphd:
@manekenpix has to manually deploy this (it doesn't build automatically), so when he has time to try it, and can go for it.  I'll update the lock file.

pnpm-lock.yaml
```diff
@@ -401,7 +401,7 @@ importers:
       merge-stream: ^2.0.0
       pm2: ^5.1.1
       shelljs: ^0.8.4
-      stream-buffers: ^3.0.2
+      streaming-cache: ^0.5.5
```
AmasiaNalbandian, line 404: Do we want the carat for streaming-cache? or should we manually keep track of it?

AmasiaNalbandian (approved):

pnpm-lock.yaml
humphd: We need to pin all of these, which is what #2725 is about.

DukeManh (approved):

tools/autodeployment/builds.js
```diff
@@ -1,13 +1,26 @@
...
+const buildLogCache = new Cache();
 
 class Build {
   constructor(type, githubData) {
     this.type = type;
     this.githubData = githubData;
     this.startedDate = new Date();
+
+    // Remember the git SHA for this build
+    this.sha = githubData.after;
```
manekenpix, line 16: I think this won't work in production. I left a [comment in 2650](https://github.com/Seneca-CDOT/telescope/issues/2650#issuecomment-1018045662) for a possible fix.

manekenpix:
I'm not sure if this is work fixing here, but [this](https://github.com/Seneca-CDOT/telescope/blob/ef163e4e87e93eeb8cba4ffc11e8042b99c5fe2a/tools/autodeployment/builds.js#L40) is preventing the `autodeployment` server from sending the status code when the build succeeds. When the status code is 0, it's not copied into the `build` object, and `code` is missing in the data fetched by the dashboard.

humphd:
I've reworked this, and it's ready for another review.  @manekenpix, thanks for testing it on staging.  I've fixed the crashes you found.  I also needed to switch the streaming cache logic to use a re-readable stream, since the way I was doing it was still consuming the stream after the first read.

I've tested this with a hacked up local autodeployment server, and I think it's working. now.  If you could test it again on staging, I'd appreciate it.

humphd:
Prettier failures are due to issues on `master` that will get fixed and I'll rebase.

DukeManh:
@humphd, can you rebase, please.

humphd:
Squashed and Rebased.

manekenpix (approved):

manekenpix:
@humphd I'm ok if you want to rebase and merge.

humphd:
Hurray 🎆

Issue #2618, opened 2021-12-16
Refactor the search to use the advanced search route


AmasiaNalbandian:
**What would you like to be added**:
[The advanced search route backend is being implemented](https://github.com/Seneca-CDOT/telescope/pull/2617). However the react context written for the search needs to be refactored to change the route we push. Right now it's very static and not too flexible.  

**Why would you like this to be added**:
Refactor all of the search queries to use the one query we have. the new advancedsearch is quite flexible in what we can search for posts.

AmasiaNalbandian:
This is completely blocked by: https://github.com/Seneca-CDOT/telescope/issues/2755

PR #2699, opened 2022-01-20
2698: Remove advanced search validation breaking production search

### LINKED ISSUE #2698: ElasticSearch Advanced Route is breaking Search in Production
**What happened**:
A new route for search was introduced in https://github.com/Seneca-CDOT/telescope/pull/2617. However one of the files which validates the query was altered so that we can do either advanced search or regular search until we refactor everything in https://github.com/Seneca-CDOT/telescope/issues/2618

```
const validateQuery = () => {
    const rules = req.baseUrl === '/' ? queryValidationRules : advancedQueryValidationRules;

    await Promise.all(rules.map((rule) => rule.run(req)));

    const result = validationResult(req);
	@@ -45,4 +96,4 @@ const validateQuery = (rules) => {
  };
};
```
The way we are using req.base.Url is definitely incorrect and the error in console states that the query cant be validated - this should not have happened.

Searches should go through completely fine.

Go to https://telescope-ve8uhdv17-humphd.vercel.app/search/ and type a query - it returns no results.

This was working in API, but in the search bar it is not..... explore why

**Environment**:

- OS:
- Browser:

AmasiaNalbandian:
## Issue This PR Addresses
This PR fixes #2698

## Type of Change
I removed the inline conditional statement to determine which validation rules to use and replaced it with a function which will return the rules.

Once we complete the refactor in #2618, we can remove the change or keep it depending if we want the route search types. 

## Description
I removed the inline conditional statement to determine which validation rules to use. 
The rules are not required for advanced search because we have not implemented any of the work yet and will be reintroduced in the refactor #2618 

## How to test
Because previously this slipped up - I will go through this thoroughly to provide why this should work:

1. delete redis-data file if it exists
2. `cp config/env.development .env`
3. `docker-compose --env-file ./config/env.development up --build search elasticsearch redis posts`
4. in another terminal run `pnpm start` to start the feed parser. Wait a bit for some posts to populate.
5. in another terminal run `pnpm run dev` to start the front end app. 

OK there's two things to test here:
_1. the backend api still works for both searches_
you can test queries using the following link on your local machine:
**Advanced queries:**
```http://localhost/v1/search/advanced/?post=elasticsearch&author=amasia```
feel free to add more filters such as author, title, from, to and post
**Regular Queries:**
`http://localhost/v1/search/?text=amasia&filter=author` OR
`http://localhost/v1/search/?text=elasticsearch&filter=post`

_2. the front end search works and ONLY uses the regular query (not advanced search)_
you can test this by going to localhost:8000 and trying any search. This is the bug this PR aims to fix. I have attached screenshots to show myself searching locally and the network tab to indicate the query was to the correct path. 

**search author in telescope UI:**
chrome_ndpIBu3bPr

**search post in telescope UI:**
chrome_kCMmwHQIHw

Both of these are now working, which is the most important. Kindly check the UI search after building the containers as described above.
## Checklist
We need to check the search works in the hosting for this PR, then it should be fine. 

### DIFF
--- a/src/api/search/src/bin/validation.js
+++ b/src/api/search/src/bin/validation.js
@@ -82,8 +82,7 @@ const advancedQueryValidationRules = [
  */
 const validateQuery = () => {
   return async (req, res, next) => {
-    const rules = req.baseUrl === '/' ? queryValidationRules : advancedQueryValidationRules;
-
+    const rules = req.route.path === '/' ? queryValidationRules : advancedQueryValidationRules;
     await Promise.all(rules.map((rule) => rule.run(req)));
 
     const result = validationResult(req);

### CODE REVIEW

File: src/api/search/src/bin/validation.js [hunk 1]
```diff
@@ -82,7 +82,7 @@ const advancedQueryValidationRules = [
  */
 const validateQuery = () => {
   return async (req, res, next) => {
-    const rules = req.baseUrl === '/' ? queryValidationRules : advancedQueryValidationRules;
```
HyperTHD, line 85: Probably want to remove the variable on line 38 as well so you get ESLint fixed as well

HyperTHD (changes requested):

src/api/search/src/bin/validation.js
AmasiaNalbandian: We need it for the advanced search... Maybe I should just comment it all out. Technically we are going to remove out the basic search

AmasiaNalbandian: But honestly we can probably fix this now anyways. The req.baseURL isn't working like it was when I built and tested it :(

AmasiaNalbandian: Do you think if I make the function take a parameter it will be better? Do we see ourselves expanding to more search types?

HyperTHD: If you need it, I would just comment it out. Since it's probably going to come back, I would mention that it's gonna be commented out until the fix comes with a link to whatever issue is filed that deals with it

humphd: I would avoid using code comments to do project management, since they get forgotten and lost over time.  The best move would be to rip out the code, delete what we don't use, and file an issue that has the code you want to replace it with later, linking to some old commit that shows what you did.  If/when we fix this in the future, we can resurrect it from git/issues.

src/api/search/src/bin/validation.js
```diff
@@ -75,14 +75,30 @@ const advancedQueryValidationRules = [
...
+  switch (searchType) {
+    case '/':
+      rules = queryValidationRules;
+      break;
+    case '/advanced':
+      rules = advancedQueryValidationRules;
+      break;
+
+    default:
+      break;
```
humphd, line 89: I don't like that we can fall through this and get to a situation where `rules` is undefined.  What is the reason for not using a ternary or if/else?

humphd (changes requested):

src/api/search/src/bin/validation.js
AmasiaNalbandian: We could you're right. I can revert it back and use route.path - should work the same. I got carried away with the previous solution.

humphd: Can I suggest that if we have no users of the advanced route, we don't do this at all until we need it?

Let's learn from our time on this, and prioritize getting the tests landed first, then add the implementation later.

AmasiaNalbandian: i'm ok with that, however the front end UI will need this. I guess I can put both in together then?

humphd (approved):

humphd:
Let's get this landed, tested on staging, and shipped in `2.5.1` today.  I need more reviews on it.

manekenpix (approved):

PR #2721, opened 2022-01-24
Fix #1778: Add support for Turborepo

### LINKED ISSUE #1778: Improve monorepo tooling with Rushjs or Turborepo

menghif:
## Issue This PR Addresses

Fixes #1778

## Type of Change

## Description

This PR adds support for [Turborepo](https://turborepo.org/) to improve our monorepo tooling. Turborepo will create a cache locally: to test it run `pnpm build` then `pnpm build` again. The first time it will take time to create the build; the second time, since nothing has changed, it will use the cache and show this:

Screen Shot 2022-01-24 at 3 06 51 PM

Turborepo has a [Remote Caching](https://turborepo.org/docs/features/remote-caching) feature that will be automatically setup on Vercel if we update the Build command to use `turbo`: https://vercel.com/docs/concepts/git/monorepos#step-2:-set-the-build-command.
I'll open an issue for this based on feedback.

This is just an initial setup but there's a lot [more that we can explore](https://turborepo.org/docs/features/pipelines).

## Checklist


package.json
```diff
@@ -116,10 +116,20 @@
     "pretty-quick": "3.1.3",
     "run.env": "1.1.0",
     "supertest": "6.1.6",
-    "ts-jest": "27.1.2"
+    "ts-jest": "27.1.2",
+    "turbo": "^1.0.28"
```
humphd, line 120: This should be a pinned version `1.0.28` vs. `^1.0.28`.

package.json
```diff
@@ -7,7 +7,7 @@
     "html-elements": "tools/html-elements.js"
   },
   "scripts": {
-    "build": "pnpm build --prefix src/web --",
+    "build": "pnpm turbo run build",
```
humphd, line 10: Can you explain this change, I don't understand.  How does it know the build directory to use when running this?

humphd (changes requested):
This is very cool.  I need to spend time understanding how this works, and don't quite get it yet.  I've left a few comments, but I'd love to get on a call some time this week to have you walk me through this.

package.json
menghif: It basically goes inside all the other `package.json` specified by pnpm workspace to find a `build` command.

> Each key in the pipeline object is the name of a task that can be executed by turbo run. If turbo finds a workspace package with a package.json scripts object with a matching key, it will apply the pipeline task configuration to that NPM script during execution.

https://turborepo.org/docs/reference/configuration#pipeline

If for example we had scripts for `lint` and `test` in our microservices' `package.json`, running `pnpm turbo run lint test build` would run all those tasks in parallel when possible.

humphd: Ok, that's really helpful.  So this means that we could change the way we do things a bit.  Previously, we always thought about running stuff like lint, test, etc. from the top-level of the repo across all of the projects in the tree.  With this, we could be more specific to each sub-project, and then use turborepo to figure out what to run.

I also wonder if we could use this to `build` our Docker images?  That would be cool.  Is turborepo able to figure out when a sub-project's files have been update, or is this only happening at the level of the command being run?

menghif: I am also wondering about how to handle the Docker images with Turborepo. I haven't found much info on it and I don't have enough Docker experience yet.

> Is turborepo able to figure out when a sub-project's files have been update, or is this only happening at the level of the command being run?

Yes, it is able to see when sub-project's files have been updated.

tcvan0707 (approved):

Issue #2676, opened 2022-01-19
Allow YouTube channels to be added via the sign-up UI


humphd:
In #2675 we're adding support for YouTube feed discovery.  We need to update the UI to allow a user to include their YouTube channel, along with their blog(s).

nguyenhung15913:
Hi David,  I would like to know this issue in more details. Should we add a new form field for the user to enter the Youtube channel in the Sign Up page? Also at the moment, when I am trying to sign up for Telescope, there is a pop-up but it only has 2 buttons (previous and next)

nguyenhung15913:
![image](https://user-images.githubusercontent.com/69558780/150237220-186eadec-1ec4-4628-b5a4-3fc9c6cf1774.png)
here is the screenshot

humphd:
We have a form now for adding your blog URL(s), and we need to update the wording and verification code so it is clear you can include a YouTube channel.

nguyenhung15913:
I guess the form you are talking about is in <code>src\web\src\components\SignUp\Forms\RSSFeeds.tsx</code> right? My question is how does it look like in the browser or where can I see the form? Because the screenshot above is what I have when I press the Sign Up button.

humphd:
We've got the sign-up stuff turned off at the moment.  Talk to @DukeManh, he wrote a lot of it, and can give some ideas.

DukeManh:
I'll have a PR for fixing this Bug.

JerryHue:
@nguyenhung15913, what is the progress related to this issue? Have you managed to implement a small PR that can address it?

DukeManh:
@nguyenhung15913 , if you can make a draft PR, I can help debug the CSS with you.

nguyenhung15913:
> @nguyenhung15913 , if you can make a draft PR, I can help debug the CSS with you.

For now, I have been working on the loading posts for Mobile app, because we don't have a lot of time and the posts component is huge. If no one is working on this, I can do it

menghif:
Closed in #3505

Issue #2555, opened 2021-12-01
[3.0] Investigate Supabase


humphd:
I wonder if we could use a self-hosted https://supabase.com/, (project is on https://github.com/supabase/supabase).  https://www.youtube.com/watch?v=WiwfiVdfRIc is a good overview.

One of my longer term goals is to move toward running Telescope in the cloud.  I haven't figured out how to fund that (yet), so in the meantime we probably need to create our own cloud instead.  Since we're already slightly invested in a Firebase-type model (see Users service, thanks to @chrispinkney), we could consider going toward Supabase and self-host our own instance via docker.

Currently they have:

- Database (built on top of postgres, adding REST APIs, websocket real-time) and Object Storage (like S3)
- Auth and user management (not sure how easy it would be to hook into this with our SAML auth)
- a killer dashboard UI to monitor and control things

They are working on adding functions, which would be a killer feature for us, since we could move toward servless.  I like that it's also written in a tech stack that we could help fix (mostly TS/JS).

NOTE: I'm not saying we should do this.  I'm just thinking out loud in issues :)

DukeManh:
If I remember correctly, self-hosted Supabase doesn't have a dashboard.

humphd:
@DukeManh they open sourced it today! https://news.ycombinator.com/item?id=29401589

DukeManh:
@humphd That's awesome news.

cindyledev:
I've been struggling with Next Auth and Prisma at work for who knows how long. This issue has motivated me to give Supabase a try. It actually looks super promising. If I get Supabase working at work, I'll share my knowledge with everyone at Telescope via blog post or Slack :)

DukeManh:
I am very keen on Firebase and have worked with it for some time.
Because Supabase is a Firebase alternative I'd like to give this a try.

I'll investigate and come back here with what I learn.

humphd:
Cool.  Another option we can consider, if this ends up being too heavy-weight, is to add a db ourselves (postgres, sqlite, etc).  We should only take this on if it gives us something we need vs. makes our setup and admin too complicated.

Thanks for looking into this.  I think it might be cool to add this, esp. when they include serverless functions.

chrispinkney:
@DukeManh Might be worth investigating [Appwrite](https://github.com/appwrite/appwrite) too! But it's very early stage AFAIK, but I think recently secured some funding which is cool?

HyperTHD:
[...]

PR #2743, opened 2022-01-26
Self host Supabase service

DukeManh:
## Issue This PR Addresses

Part of #2555. This adds a Supbase self-hosted instance to Telescope.

Note: This pr also fixes the `SignUpPage` crash bug mentioned in #2676. @nguyenhung15913 , you can take a look at `AuthProvider` here as a starting point of your PR.

## Type of Change

### Testing Supabase

1. Go to the Supabase service
```
  cd src/api/supabase
```
2. Copy the env vars
```
  cp .env.example .env
```
3. Start
```
  docker-compose up
```
4. Go to [localhost:8910](http://localhost:8910) to visit the Dashboard

### Testing Supabase API in the auth service

1. Do all the steps above
2. Uncomment Sign Up link in `BannerButtons` in https://github.com/Seneca-CDOT/telescope/blob/81775c41845a34455b5fd9dde65cbe231d14d79b/src/web/src/components/BannerButtons.tsx#L111-L120
3. In `/src/api/auth/src/middleware.js`, replace line 92 to 110
https://github.com/Seneca-CDOT/telescope/blob/81775c41845a34455b5fd9dde65cbe231d14d79b/src/api/auth/src/middleware.js#L92-L110

with 

```js
      const supabase = require('./supabase');
      ...
      const { email } = req.user;
      const { user, error } = await supabase.auth.api.createUser({
        email,
        password: 'hello',
        data: req.body,
      });

      if (error) {
        logger.warn({ status: error.status, email, data: req.body }, error.message);
        next(createError(error.status, 'unable to create user'));
        return;
      }
```
4. Go to [localhost:8000](http://localhost:8000) to Sign Up using an example account:
   - username: user1
   - password: user1pass 
 5. Finish the user registration flow
 6. Vist the Supabase dashboard again at  [localhost:8910](http://localhost:8910) and check the user list. A new user is created!

## Checklist

### DIFF
--- a/config/env.development
+++ b/config/env.development
@@ -10,7 +10,7 @@ COMPOSE_PROJECT_NAME=telescope_api
 # so it will work on Windows and Unix, see
 # https://docs.docker.com/compose/reference/envvars/#compose_file
 COMPOSE_PATH_SEPARATOR=;
-COMPOSE_FILE=docker/docker-compose.yml;docker/development.yml
+COMPOSE_FILE=docker/docker-compose.yml;docker/development.yml;docker/supabase/docker-compose.yml
 
 
 # The host where the Telescope 1.0 front-end and back-end are run.
@@ -86,7 +86,6 @@ JWT_AUDIENCE=http://localhost
 
 # How long should a JWT work before it expires
 JWT_EXPIRES_IN=1h
-
 
 ################################################################################
 # Image Service
@@ -260,3 +259,55 @@ FEED_QUEUE_PARALLEL_WORKERS=1
 
 # Max number of posts per page
 MAX_POSTS_PER_PAGE=5
+
+
+################################################################################
+# Supabase Services
+################################################################################
+
+SUPABASE_URL=http://kong:8000
+
+# Secrets
+# YOU MUST CHANGE THESE BEFORE GOING INTO PRODUCTION
+
+POSTGRES_PASSWORD=your-super-secret-and-long-postgres-password
+JWT_SECRET=your-super-secret-jwt-token-with-at-least-32-characters-long
+ANON_KEY=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyAgCiAgICAicm9sZSI6ICJhbm9uIiwKICAgICJpc3MiOiAic3VwYWJhc2UtZGVtbyIsCiAgICAiaWF0IjogMTY0MTc2OTIwMCwKICAgICJleHAiOiAxNzk5NTM1NjAwCn0.dc_X5iR_VP_qT0zsiyj_I_OZ2T9FtRU2BBNWN8Bu4GE
+SERVICE_ROLE_KEY=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyAgCiAgICAicm9sZSI6ICJzZXJ2aWNlX3JvbGUiLAogICAgImlzcyI6ICJzdXBhYmFzZS1kZW1vIiwKICAgICJpYXQiOiAxNjQxNzY5MjAwLAogICAgImV4cCI6IDE3OTk1MzU2MDAKfQ.DaYlNEoUrrEn2Ig7tqibS-PHK5vgusbcbo7X36XVt4Q
+
+# Auth
+
+## General
+
+### Telescope web URL, must match with WEB_URL
+SITE_URL=http://localhost:8000
+
+ADDITIONAL_REDIRECT_URLS=
+JWT_EXPIRY=3600
+DISABLE_SIGNUP=false
+
+## Email auth
+ENABLE_EMAIL_SIGNUP=true
+ENABLE_EMAIL_AUTOCONFIRM=true
+SMTP_ADMIN_EMAIL=admin@example.com
+SMTP_HOST=mail
+SMTP_PORT=2500
+SMTP_USER=fake_mail_user
+SMTP_PASS=fake_mail_password
+SMTP_SENDER_NAME=fake_sender
+
+## Phone auth
+ENABLE_PHONE_SIGNUP=false
+ENABLE_PHONE_AUTOCONFIRM=false
+
+# Ports
+
+## Studio port
+STUDIO_PORT=8910
+
+## API endpoint ports
+KONG_HTTP_PORT=8911
+KONG_HTTPS_PORT=8912
+
+## DB port
+POSTGRES_PORT=8913
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -84,6 +84,9 @@ services:
       - ALLOWED_APP_ORIGINS
       - JWT_EXPIRES_IN
       - USERS_URL
+      # Supabase config
+      - SERVICE_ROLE_KEY
+      - SUPABASE_URL
       # Satellite authentication/authorization support
       - JWT_ISSUER
       - JWT_AUDIENCE
--- a/docker/supabase/docker-compose.yml
+++ b/docker/supabase/docker-compose.yml
@@ -1,0 +1,156 @@
+# Usage
+#   Start:          docker-compose up
+#   With helpers:   docker-compose -f docker-compose.yml -f ./dev/docker-compose.dev.yml up
+#   Stop:           docker-compose down
+#   Destroy:        docker-compose -f docker-compose.yml -f ./dev/docker-compose.dev.yml down -v --remove-orphans
+
+version: '3.8'
+
+services:
+  studio:
+    container_name: supabase-studio
+    image: supabase/studio:latest
+    restart: unless-stopped
+    ports:
+      - ${STUDIO_PORT}:3000/tcp
+    environment:
+      SUPABASE_URL: http://kong:8000
+      STUDIO_PG_META_URL: http://meta:8080
+      SUPABASE_ANON_KEY: ${ANON_KEY}
+      SUPABASE_SERVICE_KEY: ${SERVICE_ROLE_KEY}
+
+  kong:
+    container_name: supabase-kong
+    image: kong:2.1
+    restart: unless-stopped
+    ports:
+      - ${KONG_HTTP_PORT}:8000/tcp
+      - ${KONG_HTTPS_PORT}:8443/tcp
+    environment:
+      KONG_DATABASE: 'off'
+      KONG_DECLARATIVE_CONFIG: /var/lib/kong/kong.yml
+      # https://github.com/supabase/cli/issues/14
+      KONG_DNS_ORDER: LAST,A,CNAME
+      KONG_PLUGINS: request-transformer,cors,key-auth,acl
+    volumes:
+      - ./supabase/volumes/api/kong.yml:/var/lib/kong/kong.yml
+
+  auth:
+    container_name: supabase-auth
+    image: supabase/gotrue:v2.2.12
+    depends_on:
+      - db
+    restart: unless-stopped
+    environment:
+      GOTRUE_API_HOST: 0.0.0.0
+      GOTRUE_API_PORT: 9999
+
+      GOTRUE_DB_DRIVER: postgres
+      GOTRUE_DB_DATABASE_URL: postgres://postgres:${POSTGRES_PASSWORD}@db:5432/postgres?search_path=auth
+
+      GOTRUE_SITE_URL: ${SITE_URL}
+      GOTRUE_URI_ALLOW_LIST: ${ADDITIONAL_REDIRECT_URLS}
+      GOTRUE_DISABLE_SIGNUP: ${DISABLE_SIGNUP}
+
+      GOTRUE_JWT_SECRET: ${JWT_SECRET}
+      GOTRUE_JWT_EXP: ${JWT_EXPIRY}
+      GOTRUE_JWT_DEFAULT_GROUP_NAME: authenticated
+
+      GOTRUE_EXTERNAL_EMAIL_ENABLED: ${ENABLE_EMAIL_SIGNUP}
+      GOTRUE_MAILER_AUTOCONFIRM: ${ENABLE_EMAIL_AUTOCONFIRM}
+      GOTRUE_SMTP_ADMIN_EMAIL: ${SMTP_ADMIN_EMAIL}
+      GOTRUE_SMTP_HOST: ${SMTP_HOST}
+      GOTRUE_SMTP_PORT: ${SMTP_PORT}
+      GOTRUE_SMTP_USER: ${SMTP_USER}
+      GOTRUE_SMTP_PASS: ${SMTP_PASS}
+      GOTRUE_SMTP_SENDER_NAME: ${SMTP_SENDER_NAME}
+      GOTRUE_MAILER_URLPATHS_INVITE: /auth/v1/verify
+      GOTRUE_MAILER_URLPATHS_CONFIRMATION: /auth/v1/verify
+      GOTRUE_MAILER_URLPATHS_RECOVERY: /auth/v1/verify
+      GOTRUE_MAILER_URLPATHS_EMAIL_CHANGE: /auth/v1/verify
+
+      GOTRUE_EXTERNAL_PHONE_ENABLED: ${ENABLE_PHONE_SIGNUP}
+      GOTRUE_SMS_AUTOCONFIRM: ${ENABLE_PHONE_AUTOCONFIRM}
+
+  rest:
+    container_name: supabase-rest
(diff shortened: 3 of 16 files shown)


humphd (changes requested):

humphd:
Filed https://github.com/Seneca-CDOT/telescope/pull/2748 to fix the auth build

DukeManh:
Why do I have `pnpm-lock.yaml` after installing `@supabase/supabase-js`? I installed with `pnpm add` inside 'sr/api/auth'

src/api/supabase/docker-compose.yml
```diff
@@ -0,0 +1,158 @@
+# Usage
+#   Start:          docker-compose up
+#   With helpers:   docker-compose -f docker-compose.yml -f ./dev/docker-compose.dev.yml up
```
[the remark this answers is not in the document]
DukeManh: I couldn't find the docs for this, Supabase doesn't have to most inclusive documentation.
But based on `./dev/docker-compose.dev.yml`, this will also run [inbucket](https://www.inbucket.org/)  for email testing and [meta](https://github.com/supabase/postgres-meta) for managing Postgres database like running queries and add roles.

humphd:
Created https://github.com/Seneca-CDOT/telescope/pull/2752 to fix the naming of auth containers.

humphd:
Can you rebase to pick up the changes from #2752?

humphd (changes requested):
Let's move `supabase` out of `src/api` and put it in `docker/supabase`, since that's what it is (i.e., a bunch of docker and config files).

Next, let's modify https://github.com/Seneca-CDOT/telescope/blob/master/config/env.development#L13 so that it adds a third docker-compose file to our list:

```
COMPOSE_FILE=docker/docker-compose.yml;docker/development.yml;docker/supabase/docker-compose.yml
```

Next, let's move everything in `supabase/env.example` to `config/env.development` under a new `Supabase` section.  That is, let's merge the Supabase variables with ours.

Next, lets update all of the `volumes` relative paths in the Supabase docker-compose files (there are two), so that it uses `./supabase/volumes` vs. `./volumes`.  We'll have to confirm that I have this right, but basically, we need to use a relative path for the new location, which is the merged path (i.e., the first `docker-compose.yml` we have in `docker`).

We won't add any of this to staging or production yet, only for development and gitpod.  That way we can iterate on it and secure things before we ship it to staging.

DukeManh:
@humphd works for me on Gitpod.

config/env.development
```diff
@@ -87,6 +87,14 @@ JWT_AUDIENCE=http://localhost
 # How long should a JWT work before it expires
 JWT_EXPIRES_IN=1h
 
+# Example Supabase secrets for development
+# NOT INTENDED FOR PRODUCTION
+
+# Supabase connection https://supabase.com/docs/guides/database/connecting-to-postgres
+SUPABASE_API_URL=http://localhost:8911
```
humphd, line 94: In docker, should this not be `http://kong:8911`?  Running in containers locally will not allow `localhost` to work.

config/env.development
```diff
@@ -260,3 +268,50 @@ FEED_QUEUE_PARALLEL_WORKERS=1
...
+# Supabase Service
+################################################################################
+
+# Secrets
+# YOU MUST CHANGE THESE BEFORE GOING INTO PRODUCTION
+
+POSTGRES_PASSWORD=your-super-secret-and-long-postgres-password
+JWT_SECRET=your-super-secret-jwt-token-with-at-least-32-characters-long
+ANON_KEY=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyAgCiAgICAicm9sZSI6ICJhbm9uIiwKICAgICJpc3MiOiAic3VwYWJhc2UtZGVtbyIsCiAgICAiaWF0IjogMTY0MTc2OTIwMCwKICAgICJleHAiOiAxNzk5NTM1NjAwCn0.dc_X5iR_VP_qT0zsiyj_I_OZ2T9FtRU2BBNWN8Bu4GE
+SERVICE_ROLE_KEY=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyAgCiAgICAicm9sZSI6ICJzZXJ2aWNlX3JvbGUiLAogICAgImlzcyI6ICJzdXBhYmFzZS1kZW1vIiwKICAgICJpYXQiOiAxNjQxNzY5MjAwLAogICAgImV4cCI6IDE3OTk1MzU2MDAKfQ.DaYlNEoUrrEn2Ig7tqibS-PHK5vgusbcbo7X36XVt4Q
```
humphd, line 283: Already defined on line 96.  I think we should move all the Supabase related variables into this block.

config/env.development
```diff
@@ -260,3 +268,50 @@ FEED_QUEUE_PARALLEL_WORKERS=1
...
+
+# Auth
+
+## General
+SITE_URL=http://localhost:3000
```
humphd, line 288: Again, `localhost` worries me in Docker.

config/env.production
```diff
@@ -88,6 +88,10 @@ JWT_AUDIENCE=https://api.telescope.cdot.systems
 # How long should a JWT work before it expires
 JWT_EXPIRES_IN=1h
 
+# Supabase connection https://supabase.com/docs/guides/database/connecting-to-postgres
```
humphd, line 91: Let's not do anything in `env.production` or `env.staging` yet.

config/env.staging
```diff
@@ -90,6 +90,11 @@ JWT_AUDIENCE=https://dev.api.telescope.cdot.systems
 # How long should a JWT work before it expires
 JWT_EXPIRES_IN=1h
 
+# Supabase connection https://supabase.com/docs/guides/database/connecting-to-postgres
```
humphd, line 93: Remove this for now

docker/supabase/.env.example
```diff
@@ -0,0 +1,41 @@
+# Secrets
```
humphd, line 1: You can delete this file now, or leave a comment in here to see `config/env.development` and the other files.

config/env.development
humphd: Should this be `http://studio:3000`?

humphd (changes requested):

config/env.development
DukeManh: Right, it shouldn't be localhost, the supabase port is `htttp://kong:8000`. I've tested this locally and it worked.

DukeManh: I'm thinking we should publish `htttp://kong:8000` to localhost to, because I don't think `htttp://kong` works on Gitpod.

humphd: We can do that as a separate thing, but the URL we pass into the container to use has to use the docker network (i.e, container DNS).

It might be good to have an `env.gitpod` separate from `env.development` so this is easier to separate.

config/env.development
DukeManh: I think this is a docker Port for the studio, and it's mapped to port 8910.

DukeManh: It's actually the base URL of your website, so it should be localhost:8000
https://supabase.com/docs/guides/hosting/docker#update-secrets

config/env.development
```diff
@@ -87,6 +87,9 @@ JWT_AUDIENCE=http://localhost
 # How long should a JWT work before it expires
 JWT_EXPIRES_IN=1h
 
+# Example Supabase secrets for development
```
humphd, line 90: I think you can remove this.  Everything in this file isn't suitable for production.

config/env.development
```diff
@@ -260,3 +263,54 @@ FEED_QUEUE_PARALLEL_WORKERS=1
...
+POSTGRES_PASSWORD=your-super-secret-and-long-postgres-password
+JWT_SECRET=your-super-secret-jwt-token-with-at-least-32-characters-long
+ANON_KEY=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyAgCiAgICAicm9sZSI6ICJhbm9uIiwKICAgICJpc3MiOiAic3VwYWJhc2UtZGVtbyIsCiAgICAiaWF0IjogMTY0MTc2OTIwMCwKICAgICJleHAiOiAxNzk5NTM1NjAwCn0.dc_X5iR_VP_qT0zsiyj_I_OZ2T9FtRU2BBNWN8Bu4GE
+SERVICE_ROLE_KEY=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyAgCiAgICAicm9sZSI6ICJzZXJ2aWNlX3JvbGUiLAogICAgImlzcyI6ICJzdXBhYmFzZS1kZW1vIiwKICAgICJpYXQiOiAxNjQxNzY5MjAwLAogICAgImV4cCI6IDE3OTk1MzU2MDAKfQ.DaYlNEoUrrEn2Ig7tqibS-PHK5vgusbcbo7X36XVt4Q
+
+# Auth
+
+## General
+
+// Telescope base URL
```
humphd, line 286: Your comment type is wrong here.  Also, `Telescope web URL` might be more clear, and add a note that this must match `WEB_URL` defined elsewhere in this file.

config/env.development
```diff
@@ -260,3 +263,54 @@ FEED_QUEUE_PARALLEL_WORKERS=1
...
+SITE_URL=http://localhost:8000
+ADDITIONAL_REDIRECT_URLS=
+JWT_EXPIRY=3600
+DISABLE_SIGNUP=false
```
humphd, line 290: We don't want people signing up, do we?

config/env.development
```diff
@@ -260,3 +263,54 @@ FEED_QUEUE_PARALLEL_WORKERS=1
...
+
+## Email auth
+ENABLE_EMAIL_SIGNUP=true
```
humphd, line 293: Should this be false?  We don't want email signup

docker/docker-compose.yml
```diff
@@ -83,6 +83,8 @@ services:
       - ADMINISTRATORS
       - ALLOWED_APP_ORIGINS
       - JWT_EXPIRES_IN
+      - SERVICE_ROLE_KEY
```
humphd, line 86: Maybe throw a comment above this line to indicate that these are for Supabase.

docker/supabase/docker-compose.yml
```diff
@@ -0,0 +1,158 @@
...
+
+version: '3.8'
+
+services:
+  studio:
+    container_name: supabase-studio
+    image: supabase/studio:latest
+    restart: unless-stopped
+    ports:
+      - ${STUDIO_PORT}:3000/tcp
```
humphd, line 15: So we're opening port 3000.  We'll have to evaluate this in staging and production, cc @manekenpix

docker/supabase/docker-compose.yml
```diff
@@ -0,0 +1,158 @@
...
+    environment:
+      SUPABASE_URL: http://kong:8000
```
humphd, line 17: Above we call this `SUPABASE_API_URL` and they call it `SUPABASE_URL`.  Should we change to use `SUPABASE_URL` everywhere, so we don't confuse things, or misconfigure by having them be different?

docker/supabase/docker-compose.yml
```diff
@@ -0,0 +1,158 @@
...
+      STUDIO_PG_META_URL: http://meta:8080
+      SUPABASE_ANON_KEY: ${ANON_KEY}
+      SUPABASE_SERVICE_KEY: ${SERVICE_ROLE_KEY}
+
+  kong:
+    container_name: supabase-kong
+    image: kong:2.1
+    restart: unless-stopped
+    ports:
+      - ${KONG_HTTP_PORT}:8000/tcp
```
humphd, line 27: Ports 8000, 8443.  We'll have to evaluate this in staging and production, cc @manekenpix 

Also, we already occupy `8000` on `localhost` via `WEB_URL=http://localhost:8000`

docker/supabase/docker-compose.yml
```diff
@@ -0,0 +1,158 @@
...
+      PG_META_DB_HOST: db
+      PG_META_DB_PASSWORD: ${POSTGRES_PASSWORD}
+
+  db:
+    container_name: supabase-db
+    image: supabase/postgres:14.1.0
+    command: postgres -c config_file=/etc/postgresql/postgresql.conf
+    restart: unless-stopped
+    ports:
+      - ${POSTGRES_PORT}:5432
```
humphd, line 153: This seems like a bad idea.  Can we avoid exposing postgres?

humphd (changes requested):

config/env.development
DukeManh: Sounds good

docker/supabase/docker-compose.yml
DukeManh: This would allow direct connections to Postgres using a password. Do we ever want to connect to the database without using the Supabase client?

config/env.development
DukeManh: I'm not sure, don't we want to create a new user using their email?
https://supabase.com/docs/reference/javascript/auth-signup

docker/supabase/docker-compose.yml
DukeManh: Sounds good

docker/supabase/docker-compose.yml
humphd: I think it's extremely risk to open it.  The only thing I can imagine wanting to do are backups, and we can write a container-based, automated backup system to do it within the docker network later.

DukeManh:
@manekenpix, could you also take a look, please?

config/env.development
```diff
@@ -260,3 +259,55 @@ FEED_QUEUE_PARALLEL_WORKERS=1
 
 # Max number of posts per page
 MAX_POSTS_PER_PAGE=5
+
+
+################################################################################
+# Supabase Service
```
humphd, line 265: Services (plural)

config/env.development
humphd: My (very limited) reading of this is that setting this to `true` means users can use their email to sign-up to Supabase.  We don't really want to enable sign-ups at all, since we want to manage them ourselves via the Telescope SSO service, right?  Or am I misunderstanding this.

config/env.development
humphd: What was the answer to this one?

docker/docker-compose.yml
```diff
@@ -84,6 +84,9 @@ services:
       - ALLOWED_APP_ORIGINS
       - JWT_EXPIRES_IN
       - USERS_URL
+      # Supabase secrets
```
humphd, line 87: These aren't both secrets, `Subabase config`?

docker/supabase/docker-compose.yml
humphd: What's happening here?

humphd (changes requested):

config/env.development
DukeManh: I see what you mean. When we get the user info from SSO and the registration flow, we need to register them with their email so next time they sign in we know this user is both a Seneca and Telescope user. If email sign-up is disabled, is there another way to sign up?

docker/supabase/docker-compose.yml
DukeManh: @manekenpix

config/env.development
DukeManh: Please check my comment below

config/env.development
humphd: Yeah, I'm not sure.  Let's leave it like this, and we'll test it out and make changes once we have it working and get tests written.  It's hard to reason about this based solely on this config.

docker/supabase/docker-compose.yml
DukeManh: @humphd, I wonder how they still work running on the same port?

I am think if we could use Kong to replace Nginx?

humphd: We do our SSL through nginx, so I'm not really wanting to replace it.  We just need to have this work behind nginx similar to how Traefik does now.

DukeManh: Can I keep this change?

humphd: For now, sure.  When we put it staging, we'll have to rework this.

docker/supabase/docker-compose.yml
humphd: Same for now, we can leave it open while we learn and test in dev.

humphd (approved):

manekenpix (approved):
Keeping an eye on the open ports

manekenpix (approved):

humphd (approved):

PR #2748, opened 2022-01-27
Rework auth Dockerfile so it builds on M1

humphd:
I'm testing #2743 locally, and on my M1 Mac, I can't get the auth service's Dockerfile to build.  The alpine base image is missing Python for me, so the build fails.

This reworks the Dockerfile to do things in stages/layers:

- **base**: we use a full `node:lts` base to do our build, and copy in the `package.json`
- **development**: if we want to ever test the container in dev mode, this will do it.  We won't use this in production
- **production**: installs all deps in production mode for use in Alpine (Linux musl x64)
- **deploy**: copies the build production deps out and grabs the source code, runs the server.

This strategy means we end up with a very small image, but start with a full one for the build.

### DIFF
--- a/src/api/sso/Dockerfile
+++ b/src/api/sso/Dockerfile
@@ -1,12 +1,50 @@
-FROM node:lts-alpine as base
-RUN apk add dumb-init
-WORKDIR /app
-COPY --chown=node:node ./package.json ./
+## Base ########################################################################
+# Use a larger node image to do the build for native deps (e.g., gcc, pytyhon)
+FROM node:lts as base
 
-FROM base AS dependencies
-RUN npm install --only=production --no-package-lock
+# Reduce npm log spam and colour during install within Docker
+ENV NPM_CONFIG_LOGLEVEL=warn
+ENV NPM_CONFIG_COLOR=false
 
-FROM dependencies as deploy
-COPY --chown=node:node ./src ./src
+# We'll run the app as the `node` user, so put it in their home directory
+WORKDIR /home/node/app
+# Copy the package.json and lock file over
+COPY package*.json /home/node/app/
+
+## Development #################################################################
+# Define a development target that installs devDeps and runs in dev mode
+FROM base as development
+WORKDIR /home/node/app
+# Install (not ci) with dependencies, and for Linux vs. Linux Musl (which we use for -alpine)
+RUN npm install
+# Copy the source code over
+COPY --chown=node:node . /home/node/app/
+# Switch to the node user vs. root
 USER node
+# Start the app in debug mode so we can attach the debugger
+CMD ["npm", "run", "dev"]
+
+## Production ##################################################################
+# Also define a production target which doesn't use devDeps
+FROM base as production
+WORKDIR /home/node/app
+# We'll install production only deps, and target the Linux Musl x64 for alpine.
+RUN npm install --platform=linuxmusl --arch=x64 --production
+
+## Deploy ######################################################################
+# Use a smaller node image (-alpine) at runtime
+FROM node:lts-alpine as deploy
+# Install dumb-init, see:
+# https://snyk.io/blog/10-best-practices-to-containerize-nodejs-web-applications-with-docker/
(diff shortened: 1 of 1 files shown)


src/api/auth/Dockerfile
```diff
@@ -1,12 +1,50 @@
-FROM node:lts-alpine as base
-RUN apk add dumb-init
-WORKDIR /app
-COPY --chown=node:node ./package.json ./
+## Base ########################################################################
+# Use a larger node image to do the build for native deps (e.g., gcc, pytyhon)
+FROM node:lts as base
```
manekenpix, line 3: Is this better than using alpine as base and installing the dependencies that we need? I'm asking because we're not only downloading 2 versions of node for this (not the best in terms of build time), but also the full version (node: lts) we might get a lot more than just the dependencies we need.

src/api/auth/Dockerfile
```diff
@@ -1,12 +1,50 @@
...
+FROM base as development
+WORKDIR /home/node/app
+# Install (not ci) with dependencies, and for Linux vs. Linux Musl (which we use for -alpine)
+RUN npm install
+# Copy the source code over
+COPY --chown=node:node . /home/node/app/
+# Switch to the node user vs. root
+USER node
+# Start the app in debug mode so we can attach the debugger
+CMD ["npm", "run", "dev"]
```
manekenpix, line 25: how can I test this?

humphd: When you do a `build`, you specify a `--target <stage name>` and it will build that stage only:

```sh
docker build --target development -t auth:dev .
```

See https://docs.docker.com/develop/develop-images/multistage-build/#stop-at-a-specific-build-stage

src/api/auth/Dockerfile
humphd: It's pretty common to use this approach.  True, you have to download 2 base layers, but they will get cached and re-used over and over again.  Then you can have a full environment for creating the release, but run it on a much smaller layer.

src/api/auth/Dockerfile
humphd: In a development `docker-compose.yml`, you'd use it like this:

```yml
services:
  auth:
    # Override the service so it builds the Dockerfile in development mode
    build:
      context: src/api/auth
      dockerfile: Dockerfile
      target: development
```

src/api/auth/Dockerfile
manekenpix: > but they will get cached and re-used over and over again

Not for staging and production, that's why I mentioned build time. We use fresh everything for every deployment, in case there are bug fixes that have been released upstream. I've seen the 1 layer + dependencies in production more often than the 2 layers approach, but our build time is not crazy long so maybe it's ok?

humphd: We're moving toward building all the images and pushing to a registry (cc @TDDR), so this will become a non-issue, since the build won't happen on staging/production once it's done.  Until then, I don't think this is a huge problem.

TDDR (commented):
I don't know if I'm doing it wrong, but checkout out this pr and tried to test it like you said to and it would not build for me. I don't have a Mac, but I figured it should still build for PC, right?

image

cindyledev:
I tried to copypasta this in my PR #2776 and the lines below `## Production...` gave me the same error. Complained about some frontend thing.

humphd:
You need to run that command in src/api/auth, so it can find the Dockerfile
that goes with it.

manekenpix (approved):
:+1: 
Tested both builds.
For development, I added this:
```yml
build:
  target: development
```
to sso in `docker/development.yml`

humphd:
@TDDR do you want to do the other review of this?

TDDR:
> You need to run that command in src/api/auth, so it can find the Dockerfile that goes with it.

I tried this but it's giving me a different, but similar error 
image

humphd:
We very recently renamed `src/api/auth` to `src/api/sso` so we can land Supabase, which also uses an `auth` container.  If you go into `src/api/sso` and do that, does it work?

TDDR (approved):
Was able to build it on PC.

PR #2750, opened 2022-01-27
Fixing wrong build status url

### LINKED ISSUE #2747: Wrong url for fetching build status from the autodeployment server (Production)
**What happened**:
In **production**, the build log page fails to fetch build status information from the autodeployment server

image

The build log page should display commit, status and build duration

Check the build log page in telescope's dashboard (production)

I did some testing and the issue seems to be [here](https://github.com/Seneca-CDOT/telescope/blob/855b39fa8a7d6e7020f2fe7f7a345607a0da9861/src/api/status/public/js/build-log/check-build-status.js#L5).
This works for `dev.api.telescope`, which results in `dev.telescope`, but in production it fails to turn `api.telescope` into `telescope` since the text to be replaced is `.api`.

We could probably use a regex to address this.

**Environment**:

- OS: Manjaro
- Browser: Firefox

tpmai22:
## Type of Change

## Description
Fixed #2747 

You can review the `regex` [here](https://regex101.com/r/nIin6v/1)

## Checklist

### DIFF
--- a/src/api/status/public/js/build-log/check-build-status.js
+++ b/src/api/status/public/js/build-log/check-build-status.js
@@ -2,7 +2,7 @@
 // being loaded. On staging and production, remove the `api.` subdomain.
 // On localhost:1111, use localhost (drop the port).
 const autodeploymentUrl = (path) =>
-  `//${window.location.hostname.replace('.api', '')}/deploy${path}`;
+  `//${window.location.hostname.replace(/(api\.|\.api)/, '')}/deploy${path}`;
 
 const getBuildLog = async (buildName) => {
   if (!(buildName === 'current' || buildName === 'previous')) {


src/api/status/public/js/build-log/check-build-status.js
```diff
@@ -2,7 +2,7 @@
 // being loaded. On staging and production, remove the `api.` subdomain.
 // On localhost:1111, use localhost (drop the port).
 const autodeploymentUrl = (path) =>
-  `//${window.location.hostname.replace('.api', '')}/deploy${path}`;
+  `//${window.location.hostname.replace(/(?:api\.|\.api)/gm, '')}/deploy${path}`;
```
humphd, line 5: I think this can be simplified to `/(api\.|dev\.api\.)/` so it gets rid of the `api` or `dev.api` bits

humphd (changes requested):

src/api/status/public/js/build-log/check-build-status.js
humphd: Yes, you're right, so it could be simplified to `/api\./`, right?

tpmai22: @humphd Follow what I read [here](https://www.ocpsoft.org/tutorials/regular-expressions/or-in-regex/) 

> If you don’t want your grouping and alternation to interfere with other numbered groups in your expression, each “or” group must be prefixed with `?:`

I guess I try to avoid any cache if not needed, or do you think `/api\./` should be enough without any performance issue ?

humphd: There's zero performance issue for a regex on a string of this size.

src/api/status/public/js/build-log/check-build-status.js
```diff
@@ -2,7 +2,7 @@
 // being loaded. On staging and production, remove the `api.` subdomain.
 // On localhost:1111, use localhost (drop the port).
 const autodeploymentUrl = (path) =>
-  `//${window.location.hostname.replace('.api', '')}/deploy${path}`;
+  `//${window.location.hostname.replace(/(api\.|\.api)/gm, '')}/deploy${path}`;
```
humphd, line 5: You don't need `/gm` flags.

humphd (changes requested):

tpmai22:
I think we are good to go now

humphd (approved):

cindyledev (approved):

PR #2751, opened 2022-01-27
Modified regex to match user only

### LINKED ISSUE #2707: GitHub Users tagged in posts are not showing under user/organization
**What happened**:
My github profile was linked in Luigi's post, but I do not see my avatar under user/organization
image

I should see my avatar for GitHub under the user/organization. Like this post: 
image

If you add a blog with a link to a github profile, it will not show up. If you add a repository owned by a user (not organization), the user will show up under user/orgs. This is confusing, the reference was to the repository, which already has its own section on the side. 

The posts are different because the correct one includes a link to open parliament `https://github.com/michaelmulley/openparliament`, however when we look at the user/orgs we are directed to Michael Mulley's GitHub Page... This can be a bit confusing because I think that Michael Mulley is involved in the post, not his repository (the repo is already in the repository section).

Can we look into whether the users are the repositories for the users, or the actual users' page? It would be cool if we can tag each other's GitHub in the posts. I was considering this when writing mine, but wasn't too sure if it picks up on Telescope... (it doesn't). 

**Environment**:

- OS:
- Browser:

Kevan-Y:
## Issue This PR Addresses
Fixes #2707

## Type of Change

## Description

Modified the regex that filters GitHub URL to find GitHub Users tagged in posts to be shown under user/organization
### Before
image 
### After 
image
## Checklist


AmasiaNalbandian (dismissed):
Great Work @Kevan-Y  💯 

Look at this: 
The link I hover on here is from michaels repository (you can see link at bottom left):
chrome_8NfDKNEFw6

That link adds both himself under users, and his repository under respository.
chrome_FUDZFwUFP4

Kevan-Y:
Edge case found,
```
Example:
https://github.com/issues
https://github.com/marketplace
https://github.com/features/codespaces
```
Those will still be able to display in the Github info.

humphd (commented):
This is the sort of thing that is so easy to test, we should add some for `filterGitHubUrls` in a follow-up.

Kevan-Y:
Don't very have a good solution right now besides hardcoded. Like have a list of reserved names, and do a check before adding the list of repo or org.

The second solution is to call GitHub API to check one by one but might run out of the limit in a sec.

humphd:
I don't think we should call the GitHub API.  Having a list of ones we know are static endpoints would work.  It doesn't need to be perfect, just "better".

aserputov (approved):

AmasiaNalbandian (approved):

aserputov:
@humphd can we merge this one ?

aserputov (approved):

PR #2752, opened 2022-01-27
Rename auth container to sso, to prepare for supabase auth container

humphd:
We're hoping to add Supabase in #2743, and to get it integrated with our auth stuff, we have to run the containers on the same Docker network.  The problem is, both Telescope and Supabase have a container named `auth`.

This PR updates our `auth` container to be called `sso` instead.  I've changed a bunch of env variables to match.  I have left all the `/v1/auth` endpoints as they are, since changing this will break our link with Seneca's SSO, and I don't want to have to recreate it.

So our Auth service will henceforth be known as the "SSO Auth" service, and be named `sso`.  The `auth` service will be Supabase's.

I'm hoping to get @DukeManh to rebase #2743 on top of this once we land it.


DukeManh (approved):
Thanks for doing this.

HyperTHD (approved):

Issue #2621, opened 2021-12-16
Add tests for search 


AmasiaNalbandian:
**What would you like to be added**:
Once we complete the searchbar, we need to add tests to the files to ensure they remain correct when making any changes. 

Please reference: https://github.com/Seneca-CDOT/telescope/tree/master/src/api/search

**Why would you like this to be added**:
We need tests for this section of the code, currently there are none.

PR #2757, opened 2022-01-27
initial setup for search test

JiaHua-Zou:
## Issue This PR Addresses
 This PR fixes part of the #2621

## Type of Change

## Description

initial setup files for the search tests.
So far it only has 1 simple test and it test the status if no params are given. More test will be added:
- testing the number of results return.
- testing for status return of 200 and 503.
- testing for giving part of the params.
- testing for max page number
- testing for author filter

To run it: `pnpm test search`
## Checklist

### DIFF
--- a/src/api/search/jest.config.js
+++ b/src/api/search/jest.config.js
@@ -1,0 +1,9 @@
+const baseConfig = require('../../../jest.config.base');
+
+module.exports = {
+  ...baseConfig,
+  rootDir: '../../..',
+  setupFiles: ['<rootDir>/src/api/search/jest.setup.js'],
+  testMatch: ['<rootDir>/src/api/search/test/*.test.js'],
+  collectCoverageFrom: ['<rootDir>/src.api/search/src/**/*.js'],
+};
--- a/src/api/search/test/query.test.js
+++ b/src/api/search/test/query.test.js
@@ -1,0 +1,9 @@
+const request = require('supertest');
+const { app } = require('../src');
+
+describe('/query routers', () => {
+  it('return error 400 if no params given', async () => {
+    const res = await request(app).get('/');
+    expect(res.status).toBe(400);
+  });
+});
--- a/src/api/search/jest.setup.js
+++ b/src/api/search/jest.setup.js
@@ -1,0 +1,1 @@
+process.env = { ...process.env, MOCK_ELASTIC: '1' };


humphd:
@JiaHua-Zou can you add a proper description for this, discuss what it does, how to test it, what it's missing (e.g., what's coming next), which issues this fixes or is part of, etc.

src/api/search/test/query.test.js
```diff
@@ -0,0 +1,9 @@
+const request = require('supertest');
+const { app } = require('../src');
+
+describe('Testing routers', () => {
```
humphd, line 4: Name this `'/query routes'`

src/api/search/test/query.test.js
```diff
@@ -0,0 +1,9 @@
...
+  it('return error 400', async () => {
```
humphd, line 5: Say something like `'should return a 400 if no params are given'`

humphd (changes requested):
Looks pretty close, a couple small bits to fix.

rclee91 (dismissed):

src/api/search/jest.config.js
```diff
@@ -0,0 +1,8 @@
+const baseConfig = require('../../../jest.config.base');
+
+module.exports = {
+  ...baseConfig,
+  rootDir: '../../..',
```
humphd, line 5: I might be wrong, but I think you have to specify the `jest.setup.js` file here, like this: https://github.com/Seneca-CDOT/telescope/blob/19e4dfc9587d17b4948a7229ff4f5b6d5c0f02e6/test/jest.config.js#L5

We need to do the same for the posts.  If this is working already, I'm not clear why it works without setting this.  The [docs](https://jestjs.io/docs/configuration#setupfiles-array) don't lead me to believe this would happen automatically.

rclee91: Might be because we don't have Elasticsearch mocks in the tests yet to see if its setup correctly.

humphd: But why do the Posts tests work, which also mock redis this way, but don't include the setup script in the config?  Seems like a bug to me.

JiaHua-Zou: So I checked all of the other config test. All of them did not have to specify the `jest.setup.js`

humphd: Right, we only need it very rarely.  I'm saying that I think we are doing it wrong everywhere else.

rclee91: I think somehow `MOCK_REDIS` and `MOCK_ELASTIC` are already set to `'1'`
I printed out `process.env` on the master branch locally (without the tests for search)
image

humphd: You're right, it's happening here https://github.com/Seneca-CDOT/telescope/blob/master/jest.config.base.js#L4.

I think it's a good idea to not depend on this, so we should be adding the `setupFiles` key to your search test config.  In another PR, we should do the same in the posts tests.  Can one of you do that too?

rclee91: Yup, I'll do it for Posts

aserputov: @rclee91 if I understood correctly for this PR we need to include `setupFiles: ['<rootDir>/src/api/search/jest.setup.js']`?

rclee91: Yup, looks like it

humphd: We should get this finished, as it blocks all other test work from landing.  @JiaHua-Zou can you wrap it up?

JiaHua-Zou: sorry didn't see this message. I just push a new PR

rclee91 (approved):
LGTM

HyperTHD (commented):
Looks good. mdx being broken means master will break if we merge this in which is why I wanna wait to approve it

humphd (approved):

aserputov (approved):

PR #2763, opened 2022-01-28
Fixes #2671: Add vite to dashboard

### LINKED ISSUE #2671: Add a bundler to dashboard
Options:
- [Vite](https://vitejs.dev/)
- [Snowpack](https://www.snowpack.dev/)
- [Parcel](https://parceljs.org/)

These bundlers work out-of-the-box. For simple use cases, they require little to zero configs. Speed is a plus, but we only have vanilla js so it is not really an issue.

- Mix `import` and `require`
- Polyfill cutting-edge syntax.
- Browsers only download one js file instead of multiple modules (because of how js native modules work).

Andrewnt219:
## Issue This PR Addresses

Closes #2671

## Type of Change

## Description

Bundle client-side javascript with [Vite](https://vitejs.dev/). This also allows mixing `.cjs` and `.js`.

**Changelog**
- Setup vite.config.js
- Update build script and dev script
- Remove global "toast.js" import
- Update path to the entry js file of each page
- Use logger for server errors
- Stop page render on server's failures

**Test plan**
- Run `pnpm dev`
- Everything should work as before

## Checklist


src/api/status/src/views/builds.hbs
```diff
@@ -92,4 +92,4 @@
 <script src="https://cdn.jsdelivr.net/npm/xterm-addon-fit@0.5.0/lib/xterm-addon-fit.min.js"></script>
 <script src="https://cdn.jsdelivr.net/npm/xterm-addon-web-links@0.4.0/lib/xterm-addon-web-links.min.js"></script>
 <script src="https://cdn.jsdelivr.net/npm/xterm-addon-webgl@0.11.3/lib/xterm-addon-webgl.min.js"></script>
```
humphd, line 94: All this other JS should get done in the bundle step vs. included as separate files.  You can do that in a follow-up if you like, but you need to file it now if so.

src/api/status/src/views/status.hbs
```diff
@@ -298,4 +298,4 @@
 <script src="assets/js/material-dashboard.min.js?v=3.0.0"></script>
```
humphd, line 298: Same deal here, we should bundle this.

src/api/status/vite.config.js
```diff
@@ -0,0 +1,16 @@
+import { defineConfig } from 'vite';
+
+// vite.config.js
+export default defineConfig({
+  build: {
+    outDir: './public/dist',
+    rollupOptions: {
+      output: {
+        // Prevent hashed file names
```
humphd, line 9: Why prevent this?

humphd (changes requested):
Fantastic work!  I'm so excited to see this coming along.

src/api/status/vite.config.js
Andrewnt219: With hashed file name (e.g. dashboard.abc123.js), the server has to read from vite manifest.json file to get the correct file name. It's simple, I did that initially but I thought that was unnecessary. Maybe it can cause a problem with browser's caching?

src/api/status/src/views/builds.hbs
Andrewnt219: Yeah I'll do that in a follow-up. There'll be a `main.js`, and it also utilizes Vite for sass so we won't need a separate scss compiler anymore.

src/api/status/package.json
```diff
@@ -4,8 +4,11 @@
   "version": "1.0.0",
   "description": "A service for getting the status of our services",
   "scripts": {
-    "dev": "npm run compile:scss && env-cmd -f env.local nodemon src/server.js",
+    "dev": "concurrently \"npm run watch:js\"  \"npm run compile:scss\" \"env-cmd -f env.local nodemon src/server.js\"",
```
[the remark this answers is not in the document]
Andrewnt219: Ok, let me set this up.

src/api/status/src/views/builds.hbs
humphd: Great, file an issue so we don't lose track.

src/api/status/vite.config.js
humphd: It enables browser caching. When you include the hash in the filename, the browser can cache that bundle forever, since it will never change (i.e., a new build will include a new hash).  In general, you want this.

Kevan-Y (dismissed):
Very nice feature.

src/api/status/src/server.js
```diff
@@ -114,8 +136,20 @@ service.router.get(process.env.PATH_PREFIX || '/', async (req, res) => {
...
 // Get Build page
-service.router.get(`${process.env.PATH_PREFIX || ''}/build`, (req, res) => {
-  res.render('builds', { active_build_log: true, headers: { title: 'Telescope Build log' } });
+service.router.get(`${process.env.PATH_PREFIX || ''}/build`, async (req, res) => {
+  let distJsPath = '';
+
+  try {
+    distJsPath = await getDistJsPath('public/js/pages/build.js');
+  } catch (error) {
+    console.error(error);
```
humphd, line 145: This is a bit weak.  First, use `logger` from Satellite and log it properly.  Second, if we don't have this value, should we have crashed earlier with an error (e.g., on startup)?

humphd (changes requested):

src/api/status/src/server.js
Andrewnt219: It won't crash, just 404 the script file. How about
```js
try {
 const distJsPath = await getDistJsPath();
 return res.render();
} catch(error) {
  log(error);
}

return res.send("Fail to render /build");
```

humphd: `logger.warn({ error }, 'unable to get dist/js path')`

Andrewnt219: For the crashing if it fails. Can we just send back a response instead of rendering the page? (like in the snipppet)

humphd: Sure, just so it doesn't take the server down.

humphd (approved):

Kevan-Y (approved):
LGTM

humphd:
Fantastic

Andrewnt219:
I was expecting the staging to break, too bad.

Andrewnt219:
Just realize if something goes wrong, the logger will bring down the whole server
https://github.com/Seneca-CDOT/telescope/blob/ed156ac1051a55b7b90715cf137b4e021e40c1a4/src/api/status/src/server.js#L127

humphd:
Why will that bring down the whole server?

Andrewnt219:
Shouldn't it be `logger.warn()` instead? Sending a patch now.

humphd:
Ah, yes.  Can you fix in a follow-up?

PR #2766, opened 2022-01-28
Bump expo deps in src/mobile

humphd:
I tried @menghif's work on `src/mobile` locally, and it works!

Screen Shot 2022-01-28 at 10 30 57 AM

While I was playing with it, I noticed that the deps need to be updated.  We also have a `node-fetch` issue like before with 2.6.6 that still needs to be fixed somehow.

### DIFF
--- a/src/mobile/package.json
+++ b/src/mobile/package.json
@@ -10,15 +10,15 @@
     "eject": "expo eject"
   },
   "dependencies": {
-    "expo": "44.0.0",
+    "expo": "44.0.5",
     "expo-status-bar": "1.2.0",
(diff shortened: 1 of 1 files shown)


DukeManh:
@humphd 
>  We also have a node-fetch issue like before with 2.6.6 that still needs to be fixed somehow.

Is that the expo's peer dep?

humphd:
Not sure, I didn't spend too much time digging into it.

DukeManh (approved):

HyperTHD (approved):

PR #2769, opened 2022-01-28
Adapt syntax highlighting on dark theme

### LINKED ISSUE #2377: Improve dark mode code colours
Our dark mode isn't great for `<code>...</code>` (i.e., inline) and `<pre><code>...</code></pre>` (i.e., block):

Screen Shot 2021-10-21 at 12 30 18 PM

We should probably take a look at how GitHub and other sites render code in dark mode, and see what we can borrow.

dbelokon:
## Issue This PR Addresses

Fixes #2377.

## Type of Change

## Description

All code blocks that the Telescope back-end sends us are annotated with highlight.js classes, so the front-end has to provide the stylesheet that defines these.

Since the stylesheet is global, we have to link both stylesheets for light and dark theme, and disable either depending on the current theme.

## Checklist

- [ ] **Tests**: This PR does not includes tests as it does not add new functionality or change it.
- [X] **Screenshots**: This PR includes screenshots or GIFs of the changes made: 
image
image

- [ ] **Documentation**: This PR does not include documentation as it does not add user exposed functionality. However, a couple of comments were provided in the code as well in the commit message to explain certain changes.

### DIFF
--- a/src/web/public/styles/github-dark.min.css
+++ b/src/web/public/styles/github-dark.min.css
@@ -1,0 +1,10 @@
+pre code.hljs{display:block;overflow-x:auto;padding:1em}code.hljs{padding:3px 5px}/*!
+  Theme: GitHub Dark
+  Description: Dark theme as seen on github.com
+  Author: github.com
+  Maintainer: @Hirse
+  Updated: 2021-05-15
+
+  Outdated base version: https://github.com/primer/github-syntax-dark
+  Current colors taken from GitHub's CSS
+*/.hljs{color:#c9d1d9;background:#0d1117}.hljs-doctag,.hljs-keyword,.hljs-meta .hljs-keyword,.hljs-template-tag,.hljs-template-variable,.hljs-type,.hljs-variable.language_{color:#ff7b72}.hljs-title,.hljs-title.class_,.hljs-title.class_.inherited__,.hljs-title.function_{color:#d2a8ff}.hljs-attr,.hljs-attribute,.hljs-literal,.hljs-meta,.hljs-number,.hljs-operator,.hljs-selector-attr,.hljs-selector-class,.hljs-selector-id,.hljs-variable{color:#79c0ff}.hljs-meta .hljs-string,.hljs-regexp,.hljs-string{color:#a5d6ff}.hljs-built_in,.hljs-symbol{color:#ffa657}.hljs-code,.hljs-comment,.hljs-formula{color:#8b949e}.hljs-name,.hljs-quote,.hljs-selector-pseudo,.hljs-selector-tag{color:#7ee787}.hljs-subst{color:#c9d1d9}.hljs-section{color:#1f6feb;font-weight:700}.hljs-bullet{color:#f2cc60}.hljs-emphasis{color:#c9d1d9;font-style:italic}.hljs-strong{color:#c9d1d9;font-weight:700}.hljs-addition{color:#aff5b4;background-color:#033a16}.hljs-deletion{color:#ffdcd7;background-color:#67060c}
--- a/.prettierignore
+++ b/.prettierignore
@@ -7,6 +7,7 @@ certs
 .next
 out
 pnpm-lock.yaml
+*.min.css
 
 # We don't maintain these files
 src/api/status/public/assets
--- a/src/web/public/styles/github.min.css
+++ b/src/web/public/styles/github.min.css
@@ -1,0 +1,10 @@
+pre code.hljs{display:block;overflow-x:auto;padding:1em}code.hljs{padding:3px 5px}/*!
+  Theme: GitHub
+  Description: Light theme as seen on github.com
+  Author: github.com
+  Maintainer: @Hirse
+  Updated: 2021-05-15
+
+  Outdated base version: https://github.com/primer/github-syntax-light
+  Current colors taken from GitHub's CSS
+*/.hljs{color:#24292e;background:#fff}.hljs-doctag,.hljs-keyword,.hljs-meta .hljs-keyword,.hljs-template-tag,.hljs-template-variable,.hljs-type,.hljs-variable.language_{color:#d73a49}.hljs-title,.hljs-title.class_,.hljs-title.class_.inherited__,.hljs-title.function_{color:#6f42c1}.hljs-attr,.hljs-attribute,.hljs-literal,.hljs-meta,.hljs-number,.hljs-operator,.hljs-selector-attr,.hljs-selector-class,.hljs-selector-id,.hljs-variable{color:#005cc5}.hljs-meta .hljs-string,.hljs-regexp,.hljs-string{color:#032f62}.hljs-built_in,.hljs-symbol{color:#e36209}.hljs-code,.hljs-comment,.hljs-formula{color:#6a737d}.hljs-name,.hljs-quote,.hljs-selector-pseudo,.hljs-selector-tag{color:#22863a}.hljs-subst{color:#24292e}.hljs-section{color:#005cc5;font-weight:700}.hljs-bullet{color:#735c0f}.hljs-emphasis{color:#24292e;font-style:italic}.hljs-strong{color:#24292e;font-weight:700}.hljs-addition{color:#22863a;background-color:#f0fff4}.hljs-deletion{color:#b31d28;background-color:#ffeef0}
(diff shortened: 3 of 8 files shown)


src/web/src/pages/_app.tsx
```diff
@@ -27,11 +27,27 @@ const App = ({ Component, pageProps }: AppProps) => {
     if (jssStyles) {
       jssStyles.parentElement?.removeChild(jssStyles);
     }
-  }, []);
+
+    // Enable the stylesheet for the syntax highlighting depending on the theme
+    if (preferredTheme === 'dark') {
+      document.querySelector('#light-stylesheet')?.setAttribute('disabled', 'disabled');
```
humphd, line 33: `.disabled = true` see https://developer.mozilla.org/en-US/docs/Web/API/StyleSheet/disabled

humphd: Question: should this happen within a `useEffect` @DukeManh?

src/web/src/pages/_app.tsx
```diff
@@ -27,11 +27,27 @@ const App = ({ Component, pageProps }: AppProps) => {
...
+      document.querySelector('#dark-stylesheet')?.setAttribute('disabled', 'disabled');
+    }
+  });
 
   // Switch the active theme, and also store it for next load
+  // We also disable the current stylesheet and enable the other one.
   const toggleTheme = () => {
-    setPreferredTheme(preferredTheme === 'dark' ? 'light' : 'dark');
+    if (preferredTheme === 'dark') {
+      document.querySelector('#light-stylesheet')?.removeAttribute('disabled');
```
humphd, line 43: We're repeating code here, let's consider moving this enable/disable logic to function(s), and maybe even put it in `hooks/use-preferred-theme` so it happens when we get/set the theme.

humphd (changes requested):
I'm thrilled to see us fixing this, thank you!

A couple of things to consider.

src/web/src/pages/_app.tsx
DukeManh: Yeah, it should be useEffect of hooks/use-preferred-theme

DukeManh (changes requested):
Why haven't we done this earlier. Thank @dbelokon.

One thing, let's use the minified css files to save loading time. https://highlightjs.org/download/

TueeNguyen:
I like these inline format more than the old one 
image

src/web/public/styles/github-dark.min.css
```diff
@@ -0,0 +1,10 @@
+pre code.hljs{display:block;overflow-x:auto;padding:1em}code.hljs{padding:3px 5px}/*!
```
humphd, line 1: Is it possible for us to get these via `node_modules/` vs. embedding directly in our code?  Then we can get updates.

humphd (changes requested):

src/web/public/styles/github-dark.min.css
dbelokon: So, as I understand how `next.js` deals with static files and CSS style sheets in particular, it might not be possible with the tools we have available...

If your style sheets were in `src/web/src/styles`, you can `@import` a style sheet found in `node_modules`, but `next.js` modifies the original file so that the `@import` rule is not there anymore. Thus, we cannot access the `github.css` and `github-dark.css` files from the `@import` rules as shown in [CSSImportRule](https://developer.mozilla.org/en-US/docs/Web/API/CSSImportRule). And, even if `next.js` kept the `@import` rule, we wouldn't be able to find the original style sheet with an `id`, because the `style` element generated by `next.js` does not have an `id`.

If you place the style sheets in `src/web/public`, you cannot `@import` a style sheet found in `node_modules`.

So, with my current understanding, it might not be possible. If someone knows a lot about `next.js`, please let us know if you can think of a way to make this possible.

dbelokon: In the worst case, we would have to use a dependency that copies certain files in `node_modules` during `next.js` compilation time for this specific case.

src/web/src/pages/_app.tsx
```diff
@@ -27,7 +27,7 @@ const App = ({ Component, pageProps }: AppProps) => {
     if (jssStyles) {
       jssStyles.parentElement?.removeChild(jssStyles);
     }
-  }, []);
```
humphd, line 30: Why remove this?  Won't this mean it runs on every (re)render?

humphd (commented):
I'm surprised there's no way to do this with next.js, but if it can't be done, we can go this route.  One question.

src/web/src/pages/_app.tsx
dbelokon: Yeah, I deleted by accident. Fixed now!

dbelokon:
> I'm surprised there's no way to do this with next.js, but if it can't be done, we can go this route.

I am not very experienced with `next.js`, so I do not know the limits of it and I cannot research more because Google is giving me the same pages I already checked. I could try to ask the `next.js` developers about something like this, if we do not have someone else that might have an idea

humphd:
> > I'm surprised there's no way to do this with next.js, but if it can't be done, we can go this route.
> 
> I am not very experienced with `next.js`, so I do not know the limits of it and I cannot research more because Google is giving me the same pages I already checked. I could try to ask the `next.js` developers about something like this, if we do not have someone else that might have an idea

cc @DukeManh, @Andrewnt219 for thoughts

aserputov:
@dbelokon I can help you. Let's ask people in the chat, otherwise we will think something.

Andrewnt219:
hey @dbelokon, have you tried this?

1. Add class 'dark' to the `body` tag
2. Switch CSS style base on theme
```css
.token {
 color: black
{

.dark .token {
 color: white
}
```  

You can add the script to add toggle theme class like this in `_document.tsx`
https://github.com/Andrewnt219/blog-andrewnt/blob/main/src/pages/_document.tsx
And the script is
https://github.com/Andrewnt219/blog-andrewnt/blob/main/src/modules/color-theme/ColorThemeScript.tsx

DukeManh:
@dbelokon, I think you misunderstood the import CSS part. In React/Next, CSS files can be imported like a module in JS like so: `import 'highlight.js/styles/github.css`. What you are talking about is importing a CSS file into another, correct?

@humphd, I don't think importing CSS files from node_modules would work for us because we want to give them `ids` so we can select and disable them later. I don't know how to give ids to imported CSS files.

humphd:
What if you do dynamic imports and only pull in the one you need, when you need it?

DukeManh:
Would it work for CSS file? 
Why don't you give it a try @dbelokon, https://nextjs.org/docs/advanced-features/dynamic-import

dbelokon:
So, it seems that `dynamic import` might [not work](https://github.com/vercel/next.js/discussions/17068) either :(

As I understand, it seems `next.js` treats the `import 'path/to/style.css;` statements as an exception more than a rule. That means, `import()` will not work.

The main problem is that the CSS style sheet has to be a global one, because otherwise the elements that are annotated with the class names won't get styled at all. So, if you have two global sheets that apply to the same class names, you need to disable one of them for the other to work. If the styling was bundled in a React component, this would be much easier, but of course, we wouldn't have the problem of the style sheet being global anyway.

humphd (dismissed):
@dbelokon I feel like we're holding up "good" work in an effort to have "perfect," and this is the wrong approach: we should take what you've done, which is awesome, and iterate on it later.

Thanks for entertaining these discussions as we look for a proper way to solve this.   It was worth researching.

aserputov (dismissed):

dbelokon:
Hahaha thank you @humphd!!

Going to rebase it shortly

dbelokon:
Rebased and squashed without conflicts!

humphd (commented):
Why is the light background gray now vs. white?

Screen Shot 2022-02-01 at 4 40 01 PM

dbelokon:
> Why is the light background gray now vs. white?

I did it because it looked a little bit better, especially with GitHub's syntax highlighting. This is my opinion, though 😅 

I can change it back to white background, if you don't mind

dbelokon:
It looks a bit more "separated" from the other content when the code has grey background color vs. white.

Dark theme:

image


White theme:

image

humphd:
For a "light" theme, I think white is the way to go.  It helps with text contrast with accessibility.

aserputov (approved):

humphd (changes requested):
This is only partially fixed: inline `<code>` is still gray:

Screen Shot 2022-02-01 at 5 58 28 PM

humphd (approved):

aserputov (approved):

Issue #1743, opened 2021-02-14
Automate Docker build of Feed-Discovery service with GitHub Actions


humphd:
We're moving toward containerizing everything, and it might be time to consider using an [automated container build setup](https://github.com/docker/build-push-action) with Actions so we can build and push to a registry, then automate our deployments to pull for zero-downtime updates.

[GitHub has a new Docker registry](https://docs.github.com/en/actions/guides/publishing-docker-images), but it looks like the current Seneca-CDOT account [doesn't support it](https://github.com/orgs/Seneca-CDOT/packages)?  @ctyler do you know anything about CDOT's GitHub org being a "legacy" account?

![Screen Shot 2021-02-14 at 3 08 21 PM](https://user-images.githubusercontent.com/427398/107887781-7fbbc180-6ed6-11eb-8961-4f72a852fcaf.png)

I'm not sure if this is fixable or not, but it would be cool to use GitHub's storage for this.  I don't think there's another registry that will let us host all these containers for free?

birtony:
I'd be keen to look into this

humphd:
https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry

humphd:
See https://github.com/Seneca-CDOT/telescope/discussions/2683

humphd:
I updated this to be about getting the feed-discovery service built and the image uploaded to the GitHub Package Registry.  Once we figure this out for feed-discovery, we can move onto the others.

humphd:
Some useful background reading:

- [How to build and push to GitHub's Package Registry with GitHub Actions](https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions)
- [Example of building multiple Docker images in a monorepo](https://ianbelcher.me/tech-blog/using-github-actions-and-package-registry-to-build-store-and-deploy-your-containers)
- [Discussion of how to re-use Docker code in GitHub Actions workflows with the build matrix](https://stackoverflow.com/questions/68608324/github-actions-multiple-workflows-for-parallel-image-build)

humphd:
@TDDR has done great work getting us to the point of being able to do this.  However, I've discovered that the CDOT org isn't allowed to publish to the GitHub Package Registry

As I see it, our options are:

[...]

PR #2775, opened 2022-01-29
2nd draft of workflow

TDDR:
## Issue This PR Addresses
Part Of: #1743  

## Type of Change

- [x]  **New Feature**: Change which adds functionality

## Description

This is an attempt to add a workflow that will build and push our microservice image to the GitHub Container Registry. I am unsure of a couple of lines that I will make a comment on. 

I also do not know how I would test locally. Would I not have to test it right in staging in order for the workflow to trigger?

### DIFF
--- a/.github/workflows/deploy-feed-discovery-image.yaml
+++ b/.github/workflows/deploy-feed-discovery-image.yaml
@@ -1,0 +1,26 @@
+name: build-and-push-feed-discovery
+
+on:
+  push:
+    branches:
+      - 'issue#1743'
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          context: src\api\feed-discovery
+          dockerfile: Dockerfile
+          push: true
+          tags: user/app:latest
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -422,6 +422,14 @@ importers:
       rereadable-stream: 1.4.13
       shelljs: 0.8.5
 
+  tools/migrate:
+    specifiers:
+      jsdom: ^18.0.1
+      node-fetch: 2.6.7
+    dependencies:
+      jsdom: 18.1.1
+      node-fetch: 2.6.7
+
 packages:
 
   /@apideck/better-ajv-errors/0.3.2_ajv@8.9.0:
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -5,4 +5,4 @@ packages:
   # microservices
   - 'src/api/*'
   # autodeployment
-  - 'tools/autodeployment'
+  - 'tools/*'
--- a/src/web/src/hooks/use-preferred-theme.ts
+++ b/src/web/src/hooks/use-preferred-theme.ts
@@ -5,7 +5,7 @@ import { useLocalStorage, useMedia } from 'react-use';
  * loads of the app, and initial logic to get the browser's preferred colour.
  */
 export default function usePreferredTheme() {
-  const isDarkThemePreferred = useMedia('(prefers-color-scheme: dark)');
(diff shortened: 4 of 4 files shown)


.github/workflows/deploy-feed-discovery-image.yaml
```diff
@@ -0,0 +1,25 @@
+name: Build and publish a feed-discovery image to ghcr.io
+on:
+  release:
+    types: [published]
+
+  push:
+    branches:
+      - master
```
manekenpix, line 8: Can you try to change this to the name of the branch you're using and check if this works on your fork?

.github/workflows/deploy-feed-discovery-image.yaml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  docker_publish:
+    runs-on: 'ubuntu-20.04'
+
+    steps:
+      - uses: actions/checkout@v2
+
+      # https://github.com/marketplace/actions/push-to-ghcr
+      - name: Build and publish a Docker image for ${{ github.repository }}
+        uses: macbre/push-to-ghcr@master
```
humphd, line 19: I don't think this action is going to work.  They have a bug with uppercase letters in repo names.

Let's use the official Docker one, https://github.com/docker/build-push-action, and we can use https://github.com/docker/login-action to login to ghcr.io

humphd (changes requested):

TDDR:
Not really sure why these two commits are trying to be added again. 
image

I thought the rebase went fine 
image

but the log shows that the new commits got put on top of mine still
image

.github/workflows/deploy-feed-discovery-image.yaml
```diff
@@ -0,0 +1,26 @@
...
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          context: src\api\feed-discovery
```
humphd, line 23: Prefer Unix style paths `src/api/feed-discovery`

.github/workflows/deploy-feed-discovery-image.yaml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push
+        id: docker_build
```
humphd, line 20: I don't think we need an `id`

.github/workflows/deploy-feed-discovery-image.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+on:
+  push:
+    branches:
+      - 'issue#1743'
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
```
humphd, line 11: Let's also include `buildx`, which gives us some benefits when building multi-stage builds:

```
        name: Set up Docker Buildx
        uses: docker/setup-buildx-action@v1
```

See https://docs.docker.com/buildx/working-with-buildx/ for more info.

.github/workflows/deploy-feed-discovery-image.yaml
```diff
@@ -0,0 +1,26 @@
...
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          context: src\api\feed-discovery
+          dockerfile: Dockerfile
+          push: true
+          tags: user/app:latest
```
humphd, line 26: Let's tag this slightly different, so we update the `latest` tag for `feed-discovery` to this new image:

```
tags: feed-discovery:latest
```

I also wonder if we should use `feed-discovery:${{github.sha}}` later on.  But we can add that if we need to go back in time to get old versions.

.github/workflows/deploy-feed-discovery-image.yaml
```diff
@@ -0,0 +1,26 @@
+name: build-and-push-feed-discovery
+
+on:
```
humphd, line 3: We could probably add a path section in here to isolate this to changes in `src/api/feed-discovery` only.  cc @menghif, who just did this for our docs workflows.

humphd (commented):
I see you closed this, but wanted to give some feedback.  It's looking pretty close.  Nice work.

PR #2784, opened 2022-01-31
2nd draft for Registering an Image

### LINKED ISSUE #1743: Automate Docker build of Feed-Discovery service with GitHub Actions

TDDR:
## Issue This PR Addresses

Fixes #1743 

## Type of Change

## Description

This is my second draft of this after attempting to use a different workflow in a [previous iteration](https://github.com/Seneca-CDOT/telescope/pull/2775). I have tried multiple different ways to try and get the buider to go into a sub directory instead of using the Dockerfile in the root with varying results 
image

I think the version in this PR is the proper way to go about it, but it's not supported yet 
image
image

I could be wrong but I'm not sure what else to try.

### DIFF
--- a/.github/workflows/deploy-feed-discovery-image.yaml
+++ b/.github/workflows/deploy-feed-discovery-image.yaml
@@ -1,0 +1,29 @@
+name: Build and push feed-discovery service to ghcr
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'src/api/feed-discovery/**'
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          context: src/api/feed-discovery
+          push: true
+          tags: feed-discovery:latest


.github/workflows/deploy-feed-discovery-image.yaml
```diff
@@ -0,0 +1,25 @@
+name: Build and push feed-discovery service to ghcr
+
+on:
+  push:
+    branches:
+      - 'issue-1743'
```
cindyledev, line 6: I'm very interested in this GitHub Container Registry stuff. Were you able to push an image from your personal issue-1743 branch?

TDDR: Not yet, but having that allows the workflow to trigger when I push it to my fork for testing purposes.

TDDR: A couple if the iterations went through the entire process but failed on the final push, but it was using the wrong Dokerfile anyway. I figure if I can get it to do this on the proper file then when I switch that to master and merge it it *should* work 
image

humphd (changes requested):

.github/workflows/deploy-feed-discovery-image.yaml
```diff
@@ -0,0 +1,25 @@
...
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          context: '{{defaultContext}}:src/api/feed-discovery/Dockerfile'
```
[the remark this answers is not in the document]
humphd: Also, the default `file` value is always `{{context}}/Dockerfile` so you shouldn't need to include that in the context.  I think this should be:

```yaml
context: ./src/api/feed-discovery
```

humphd (changes requested):

.github/workflows/deploy-feed-discovery-image.yaml
```diff
@@ -0,0 +1,27 @@
+name: Build and push feed-discovery service to ghcr
+
+on:
+  push:
+    branches:
+      - 'issue-1743'
```
humphd, line 6: Let's isolate this to changes in the feed-discovery service:

```yaml
on:
  push:
    branches:
      - master
    paths:
      - 'src/api/feed-discovery/**'
```

cc @menghif who just did this for doc changes in https://github.com/Seneca-CDOT/telescope/commit/818ad8e257a66b288f77cbb6c4c4ca07066b4665

humphd (changes requested):
This is looking great.  One more thing I notice.

humphd (dismissed):
Let's try this and see what it does.

NOTE: we'll need to make a change within `src/api/feed-discovery/**` for this to do anything once it lands.  Maybe update one of the deps?

humphd:
@TDDR can you squash, rebase, and reword this?

```
git checkout master
git pull upstream master
git checkout issue-1743
git rebase -i master
...change first commit to reword, second to fixup, save, then rewrite commit message, save again
git push origin issue-1743 -f
```

manekenpix (dismissed):

cindyledev (dismissed):

humphd (approved):

humphd (approved):

TueeNguyen (approved):

aserputov (approved):

PR #2797, opened 2022-02-02
Set up back end to recollect information from pnpm

### LINKED ISSUE #2738: Back-end code to gather dependencies from pnpm data

JerryHue:
## Issue This PR Addresses

Fixes #2738 

## Type of Change

## Description

The service offers a single route: `/projects`, where it returns a list of all project names that Telescope depends on. The service uses a text file to load this information. The file is named `deps.txt`. The one I have copied there is for testing purposes, so that you don't have to run the tool if you can't.

The tool that generates this file is a one-line `bash` script, which calls `pnpm list` and does some stream processing to reduce the information needed.  The tool can be found in the `tools` folder. Since this is a `bash` script, Windows users cannot test it. A PowerShell script may be necessary.

## Steps to test the PR

As of right now, this service is not integrated completely with Docker. To start it up, you have to do the following:

1. `pnpm install` to get all dependencies
2. Navigate to the `src/api/dependency-graph` folder
3. `pnpm start` to start up the application

## Checklist

### DIFF
--- a/src/api/dependency-discovery/src/dependency-list.js
+++ b/src/api/dependency-discovery/src/dependency-list.js
@@ -1,0 +1,14 @@
+const { readFile } = require('fs/promises');
+const { join } = require('path');
+const { cwd } = require('process');
+
+async function getDependencies() {
+  const dependencies = await readFile(join(cwd(), 'deps.txt'), 'utf8');
+
+  // The order of the alternatives is important!
+  // The regex engine will favor the first pattern on an alternation
+  // even if the other alternatives are subpatterns
+  return dependencies.split(/\r\n|\n|\r/).filter((line) => line !== '');
+}
+
+module.exports = getDependencies();
--- a/src/api/dependency-discovery/src/server.js
+++ b/src/api/dependency-discovery/src/server.js
@@ -1,0 +1,5 @@
+const service = require('.');
+
+const port = parseInt(process.env.DEP_DISCOVERY_PORT || 10500, 10);
+
+service.start(port);
--- a/tools/pnpm-deps-gen.sh
+++ b/tools/pnpm-deps-gen.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+# First regex can be read as the following:
+# .*\.pnpm\/(@?[^@]+).*
+# We use the literal .pnpm to find the start of the package name
+# After that, we use the pattern "@?[^@]+" to search for the name
+# that appears first, instead of ".+".
+#
+# Sometimes we could have foo@1.0.0_bar@1.0.0, in which case we want
+# to capture foo. With ".+", we could capture foo@1.0.0_bar.
+# Instead, we search for the string of characters that do not include
+# "@", which translates to "[^@]+". However, some package names include
+# @ at the start, so we cover that case, as well.
+
+set -e
+
+option=$1
+filename=$2
+
+if [[ $# -eq 2 && $option = "-o" && $filename ]]; then
+  echo "Start collecting dependencies"
+  pnpm list -w -r --parseable --depth=0 | sed -n -r -e "s|.*\.pnpm/(@?[^@]+).*|\1|p" -e "s/+/\//" | sort -u > $filename
+  echo "Dependencies collected"
+else
+  echo "command usage: $BASH_SOURCE -o filename"
+fi
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -169,6 +169,17 @@ importers:
       ts-jest: 27.1.2_ac3f9d4a26c88540c3f2823ef15b8533
       turbo: 1.0.28
 
+  src/api/dependency-discovery:
+    specifiers:
+      '@senecacdot/satellite': ^1.x
+      env-cmd: 10.1.0
+      nodemon: 2.0.15
+    dependencies:
+      '@senecacdot/satellite': 1.17.0
+    devDependencies:
+      env-cmd: 10.1.0
+      nodemon: 2.0.15
+
   src/api/feed-discovery:
     specifiers:
       '@senecacdot/satellite': ^1.x
@@ -13466,7 +13477,7 @@ packages:
     dependencies:
       debug: 4.3.3
       module-details-from-path: 1.0.3
-      resolve: 1.21.0
+      resolve: 1.22.0
     transitivePeerDependencies:
       - supports-color
     dev: false
--- a/src/api/dependency-discovery/env.local
+++ b/src/api/dependency-discovery/env.local
@@ -1,0 +1,6 @@
+# Environment variables needed if you want to run `npm run dev`
+# and run the server locally, outside of Docker. For all the various
+# ways we run this using Docker, the env is defined by one of the
+# config/env.* files in the root.  This requires that the other
+# services are running, particularly the login and users services.
+DEP_DISCOVERY_PORT=10500
--- a/src/api/dependency-discovery/package.json
+++ b/src/api/dependency-discovery/package.json
@@ -1,0 +1,28 @@
+{
+  "name": "dependency-discovery",
+  "private": true,
+  "version": "0.1.0",
+  "description": "A dependency graph compilation service for Telescope",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "start": "node src/server.js",
+    "dev": "env-cmd -f env.local nodemon src/server.js"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/Seneca-CDOT/telescope.git"
+  },
+  "author": "",
+  "license": "BSD-2-Clause",
+  "bugs": {
+    "url": "https://github.com/Seneca-CDOT/telescope/issues"
+  },
+  "homepage": "https://github.com/Seneca-CDOT/telescope#readme",
+  "dependencies": {
+    "@senecacdot/satellite": "^1.x"
+  },
+  "devDependencies": {
+    "env-cmd": "10.1.0",
+    "nodemon": "2.0.15"
+  }
+}
--- a/src/api/dependency-discovery/src/index.js
+++ b/src/api/dependency-discovery/src/index.js
@@ -1,0 +1,9 @@
+const { Satellite } = require('@senecacdot/satellite');
+
+const dependenciesRoute = require('./router');
+
+const service = new Satellite();
+
+service.router.use('/', dependenciesRoute);
+
+module.exports = service;
--- a/src/api/dependency-discovery/src/router.js
+++ b/src/api/dependency-discovery/src/router.js
@@ -1,0 +1,15 @@
+const { Router } = require('@senecacdot/satellite');
+const dependencyList = require('./dependency-list');
+
+const router = Router();
+
+router.get('/projects', async (req, res, next) => {
+  try {
+    res.set('Cache-Control', 'max-age=3600');
+    res.status(200).json(await dependencyList);
+  } catch (err) {
+    next(err);
+  }
+});
+
+module.exports = router;


src/api/dependency-graph/src/graph-collector/graph.js
```diff
@@ -0,0 +1,132 @@
...
+   * @param {string} name the name of the project
+   * @param {string} version the version tag
+   * @param {object} peerDependencies collection of dependencies
+   */
+  appendPeerDepsToProjectRelease(name, version, peerDependencies) {
+    const project = this.projectNodes[name];
+
+    const projectPeerDeps = project.releases[version].peerDependencies;
+
+    for (const peerDep in peerDependencies) {
```
TueeNguyen, line 110: use `peerDependencies.forEach()` instead.

TueeNguyen:
I notice a few Eslint issues, fix them first

TueeNguyen (changes requested):

DukeManh (changes requested):
I don't think this is a great way to do it. The dependencies list and their information can be extracted using pnpm list and pnpm view cli, and we can use `child_process.exec` or other libraries to call the cli.
For example

```js
const { exec } = require('child_process');
exec("pnpm list --dep  2 --json", (error, stdout) => {
 stdout.forEach((dep) => {
     exec(`pnpm view ${dep} repository.url`")
})
})
```

That's untested code, hope that helps.
See 
https://docs.npmjs.com/cli/v7/commands/npm-view
https://pnpm.io/cli/list

src/api/dependency-graph/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "dependency-graph",
```
humphd, line 2: Can we rename this?  It's not a `graph` so much as a service for inspecting and discovering our dependencies.  What about calling it `dependency-discovery` and changing the directory name too?

src/api/dependency-graph/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "private": true,
+  "version": "0.1.0",
+  "description": "A dependency graph compilation service for Telescope",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
```
humphd, line 7: We'll need tests for this, either here, or an issue filed to add them.

src/api/dependency-graph/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "start": "node src/server.js"
```
humphd, line 8: Let's also add `nodemon` like the other services do for dev, and include a way to pass a local set of environment variables.  Steal code from another service that does this.

src/api/dependency-graph/package.json
```diff
@@ -0,0 +1,24 @@
...
+  },
+  "author": "",
+  "license": "BSD-2-Clause",
+  "bugs": {
+    "url": "https://github.com/Seneca-CDOT/telescope/issues"
+  },
+  "homepage": "https://github.com/Seneca-CDOT/telescope#readme",
+  "dependencies": {
+    "@senecacdot/satellite": "^1.x",
+    "js-yaml": "4.1.0"
```
humphd, line 22: I don't think we need this, since `pnpm list` can give us JSON.

src/api/dependency-graph/src/graph-collector/graph.js
```diff
@@ -0,0 +1,132 @@
...
+ * @param {string} version the version tag
+ * @param {object} allDependencies collection of all dependencies
+ */
+function Project(name, description, repositoryUrl, version, allDependencies) {
+  this.name = name;
+  this.description = description;
+  this.repositoryUrl = repositoryUrl;
+  this.releases = {};
+
+  this.releases[version] = new Release(
```
humphd, line 32: Do we care about historical releases?  Should we focus on the dependencies for the current version only, since you're going to snapshot the Telescope dependencies vs. try to keep them in sync?

src/api/dependency-graph/src/graph-collector/graph.js
```diff
@@ -0,0 +1,132 @@
...
+   * @param {object} allDependencies collection of all types of dependencies
+   */
+  addProject(name, description, repoUrl, version, allDependencies) {
+    const newProject = new Project(name, description, repoUrl, version, allDependencies);
+    this.projectNodes[name] = newProject;
+  }
+
+  /**
+   * @param {string} name
+   * @returns whether the project of given name is registered in the graph
```
humphd, line 69: `@returns {boolean} ...`

src/api/dependency-graph/src/graph-collector/index.js
```diff
@@ -0,0 +1,136 @@
+const yaml = require('js-yaml');
+const fs = require('fs');
+const { logger, fetch } = require('@senecacdot/satellite');
+const Graph = require('./graph');
+
+function extractRepositoryUrl(repository) {
```
humphd, line 6: Instead of doing this, why don't we use the package name you already have from `pnpm` to query the npmjs registry programmatically?  For example, you can get the package info for React using https://registry.npmjs.org/react.  This will return everything we need in JSON format.

Here are some docs:

- https://github.com/npm/registry/blob/master/docs/REGISTRY-API.md
- https://www.edoardoscibona.com/exploring-the-npm-registry-api
- https://www.npmjs.com/package/query-registry
- https://www.jsdocs.io/package/query-registry

src/api/dependency-graph/src/router.js
```diff
@@ -0,0 +1,13 @@
+const { Router } = require('@senecacdot/satellite');
+
+module.exports = function (graphConstructor) {
```
humphd, line 3: Don't pass in the constructor here, `require` it above.

src/api/dependency-graph/src/router.js
```diff
@@ -0,0 +1,13 @@
...
+  const graph = graphConstructor();
+
+  const router = Router();
+
+  router.get('/project/:id', async (req, res) => {
+    res.status(200).json(graph.getProject(req.params.id));
```
humphd, line 9: Let's add `cache-control` headers for this, since it won't update very often.  We can probably cache for 30 mins or an hour?

humphd: You also need to deal with error cases.  If `getProject` throws, your server will crash.

src/api/dependency-graph/src/router.js
```diff
@@ -0,0 +1,13 @@
+const { Router } = require('@senecacdot/satellite');
+
+module.exports = function (graphConstructor) {
+  const graph = graphConstructor();
+
+  const router = Router();
+
+  router.get('/project/:id', async (req, res) => {
```
humphd, line 8: There's no `await` in this function, do you need `async`?

src/api/dependency-graph/src/graph-collector/index.js
```diff
@@ -0,0 +1,136 @@
...
+    const possibleVersionWithPeer = dependenciesObj[key];
+    const { version } = extractVersionAndPeerDependency(possibleVersionWithPeer);
+    dependenciesObj[key] = version;
+  });
+}
+
+async function createGraphFromSources() {
+  const graph = new Graph();
+
+  const pnpmLock = yaml.load(
```
humphd, line 79: Let's not do this.  @DukeManh's idea of calling `pnpm list` in a child-process, and having it return JSON is probably the better route.

humphd (changes requested):

JerryHue:
So, the only problem with `pnpm list` is that it requires the whole `node_modules` folder installed, so you would need to include the whole repo in the Docker image, run an installation, and then run the `pnpm list` to get the whole list. That is why I went with the approach of reading the `pnpm-lock.yml`.

So, if that won't make the image or the container super big, then I can switch it to that approach. Could someone that knows more about Docker let me know if this is true?

cc @humphd, @DukeManh

src/api/dependency-graph/src/router.js
JerryHue: I am sure that I don't need it for the router callback, but I am not sure how to resolve the `async` chain that I have with the whole `graph-collector` module. I just wanted to use `async` because of the `fetch` request, and then it got out of control

humphd:
Yes, this will complicate things.

I think we're going to have to add a script in `tools/` that runs `pnpm list` the way you want in the root (when the `node_modules/` is already there) and copies the result over to your microservice directory so that it can be included in the build context we copy in.

So maybe focus on writing a node.js script in `tools/` to generate your JSON formatted dependency file from `pnpm list`, and we'll figure out how to get that into the service once it exists, maybe as part of https://github.com/Seneca-CDOT/telescope/blob/master/tools/autodeployment/deploy.sh.

@manekenpix might have a better idea.

src/api/dependency-graph/src/router.js
humphd: All you need to worry about is adding a `try/catch` here, as well as `next` to deal with any promise rejections:

```js
router.get('/projects/:id', async (req, res, next) = {
  const { id } = req.params;
  try {
    res.status(200).json(await graph.getProject(id));
  } catch(err) {
    next(err);
  }
});
```

Also note: for a REST API endpoint, we always use plural forms: `/projects` vs. `/project`

src/api/dependency-graph/src/graph-collector/graph.js
JerryHue: So, this code does not worry about historical releases of Telescope, it worries about all the dependencies that it is currently using, where some of these dependencies could feature different releases while still being the same project. For example, the `@babel/runtime` appears in the dependency tree as `7.16.7`, and again as `7.5.0`, but these releases belong to different projects in the tree.

JerryHue:
Is it possible to run `pnpm list` to provide all the transitive dependencies of Telescope? That includes not only the direct ones mentioned in `package.json` files, but also the dependencies of those dependencies, and so on.

I would love it to also display a flat list just like how `pnpm list --parseable` does, so that the `json` output doesn't repeat so much information.

humphd:
> Is it possible to run `pnpm list` to provide all the transitive dependencies of Telescope? That includes not only the direct ones mentioned in `package.json` files, but also the dependencies of those dependencies, and so on.
> 
> I would love it to also display a flat list just like how `pnpm list --parseable` does, so that the `json` output doesn't repeat so much information.

You just need to specify a depth, and it will do deps of deps.

JerryHue:
> > Is it possible to run `pnpm list` to provide all the transitive dependencies of Telescope? That includes not only the direct ones mentioned in `package.json` files, but also the dependencies of those dependencies, and so on.
> > I would love it to also display a flat list just like how `pnpm list --parseable` does, so that the `json` output doesn't repeat so much information.
> 
> You just need to specify a depth, and it will do deps of deps.

From the [discussion in the pnpm repo](https://github.com/pnpm/pnpm/discussions/3888?sort=top), it seems I can specific `Infinity` to get all dependencies. 

One small clarification: do we want to lump all dependencies and say they are part of Telescope, instead of dividing them into the microservices projects? If we do it with the json/parseable way, we will have to group all dependencies this way

humphd:
> From the [discussion in the pnpm repo](https://github.com/pnpm/pnpm/discussions/3888?sort=top), it seems I can specify `Infinity` to get all dependencies.

When I tried running it with any depth over 5, it returned one billion responses and took forever.  I'm not sure that we have to go that deep?

I don't think we care which part of Telescope uses them, or whether they are a direct or indirect dep.  We just want to figure out where to look for things to work on.

JerryHue:
> > From the [discussion in the pnpm repo](https://github.com/pnpm/pnpm/discussions/3888?sort=top), it seems I can specify `Infinity` to get all dependencies.
> 
> When I tried running it with any depth over 5, it returned one billion responses and took forever. I'm not sure that we have to go that deep?

I think if you ran it with `--parseable`, some lines were repeated. The same happens with `--json`. 

I had the intention to collect all dependencies of Telescope, and the easy way was by going through the `pnpm-lock.yaml`. That is why I would like to go as deep as possible into the dependency tree.

humphd:
Pipe the output through `uniq` and you can remove the duplicates:

```sh
❯ pnpm list --parseable --depth 5 | wc -l
   97762
❯ pnpm list --parseable --depth 5 | uniq | wc -l
    2466
~/repos/telescope sso-dockerfile-fix
```

humphd (changes requested):

src/api/dependency-discovery/src/dependency-list.js
```diff
@@ -0,0 +1,11 @@
+const { readFile } = require('fs/promises');
+const { join } = require('path');
+const { cwd } = require('process');
+
+async function getDependencies() {
+  const dependencies = await readFile(join(cwd(), 'deps.txt'), 'utf8');
```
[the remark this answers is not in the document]
JerryHue: Since `getDependencies()` returns a Promise, and Promises can [be resolved several times with the same result](https://stackoverflow.com/questions/20328073/is-it-safe-to-resolve-a-promise-multiple-times), I think caching might not be necessary.

Also, since `getDependencies()` is private to the module, we shouldn't have to worry that another module imports it and calls it again.

tools/pnpm-deps-gen.sh
```diff
@@ -0,0 +1,15 @@
...
+# After that, we use the pattern "@?[^@]+" to search for the name
+# that appears first, instead of ".+".
+#
+# Sometimes we could have foo@1.0.0_bar@1.0.0, in which case we want
+# to capture foo@1.0.0. With ".+", we could capture foo@1.0.0_bar.
+# Instead, we search for the string of characters that do not include
+# "@", which translates to "[^@]+". However, some package names include
+# @ at the start, so we cover that case, as well.
+
+pnpm list -r --parseable --depth=Infinity $1 | sed -n "s/.*\.pnpm\/\(@\?[^@]\+\).*/\1/p" | sort | uniq | sed "s/+/\//"
```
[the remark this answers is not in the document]
JerryHue: > Should this script also copy the file into `src/api/dependency-discovery/deps.txt` when it finishes?

We can do it like that, but then redirection might be more difficult to do. Maybe we could add a script command in `package.json` called `discover-deps` to call that script and redirect the result to the file?

src/api/dependency-discovery/src/dependency-list.js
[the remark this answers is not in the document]
humphd: But you're not returning a promise.  You're resolving the `readFile` promise, and it will do this every time you call this function.

tools/pnpm-deps-gen.sh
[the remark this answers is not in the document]
humphd: Yeah, that works too.  Or have your script accept an `-o` arg with a filepath of where to write the data when it completes vs. stdout.

src/api/dependency-discovery/src/dependency-list.js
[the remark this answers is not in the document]
JerryHue: > But you're not returning a promise.

`getDependencies()` is labelled as `async`, so that means it is returning a Promise that contains an array of strings.

In fact, if we were to log inside `getDependencies()` and hit `/projects` as many times as we wanted, it would only appear once.

While we are resolving the promise of `readFile()`, this is just another Promise that precedes the Promise returned by `getDependencies()`.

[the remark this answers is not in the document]
humphd: Indeed, I'm misreading you as returning `getDependencies` vs. `getDependencies()` from the module.

tools/pnpm-deps-gen.sh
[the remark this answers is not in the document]
JerryHue: That one sounds like a better idea. I will try to implement it

src/api/dependency-discovery/src/server.js
```diff
@@ -0,0 +1,5 @@
+const service = require('.');
+
+const port = parseInt(process.env.DEPENDENCY_GRAPH_PORT || 10500, 10);
```
humphd, line 3: We should create an `env.local` like we do for other services (e.g., https://github.com/Seneca-CDOT/telescope/blob/master/src/api/sso/env.local).    Then pass it into `dev` like this: https://github.com/Seneca-CDOT/telescope/blob/master/src/api/sso/package.json#L7.

humphd (changes requested):
We're going to need to Dockerize this as well, and figure out the logic for calling your script at build-time when we deploy in a follow-up.

But I think this gives us a beginning worth having.

tools/pnpm-deps-gen.sh
```diff
@@ -0,0 +1,24 @@
...
+# Instead, we search for the string of characters that do not include
+# "@", which translates to "[^@]+". However, some package names include
+# @ at the start, so we cover that case, as well.
+
+option=$1
+filename=$2
+
+if [[ $# -eq 2 && $option = "-o" && $filename ]]; then
+  echo "Start collecting dependencies"
+  pnpm list -r --parseable --depth=Infinity | sed -n "s/.*\.pnpm\/\(@\?[^@]\+\).*/\1/p" | sort -u | sed "s/+/\//" > $filename
```
DukeManh, line 20: If you want to make this more readable, you can use extended regular expression and choose a custom delimiter to avoid escaping characters.
```suggestion
  pnpm list -r --parseable --depth=Infinity | sed -n -r -e "s-.*\.pnpm/(@?[^@]+).*-\1-p" | sort -u | sed "s/+/\//" >  $filename
```

  Also, are we pursuing depth=Infinity, seems too deep to me. The whole script took 20s to execute.
cc @humphd

DukeManh (changes requested):

src/api/dependency-discovery/src/dependency-list.js
```diff
@@ -0,0 +1,11 @@
...
+
+  return [...new Set(dependencies.split('\n').filter((line) => line !== ''))];
```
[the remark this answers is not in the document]
DukeManh: you don't need to use Set here as these are already uniquely sorted by `sort -u`

DukeManh (changes requested):

src/api/dependency-discovery/src/dependency-list.js
```diff
@@ -0,0 +1,11 @@
...
+}
+
+module.exports = getDependencies();
```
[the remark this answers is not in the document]
JerryHue: Because the `await` keyword is not valid in the top level of a file. When you run the project with `pnpm start`, it will show this error:

```
SyntaxError: await is only valid in async functions and the top level bodies of modules
```

That is why we have to annoyingly await every time we access the object.

src/api/dependency-discovery/src/dependency-list.js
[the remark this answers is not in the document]
JerryHue: I didn't cover that case. Should I also cover the case when it is `\r` only? I think MacOS uses `\r` but I am not sure. I am going to add it just in case.

tools/pnpm-deps-gen.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
```
humphd, line 2: Let's add:

```sh
set -e
```

So we exit on errors

humphd (dismissed):
A few minor nits.

tools/pnpm-deps-gen.sh
```diff
@@ -0,0 +1,24 @@
...
+# Instead, we search for the string of characters that do not include
+# "@", which translates to "[^@]+". However, some package names include
+# @ at the start, so we cover that case, as well.
+
+option=$1
+filename=$2
+
+if [[ $# -eq 2 && $option = "-o" && $filename ]]; then
+  echo "Start collecting dependencies"
+  pnpm list -w -r --parseable --depth=0 | sed -n -r -e "s|.*\.pnpm/(@?[^@]+).*|\1|p" | sort -u | sed "s/+/\//" > $filename
```
DukeManh, line 20:
```suggestion
  pnpm list -w -r --parseable --depth=0 | sed -n -r -e "s|.*\.pnpm/(@?[^@]+).*|\1|p" -e "s/\+/\//" | sort -u  > $filename
```

DukeManh (approved):
Just 1 nit: you can chain sed expressions

DukeManh (approved):
Fantastic work. 👍

humphd (approved):
Looks good.  I'll file follow-up issues on things I notice that still need to happen.

PR #2809, opened 2022-02-02
Use buildx in docker-compose

humphd:
This forces docker to prefer [`buildx`](https://docs.docker.com/buildx/working-with-buildx/) when building with `docker-compose`. I've only tested on macOS.

Not critical for 2.6.

### DIFF
--- a/config/env.development
+++ b/config/env.development
@@ -6,6 +6,11 @@
 
 COMPOSE_PROJECT_NAME=telescope_api
 
+# Use buildx in docker-compose, see https://docs.docker.com/buildx/working-with-buildx
+# With buildx, you get a more efficient build, better caching, and more.
+COMPOSE_DOCKER_CLI_BUILD=1
+DOCKER_BUILDKIT=1
+
 # Compose files to use together during development. NOTE: we specify separator below
 # so it will work on Windows and Unix, see
 # https://docs.docker.com/compose/reference/envvars/#compose_file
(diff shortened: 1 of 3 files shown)


config/env.development
```diff
@@ -6,6 +6,10 @@
 
 COMPOSE_PROJECT_NAME=telescope_api
 
+# Use buildx in docker-compose
```
DukeManh, line 9: Maybe include a link to what buildx is here?

I'm just curious, do you talk about this your CCP55 teachings?

DukeManh (changes requested):

Kevan-Y:
Not sure if I tested correctly.

Steps: 
`cp .\config\env.development .env`
`pnpm services:clean`
`pnpm services:start`

But I'm getting this, no sure if related to `buildx`

image

Kevan-Y (approved):
Ignore my previous comment, I just had another image on my docker container.
It's working on windows 10

aserputov (approved):

config/env.development
humphd: Sure, I'll add one. `buildx` adds a bunch of goodies to `build` (e.g., skipping layers that aren't being used, allow cross-compiling, etc.  It's the default on some systems to use it, and on some it isn't.

humphd:
Add a comment, rebased.

aserputov (approved):

DukeManh (approved):

Issue #2830, opened 2022-02-04
Dockerize dependency service


humphd:
In https://github.com/Seneca-CDOT/telescope/pull/2797 we are adding a dependency service, and we need to dockerize it, and connect it to our Traefik API gateway for routing.

There are lots of examples of how to do this with other services.

JerryHue:
The `dependency-discovery` service is a little bit special:

it requires a text file called `deps.txt`, which gets generated by `toosl/pnpm-deps-gen.sh`. However, to run this file you need:

* `pnpm` installed
* the whole Telescope project
* `node_modules` for each one of these projects

to run it successfully.

The way I am thinking to generate this is:

* copy the whole Telescope project into the Docker container
* run the script to generate the file in `src/api/dependency-discovery`

However, this might create a humongous container (if my understanding of how Docker works is right). There might be a better way to do this.

humphd:
This is an interesting problem.  I think we should solve it in a simple way to begin: run the command outside the Docker container and copy the `deps.txt` file in.

I'd run your script in our `deploy.sh`, maybe here https://github.com/Seneca-CDOT/telescope/blob/master/tools/autodeployment/deploy.sh#L12, and put the file in the right location so that it can be found when we build the dependency service image.

cc'ing @manekenpix and @raygervais in case they have a better solution.

JerryHue:
That approach would be the most ideal, but sadly I think it only fits for deploying.

For development, we would have change other files to generate `deps.txt` prior to the docker building: maybe changing `tools/services.js` to call this script.

humphd:
Another option we could consider is adding this script to a `preversion` step, and store the file in git when we `tag` a new version.  So we'd focus on deps for our latest deployment, but you can always re-generate it when we deploy.  That way it would be in the tree, but also constantly updated, too.

I think that running this any more often than when we tag might be too much.

JerryHue:
That sounds wonderful. I don't think the `depedendency-discovery` service has to keep track of the dependencies added in between release, just right before the release, so this fits perfectly.

humphd:
So the only issue I see with this approach, is that your script depends on being run in a POSIX shell.  If students are trying to do the release on Windows, it will fail.

We could run your script in our [release workflow](https://github.com/Seneca-CDOT/telescope/blob/master/.github/workflows/release.yml), but by then the git `tag` is already created (it's too late).

[...]

PR #2838, opened 2022-02-04
Write portable dependency script

JerryHue:
## Issue This PR Addresses

Somewhat related to #2830 

## Type of Change

## Description

In #2830 we were discussing how to dockerize the service, and one major problem is when we should generate `deps.txt` which the service depends on.

We decided to write a script that will be run before a release so that it can update it and the file won't go stale.

We also write it to be a JavaScript script so that it can be run on Windows, Linux, and MacOS without much problem.

Notes:

I generated `deps.txt` and placed it in the dependency tree to start with something, but this file may change for release 2.7.

This is also a bug fix because the original `bash` script was not working. I think I changed it for a review in #2797 and I didn't check that it worked fine :(

I also need confirmation from a MacOS user and a Windows user that the script is working fine. We can never be too sure :)

## Steps to test the PR

1. Fetch the changes (`git remote add JerryHue https://github.com/JerryHue/telescope.git && git fetch JerryHue`)
2. Go to the branch (`git checkout portable-pnpm-script` or `git checkout JerryHue/portable-pnpm-script` to avoid creating a new branch).
3. Run the script on the root of the project by writing `node tools/collect-dependencies.js -- -o <filepath>` where `filepath` can be anywhere you want.

## Checklist


package.json
```diff
@@ -26,6 +26,7 @@
     "server": "node src/backend/web/server",
     "test-ci": "run-s prettier-check test",
     "pre-commit": "pretty-quick --staged",
+    "preversion": "node tools/collect-dependencies.js -- -o src/api/dependency-discovery/deps.txt && git add src/api/dependency-discovery/deps.txt",
```
humphd, line 29: Question: will this get committed by `version`?  I assume so, but haven't tested.

humphd: We better add some docs to the `README.md` in your dependency service to explain this logic and how/where this file comes from.

JerryHue: > Question: will this get committed by `version`? I assume so, but haven't tested.

I wanted to check whether that is true, so I head to the [documentation of `npm version`](https://docs.npmjs.com/cli/v8/commands/npm-version#description) to see if this was mentioned.

It seems that it will committed as long as you add added them to staging with `git add` before the commit done by `npm version`.

src/api/dependency-discovery/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Install
+
+Run `pnpm install` from the root of the Telescope project.
+
+## Usage
+
+```
+# For development mode
+pnpm run dev
```
humphd, line 14: Drop the `run` with `pnpm`

humphd (commented):
One small fix, looks good otherwise.

tools/collect-dependencies.js
```diff
@@ -0,0 +1,34 @@
+const { execSync } = require('child_process');
+const { writeFileSync } = require('fs');
+const { argv, exit } = require('process');
+
+const option = argv[3];
+const filename = argv[4];
+
+if (option !== '-o' || !filename) {
+  console.log(`command usage: ${__filename} -o filename`);
+  exit(-1);
```
DukeManh, line 10: I think '9' is more appropriate here. https://nodejs.org/api/process.html#process_exit_codes
```suggestion
  exit(9);

File: tools/collect-dependencies.js [hunk 4]
```diff
@@ -0,0 +1,34 @@
...
+}
+
+console.log('Starting to collect dependencies');
+
+console.log('Invoking pnpm list...');
+const output = execSync('pnpm list -w -r --parseable --depth=0').toString('utf8');
```
DukeManh, line 16: We are collecting dep=0 now?

tools/collect-dependencies.js
```diff
@@ -0,0 +1,34 @@
...
+console.log('Processing file...');
+const packageNames = new Set(
+  [...output.matchAll(regex)].map((results) => results[1].replaceAll('+', '/')).sort()
+);
+
+let buf = '';
+
+packageNames.forEach((packageName) => {
+  buf += `${packageName}\n`;
+});
```
DukeManh, line 29:
```suggestion
const buff =  packageNames.join('\n');
```

DukeManh (changes requested):

manekenpix:
@JerryHue @DukeManh @dbelokon @Kevan-Y @aserputov 
I'd like you all to take a look at [my comment](https://github.com/Seneca-CDOT/telescope/issues/2830#issuecomment-1030484203) and provide some feedback before more work is put into this.

humphd:
> @JerryHue @DukeManh @dbelokon @Kevan-Y @aserputov
> I'd like you all to take a look at [my comment](https://github.com/Seneca-CDOT/telescope/issues/2830#issuecomment-1030484203) and provide some feedback before more work is put into this.

@manekenpix we discussed this a bit in the meeting, and I think we should consider doing a global script first via `deploy.sh`, so that we can get started on the rest of the dependency stuff, and refactor to use the Dockerfile solution once we have a clear path of how to support non-Satellite sub-projects (maybe you already do, but I don't).

tools/collect-dependencies.js
humphd: Agree, I would go deeper than 0, but not to Infinity.  2?

humphd (changes requested):
Let's move this forward as a first strategy, and refactor to something more clever with Docker later.

PR #2847, opened 2022-02-06
Resize terminal depending on window height

### LINKED ISSUE #2842: Build log not filling vertical space
The build log dashboard isn't filling the whole vertical space for me:

Screen Shot 2022-02-05 at 10 17 08 AM

JerryHue:
## Issue This PR Addresses

Fixes #2842

## Type of Change

## Description

We resize the height of the terminal for the build log in the dashboard by resizing it with the API offered by `xterm.js`.

## Steps to test the PR

1. Get the changes from the branch
2. Install dependencies (`pnpm install`)
3. Run the dashboard (`cd src/api/status && pnpm dev`)

## Checklist

- [X] **Quality**: This PR builds and passes our npm test and works locally.
- [ ] **Tests**: This PR does not include tests because it does not add a new feature.
- [X] **Screenshots**: This PR includes screenshots image
- [ ] **Documentation**: This PR does not include documentation because no user-exposed functionality was added/changed.

### DIFF
--- a/src/api/status/public/js/build-log/terminal.js
+++ b/src/api/status/public/js/build-log/terminal.js
@@ -66,4 +66,8 @@ const copyToClipboard = async () => {
 };
 copyToClipboardButton.addEventListener('click', () => copyToClipboard());
 
+window.addEventListener('resize', () => {
+  fitAddon.fit();
+});
+
 export default terminal;
--- a/src/api/status/src/views/layouts/main.hbs
+++ b/src/api/status/src/views/layouts/main.hbs
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   {{> header }}
-<body class="g-sidenav-show bg-gray-200">
+<body class="g-sidenav-show bg-gray-200 vh-100">
     {{> _toast}}
     {{{body}}}
 </body>
--- a/src/api/status/src/views/builds.hbs
+++ b/src/api/status/src/views/builds.hbs
@@ -1,6 +1,6 @@
 {{> sidebar}}
 
-<main class="main-content position-relative max-height-vh-100 h-100 border-radius-lg display:flex; flex-direction:column">
+<main class="main-content position-relative max-height-vh-100 h-100 border-radius-lg d-flex flex-column align-items-stretch">
   <!-- Navbar -->
   <nav class="navbar navbar-main navbar-expand-lg px-0 shadow-none border-radius-xl" id="navbarBlur"
     navbar-scroll="true">
@@ -31,7 +31,7 @@
       </nav>
     </div>
   </nav>
-  <div style="width: 100%; max-height: 100%; ">
+  <div class="d-flex flex-column flex-fill" style="width: 100%; min-height: 0;">
     <div id="build-header" class="card bg-white">
      <div class="card-body">
         <h6 id="build-header-title" class="mb-0 text-lg text-center">
@@ -73,9 +73,9 @@
         </div>
      </div>
     </div>
-    <div class="position-relative">
+    <div class="position-relative d-flex flex-fill" style="min-height: 0">
     <button id="copy-to-clipboard-button" class="btn btn-dark">Copy<i class="material-icons">content_copy</i></button>
-    <div id="terminal"></div>
+    <div id="terminal" class="flex-fill h-100"></div>
     </div>
   </div>
 </main>


DukeManh:
@JerryHue, can I make one suggestion? I just notice we are using the fit-addon to fit rows and columns of the terminal to a container. How about editing the CSS of `#terminal` so it uses the full height, then the `xterm` will just adjust itself to fit its container.

src/api/status/public/js/build-log/terminal.js
```diff
@@ -66,4 +65,28 @@ const copyToClipboard = async () => {
...
+ * Resize the terminal depending on the height of the window
+ */
+const resizeTerminal = (function () {
+  // Since the terminal will be resized, we need to store these
+  const rect = terminal.element.getBoundingClientRect();
+  const rowHeight = rect.height;
+  const rectTop = rect.y;
+  const bottomMargin = 16;
+
+  return (windowHeight) => {
```
Andrewnt219, line 78: Question: why do we have to create a closure instead of re-calculating everything.
```js
const resizeTerminal = function (windowHeight) {
  // Since the terminal will be resized, we need to store these
  const rect = terminal.element.getBoundingClientRect();
  const rowHeight = rect.height;
  const rectTop = rect.y;
  const bottomMargin = 16;

  const rowNumber = Math.floor((windowHeight - rectTop - bottomMargin) / rowHeight);

  terminal.resize(terminal.cols, rowNumber);
};
```

Andrewnt219 (commented):
This's brilliant! Thanks for the PR.

JerryHue:
> @JerryHue, can I make one suggestion? I just notice we are using the fit-addon to fit rows and columns of the terminal to a container. How about editing the CSS of `#terminal` so it uses the full height, then the `xterm` will just adjust itself to fit its container.

I tried this approach, but I struggled a little with the CSS of the page. I could try to do it this way one more time, but I will have to ask for some assistance with the CSS

src/api/status/public/js/build-log/terminal.js
JerryHue: The `terminal.element` will get updated the moment the `terminal` is resized, so if we only capture the `terminal` reference during function creation, the `y` and `rowHeight` values will change, as well.

The closure is for mainly storing the `y` and `rowHeight` so that they don't get destroyed the moment the function `resizeTerminal()` returns.

src/api/status/public/js/build-log/terminal.js
```diff
@@ -66,4 +65,28 @@ const copyToClipboard = async () => {
 };
 copyToClipboardButton.addEventListener('click', () => copyToClipboard());
 
+/**
+ * Resize the terminal depending on the height of the window
+ */
+const resizeTerminal = (function () {
```
TueeNguyen, line 71: Can we do this without using `currying` (not sure if it's the correct name for this pattern). It introduces unnecessary complexity

DukeManh:
>  but I will have to ask for some assistance with the CSS

I'll be happy to help you with that. 

What I think of right now is to have a flex-column layout, the container below the build-header has `flex:1` properties so that it stretches to available space. Let me know of any issues.

src/api/status/public/js/build-log/terminal.js
JerryHue: I think you meant using a closure. I am going to try with the CSS approach mentioned by @DukeManh, and if it does not work, I will rewrite this to be a little bit simpler.

JerryHue:
> What I think of right now is to have a flex-column layout, the container below the build-header has `flex:1` properties so that it stretches to available space. Let me know of any issues.

So, I managed to make it work, but if you resize to a smaller screen, it won't decrease the size. I am not really sure why, but the terminal seems to only grow, but not shrink.

TueeNguyen:
When I resize the window, it overflows, it should not right? 

https://user-images.githubusercontent.com/58532267/152700197-33efe506-3ed6-45f8-a302-42b06a52d3ab.mp4

TueeNguyen (commented):
When I resize the window, it overflows, it should not right? 

https://user-images.githubusercontent.com/58532267/152700197-33efe506-3ed6-45f8-a302-42b06a52d3ab.mp4

JerryHue:
> When I resize the window, it overflows, it should not right?

It shouldn't, but for some reason, the container does not shrink. I am not really that knowledgeable in CSS, so I am not really how to figure this one out.

DukeManh:
@JerryHue, adding `align-items: stretch` to 'main' would do it. 
After that, if it still overflows on resize, you can add an event listener on window resizing and call `terminal.fit()`

JerryHue:
> @JerryHue, adding `align-items: stretch` to 'main' would do it. After that, if it still overflows on resize, you can add an event listener on window resizing and call `terminal.fit()`

Hmm... maybe I am doing something wrong, because it still doesn't resize the terminal properly. I will push my changes for the sake of somebody else testing, too...

src/api/status/src/views/layouts/main.hbs
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   {{> header }}
-<body class="g-sidenav-show bg-gray-200 vh-100">
+<body class="g-sidenav-show bg-gray-200 vh-100" style="display: grid; align-items: stretch;">
```
DukeManh, line 4:
```suggestion
<body class="g-sidenav-show bg-gray-200 vh-100">
```

DukeManh (changes requested):
I feel like I'm stopping good work from merging, but we're here to learn and find the best solution right: Here's what I would do in addition to your current change.

https://github.com/Seneca-CDOT/telescope/blob/da0b0f664743f54a22e0f2605aba82aed11d0425/src/api/status/src/views/builds.hbs#L3

```diff
--<main class="main-content position-relative max-height-vh-100 h-100 border-radius-lg display:flex; flex-direction:column">
++<main class="main-content position-relative max-height-vh-100 h-100 border-radius-lg d-flex flex-column align-items-stretch">
```
https://github.com/Seneca-CDOT/telescope/blob/da0b0f664743f54a22e0f2605aba82aed11d0425/src/api/status/src/views/builds.hbs#L34

```diff
-- <div style="width: 100%; max-height: 100%; ">
++  <div class="d-flex flex-column flex-fill" style="width: 100%; min-height: 0">
```

https://github.com/Seneca-CDOT/telescope/blob/da0b0f664743f54a22e0f2605aba82aed11d0425/src/api/status/src/views/builds.hbs#L76-L78

```diff
--  <div class="position-relative">
-- <button id="copy-to-clipboard-button" class="btn btn-dark">Copy<i class="material-icons">content_copy</i></button>
-- <div id="terminal"></div>
++  <div class="position-relative flex-fill" style="min-height: 0">
 ++ <button id="copy-to-clipboard-button" class="btn btn-dark">Copy<i class="material-icons">content_copy</i></button>
++ <div id="terminal" class="flex-fill h-100"></div>
```

DukeManh:
The magic here is adding `min-height: 0`. 
I found this explanation for how it helps prevent flex items from overflowing their parent.

https://stackoverflow.com/questions/36230944/prevent-flex-items-from-overflowing-a-container

JerryHue:
Great! Now it doesn't overflow! Thank you so much @DukeManh. How did you even come up with this solution? I would like to improve my understanding of CSS.

src/api/status/public/js/build-log/terminal.js
```diff
@@ -66,4 +66,8 @@ const copyToClipboard = async () => {
 };
 copyToClipboardButton.addEventListener('click', () => copyToClipboard());
 
+window.addEventListener('resize', () => {
+  fitAddon.fit();
```
humphd, line 70: Nice

humphd (approved):

TueeNguyen (approved):
Mostly, we're changing these containers to flex-display. I did try add the CSS in dev tools (staging telescope) without the `fitAddon.fit()`, it didn't work. So, the magic comes from `fitAddon.fit()` and `min-width: 0`.  The PR seems small but quite difficult

DukeManh (approved):
Nice. 

@JerryHue I didn't come up with the solution https://dev.to/dukemanh/week-4-notes-3c78. I would say you should continue to build things with CSS your skills will inevitably improve.

JerryHue:
> Mostly, we're changing these containers to flex-display. I did try add the CSS in dev tools (staging telescope) without the `fitAddon.fit()`, it didn't work. So, the magic comes from `fitAddon.fit()` and `min-width: 0`. The PR seems small but quite difficult

Yeah, I try to also just do the solution in the dev tools, but it seems that the terminal itself doesn't resize unless you tell it resize (whether that be with the addon or another way).

PR #2856, opened 2022-02-07
Default health check to master in development

### LINKED ISSUE #2854: Latest commit remains undefined in local development
On the dev build of Telescope, the commit link still shows undefined, which is strange considering that we fixed this awhile ago. The reasoning for this lies [here](https://github.com/Seneca-CDOT/telescope/blob/master/src/backend/web/routes/health.js#L7).

When building Telescope locally, the frontend uses the legacy backend's url as the API_URL. It looks for a health check route that exists. This route tried to obtains the sha variable via an env value that doesn't exist anymore. This is not an issue on staging or production since the sha is obtained differently than how it is here. 

Since I'm pretty sure we don't plan on adding dev checks to the status service, this line will need to be fixed up.

HyperTHD:
## Issue This PR Addresses
This PR closes #2854 

## Type of Change

## Description

When looking at Telescope in development build, the git `sha` remained undefined, even though we fixed it for staging and production. The health route in the backend was the root cause, since it's the backend that's used as the `API` route in the local telescope build. This PR defaults the sha to point to the current commit on the master branch.

## Steps to test the PR

The link that points to the current `sha` should point to the current commit on master. It shouldn't have any specific sha hash, but should refer to master only.

## Checklist

### DIFF
--- a/src/api/status/package.json
+++ b/src/api/status/package.json
@@ -30,7 +30,11 @@
     "xterm": "4.15.0",
     "xterm-addon-fit": "0.5.0",
     "xterm-addon-web-links": "0.4.0",
-    "xterm-addon-webgl": "0.11.3"
+    "xterm-addon-webgl": "0.11.3",
+    "get-repo-package-json": "2.0.0",
+    "npm-run-all": "4.1.5",
+    "sass": "1.45.2",
+    "vite": "2.7.13"
   },
   "engines": {
     "node": ">=12.0.0"
--- a/src/api/status/src/server.js
+++ b/src/api/status/src/server.js
@@ -3,6 +3,7 @@ const { static: serveStatic } = require('express');
 const path = require('path');
 const { engine } = require('express-handlebars');
 const fs = require('fs/promises');
+const getPackage = require('get-repo-package-json');
 const { check } = require('./services');
 const getGitHubData = require('./js/github-stats');
 const getFeedCount = require('./js/feed-stats');
@@ -47,6 +48,19 @@ const satelliteOptions = {
         imgSrc: ["'self'", 'data:', 'https:'],
       },
     },
+  },
+  healthCheck: async (req, res) => {
+    const sha = process.env.GIT_COMMIT || 'master';
+    const gitHubUrl = `https://github.com/Seneca-CDOT/telescope/commit/${sha}`;
+    const { version } = await getPackage('https://github.com/Seneca-CDOT/telescope/blob/master/');
+
+    res.set('Cache-Control', 'public, max-age=300');
+
+    return {
+      version,
+      sha,
+      gitHubUrl,
+    };
   },
 };
 
--- a/src/backend/web/routes/health.js
+++ b/src/backend/web/routes/health.js
@@ -4,7 +4,7 @@ const healthcheck = require('express-healthcheck');
 // We want the current version from package.json
 const { version } = require('../../../../package.json');
 // As well as the current git sha
-const sha = process.env.GIT_COMMIT || 'undefined';
+const sha = process.env.GIT_COMMIT || 'master';
 // And a github URL to this commit
 const gitHubUrl = `https://github.com/Seneca-CDOT/telescope/commit/${sha}`;
 
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -317,6 +317,7 @@ importers:
       env-cmd: 10.1.0
       express: 4.17.2
       express-handlebars: 6.0.2
+      get-repo-package-json: 2.0.0
       nodemon: 2.0.15
       npm-run-all: 4.1.5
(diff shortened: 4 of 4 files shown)


src/backend/web/routes/health.js
```diff
@@ -4,7 +4,7 @@ const healthcheck = require('express-healthcheck');
 // We want the current version from package.json
 const { version } = require('../../../../package.json');
 // As well as the current git sha
-const sha = process.env.GIT_COMMIT || 'undefined';
+const sha = 'master';
```
manekenpix, line 7:
```suggestion
const sha = process.env.GIT_COMMIT || 'master';
```

manekenpix:
I don't know if you'd like to solve the whole issue here or in a follow-up PR, but the thing is that the `backend` folder will probably be gone some time in the future, so we should move the health check to the `status` microservice (for example). The change would imply duplicating the `health` endpoint in `status` to serve the `sha` and version, and also making the front-end fetch the `sha` and version from `status`.

src/backend/web/routes/health.js
HyperTHD: This env value doesn't exist anymore in any of our env files, should we really be referring to this still?

HyperTHD:
I don't mind solving it here, since it makes alot of sense to grab it from the status service instead of the backend.

HyperTHD:
I made a simple health check route that does what the health route should do. There's some duplicate info that I'm confused with in terms of how it's appearing. Also, ik that the route needs to be asynchronous which is not really what this new route should do. Looking for some feedback as always!

src/backend/web/routes/health.js
manekenpix: We add it to the `env` file used for deployment at build time [here](https://github.com/Seneca-CDOT/telescope/blob/da0b0f664743f54a22e0f2605aba82aed11d0425/tools/autodeployment/deploy.sh#L34)

HyperTHD: Ahh, figured it gets used in deployment. Will add it back

src/api/status/src/server.js
```diff
@@ -8,6 +8,7 @@ const getGitHubData = require('./js/github-stats');
 const getFeedCount = require('./js/feed-stats');
 const getPostsCount = require('./js/posts-stats');
 const getJobCount = require('./js/queue-stats');
+const { version } = require('../../../../package.json');
```
manekenpix, line 11: I don't think this is going to work. The legacy back-end has access to this file, but microservices are not really aware of what's outside of their container, so the main `package.json` won't be available.
We need to find a different way to get Telescope's version. Also, I think the front-end would need some changes. If you think it's too much for this PR, you can do it in a follow-up.

HyperTHD: Can definitely do the frontend portion in a follow up, also forgot the purpose of microservices, I'll figure out how to grab the version another way

Kevan-Y:
@HyperTHD , the code looks good. But I don't think there is a way to test it until we land it right?

src/api/status/package.json
```diff
@@ -30,7 +30,11 @@
     "xterm": "4.15.0",
     "xterm-addon-fit": "0.5.0",
     "xterm-addon-web-links": "0.4.0",
-    "xterm-addon-webgl": "0.11.3"
+    "xterm-addon-webgl": "0.11.3",
+    "get-repo-package-json": "^2.0.0",
```
humphd, line 34: `2.0.0`, no `^`

src/api/status/src/server.js
```diff
@@ -48,6 +49,17 @@ const satelliteOptions = {
       },
     },
   },
+  healthCheck: async () => {
+    const sha = process.env.GIT_COMMIT || 'master';
+    const gitHubUrl = `https://github.com/Seneca-CDOT/telescope/commit/${sha}`;
+    const { version } = await getPackage('https://github.com/Seneca-CDOT/telescope/blob/master/');
```
humphd, line 55: Can we cache this result?  Will it call it every time we hit the health check?  Seems like we should do it once only.

humphd (dismissed):

Kevan-Y (dismissed):
Need to squash those commits then good to go. Thanks @HyperTHD

humphd (dismissed):

TueeNguyen (approved):

TueeNguyen:
@HyperTHD last rebase!

HyperTHD:
Will merge in once rebase and squash is finished. Also: 

> I don't think this is going to work. The legacy back-end has access to this file, but microservices are not really aware of what's outside of their container, so the main `package.json` won't be available. We need to find a different way to get Telescope's version. Also, I think the front-end would need some changes. If you think it's too much for this PR, you can do it in a follow-up.

I'll look into this further and see what changes, if any, will be needed and will do them in a follow-up PR

PR #2891, opened 2022-02-11
Fix missing import of @docusaurus/types

humphd:
VSCode is unhappy with how we import these types.  This adds the type definitions.


Kevan-Y:
Oh, I didn't notice VSCode didn't like how we import @docusaurus/types. Nice catch

Kevan-Y (approved):

joelazwar (approved):

PR #2927, opened 2022-02-14
add testing for normal search

JiaHua-Zou:
## Issue This PR Addresses
#2621 

## Type of Change

- [x] **Tests**: Add testing for the normal search function.

## Description

Adding tests for the normal search function.
The test has cover all of the normal search function.
image

## Steps to test the PR
`pnpm install`
`pnpm test search`

## Checklist

### DIFF
--- a/src/api/search/jest.setup.js
+++ b/src/api/search/jest.setup.js
@@ -1,1 +1,1 @@
-process.env = { ...process.env, MOCK_ELASTIC: '1' };
+process.env = { ...process.env, MOCK_ELASTIC: '1', LOG_LEVEL: 'silent' };
--- a/src/api/search/test/query.test.js
+++ b/src/api/search/test/query.test.js
@@ -4,10 +4,177 @@ const { app } = require('../src');
 
 const { POSTS_URL } = process.env;
 
+console.log(
+  '****************************************\n' +
+    'The logs are in slient mode.\nTo change it, go to ../jest-setup.js\n' +
+    '**************************************'
+);
 describe('/query routers', () => {
+  // Create the mock client.
+  const { mock } = new Elastic();
+
+  afterEach(() => {
+    mock.clearAll();
+  });
+
   it('return error 400 if no params given', async () => {
     const res = await request(app).get('/');
-    expect(res.status).toBe(400);
+    expect(res.statusCode).toBe(400);
+  });
+
+  it('return error 400 if no filter is given', async () => {
+    mock.add(
+      {
+        method: ['POST', 'GET'],
+        path: '/posts/post/_search',
+      },
+      () => {
+        return {};
+      }
+    );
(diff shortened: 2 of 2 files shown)


humphd (changes requested):
Do you want to do the same thing here with the version so it can get the mock stuff from Satellite?

src/api/search/test/query.test.js
```diff
@@ -1,9 +1,173 @@
...
+  });
+
+  it('return 503 if server has error', async () => {
+    mock.add(
+      {
+        method: ['POST', 'GET'],
+        path: '/posts/post/_search',
+      },
+      () => {
+        return { statusCode: 503 };
```
rclee91, line 167: I mentioned in my tests as well, it's not the `statusCode` that is being returned by ElasticSearch that is returning the 503 error. 
It's failing because Search is trying to do [these two lines](https://github.com/Seneca-CDOT/telescope/blob/f5785c6c33e4747260aab267d9697aa105fb32bc/src/api/search/src/search.js#L75-L78) here in the return.
It's trying to read `hits.total` and `hits.hits`,  but the `hits` object is undefined here because its not being returned by ElasticSearch. 
For example, changing it to `return { statusCode: 404 }` will also pass this test. 
I think we might need to rephrase the description for this test, so its not as misleading

src/api/search/jest.setup.js
```diff
@@ -1 +1 @@
-process.env = { ...process.env, MOCK_ELASTIC: '1' };
+process.env = { ...process.env, MOCK_ELASTIC: '1', LOG_LEVEL: 'error' };
```
rclee91, line 1: I didn't add the `LOG_LEVEL` in my PR, but commented that anyone could do this locally. 
I guess its up for discussion if we wanted to keep this or not.

rclee91 (changes requested):

src/api/search/jest.setup.js
cindyledev: I appreciated the `LOG_LEVEL` hint in your PR. My console was flooded with stuff until I changed that variable

humphd: What I usually do is add a `console.log()` to say, `we have our logs on silent, but you can change in file x/y/z to error or debug if you want`, and then make this `silent`, or it spams too much.

src/api/search/test/query.test.js
JiaHua-Zou: Should I try to return something that fails to read the `hits` object?

src/api/search/jest.setup.js
JiaHua-Zou: Yea I think that is a good idea. Even in `error` mode. It still print a whole bunch of stuff.

rclee91 (approved):

humphd (approved):
Great stuff.

Issue #2791, opened 2022-02-01
Expose Supabase via Traefik, nginx


humphd:
In #2743 we added a bunch of Supabase containers to our docker network.  This includes a few that we need to expose publicly:

- Kong API Gateway (access to all other internal containers) - http://kong:8000
- Web Studio (console) - http://studio:3000

You can start everything with `pnpm services:start` and then browse to http://localhost:8910 and http://localhost:8911, or do something similar on GitPod.  This is not going to work in production.

We discussed the need to be careful with exposing ports in staging and production: we don't want to open any ports beyond 80/443 that we can avoid.  Instead, we need to route all traffic through our various routers and proxies:

- nginx - manages our SSL certs and serves as our main ingress point from outside traffic.  Everything in Telescope goes through nginx.
- Traefik - sits behind nginx and manages access to our services via URL routing at `/v1/*` endpoints.  So the Posts service lives at `/v1/posts`, and search lives at `/v1/search`, etc.

I think the logical approach here is to do the following:

1. Expose the Kong API Gateway via Traefik at `/v1/supabase/*`, which will route to `http://kong:8000` within docker
2. Expose the Web Studio via nginx at `/supabase`.  This is similar to how we run `/portainer` right now.

Here are some examples of how to do each of these:

1. Modify `docker/supabase/docker-compose.yml` to add `labels`, with Traefik routing details, see https://github.com/Seneca-CDOT/telescope/blob/master/docker/docker-compose.yml#L213-L223 for an example, and do it here: and https://github.com/Seneca-CDOT/telescope/blob/master/docker/supabase/docker-compose.yml#L22-L36
2. Modify `config/nginx.conf.template` to do something similar to how we expose `/portainer/` and `/deploy/`.  See https://github.com/Seneca-CDOT/telescope/blob/master/config/nginx.conf.template#L228-L239

Once we do this, we should be able to modify https://github.com/Seneca-CDOT/telescope/blob/master/config/env.staging#L13 to include the Supabase docker-compose file like this: https://github.com/Seneca-CDOT/telescope/blob/master/config/env.development#L13, which would turn it on in staging.

cc'ing our "docker" folks, so one or more of you can pick this up: @TDDR, @aserputov, @Kevan-Y, @DukeManh, @sirinoks, @TueeNguyen, @menghif, @rclee91, @tcvan0707, @tpmai22.  This could be split in two, if two people want to do it.

humphd:
[...]

PR #2931, opened 2022-02-15
Expose kong via Traefik

Kevan-Y:
## Issue This PR Addresses
Fixes: #2791 

## Type of Change

## Description
Expose kong to `v1/supabase`

## Steps to test the PR
    - `gh pr checkout 2931`
    - `pnpm services:start`

## Checklist

### DIFF
--- a/docker/supabase/docker-compose.yml
+++ b/docker/supabase/docker-compose.yml
@@ -23,18 +23,30 @@ services:
     container_name: supabase-kong
     image: kong:2.1
     restart: unless-stopped
-    ports:
-      - ${KONG_HTTP_PORT}:8000/tcp
-      - ${KONG_HTTPS_PORT}:8443/tcp
     environment:
       KONG_DATABASE: 'off'
       KONG_DECLARATIVE_CONFIG: /var/lib/kong/kong.yml
       # https://github.com/supabase/cli/issues/14
       KONG_DNS_ORDER: LAST,A,CNAME
       KONG_PLUGINS: request-transformer,cors,key-auth,acl
+      JWT_ISSUER: ${JWT_ISSUER}
+      JWT_AUDIENCE: ${JWT_AUDIENCE}
+      SECRET: ${SECRET}
     volumes:
       - ./supabase/volumes/api/kong.yml:/var/lib/kong/kong.yml
-
+    depends_on:
+      - traefik
+    labels:
+      # Enable Traefik
+      - 'traefik.enable=true'
+      # Traefik routing for the kong service at /v1/supabase
+      - 'traefik.http.routers.kong.rule=PathPrefix(`/${API_VERSION}/supabase`)'
+      # Specify the kong service port
+      - 'traefik.http.services.kong.loadbalancer.server.port=8000'
+      # Add middleware to this route to strip the /v1/supabase prefix
+      - 'traefik.http.middlewares.strip_kong_prefix.stripprefix.prefixes=/${API_VERSION}/supabase'
+      - 'traefik.http.middlewares.strip_kong_prefix.stripprefix.forceSlash=true'
+      - 'traefik.http.routers.kong.middlewares=strip_kong_prefix'
   auth:
     container_name: supabase-auth
     image: supabase/gotrue:v2.2.12
--- a/config/env.development
+++ b/config/env.development
@@ -270,7 +270,7 @@ MAX_POSTS_PER_PAGE=5
 # Supabase Services
 ################################################################################
 
-SUPABASE_URL=http://kong:8000
+SUPABASE_URL=http://localhost/v1/supabase
 
 # Secrets
 # YOU MUST CHANGE THESE BEFORE GOING INTO PRODUCTION
--- a/docker/development.yml
+++ b/docker/development.yml
@@ -87,9 +87,9 @@ services:
 
   sso:
     environment:
-      # We development and testing, the Auth service needs to contact the users
-      # service directly via Docker vs through the http://localhost domain.
-      - USERS_URL=http://users:7000
+      # In development and testing, the SSO service needs to contact the Supabase
+      # service directly via Docker vs through the http://localhost/v1/supabase domain.
+      - SUPABASE_URL=http://kong:8000
     depends_on:
       - test-web-content
 
--- a/src/api/sso/test/e2e/utils.js
+++ b/src/api/sso/test/e2e/utils.js
@@ -4,18 +4,9 @@ const { hash } = require('@senecacdot/satellite');
 
 const { createClient } = require('@supabase/supabase-js');
 
-const { SERVICE_ROLE_KEY } = process.env;
-
-// From the browser, we need to access Supabase outside the docker container, so
-// we can't use the normal SUPABASE_URL.  We also use the SERVICE_ROLL_KEY for
-// admin access, which isn't acceptable outside of testing.
-const SUPABASE_URL = 'http://localhost:8911';
+const { SUPABASE_URL, SERVICE_ROLE_KEY } = process.env;
 
 const supabase = createClient(SUPABASE_URL, SERVICE_ROLE_KEY);
-
-// In tests, we need to hit this via a public URL like the front-end would
-// vs. the internal Docker URL, which the services use from the regular env value.
-const USERS_URL = `http://localhost/v1/users`;
 
 const createTelescopeUsers = (users) =>
   Promise.all(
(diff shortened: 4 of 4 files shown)


config/nginx.conf.template
```diff
@@ -231,6 +231,12 @@ http {
       proxy_pass http://portainer:9000/;
     }
 
+    location /studio/ {
```
humphd, line 234: Let's not do this one yet, since we don't have a way to secure it (e.g., our instance is wide open to anyone).  Only change the development template.

docker/supabase/docker-compose.yml
```diff
@@ -32,9 +32,25 @@ services:
...
+      - 'traefik.enable=true'
+      # Traefik routing for the kong service at /v1/supabase
+      - 'traefik.http.routers.kong.rule=PathPrefix(`/${API_VERSION}/supabase`)'
+      # Specify the kong service port
+      - 'traefik.http.services.kong.loadbalancer.server.port=8000'
 
+      # Add middleware to this route to strip the /v1/image prefix
+      - 'traefik.http.middlewares.strip_kong_prefix.stripprefix.prefixes=/${API_VERSION}/supabase'
+      - 'traefik.http.middlewares.strip_kong_prefix.stripprefix.forceSlash=true'
+      - 'traefik.http.routers.image.middlewares=strip_kong_prefix'
```
humphd, line 53: copy/paste error here: `routers.image`.  Should be `routers.kong`.

humphd (changes requested):

docker/supabase/docker-compose.yml
Kevan-Y: This is working

Kevan-Y:
Error I'm getting for superbase studio
image

docker/supabase/docker-compose.yml
```diff
@@ -32,9 +32,25 @@ services:
...
+      - traefik
+    labels:
+      # Enable Traefik
+      - 'traefik.enable=true'
+      # Traefik routing for the kong service at /v1/supabase
+      - 'traefik.http.routers.kong.rule=PathPrefix(`/${API_VERSION}/supabase`)'
+      # Specify the kong service port
+      - 'traefik.http.services.kong.loadbalancer.server.port=8000'
 
+      # Add middleware to this route to strip the /v1/kong prefix
```
humphd, line 50: `/v1/supabase`

docker/supabase/docker-compose.yml
```diff
@@ -32,9 +32,25 @@ services:
...
       - ./supabase/volumes/api/kong.yml:/var/lib/kong/kong.yml
+    depends_on:
+      - traefik
+    labels:
+      # Enable Traefik
+      - 'traefik.enable=true'
+      # Traefik routing for the kong service at /v1/supabase
+      - 'traefik.http.routers.kong.rule=PathPrefix(`/${API_VERSION}/supabase`)'
+      # Specify the kong service port
+      - 'traefik.http.services.kong.loadbalancer.server.port=8000'
```
humphd, line 49: Remove this blank line please

docker/supabase/docker-compose.yml
```diff
@@ -32,9 +32,25 @@ services:
...
```
humphd, line 48: Above, we can get rid of this:

```
    ports:
      - ${KONG_HTTP_PORT}:8000/tcp
      - ${KONG_HTTPS_PORT}:8443/tcp
```

config/nginx.conf.development.template
```diff
@@ -151,6 +151,12 @@ http {
       proxy_pass http://planet:9876;
     }
 
+    location /studio/ {
```
humphd, line 154: I can't remember the rules about the trailing slash here, @AmasiaNalbandian or @manekenpix might remember.

humphd (changes requested):

humphd (changes requested):
I pulled this PR locally and tested.  The problem with the Supabase studio not loading static assets is that their next.js site uses URLs like `/css/fonts.css`, see https://github.com/supabase/supabase/blob/03eed357949495a8cd769a2f302e7f738adfdaa6/studio/pages/_app.tsx#L45.  We need to tell their app to [use a different prefix for asset loading](https://nextjs.org/docs/api-reference/next.config.js/cdn-support-with-asset-prefix) when it builds the next.js site.

We need to overwrite the `next.config.js` file used in their build to add our own `assetPrefix`.   I attached to the running container, and their file is at `/usr/src/app/next.config.js` and looks like this:

```js
const { withSentryConfig } = require('@sentry/nextjs')
const withPlugins = require('next-compose-plugins')
const withBundleAnalyzer = require('@next/bundle-analyzer')({
  enabled: process.env.ANALYZE === 'true',
})

// This file sets a custom webpack configuration to use your Next.js app
// with Sentry.
// https://nextjs.org/docs/api-reference/next.config.js/introduction
// https://docs.sentry.io/platforms/javascript/guides/nextjs/

const nextConfig = {
  async headers() {
    return [
      {
        source: '/(.*?)',
        headers: [
          {
            key: 'X-Frame-Options',
            value: 'DENY',
          },
          {
            key: 'Referrer-Policy',
            value: 'strict-origin-when-cross-origin',
          },
        ],
      },
      {
        source: '/img/:slug*',
        headers: [{ key: 'cache-control', value: 'max-age=2592000' }],
      },
      {
        source: '/fonts/:slug*',
        headers: [{ key: 'cache-control', value: 'max-age=2592000' }],
      },
    ]
  },
}

// Export all config
const moduleExports = withPlugins([[withBundleAnalyzer({})]], nextConfig)

const sentryWebpackPluginOptions = {
  // Additional config options for the Sentry Webpack plugin. Keep in mind that
  // the following options are set automatically, and overriding them is not
  // recommended:
  //   release, url, org, project, authToken, configFile, stripPrefix,
  //   urlPrefix, include, ignore

  silent: true, // Suppresses all logs

  // For all available options, see:
  // https://github.com/getsentry/sentry-webpack-plugin#options.
}

// Make sure adding Sentry options is the last code to run before exporting, to
// ensure that your source maps include changes from all other Webpack plugins
module.exports = process.env.NEXT_PUBLIC_IS_PLATFORM === 'true'
  ? withSentryConfig(moduleExports, sentryWebpackPluginOptions)
  : nextConfig
```

The file comes from https://github.com/supabase/supabase/blob/03eed357949495a8cd769a2f302e7f738adfdaa6/studio/next.config.js.

A quick fix that we can do now is to copy this file into `docker/supabase/volumes/studio/next.config.js` and modify it so that it uses a proper prefix for our build case.  We'll have to pass the `WEB_URL` environment variable into the `studio` container so that we can use `assetPrefix: process.env.WEB_URL` in the config.

An even better upstream fix would be for them to allow us to pass this to the container as an environment variable, and have the next config automatically use that prefix when set.

cc'ing @DukeManh

DukeManh:
@humphd, While waiting for Supabase to reply, I'm thinking if could should configure nginx to fix it.

So, the studio is loading assets from http://locahost:8000/_next/* instead of http://locahost:8000/studio/_next/*, what if we add a Nginx location that resolves `_next/*` to 'studio/_next`? 
```
 location /_next/ {
     proxy_pass http://studio:3000/_next/;
    ...
}
```

humphd:
> @humphd, While waiting for Supabase to reply, I'm thinking if we could/should configure nginx to fix it.
> 
> So, the studio is loading assets from http://locahost:8000/_next/* instead of http://locahost:8000/studio/_next/*, what if we add a Nginx location that resolves `_next/*` to 'studio/_next`?
> 
> ```
>  location /_next/ {
>      proxy_pass http://studio:3000/_next/;
>     ...
> }
> ```

I'm testing a similar approach right now.  I'll report back when I get it built...

humphd:
Fixed the e2e tests in https://github.com/Kevan-Y/telescope/pull/1.

I wonder if we should remove the nginx changes from this PR, and land the rest.  The studio part we can do in a follow-up.

Kevan-Y:
@humphd Not sure why, test case timeout. I tried to increase the timeout but it's not working.

humphd:
> @humphd Not sure why, test case timeout. I tried to increase the timeout but it's not working.

Fixed in https://github.com/Kevan-Y/telescope/pull/2

DukeManh (dismissed):

humphd (dismissed):

tcvan0707 (approved):

TueeNguyen (approved):

PR #2949, opened 2022-02-16
[Chore] Remove firebase and users microservice

manekenpix:
## Description

Since we're going to use `supabase` as telescope's main database, there's no need for a `firebase` emulator in development.
This PR removes all references to the emulator and the `users` service.

## Steps to test the PR
- Run `docker-compose --env-file ./config/env.development up --build`
- Make sure `telescope` runs as expected

## Checklist


humphd (dismissed):
I actually think we can remove the Users service in a follow-up too.  Supabase will be the Users service going forward.

The only thing we need to consider is what our "admin" story is.

DukeManh (dismissed):

DukeManh:
Oh, wait, removing this would break the users test because it uses the firebase emulator. @manekenpix, we should remove the users service in this PR as well.

docker/docker-compose.yml
```diff
@@ -198,34 +198,6 @@ services:
       - 'traefik.http.middlewares.strip_feed_discovery_prefix.stripprefix.forceSlash=true'
       - 'traefik.http.routers.feed-discovery.middlewares=strip_feed_discovery_prefix'
 
-  # users service
```
humphd, line 201: Line 100 above, remove the `users` service from the `sso`'s depends:

```
Service 'sso' depends on service 'users' which is undefined.
```

humphd (changes requested):
Looks good, one thing.

menghif (approved):

humphd:
Rebased and fixed in https://github.com/Seneca-CDOT/telescope/commit/aca9f75e2c1359bfa02dfd38f4eb6947f1a05644.

PR #2950, opened 2022-02-16
Move Portainer behind Traefik instead of nginx

humphd:
We are trying to figure out how to organize our various URLs and domains.  We have:

- Web app at `telescope.cdot.systems`
- API servers at `api.telescope.cdot.systems`

And we have a growing set of "admin" apps: web apps that our developers need to access, separate from our web/api.

Trying to mix the "admin" apps into the current nginx/traefik web+api structure is proving difficult.  We use a path-based routing system vs. an sub-domain routing system, and it's causing problems with web apps that assume they live at the `/` of the web server.

I want to move our admin apps to their own sub-domains: `<admin-app-name>.telescope.cdot.systems`.  I wanted to try this with Portainer first, since we already run it in staging/production, and should be able to test.

This change updates our [Traefik routers](https://doc.traefik.io/traefik/routing/routers/) to use the host as well as path prefix when making decisions about which container should handle a request.

I've also added a new environment variable for the `PORTAINER_HOST` name, which we will use in our nginx config to define another server name.

This can't be tested locally, beyond making sure that it doesn't break local development (it seems fine for me).  We probably need to run it on staging to see how it behaves.

### DIFF
--- a/docker/production.yml
+++ b/docker/production.yml
@@ -139,11 +139,11 @@ services:
     restart: unless-stopped
     depends_on:
       - telescope
-    # This makes nginx swap only the 2 specified environment variables in our template (see:
+    # This makes nginx swap only the specified environment variables in our template (see:
     # https://github.com/docker-library/docs/issues/496) and then reload this configuration
     # (and the associated certificates) every six hours in the background and launches nginx
     # in the foreground.
-    command: /bin/sh -c "envsubst '$$TELESCOPE_HOST $$API_HOST' < /etc/nginx/nginx.conf.template > /etc/nginx/nginx.conf && while :; do sleep 6h & wait $${!}; nginx -s reload; done & nginx -g 'daemon off;'"
+    command: /bin/sh -c "envsubst '$$TELESCOPE_HOST $$API_HOST $$PORTAINER_HOST' < /etc/nginx/nginx.conf.template > /etc/nginx/nginx.conf && while :; do sleep 6h & wait $${!}; nginx -s reload; done & nginx -g 'daemon off;'"
 
   # API Gateway running on 8888. We use nginx in front of this for secure/compressed/cached responses
   traefik:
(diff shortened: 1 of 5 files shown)


docker/production.yml
```diff
@@ -139,11 +139,11 @@ services:
     restart: unless-stopped
     depends_on:
       - telescope
-    # This makes nginx swap only the 2 specified environment variables in our template (see:
+    # This makes nginx swap only the specified environment variables in our template (see:
     # https://github.com/docker-library/docs/issues/496) and then reload this configuration
     # (and the associated certificates) every six hours in the background and launches nginx
     # in the foreground.
-    command: /bin/sh -c "envsubst '$$TELESCOPE_HOST $$API_HOST' < /etc/nginx/nginx.conf.template > /etc/nginx/nginx.conf && while :; do sleep 6h & wait $${!}; nginx -s reload; done & nginx -g 'daemon off;'"
+    command: /bin/sh -c "envsubst '$$TELESCOPE_HOST $$API_HOST $$PORTAINER_HOST' < /etc/nginx/nginx.conf.template > /etc/nginx/nginx.conf && while :; do sleep 6h & wait $${!}; nginx -s reload; done & nginx -g 'daemon off;'"
```
Kevan-Y, line 146: I don't quite understand this part.
Are we loading `env` from `/etc/nginx/nginx.conf.template` into `/etc/nginx/nginx.conf ` every 6h?

Kevan-Y (approved):
Let's try on staging

docker/production.yml
humphd: For some reason, we're restarting nginx every 6 hours (maybe for a memory leak, I can't remember. @manekenpix might remember?), and when we do, we are taking what's in the env and using those variables to replace things in the template files in order to create the conf files again.

This seems unnecessary?  But I can't remember why we did it.

manekenpix: If I remember correctly, we restart `nginx` every 6 hours in case our `ssl` certificates get renewed. This way, we can reload them and avoid a [Microsoft Exchange](https://www.bleepingcomputer.com/news/microsoft/microsoft-exchange-admin-portal-blocked-by-expired-ssl-certificate/) situation.

humphd: I should add a comment to this effect, I totally forgot.

humphd:
I'd say this didn't work, https://portainer.dev.telescope.cdot.systems/ isn't resolving, I get:

```
This site can’t be reached
Check if there is a typo in portainer.dev.telescope.cdot.systems.
DNS_PROBE_FINISHED_NXDOMAIN
```

I think we have more config we have to do in order to support this CNAME.

I'll file an issue and talk to Chris.

PR #2953, opened 2022-02-16
Stop the dashboard UI from stutter by checking the build sha

### LINKED ISSUE #2773: Dashboard build log updates previous build log too often
The build log page polls the autodeployment server constantly, and it makes the UI stutter as we constantly reload the same data over and over.  We need to get smarter about what we do when we get back the identical data to the previous run (e.g., we shouldn't bother re-rendering if it's the same).

Screen Shot 2022-01-29 at 9 20 17 AM

tpmai22:
## Issue This PR Addresses
Fix #2773 

## Type of Change

## Description
Checking the `sha` of the build and stop re-rendering if we have the same `sha` displayed in the `build log` and the `build header` already

## Checklist

### DIFF
--- a/src/api/status/public/js/build-log/check-for-build.js
+++ b/src/api/status/public/js/build-log/check-for-build.js
@@ -3,11 +3,19 @@ import terminal from './terminal.js';
 import buildHeader from './build-header.js';
 import showToast from '../utils/toast.js';
 
+let build;
+
 export default async function checkForBuild() {
   const status = await checkBuildStatus();
 
   // Prefer the current build, but fallback to the previous one
-  const build = status.current ?? status.previous;
+  const buildData = status.current ?? status.previous;
+
+  if (build && build.sha === buildData.sha) {
+    return;
+  }
+
+  build = buildData;
 
   // Render the build header info
   buildHeader(build);
--- a/src/api/status/public/js/pages/build.js
+++ b/src/api/status/public/js/pages/build.js
@@ -3,5 +3,5 @@ import './main.js';
 
 window.addEventListener('load', () => {
   checkForBuild();
-  setInterval(checkForBuild, 5000);
+  setInterval(checkForBuild, 10000);
 });


src/api/status/public/js/pages/build.js
```diff
@@ -3,5 +3,5 @@ import './main.js';
 
 window.addEventListener('load', () => {
   checkForBuild();
-  setInterval(checkForBuild, 5000);
+  setInterval(checkForBuild, 30000);
```
humphd, line 6: This can probably be every 10s, since you'll ignore it if the build hasn't changed.

humphd (approved):

TueeNguyen (approved):

src/api/status/public/js/build-log/check-for-build.js
```diff
@@ -3,11 +3,19 @@ import terminal from './terminal.js';
 import buildHeader from './build-header.js';
 import showToast from '../utils/toast.js';
 
+let build;
```
tcvan0707, line 6: Should we put this variable inside the function `checkBuildStatus()`?

DukeManh: Putting it here is ok, we want to have a variable here containing the build being shown in the build logs. This is compared against upcoming builds, if the same, skip writing a new log to the terminal.

DukeManh (approved):

tcvan0707 (approved):

src/api/status/public/js/build-log/check-for-build.js
tpmai22: I mean the best practice is to avoid global variable @humphd can I have your thought on this ?

humphd: It's a module-level variable vs. global, so it's not terrible.  If it really bothered you, you could wrap this data + your function below in a class, so they go together.  But this is no worse than `showToast`, which is also being defined at the module level.

In general I agree with you: fewer variables, reduce their scope.  But this is hard to avoid completely.

tpmai22: Understood !

TueeNguyen (approved):

PR #2976, opened 2022-02-17
Set packages to private:true so they don't get published by pnpm

### LINKED ISSUE #2974: Stop pnpm from publishing autodeployment
Our `pnpm publish` script is working a little too well, and it pushed out a new package: https://www.npmjs.com/package/@senecacdot/autodeployment.

I think we need `"private": true` in https://github.com/Seneca-CDOT/telescope/blob/master/tools/autodeployment/package.json to stop that.

humphd:
Fixes #2974 

This should stop `pnpm publish` from publishing these packages to npm.

### DIFF
--- a/tools/autodeployment/package.json
+++ b/tools/autodeployment/package.json
@@ -1,5 +1,6 @@
 {
   "name": "@senecacdot/autodeployment",
+  "private": true,
   "version": "0.0.1",
   "description": "A tool for automatic deployment",
   "main": "server.js",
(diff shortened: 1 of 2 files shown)


TueeNguyen (approved):

TueeNguyen:
Should we do the same for the eslint one?

DukeManh (approved):

humphd:
No, eslint we want to publish.

Issue #2860, opened 2022-02-08
Migrate Planet Feed to Supabase db


joelazwar:
**What would you like to be added**:
Part of the Supabase implementation, we need to move the list of feeds from the Seneca CDOT wiki page to new table(s)

We need to figure out the schema to store and relate the tables.

Note: Duc has made a PR regarding this "[Handle Telescope user registration and login](https://github.com/Seneca-CDOT/telescope/pull/2844/files)", but we plan to split the PR into smaller manageable parts.

**Why would you like this to be added**:

To further progress implementing Supabase to Telescope

humphd:
The code in https://github.com/Seneca-CDOT/telescope/tree/master/tools/migrate can help with this.

In our discussions today, we said that we'd untangle the idea of a user from feeds, so that we can create feeds in our database, but not necessarily associate them with a user.  This could be done via an optional foreign key value that links to a user; or we could use a third table to link `users <-> user_feeds <-> feeds`.  It wasn't clear which way to go, but the group seemed to lean toward making the ownership field for a feed optional.

JerryHue:
For the sake of simplicity, I suggest we go with the option that @DukeManh proposed: an optional field.

For this to work, we need to keep two facts: 
- feed url
- author name

The `feeds` table already has a column to store the URL, so where does the author name go?

I suggest we violate some rules of database design for the sake of simplicity and add a field that stores the author name (call it `author_name`). The author name refers to the name stored in the planet feed list. However, on the application, we will choose by default the `author_name` UNLESS there is an `user_id` key defined, which means we have to use the information from the `telescope_profiles` table.

humphd:
I like this.  We would end up with:

|Column | Notes |
|--------|-------|
|url         | Unique, primary key for a feed|
|author_name| Optional value from planet feed wiki, unused for future records |
|telescope_user_id| Optional foreign key to a Telescope user who owns the feed|

We still need to add some more data here.  Our concept of a `Feed` is a bit more complex than just having `url` and `author`: https://github.com/Seneca-CDOT/telescope/blob/master/src/backend/data/feed.js#L27-L44

For example, we still need:

- `link` - the URL of the actual thing the feed refers to.  This might be a blog URL (vs. a feed URL), a YouTube channel, a Twitch channel, etc.
[...]

PR #2977, opened 2022-02-17
Update telescope_profiles and feeds tables

DukeManh:
## Issue This PR Addresses
Part of #2860

This updates `feeds` table schema in preparation for migrating [Planet feed list](https://wiki.cdot.senecacollege.ca/wiki/Planet_CDOT_Feed_List) to Supabase. 

Schema
|  Field | Description  | 
|---|---|
|  url |   text -  primary key - unique feed url|
|   user_id |  text - optional foreign key - A register can claim an existing feed when they register for Telescope|
|  author_name |  text - author name get from Planet, will be unused when the feed is linked with a user |
|  html_url |  text optional - Actually blog URL of the RSS feed |
|  type |  'blog' \| 'youtube' \| 'twitch' |
|  flagged |  boolean |
|  Invalid |  boolean |

## Steps to test the PR

1. Pick up the new schema script
   ```
    docker-compose stop db && docker-compose rm db
    docker-compose up db -d
   ```
2. Rebuild the SSO service
   ```
    docker-compose build sso
   ```
3. Start all services
   ``` 
    pnpm  dev
    pnpm services:start
   ```
4. Register Telescope user at http://localhost:8000/signup
5. If a feed entered is already associated with a user, registration should fail


DukeManh:
@tcvan0707 This made a slight change to how a feed is considered to be invalid. A user can add an existing feed if it is not already linked with a Telescope user, or the user_id is null.

DukeManh:
@humphd and @JerryHue, I might miss some properties as suggested here https://github.com/Seneca-CDOT/telescope/issues/2860#issuecomment-1032927389, let me know.

TueeNguyen:
@JerryHue said to change `link` to `source` or `src_link`. I don't think `link` is descriptive enough but for me it's fine

DukeManh:
> html_url is probably better than link, such that url is a feed URL, and html_url is for HTML.

I changed link -> html_url

cindyledev:
I noticed you initially created the `telescope_profiles` table without explicitly passing the schema which means Postgres creates the table in the `public` schema. Would this be a problem in the future?

DukeManh:
@cindyledev, we want to put them public schema so they are exposed to the client.
When you write `supabase.from('telescope_profiles').select('*')`, supabase is going to look for the table in the public schema. There shouldn't be a problem if we have [RLS](https://app.supabase.io/project/rrvwnhzlserkxkyjtutq/editor/16825)  to give/deny access for certain users.

docker/supabase/volumes/db/init/04-telescope-schema.sql
```diff
@@ -0,0 +1,41 @@
...
+  github_username citext UNIQUE NOT NULL,
+  github_avatar_url text NOT NULL
+);
+
+-- Planet CDOT Feed List, see https://wiki.cdot.senecacollege.ca/wiki/Planet_CDOT_Feed_List
+CREATE TABLE IF NOT EXISTS feeds (
+  url text PRIMARY KEY,
+  user_id text references telescope_profiles ON DELETE CASCADE, -- optional, a user can claim an existing feed when they register
+  author_name text, -- owner of a feed, maybe unused when the feed is linked with an actual user
+  htm_url text, -- actual URL the feed refers to, could be a blog URL, a Youtube or Twitch channel
```
humphd, line 22: `html_url` (missing `l`)

docker/supabase/volumes/db/init/04-telescope-schema.sql
```diff
@@ -0,0 +1,41 @@
...
+  email citext UNIQUE NOT NULL,
+  github_username citext UNIQUE NOT NULL,
+  github_avatar_url text NOT NULL
+);
+
+-- Planet CDOT Feed List, see https://wiki.cdot.senecacollege.ca/wiki/Planet_CDOT_Feed_List
+CREATE TABLE IF NOT EXISTS feeds (
+  url text PRIMARY KEY,
+  user_id text references telescope_profiles ON DELETE CASCADE, -- optional, a user can claim an existing feed when they register
+  author_name text, -- owner of a feed, maybe unused when the feed is linked with an actual user
```
humphd, line 21: Because `author_name` and `html_url` are only relevant for Wiki/legacy feeds, should we prefix the names (e.g., `wiki_author_name` and `wiki_html_url`)?  So that it's clear that these don't cross-over with the user info?

humphd (changes requested):
Looks good, a few small things.

docker/supabase/volumes/db/init/04-telescope-schema.sql
DukeManh: @humphd, `wiki_author_name` sounds good, do we have the html_url (the main blog url) of a feed in the Planet Feed?

Kevan-Y:
@DukeManh followed the steps. I'm getting this 🔽
image
image
I have the user stored in the database, but not for the feed.

DukeManh:
You'll need to run the frontend server with `pnpm run dev`, I've updated the instructions. Can you try again?

humphd (approved):
Good point, I'm wrong about html url from wiki.  Looks good

docker/supabase/volumes/db/init/04-telescope-schema.sql
```diff
@@ -0,0 +1,41 @@
...
+CREATE TABLE IF NOT EXISTS telescope_profiles (
+  id text PRIMARY KEY,
+  registered_at timestamptz DEFAULT timezone('utc'::text, now()) NOT NULL,
+  updated_at timestamptz DEFAULT timezone('utc'::text, now()) NOT NULL,
+  display_name citext UNIQUE NOT NULL,
+  first_name text NOT NULL,
+  last_name text NOT NULL,
+  email citext UNIQUE NOT NULL,
+  github_username citext UNIQUE NOT NULL,
+  github_avatar_url text NOT NULL
```
JerryHue, line 14: Would we want to bake in the GitHub information inside the Telescope profile?

docker/supabase/volumes/db/init/04-telescope-schema.sql
```diff
@@ -0,0 +1,41 @@
...
+);
+
+-- Planet CDOT Feed List, see https://wiki.cdot.senecacollege.ca/wiki/Planet_CDOT_Feed_List
+CREATE TABLE IF NOT EXISTS feeds (
+  url text PRIMARY KEY,
+  user_id text references telescope_profiles ON DELETE CASCADE, -- optional, a user can claim an existing feed when they register
```
JerryHue, line 20: The `ON DELETE CASCADE` means that when a user is deleted, their feed would also be deleted, or is it the other way around?

docker/supabase/volumes/db/init/04-telescope-schema.sql
cindyledev: We assume all Telescope users would have a GitHub profile. Thinking aloud. An alternative approach could be having a separate table for GitHub information and forming a relationship between the two tables.

```
CREATE TABLE IF NOT EXISTS telescope_profiles (
  id text PRIMARY KEY,
  registered_at timestamptz DEFAULT timezone('utc'::text, now()) NOT NULL,
  updated_at timestamptz DEFAULT timezone('utc'::text, now()) NOT NULL,
  display_name citext UNIQUE NOT NULL,
  first_name text NOT NULL,
  last_name text NOT NULL,
  email citext UNIQUE NOT NULL
);

CREATE TABLE IF NOT EXISTS github_info (
  github_username citext UNIQUE NOT NULL,
  github_avatar_url text NOT NULL,
  CONSTRAINT fk_telescope_profiles FOREIGN KEY(github_username) REFERENCES telescope_profiles(id)
);
```
This way when a user signs up on Telescope, they still have to provide a GitHub account but the GitHub data is kept in another table.

I believe GitHub usernames can be changed so we probably need to use something else the primary key.

Untested so this is just a thought.

docker/supabase/volumes/db/init/04-telescope-schema.sql
cindyledev: `ON DELETE CASCADE` would mean if we delete a telescope profile, the feed connected to that profile will also be deleted. So a hypothetical scenario, I sign up as a Telescope user but I'm a baddie and I claim a feed that's obviously not mine, such as:

```
[https://epicgamercharlie.wordpress.com/feed/]
name = Charlie Le
```

then I delete my telescope profile and Charlie's feed would get deleted as well. 

I pick on Charlie because we had a couple of classes together and I was `cle15` and he was `cle17`. My arch-nemesis.

DukeManh: This means feeds will be deleted if a user is deleted. The default behavior would prevent user deletion if a feed is referencing its primary key.

docker/supabase/volumes/db/init/04-telescope-schema.sql
DukeManh: I'm not sure what you mean here.

docker/supabase/volumes/db/init/04-telescope-schema.sql
DukeManh: @cindyledev, anyone can go and claim feeds + GitHub account that's not theirs, it's built on trust pretty much. Maybe we should file an issue to prevent 'baddies' in the future.

cindyledev: A great lazy way would be to only allow admins to delete feeds because who will stop me from committing such a sin? >:D

docker/supabase/volumes/db/init/04-telescope-schema.sql
DukeManh: I don't see any benefits of managing an extra table to store GitHub data.

Having another table we would have to do more queries to register a user, get user info, and manage additional RLS policies for that table.

docker/supabase/volumes/db/init/04-telescope-schema.sql
humphd: Every user is already a Seneca user, and if someone does something bad, we'll deal with them directly.

I don't think we have to worry about anonymous trolls doing weird things here.

DukeManh: When this PR is in, there's will be another PR for RLS of these two tables which restrict certain queries can be done on by users.

docker/supabase/volumes/db/init/04-telescope-schema.sql
cindyledev: Right now, we're only using the github username and avatar URL, but what if we want to use more github user info in the future? Say we want to have a User Profile page on Telescope and want to display a few more interesting data points from the user's github like how many followers they have, if they're hireable, how long they've been on GitHub, etc.

DukeManh: I don't think it's a good idea to manage that, have to keep it in sync with GitHub, etc. Why don't we use GitHub APIs to get more user info if we ever want to?

cindyledev: I just read the GitHub Docs, you're right, it would make sense to just use the GitHub API in this case.

docker/supabase/volumes/db/init/04-telescope-schema.sql
cindyledev, line 20: Can `telescope_profiles` be referenced like this or does it need a column too?

```
user_id text references telescope_profiles(id) ON DELETE CASCADE, 
```

vs

```
user_id text references telescope_profiles ON DELETE CASCADE, 
```

DukeManh: They are equivalent.
[Table constraints](https://www.postgresql.org/docs/9.2/ddl-constraints.html)

>  in absence of a column list the primary key of the referenced table is used as the referenced column(s).

cindyledev: Wow Supabase is really sophisticated

DukeManh: It's actually a Postgres Rule 😜

cindyledev: I stand corrected XD

DukeManh: Feel free to have any question, I learned answering questions too.

Kevan-Y (approved):

DukeManh:
Thank you, everyone.

PR #2978, opened 2022-02-17
Use random UUID for JWT_TOKEN on every deploy

humphd:
This injects a random UUID string as our JWT_TOKEN when we deploy.  It will mean that our JWT_TOKEN won't be empty when we launch the containers on staging and production.

Later, we should employ a better secrets management technique.

### DIFF
--- a/tools/autodeployment/deploy.sh
+++ b/tools/autodeployment/deploy.sh
@@ -31,7 +31,19 @@ else
     OLD="green"
 fi
 
+
+# Include the current commit sha that we're building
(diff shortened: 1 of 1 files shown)


manekenpix (approved):
:+1:

joelazwar (approved):

rclee91 (approved):

humphd:
Filed https://github.com/Seneca-CDOT/telescope/issues/2979 for later to solve this in a better way.

PR #3008, opened 2022-02-20
Increase buffer size for the dependency script

### LINKED ISSUE #3006: Dependency-discovery gives empty array in staging and production
**What happened**: The `dependency-discovery` service gives an empty array for the list of projects that Telescope has a dependencies

**What should have happened**: It should have given an array with several project names.

**How to reproduce it (as precise as possible)**: Go to `https://api.telescope.cdot.systems/v1/dependency-discovery/projects`

It seems that the `deps.txt` the `dependency-discovery` service depends on got deleted in d3e09d74703b519c0c2919b69d9c3eeb3f901369.

After @manekenpix run the script that generates `deps.txt`, we noticed it might be an issue related to the buffer size being too small.

JerryHue:
## Issue This PR Addresses

Fixes #3006 

## Type of Change

## Description

This may fix the issue, however, I am not sure that it will completely fix it. The issue that @manekenpix showed is related to the output being too big so it cannot hold it in the buffer. This PR just increases the maximum size of the buffer.

The reason I am unsure that these two errors are linked is because the error happens before writing to the file, however, the file itself looks like it was truncated and did not receive any output.

## Steps to test the PR

1. Get my changes
2. You can run the script with `node tools/collect-dependencies.js -- -o src/api/dependency-discovery/deps.txt`.

We might need to test the `preversion` script. I would need some assistance for this.

## Checklist

### DIFF
--- a/tools/collect-dependencies.js
+++ b/tools/collect-dependencies.js
@@ -13,7 +13,13 @@ if (option !== '-o' || !filename) {
 console.log('Starting to collect dependencies');
 
 console.log('Invoking pnpm list...');
-const output = execSync('pnpm list -w -r --parseable --depth=2').toString('utf8');
+
+// Warning: the maxBuffer has been increased to 10 MB as a quick solution.
+// In the case we hit the limit of the buffer again, we shall rewrite the
+// script for a better way to handle it.
+const output = execSync('pnpm list -w -r --parseable --depth=2', {
+  maxBuffer: 1024 * 1024 * 10, // approx. 10 Megabyte = 1024 bytes per Kilobyte * 1024 Kilobytes per Megabyte * 10 Megabytes
+}).toString('utf8');
 
 const regex = /.*\.pnpm\/(@?[^@]+).*/g;
 


manekenpix (approved):

cindyledev (approved):

tools/collect-dependencies.js
```diff
@@ -13,7 +13,13 @@ if (option !== '-o' || !filename) {
 console.log('Starting to collect dependencies');
 
 console.log('Invoking pnpm list...');
-const output = execSync('pnpm list -w -r --parseable --depth=2').toString('utf8');
+
+// Warning: the maxBuffer has been increased to 10 MB as a quick solution.
+// In the case we hit the limit of the buffer again, we shall rewrite the
+// script for a better way to handle it.
+const output = execSync('pnpm list -w -r --parseable --depth=2', {
```
humphd, line 20: @JerryHue A better fix for this would be to not buffer at all, and use the `stdout` stream you get back from `spawn` to `pipe` the output to an writable file stream.  There's no point holding anything in memory, just redirect it to disk.

JerryHue: But we need to process the input from `stdout` to get the name of the packages, as well as only have unique project names (some project names appear several times due to some places using different versions).

Issue #2897, opened 2022-02-12
Twitch iframes are filtered out by sanitizer


humphd:
<img width="1033" alt="Screen Shot 2022-02-12 at 9 29 50 AM" src="https://user-images.githubusercontent.com/427398/153715345-1ef5b9f3-4569-40cc-82b2-996f1bf46315.png">

We need to update the sanitizer to allow the twitch iframe origin.

humphd:
That's for the front-end, we also need to fix https://github.com/Seneca-CDOT/telescope/blob/master/src/backend/utils/html/sanitize.js#L50-L57 in the backend.  And the parser service would need this too.

HyperTHD:
We also need to update the tests as well to ensure that this is working as intended. This is where they should go https://github.com/Seneca-CDOT/telescope/blob/master/test/sanitize-html.test.js

HyperTHD:
Taking this

humphd:
I'm seeing this for a second time, still unfixed.  @HyperTHD do you have time to work on this, or should we give to someone else?

HyperTHD:
> I'm seeing this for a second time, still unfixed. @HyperTHD do you have time to work on this, or should we give to someone else?

Been pretty busy lately, but got some time. I'll have a PR up today hopefully for this

manekenpix:
The issue is still there
![image](https://user-images.githubusercontent.com/23108901/155035782-ae8ae4db-92e6-4241-90ff-b95149ef5db2.png)

humphd:
Won't it need to re-index for this to get applied to that post?

manekenpix:
@humphd or, I forgot, let me make the changes on staging to see if it solves the issue

manekenpix:
No, still the same issue
![image](https://user-images.githubusercontent.com/23108901/155037855-29f4703e-f56f-4696-8400-42cafc2ea1ed.png)

manekenpix:
I think this is probably related to @2306, I see the same behaviour in both cases.

DukeManh:
Could be because the hostname is supposed to be `player.twitch.tv` vs `twtich.tv`
@HyperTHD

DukeManh:
This is Alex's twitch iframe


```html
<iframe src="https://player.twitch.tv/?video=1303803789&amp;parent=dev.to&amp;autoplay=false" allowfullscreen="true" scrolling="no" width="710" height="399" frameborder="0">
</iframe>
```

[...]

PR #2846, opened 2022-02-06
Making parser service work

TueeNguyen:
## Issue This PR Addresses
Fixes #2111 
This is a big change so I'd rather have it looked at early

## Type of Change

- [ ] **Bugfix**: Change which fixes an issue
- [x] **New Feature**: Change which adds functionality
- [ ] **Documentation Update**: Change which improves documentation
- [ ] **UI**: Change which improves UI

## Description
- Add changes based on backend and #2706, #2581
- Export Redis constructor instead of Redis instance
- Bump node-fetch to 2.6.7

## Steps to test the PR
1. `pnpm install` in root 
2. `pnpm services:start redis elasticsearch traefik posts`
4. `npm run dev` and go to `localhost:8000` or look at the terminal log
**Notes**: I have not tested this for Youtube feeds, will edit this with test result
## Next steps
Should I open small issues for these steps?
1. Open a proper PR or keep this 
2. Add unit and e2e tests
3. Dockerize parser service

## Checklist

- [ ] **Quality**: This PR builds and passes our npm test and works locally
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not (if applicable)
- [ ] **Documentation**: This PR includes updated/added documentation to user exposed functionality or configuration variables are added/changed or an explanation of why it does not(if applicable)

PR #3023, opened 2022-02-22
Switched twitch iframe from `twitch.tv` to `player.twitch.tv`

HyperTHD:
I added `twitch.tv` to our list of allowed urls to not be sanitized, however, it appears that twitch instead uses a different url for embedded content. I switched the url source, so hopefully this should solve #2897 

cc @humphd @manekenpix


manekenpix:
Firefox
image

Chromium
image

DukeManh (approved):

cindyledev:
Not sure if I'm looking at the right thing here but from Alex's blog on dev.to, the iframe is
```
<iframe src="https://player.twitch.tv/?video=1303803789&parent=dev.to&autoplay=false" height="399" width="710" allowfullscreen="true" frameborder="0" scrolling="no">
</iframe>
```

If I tried to take only
the `https://player.twitch.tv/?video=1303803789` part so the iframe would look like this

```
<iframe src="https://player.twitch.tv/?video=1303803789" height="399" width="710" allowfullscreen="true" frameborder="0" scrolling="no">
</iframe>
```

I get this interesting error:
2022-02-22 10_20_32-Window

If I include the `parent` attribute but change it to `localhost` (because I'm on `localhost`) so the iframe would look like this:

```
<iframe src="https://player.twitch.tv/?video=1303803789&parent=localhostautoplay=false">
</iframe>
```

The video shows up
2022-02-22 10_32_11-Window

cindyledev:
I think this is more than just sanitizing the URL on our side, we probably have to include the `parent` attribute as well

humphd:
When it doubt, read the docs! https://dev.twitch.tv/docs/embed/video-and-clips/

src/backend/web/app.js
```diff
@@ -23,7 +23,7 @@ app.use(
             directives: {
               defaultSrc: ["'self'"],
               fontSrc: ["'self'", 'https:', 'data:'],
-              frameSrc: ["'self'", '*.youtube.com', '*.vimeo.com', '*.twitch.tv'],
+              frameSrc: ["'self'", '*.youtube.com', '*.vimeo.com', '*.player.twitch.tv'],
```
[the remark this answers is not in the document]
HyperTHD: When looking at Alex's iframe on her blog, there was a `blob` placed before the https url. Not sure if it's something that's included in all twitch links but if it's not, can be removed

humphd:
Left a note in #2846 about not losing track of this fix as we land both.

HyperTHD:
Wtih #2846 in, I'll update the `sanitizehtml.js` file inside the parser service as well, so @TueeNguyen don't gotta worry about that

test/sanitize-html.test.js
```diff
@@ -145,7 +145,11 @@ describe('Sanitize HTML', () => {
   });
 
   test('twitch.tv embedded content should not be removed', () => {
-    const data = sanitizeHTML('<iframe src="https://www.twitch.tv/0pensrc"></iframe>');
-    expect(data).toBe('<iframe src="https://www.twitch.tv/0pensrc"></iframe>');
+    const data = sanitizeHTML(
+      '<iframe src="https://player.twitch.tv/?video=1303803789&parent=dev.to"></iframe>'
```
DukeManh, line 149: According to [this](https://dev.twitch.tv/docs/embed/video-and-clips/), `parent` is the domain that will embed the stream. So this should be telescope.cdot.systems vs dev.to. I'd say we have to take a twitch iframe, and manually add a `parent` attribute to it.
Probably will solve the problem @manekenpix  mentioned [here](https://github.com/Seneca-CDOT/telescope/pull/3023#issuecomment-1047363874)

DukeManh (changes requested):

test/sanitize-html.test.js
HyperTHD: Will this also affect the host domain if ran in a different environment like local or staging? Was thinking of just grabbing that from the `.env` and using that here. Thoughts?

DukeManh: That's what I think too, we can grab the value it from .env

src/backend/web/app.js
```diff
@@ -23,7 +23,7 @@ app.use(
             directives: {
               defaultSrc: ["'self'"],
               fontSrc: ["'self'", 'https:', 'data:'],
-              frameSrc: ["'self'", '*.youtube.com', '*.vimeo.com', '*.twitch.tv'],
+              frameSrc: ["'self'", '*.youtube.com', '*.vimeo.com', 'player.twitch.tv'],
```
TueeNguyen, line 26: We don't run this web server anymore right?

docker/docker-compose.yml
```diff
@@ -212,6 +212,7 @@ services:
   #     - JWT_ISSUER
   #     - JWT_AUDIENCE
   #     - JWT_SECRET
+  #     - WEB_URL
```
TueeNguyen, line 215: thanks!

TueeNguyen (approved):

tcvan0707 (dismissed):

src/api/parser/src/utils/html/sanitize.js
```diff
@@ -50,6 +51,8 @@ module.exports = (dirty) => {
     allowedIframeHostnames: [
       'www.youtube.com',
       'player.vimeo.com',
+      'player.twitch.tv',
+      WEB_URL,
```
humphd, line 55: Why do we need this here?  Won't this only show-up on the URL via `parant=`?  We're not actually hosting content at this URL?

test/sanitize-html.test.js
```diff
@@ -145,7 +146,12 @@ describe('Sanitize HTML', () => {
...
 
   test('twitch.tv embedded content should not be removed', () => {
-    const data = sanitizeHTML('<iframe src="https://www.twitch.tv/0pensrc"></iframe>');
-    expect(data).toBe('<iframe src="https://www.twitch.tv/0pensrc"></iframe>');
+    // The parent domain also embeds the stream, so include the web url used here as well
+    const data = sanitizeHTML(
+      `<iframe src="https://player.twitch.tv/?video=1303803789&parent=${WEB_URL}"></iframe>`
+    );
+    expect(data).toBe(
+      `<iframe src="https://player.twitch.tv/?video=1303803789&amp;parent=${WEB_URL}"></iframe>`
```
humphd, line 154: `encodeURIComponent(WEB_URL)` in both cases.

humphd (changes requested):

src/api/parser/src/utils/html/sanitize.js
cindyledev, line 55: I don't think this is where we want the `WEB_URL`, we need to strip put the `dev.to` from `parent='dev.to` and replace it with `parent={WEB_URL}`

cindyledev (changes requested):

src/api/parser/src/utils/html/sanitize.js
humphd: We need to make a `new URL()` out of the iframe URL in the parser, then break down its query params and modify the `parent` value, then get the `.href` to get the full URL again and put that in the iframe.

src/backend/web/app.js
HyperTHD: Nope, can be removed later on

test/sanitize-html.test.js
```diff
@@ -145,7 +146,12 @@ describe('Sanitize HTML', () => {
   });
 
   test('twitch.tv embedded content should not be removed', () => {
-    const data = sanitizeHTML('<iframe src="https://www.twitch.tv/0pensrc"></iframe>');
-    expect(data).toBe('<iframe src="https://www.twitch.tv/0pensrc"></iframe>');
+    // The parent domain also embeds the stream, so include the web url used here as well
+    const data = sanitizeHTML(
+      `<iframe src="https://player.twitch.tv/?video=1303803789&parent=${WEB_URL}"></iframe>`
```
DukeManh, line 151: Is this ready to be reviewed? You will need to [transfrom](https://github.com/apostrophecms/sanitize-html#transformations) an iframe we get from dev.to or other websites, remove the parent attribute on it, and add our own following Dave's suggestion.

DukeManh (changes requested):

HyperTHD:
Not too familiar with the text transform function of sanitize html, so I'm pretty sure I messed up somewhere

src/backend/utils/html/sanitize.js
```diff
@@ -49,12 +52,21 @@ module.exports = function (dirty) {
...
       'www.youtube.com',
-      'www.twitch.tv',
+      'player.twitch.tv',
       'player.vimeo.com',
       'giphy.com',
       'cdn.embedly.com',
       'open.spotify.com',
       'medium.com',
     ],
+    transformTag: {
```
humphd, line 62: OK, this is really cool, nice work @HyperTHD, I didn't know we could do this.

Reading their docs, I think you want something like this (untested code!):

```js
transformTags: {
  iframe: (tagName, attribs) => {
    // Swap the parent param for Twitch iframe URLs. For example, use a new host for dev.to in:
    // https://player.twitch.tv/?video=1303803789&parent=dev.to&autoplay=false
    const swapParent = (parent, host) => {
      try {
        const urlObj = new URL(attribs.src);
        // Only bother if this is a Twitch embed iframe
        if(urlObj.host === 'player.twitch.tv' && urlObj.searchParams.get('parent')) {
          urlObj.searchParams.set('parent', host);
        }
        return urlObj.href;
      } catch {
        // Return the original untouched
        return url;
      }
    };

    // Merge all the other attributes, but fix src
    return {
      tagName,
      attribs: {
        ...attribs,
        src: swapParent(attribs.src, new URL(WEB_URL).host),
      },
    },
  }
}
```

humphd (changes requested):

src/backend/utils/html/sanitize.js
HyperTHD: Worth a shot, hard to test locally since I'd have to re-index.

aserputov:
@HyperTHD you need to rebase.

src/api/parser/src/utils/html/sanitize.js
humphd: Remove this

test/sanitize-html.test.js
```diff
@@ -145,7 +146,16 @@ describe('Sanitize HTML', () => {
   });
 
   test('twitch.tv embedded content should not be removed', () => {
-    const data = sanitizeHTML('<iframe src="https://www.twitch.tv/0pensrc"></iframe>');
-    expect(data).toBe('<iframe src="https://www.twitch.tv/0pensrc"></iframe>');
+    // The parent domain also embeds the stream, so include the web url used here as well
+    const data = sanitizeHTML(
+      `<iframe src="https://player.twitch.tv/?video=1303803789&parent=${encodeURIComponent(
+        WEB_URL
```
humphd, line 152: This should be `encodeURIComponent(new URL(WEB_URL).host)`

test/sanitize-html.test.js
```diff
@@ -145,7 +146,16 @@ describe('Sanitize HTML', () => {
...
+      )}"></iframe>`
+    );
+    expect(data).toBe(
+      `<iframe src="https://player.twitch.tv/?video=1303803789&amp;parent=${encodeURIComponent(
+        WEB_URL
```
humphd, line 157: And here.

humphd (changes requested):

src/backend/utils/html/sanitize.js
```diff
@@ -2,6 +2,9 @@
 // <img src=x onerror=alert(1)> becomes <img src="x">
 
 const sanitizeHtml = require('sanitize-html');
+const jsdom = require('jsdom');
+const { JSDOM } = jsdom;
```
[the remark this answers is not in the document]
HyperTHD: the eslint file that was in the root directory is not there but in the docs service folder. I've been getting eslint errors regarding `const` being reserved in all files I've worked on but it doesn't affect the CI process either.

DukeManh (changes requested):

src/api/parser/src/utils/html/sanitize.js
```diff
@@ -50,10 +51,34 @@ module.exports = (dirty) => {
...
       'player.vimeo.com',
+      'player.twitch.tv',
       'giphy.com',
       'cdn.embedly.com',
       'open.spotify.com',
       'medium.com',
     ],
+    transformTag: {
+      iframe: (tagName, attribs) => {
+        const swapParent = (parent, host) => {
```
DukeManh, line 62: `parent` parameter is not used. It's better to name this `src` and use it in line 64

HyperTHD: All good!

test/sanitize-html.test.js
```diff
@@ -145,7 +146,12 @@ describe('Sanitize HTML', () => {
...
 
   test('twitch.tv embedded content should not be removed', () => {
-    const data = sanitizeHTML('<iframe src="https://www.twitch.tv/0pensrc"></iframe>');
-    expect(data).toBe('<iframe src="https://www.twitch.tv/0pensrc"></iframe>');
+    // The parent domain also embeds the stream, so include the web url used here as well
+    const data = sanitizeHTML(
+      `<iframe src="https://player.twitch.tv/?video=1303803789&parent=dev.to"></iframe>`
+    );
+    expect(data).toBe(
+      `<iframe src="https://player.twitch.tv/?video=1303803789&amp;parent=dev.to"></iframe>`
```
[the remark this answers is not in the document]
humphd: Also, it looks like you have double-encoding happeing `&amp;`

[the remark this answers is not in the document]
DukeManh: @HyperTHD, I think the option is `transformTags` with an s.

HyperTHD:
> ```
> Sanitize HTML › twitch.tv embedded content should not be removed
> 
>     expect(received).toBe(expected) // Object.is equality
> 
>     Expected: "<iframe src=\"https://player.twitch.tv/?video=1303803789&parent=dev.to\"></iframe>"
>     Received: "<iframe src=\"https://player.twitch.tv/?video=1303803789&amp;parent=localhost%3A8000\"></iframe>"
> 
>       151 |       `<iframe src="https://player.twitch.tv/?video=1303803789&parent=dev.to"></iframe>`
>       152 |     );
>     > 153 |     expect(data).toBe(
>           |                  ^
>       154 |       `<iframe src="https://player.twitch.tv/?video=1303803789&parent=dev.to"></iframe>`
>       155 |     );
>       156 |   });
> 
>       at Object.<anonymous> (test/sanitize-html.test.js:153:18)
> ```

The 

> ```
> Sanitize HTML › twitch.tv embedded content should not be removed
> 
>     expect(received).toBe(expected) // Object.is equality
> 
>     Expected: "<iframe src=\"https://player.twitch.tv/?video=1303803789&parent=dev.to\"></iframe>"
>     Received: "<iframe src=\"https://player.twitch.tv/?video=1303803789&amp;parent=localhost%3A8000\"></iframe>"
> 
>       151 |       `<iframe src="https://player.twitch.tv/?video=1303803789&parent=dev.to"></iframe>`
>       152 |     );
>     > 153 |     expect(data).toBe(
>           |                  ^
>       154 |       `<iframe src="https://player.twitch.tv/?video=1303803789&parent=dev.to"></iframe>`
>       155 |     );
>       156 |   });
> 
>       at Object.<anonymous> (test/sanitize-html.test.js:153:18)
> ```

The parent part is on me, we should be expecting the host to be ours and not `dev.to` after processing that. 
I'm not sure about the & part though

HyperTHD:
Update: Tests should be fine, struggling with transforming the `&`. My experiments with the string functions haven't been working. I can transform the ":" but not the "&". The only issue here is that this current solution will not work on local environment since we include the port number in the web url and I'm also struggling with stripping that out as well.

test/sanitize-html.test.js
```diff
@@ -105,19 +106,19 @@ describe('Sanitize HTML', () => {
 
   test('<iframe> tag to a spotify embed should not get removed', () => {
     const data = sanitizeHTML(
-      '<iframe src="https://open.spotify.com/embed/album/21wMUhXhWLew2zsWQhlYEM?referrer=https%3A%2F%2Fmedium.com%2F%40pedrofonsecadev%2Ftelescope-note-feb-22-2021-8601dfbaa1f0"></iframe>'
```
DukeManh, line 108: is this what we get from the parser?

DukeManh (approved):

test/sanitize-html.test.js
HyperTHD: Not the symbols, those should be what comes out after sanitization, not prior. I can fix that

DukeManh:
@HyperTHD how is this going, is this working locally?

HyperTHD:
> @HyperTHD how is this going, is this working locally?

Nothing yet, looking into this more this afternoon. Once I get that though, this can be safely merged in, it should work for all other environments

src/api/parser/src/utils/html/sanitize.js
```diff
@@ -50,10 +51,41 @@ module.exports = (dirty) => {
...
+            let updatedUrl;
+            if (urlObj.host === 'player.twitch.tv' && urlObj.searchParams.get('parent')) {
+              // Set the parent to localhost in development only
+              if (process.env.NODE_ENV === 'development') {
+                urlObj.searchParams.set('parent', 'localhost');
+              } else {
+                urlObj.searchParams.set('parent', host);
+              }
+            }
+            updatedUrl = urlObj.href.replace('%3A', ':').replace('%2F', '/');
```
humphd, line 74: This needs a comment explaining why (I know why now, but won't in the future).

test/sanitize-html.test.js
```diff
@@ -105,19 +106,19 @@ describe('Sanitize HTML', () => {
...
+      '<iframe src="https://open.spotify.com/embed/album/21wMUhXhWLew2zsWQhlYEM?referrer=https://medium.com/@pedrofonsecadev/telescope-note-feb-22-2021-8601dfbaa1f0"></iframe>'
     );
     expect(data).toBe(
-      '<iframe src="https://open.spotify.com/embed/album/21wMUhXhWLew2zsWQhlYEM?referrer=https%3A%2F%2Fmedium.com%2F%40pedrofonsecadev%2Ftelescope-note-feb-22-2021-8601dfbaa1f0"></iframe>'
+      '<iframe src="https://open.spotify.com/embed/album/21wMUhXhWLew2zsWQhlYEM?referrer=https://medium.com/@pedrofonsecadev/telescope-note-feb-22-2021-8601dfbaa1f0"></iframe>'
```
humphd, line 112: These aren't technically valid URLs (with the :// included like this).  I guess if it passes, it's OK, but it's not ideal.

humphd (changes requested):

DukeManh (dismissed):

test/sanitize-html.test.js
HyperTHD: Would it be better to leave them with their escaped characters as it was originally? For readability purposes, this is currently fine but the escaped characters is right. I can change it back quickly

humphd: I think in the escaped form is better, yes.

src/backend/utils/html/sanitize.js
```diff
@@ -1,7 +1,9 @@
 // Sanitize HTML and prevent XSS attacks
 // <img src=x onerror=alert(1)> becomes <img src="x">
 
+const { default: formatDistance } = require('date-fns/formatDistance/index');
```
DukeManh, line 4:
```suggestion
```

DukeManh (changes requested):

HyperTHD:
Ready to merge @tpmai22 @DukeManh

DukeManh (approved):

tpmai22 (approved):
Good work and thank you for stepping up !

PR #3042, opened 2022-02-24
Fix nginx config for next.js

### LINKED ISSUE #1837: Refactor nginx.conf after moving from Gatsby to Next.js
There is some code in `nginx.conf` that needs to be refactored to work with Next.js: [here](https://github.com/Seneca-CDOT/telescope/blob/master/nginx.conf#L112-L134)

cc @chrispinkney because he wanted to learn more about Nginx

humphd:
I took a shot at cleaning up our cache directives in nginx for next.js vs. Gatsby.  I can't find an authoritative document on how you're supposed to do this, so I looked at what Vercel does in their code, and live on their own sites.

Fixes #1837


manekenpix:
@humphd does it make sense to do this for the [development `nginx` config file](https://github.com/Seneca-CDOT/telescope/blob/master/config/nginx.conf.development.template)?

humphd:
Good idea, done.

manekenpix (approved):

PR #3043, opened 2022-02-24
Add RLS policies

### LINKED ISSUE #2981: Enable Row Level Security in our tables

Currently, a client, using the public `anon` key is permitted to do any queries such as SELECT, INSERT, UPDATE, DELETE. To protect our data in production, we need to write Row Level Security to restrict public access to our database.

That begins with figuring out who is allowed to do what. This is what our policies could look like. Let us know what you think.

|   Telescope Profiles | |
|---|---|
|  Read | all Seneca users  |
| Update  |  Match user_id |
| Insert, Delete  | False  |

|  Feed List | |
|---|---|
|  Read | all users  |
| Update, Insert, Delete  |  Match user_id |

This should happen before we define Supabase in staging and production

Document reference: 
https://www.postgresql.org/docs/current/ddl-rowsecurity.html
https://supabase.com/docs/learn/auth-deep-dive/auth-row-level-security

DukeManh:
## Issue This PR Addresses
Fixes #2981. This adds RLS policies to secure the `feeds` and `telescope_profiles` tables in the public schema.  

## Step to test 

Refer to https://www.youtube.com/watch?v=hU8vmrytsxM&feature=youtu.be

## Checklist


docker/supabase/volumes/db/init/04-telescope-schema.sql
```diff
@@ -39,3 +39,36 @@ CREATE TRIGGER set_timestamp
...
+
+-- Disable all operations, unless being overwritten by other policies
+CREATE POLICY profiles_policy ON telescope_profiles
+    FOR ALL
+    USING (false);
+
+-- Enable read access for Seneca users
+CREATE POLICY profiles_read_policy ON telescope_profiles
+    FOR SELECT
+    USING (((current_setting('request.jwt.claims'::text, true))::json ->> 'email'::text) IS NOT NULL);
```
humphd, line 56: A better claim here would be to use the `role`.  We (currently) set the `role` to an `Array`, and it will be one of:

- `["seneca"]`
- `["seneca", "telescope"]`
- `["seneca", "telescope", "admin"]`

Either you should check to see if `role` includes `"seneca"` (not sure how you do that), or we can also play with how we define the role to be a string (e.g., `role="telescope"` or `role="seneca"`).  

I don't think there's any value in the `"admin"` role, since we're going to use GitHub team membership instead.

src/api/sso/test/e2e/auth-flows.test.js
```diff
@@ -4,14 +4,8 @@ const { hash } = require('@senecacdot/satellite');
...
 const { SSO_URL, ALLOWED_APP_ORIGINS } = process.env;
-const {
-  login,
-  logout,
-  getTokenAndState,
-  createTelescopeUsers,
-  cleanupTelescopeUsers,
-  ensureUsers,
-} = require('./utils');
+const { login, logout, getTokenAndState } = require('./browserUtil');
```
humphd, line 7: `browser-util` please, don't rely on case in the filesystem, as not all filesystems are case sensitive.

src/api/sso/test/e2e/auth-flows.test.js
```diff
@@ -4,14 +4,8 @@ const { hash } = require('@senecacdot/satellite');
...
+const { createTelescopeUsers, cleanupTelescopeUsers, ensureUsers } = require('../supabaseUtil');
```
humphd, line 8: Same thing here: `supabase-util`

src/api/sso/test/e2e/auth-flows.test.js
```diff
@@ -82,8 +76,8 @@ describe('Authentication Flows', () => {
     expect(SSO_URL).toEqual('http://localhost/v1/auth');
   });
 
-  it('should have all expected Telescope users in Users service for test data accounts', () =>
-    ensureUsers(users));
+  it('should have all expected Telescope users in Users service for test data accounts', async () =>
+    await ensureUsers(users));
```
humphd, line 80: You don't need async/await here, just return the promise.

src/api/sso/test/e2e/signup-flow.test.js
```diff
@@ -46,8 +47,8 @@ describe('Signup Flow', () => {
     await page.close();
   });
 
-  it('should have none of the users in the Users service for test data accounts', () =>
-    ensureUsers(users, false));
+  it('should have none of the users in the Users service for test data accounts', async () =>
```
humphd, line 50: No async/await needed.

src/api/sso/test/e2e/super-user.test.js
```diff
@@ -66,8 +67,8 @@ describe('Super User Authentication', () => {
       await page.close();
     });
 
-    it('should have all expected Telescope users in Users service for test data accounts', () =>
-      ensureUsers(users));
+    it('should have all expected Telescope users in Users service for test data accounts', async () =>
```
humphd, line 70: Same here.

src/api/sso/test/e2e/super-user.test.js
```diff
@@ -111,8 +112,8 @@ describe('Super User Authentication', () => {
...
+      await ensureUsers(users));
```
humphd, line 116: And here.

src/api/sso/test/row-level-policies.test.js
```diff
@@ -0,0 +1,350 @@
...
+  },
+  {
+    url: unknowUser.feeds[0],
+    wikiAuthorName: unknowUser.displayName,
+  },
+];
+
+describe('Unauthenticated users with public anon_key', () => {
+  describe('"telescope_profiles" security policies', () => {
+    beforeEach(async () => {
```
humphd, line 74: Same thing here.  If you return a `Promise` to Jest, it will do the right thing.

src/api/sso/test/row-level-policies.test.js
```diff
@@ -0,0 +1,350 @@
...
+      await createTelescopeUsers(users);
+    });
+
+    afterEach(async () => {
+      await cleanupTelescopeUsers(users);
```
humphd, line 79: And here.

src/api/sso/test/row-level-policies.test.js
```diff
@@ -0,0 +1,350 @@
...
+  ensureUsers,
+  createTelescopeUsers,
+  cleanupTelescopeUsers,
+  addFeeds,
+  cleanFeeds,
+  ensureFeeds,
+} = require('./supabaseUtil');
+const { createToken } = require('../src/token');
+
+const supabase = createClient(SUPABASE_URL, ANON_KEY);
```
humphd, line 15: I wonder if it would be better to create this client in your tests, or `beforeEach` so it's more clear who you are when making the requests below?  I had to think about it as I read.

src/api/sso/test/row-level-policies.test.js
```diff
@@ -0,0 +1,350 @@
...
+    const accessToken = createToken(
+      email,
+      firstName,
+      lastName,
+      displayName,
+      ['seneca', 'telescope'],
+      githubAvatarUrl
+    );
+    // overwrite public anon_key with a JWT sent from the server
+    supabase.auth.setAuth(accessToken);
```
humphd, line 195: I would create a new Supabase client per-test to do this, so there isn't any global state.

src/api/sso/test/row-level-policies.test.js
```diff
@@ -0,0 +1,350 @@
...
+  });
+
+  afterAll(async () => {
+    supabase.auth.setAuth(ANON_KEY);
```
humphd, line 199: vs. doing this, where you have to undo your state change

src/api/sso/test/row-level-policies.test.js
```diff
@@ -0,0 +1,350 @@
...
+      expect(profiles[0].display_name).toBe(unknowUser.displayName);
+    });
+
+    it("should not allow users to update others' profiles", async () => {
+      const { status } = await supabase
+        .from('telescope_profiles')
+        .update({ display_name: unknowUser.displayName })
+        .match({ id: hash(johannesKepler.email) });
+
+      expect(status).toBe(404);
```
humphd, line 252: So it just doesn't find the match vs. giving an auth error?

src/api/sso/test/row-level-policies.test.js
```diff
@@ -0,0 +1,350 @@
...
+
+    it('should not allow deleting users', async () => {
+      const { status } = await supabase
+        .from('telescope_profiles')
+        .delete()
+        .match({
+          id: hash(hansLippershey.email),
+        });
+
+      expect(status).toBe(200);
```
humphd, line 263: Is your comment in 255 correct?  "should not" and 200 don't sound like the same thing.

src/api/sso/test/row-level-policies.test.js
```diff
@@ -0,0 +1,350 @@
...
+      await ensureUsers(users);
+    });
+  });
+
+  describe('"feeds" security policies', () => {
+    // Create feeds for HansLippershey and an unknown user
```
humphd, line 269: Who is the "unknown user"?

src/api/sso/test/row-level-policies.test.js
```diff
@@ -0,0 +1,350 @@
...
+      });
+
+      expect(status).toBe(200);
+      await ensureFeeds(feedList.slice(0, 1), false);
+    });
+
+    it("should not allow users to delete others' feeds", async () => {
+      const { status } = await supabase.from('feeds').delete().match({ url: unknowUser.feeds[0] });
+
+      expect(status).toBe(200);
```
humphd, line 346: "should not" and 200 again?

humphd (changes requested):
Excited to see this happening!  I did a first reading of it and left feedback/questions.

src/api/sso/test/row-level-policies.test.js
DukeManh: It returns 200 but no feeds are deleted. The line below ensures that.

src/api/sso/test/row-level-policies.test.js
DukeManh: I mean an unregistered user, I'll change it

src/api/sso/test/row-level-policies.test.js
DukeManh: The line below is to ensure no users are deleted. I didn't expect 200 either but whether deletion is successful or not, Supabase/PostgREST returns 200

src/api/sso/test/row-level-policies.test.js
DukeManh: Good to know.

docker/supabase/volumes/db/init/04-telescope-schema.sql
DukeManh: I might need your help on this if you can, I didn't manage to write a valid SQL expression to check if a role is present in the role array.

If you go to http://localhost:8910/project/default/auth/policies, to add a new policy, the SQL expression for that would look something like this
```
('seneca' in ((current_setting('request.jwt.claims'::text, true))::json #> '{role}'::text[]))
```
But that's an invalid expression.
in this case though, I think having an email is already sufficient

src/api/sso/test/row-level-policies.test.js
DukeManh: Yes, sounds good.

docker/supabase/volumes/db/init/04-telescope-schema.sql
humphd: I think we should try to solve this, and I agree it's not obvious (to me) how to do it.  I can't believe how little there is in the way of code examples online for this problem.

Based on what Supabase does, I suggest we write our own function(s) to help us accomplish this.  I want to write something like this (but it's basically pseudo-code):

```sql
create or replace function user_has_role(role text)
returns boolean
language sql stable
as $$
  IF (role IN current_setting('request.jwt.claim.roles', true))
    then return true;
  else
    return false;
  END IF;
$$;
```

Then we can use `user_has_role("seneca")` or `user_has_role("telescope")`

@JerryHue would you be willing to help out here?  We need a way to be able to extract the `Array` of `roles` (`string[]`) from the JWT, and check if a particular value is included.  I *think* using `IN` is the right way, but again, I've spent my career avoiding databases as much as possible!

humphd: For reference, here are examples of the kind of functions that Supabase is using that I want to copy:

- https://github.com/Seneca-CDOT/telescope/blob/master/docker/supabase/volumes/db/init/01-auth-schema.sql#L92-L123
- https://github.com/Seneca-CDOT/telescope/blob/master/docker/supabase/volumes/db/init/02-storage-schema.sql#L37-L93

src/api/sso/test/row-level-policies.test.js
humphd: Do we get back any object with info about how many rows were affected, beyond the status?

docker/supabase/volumes/db/init/04-telescope-schema.sql
humphd: If this is going to block us, I'm also willing to change our `roles` from being `string[]` to just `string`:

- `"roles": ["seneca"]` would become `role: "seneca"`
- `"roles": ["seneca", "telescope"]` would become `role: "telescope"`

If we do that, the existing `auth.role()` function would work right now.

DukeManh: Aright, I figured it out by luck. The key is to use [jsonb](https://www.postgresql.org/docs/current/datatype-json.html#JSON-CONTAINMENT) and its containment check operator. 
```sql
(((current_setting('request.jwt.claims'::text, true))::jsonb #> '{role}'::text[]) ? 'seneca'::text)
```

DukeManh: That converts a jwt into a `jsonb`, gets the value of the `roles`, which is a JSON array, and checks if "seneca" is in that array.

src/api/sso/test/row-level-policies.test.js
DukeManh: I'm using `count` returned now.

DukeManh:
I don't understand why CI keeps failing here but not locally.

humphd:
You're doing a unit test that is trying to access a container, and we only run containers in e2e tests.

src/api/sso/test/row-level-policies.test.js
```diff
@@ -0,0 +1,348 @@
...
+  },
+  {
+    url: unregisteredUser.feeds[0],
+    wikiAuthorName: unregisteredUser.displayName,
+  },
+];
+
+describe('Unauthenticated users with public anon_key', () => {
+  beforeAll(() => {
+    supabase = createClient(SUPABASE_URL, ANON_KEY);
```
humphd, line 70: You need a `let supabase;` above this `beforeAll` so you don't create a global.  Or even better:

```js
const supabase = () => createClient(SUPABASE_URL, ANON_KEY);
...
await supabase().from(...)
```

Not sharing state between tests is ideal.

src/api/sso/test/row-level-policies.test.js
```diff
@@ -0,0 +1,348 @@
...
+    afterEach(() => cleanupTelescopeUsers(users));
+
+    it('should return no users', async () => {
+      const { status, count } = await supabase
+        .from('telescope_profiles')
+        .select('*', { count: 'exact' });
+
+      // the request succeeds but no rows are read
+      expect(status).toBe(200);
+      expect(count).toBe(0);
```
humphd, line 85: Way better, nice!

src/api/sso/test/row-level-policies.test.js
```diff
@@ -0,0 +1,348 @@
...
+        },
+        { returning: 'minimal' }
+      );
+
+      // forbidden, users cannot be created on the client
+      expect(status).toBe(403);
+    });
+
+    it('should not allow updating users', async () => {
+      const { status, count } = await supabase
```
humphd, line 107: You request but never use `count`.  Can you add an assertion on it?  Or remove?

src/api/sso/test/row-level-policies.test.js
```diff
@@ -0,0 +1,348 @@
...
+        });
+
+      // no rows are deleted
+      expect(status).toBe(200);
+      expect(count).toBe(0);
+    });
+  });
+
+  describe('"feeds" security policies', () => {
+    // Create feeds for HansLippershey and an unregistered user
```
humphd, line 131: Should this be:

```js
// Create feeds for HansLippershey, an unregistered user
```

If not, it reads like there should be 2 users here, not one.

src/api/sso/test/row-level-policies.test.js
humphd: Adding to my idea above, what if you do:

```js
const supabase = (key = ANON_KEY) => createClient(SUPABASE_URL, key);
```

Now you can pass a different key in tests where that matters.

src/api/sso/test/row-level-policies.test.js
```diff
@@ -0,0 +1,348 @@
...
+          id: hash(hansLippershey.email),
+        });
+
+      expect(status).toBe(200);
+      expect(count).toBe(0);
+    });
+  });
+
+  describe('"feeds" security policies', () => {
+    // Create feeds for HansLippershey and an unregistered user
```
humphd, line 263: Same wording issue here.

humphd (changes requested):

src/api/sso/test/row-level-policies.test.js
DukeManh: It's meant to create feeds for 2 users, one is registered, Hanlippershey, and one is not.

src/api/sso/test/row-level-policies.test.js
humphd: I don't understand.  You're creating a global when you do this.  You have to use `let` or `const` somewhere.  What am I missing?

DukeManh: When creating separate clients, the second initialized by an accessToken still uses the old key and makes the tests fail. The test for the authorized client only works when I called setAuth.

humphd (changes requested):
This is why I think you should have the `supabase` creation happen via a function, so we never re-use.

src/api/sso/test/row-level-policies.test.js
DukeManh: What I think is Supabase doesn't expect us to create many clients, once one client is initialized, the next calls to `createClient()` returns the already initialized one vs creating a new one.

humphd: It looks like it always makes a new one to me:

https://github.com/supabase/supabase-js/blob/master/src/index.ts#L13-L22

DukeManh: I've tried a couple of things:
- Initialize clients in `beforeAll`: `const supabase = createClient(...)` -> tests fail because supabase is undefined
- Initialize at the top of each `describe` block:  Tests for the authenticated client fail, even though being initialized using an access token

DukeManh: I don't think the client connects to any services, it just initializes a class used for making HTTP request

DukeManh: I think I got it. `createClient` expects a public URL, and na API key, either a public anon key or service role key. Our JWT is not an API key so we can not pass it to `createClient`. The access token can only be used with `setAuth`

src/api/sso/test/e2e/row-level-policies.test.js
```diff
@@ -0,0 +1,350 @@
...
+  createTelescopeUsers,
+  cleanupTelescopeUsers,
+  addFeeds,
+  cleanFeeds,
+} = require('./supabase-util');
+const { createToken } = require('../../src/token');
+
+const createSupabaseClient = () => createClient(SUPABASE_URL, ANON_KEY);
+
+// registered Telescope users
```
humphd, line 15: nit: s/users/user/

src/api/sso/test/e2e/row-level-policies.test.js
```diff
@@ -0,0 +1,350 @@
...
+  displayName: 'Hans Lippershey',
+  isAdmin: false,
+  isFlagged: false,
+  feeds: ['https://imaginary.blog.com/feed/hans'],
+  githubUsername: 'hlippershey',
+  githubAvatarUrl:
+    'https://avatars.githubusercontent.com/u/33902374?s=460&u=733c50a2f50ba297ed30f6b5921a511c2f43bfee&v=4',
+};
+
+// registered Telescope users
```
humphd, line 29: nit: s/users/user/

src/api/sso/test/e2e/row-level-policies.test.js
```diff
@@ -0,0 +1,350 @@
...
+          { returning: 'minimal' }
+        );
+
+        expect(status).toBe(401);
+      });
+
+      it('should allow users to update their own profiles', async () => {
+        const { status, data: profiles } = await supabase
+          .from('telescope_profiles')
+          .update({ display_name: unregisteredUser.displayName }, { returning: 'representation' })
```
humphd, line 233: To make this less confusing, I'd avoid using `unregisteredUser` here, since you're not really doing anything with that user.  All you need is a new string:

```js
const newDisplayName = 'Updated';

.update({ display_name: newDisplayName}, { returning: 'representation' })
...
expect(profiles[0].display_name).toEqual(newDisplayName);
```

src/api/sso/test/e2e/row-level-policies.test.js
```diff
@@ -0,0 +1,350 @@
...
+
+        expect(status).toBe(200);
+        expect(profiles.length).toBe(1);
+        expect(profiles[0].display_name).toBe(unregisteredUser.displayName);
+      });
+
+      it("should not allow users to update others' profiles", async () => {
+        const { status } = await supabase
+          .from('telescope_profiles')
+          .update({ display_name: unregisteredUser.displayName })
```
humphd, line 244: Same thing here, don't confuse things by bringing this other user into this test.

src/api/sso/test/e2e/supabase-util.js
```diff
@@ -0,0 +1,106 @@
...
+          if (result.error) {
+            throw new Error(`unable to delete user from Supabase: ${result.error.message}`);
+          }
+          return result;
+        })
+    )
+  );
+
+// Given an array of users, make sure they all respond with the expected HTTP
+// result via the Users Service.
```
humphd, line 48: This comment needs an update to get rid of 'Users Service'.

src/api/sso/test/e2e/supabase-util.js
```diff
@@ -0,0 +1,106 @@
...
+      { returning: 'minimal' }
+    )
+    .then((result) => {
+      if (result.error) {
+        throw new Error(`unable to insert user into Supabase: ${result.error.message}`);
+      }
+      return result;
+    });
+
+// Delete the Telescope users we created in the Users service.
```
humphd, line 30: This comment needs an update to get rid of 'Users Service'.

src/api/sso/test/e2e/super-user.test.js
```diff
@@ -1,7 +1,8 @@
 // NOTE: you need to run the sso and login services in docker for these to work
```
humphd, line 1: Filed #3509 to get rid of this later.

humphd (changes requested):
Congrats on getting these tests to pass, that's huge!

A few more things to tidy up, and this is ready.  Awesome work.

aserputov (approved):

PR #3049, opened 2022-02-25
Update Satellite to 1.26.0 in all packages

humphd:
Updates to pick up @rclee91's fix in https://github.com/Seneca-CDOT/satellite/pull/64.

### DIFF
--- a/src/api/dependency-discovery/package.json
+++ b/src/api/dependency-discovery/package.json
@@ -18,7 +18,7 @@
   },
   "homepage": "https://github.com/Seneca-CDOT/telescope#readme",
   "dependencies": {
-    "@senecacdot/satellite": "1.25.0"
+    "@senecacdot/satellite": "1.26.0"
   },
   "devDependencies": {
     "env-cmd": "10.1.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -145,23 +145,23 @@ importers:
 
   src/api/dependency-discovery:
     specifiers:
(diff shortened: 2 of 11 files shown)


rclee91 (approved):

src/api/dependency-discovery/package.json
```diff
@@ -18,7 +18,7 @@
   },
   "homepage": "https://github.com/Seneca-CDOT/telescope#readme",
   "dependencies": {
-    "@senecacdot/satellite": "1.25.0"
+    "@senecacdot/satellite": "1.26.0"
```
DukeManh, line 21: I just talked about this with Roxanne yesterday, I think Renovate can take care of updating Satellite if we make the version `^1.26.0` and override the range strategy for it to "[bump](https://docs.renovatebot.com/configuration-options/#rangestrategy)".

> bump = e.g. bump the range even if the new version satisfies the existing range, e.g. ^1.0.0 -> ^1.1.0

humphd: Cool, someone should do a PR to make that happen.

Until then, this will get us on the latest version now across the board, and hopefully fix this crash case.

DukeManh (approved):

src/api/dependency-discovery/package.json
humphd: @rclee91 would you mind filing that ^^ for follow-up?  It doesn't have to be you who fixes, but we should get it filed, or add that comment to an existing issue (we might have one?)

PR #3122, opened 2022-03-06
Run 'redis-check-aof' before starting redis

### LINKED ISSUE #1365: Avoid Redis database corruption on power failure
When we lose power on dev or prod, we sometimes see corruption in the Redis database.  This isn't ideal, and it shouldn't be happening.

Let's figure out what we're doing wrong to cause this.  Others have reported similar situations, see:

- https://github.com/redis/redis/issues/5177
- https://groups.google.com/g/redis-db/c/Jqyk_9sHqYU?pli=1
- https://www.inovex.de/blog/redis-backup/

JerryHue:
## Issue This PR Addresses

Fixes #1365

## Type of Change

## Description

In the case of a power failure, we run the 'redis-check-aof' command to check and fix the 'aof' file that got corrupted after the power failure.

## Steps to test the PR

1. Pull my changes.
2. Ensure that you have no `redis-data` folder (delete if that's the case).
3. Start up the containers (`pnpm run services:start`)
4. Fill the redis DB with posts (`pnpm start`).
5. After running for a while, you may stop the backend.
6. Truncate the `redis-data/appendonly.aof` file (`truncate` in linux and macos, I am not sure in Windows...)
7. Start up the `redis` container now (`pnpm run services:start redis`)
8. Check the logs of the container (`pnpm run services:logs redis`)
8.1. If the log says that redis is waiting for connection, then `redis-check-aof` successfully fixed the file.

## Checklist

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -305,7 +305,12 @@ services:
   redis:
     image: redis:latest
     container_name: 'redis'
-    command: ['redis-server', '--appendonly', 'yes']
+    command:
+      [
+        'bash',
+        '-c',
+        'printf "y\n" | redis-check-aof --fix /data/appendonly.aof ; redis-server --appendonly yes',
+      ]
     volumes:
       - ../redis-data:/data
 
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -305,6 +305,9 @@ services:
   redis:
     image: redis:latest
     container_name: 'redis'
+    # The following command invokes "redis-check-aof" to fix the AOF file in the case
+    # of corruption due to a power outage.
+    # We also need the "printf" command to confirm the prompt of "redis-check-aof"
     command:
       [
         'bash',


docker/docker-compose.yml
```diff
@@ -305,7 +305,12 @@ services:
   redis:
     image: redis:latest
     container_name: 'redis'
-    command: ['redis-server', '--appendonly', 'yes']
+    command:
```
humphd, line 308: Let's add a comment here so we don't forget why we did this, or what it's for.

docker/docker-compose.yml
```diff
@@ -305,7 +305,12 @@ services:
...
+      [
+        'bash',
+        '-c',
+        'printf "y\n" | redis-check-aof --fix /data/appendonly.aof ; redis-server --appendonly yes',
```
manekenpix, line 312: Do we need that `printf "y/n"` part?

JerryHue: It seems to be the case, since I ran it without it, and the `redis-check-aof` would prompt for confirmation. I checked the logs to confirm it.

humphd (approved):

docker/docker-compose.yml
manekenpix: I'm a bit confused. I can't reproduce this (`redis-check-aof` asking for confirmation), shouldn't we just print `y` instead of the 2 options `y/n`? I think we could just pass `y`like this `y | redis-check-aof`.

DukeManh (approved):

tpmai22:
@JerryHue Let's rebase and I think this is good to go

manekenpix (changes requested):
I don't know why this is getting approvals when it doesn't work. I tested this yesterday with a truncated `aof` file and I got the exact same behaviour that we're getting without this change. I had a brief chat with @JerryHue, this might need a different approach.

humphd:
@JerryHue @manekenpix can you two report your findings in here?

JerryHue:
So, @manekenpix told me that to replicate the corruption of the AOF file during a power outage, we can truncate by a certain amount of bytes. I did some tests where I truncated the file by 1KB or so, and it seemed to do a good job at fixing it. However, @manekenpix truncated the file by over 50MB, and the `redis-check-aof` utility didn't seem to be able to fix it.

We might have to go with another approach. Instead of using the AOF as the backup, we use the `RDB` file as the actual backup, and load it in the case of the AOF being corrupted. I am not sure how difficult this will be. I will be researching ways to make this possible.

cc @humphd

humphd:
See https://redis.io/topics/persistence.  Maybe we should switch persistence methods?

JerryHue:
Yeah. I am currently searching how to use the "rdb" file during initialization. We need to figure out that part, as well as how to properly generate this backup.

JerryHue:
Since this PR does not solve the issue at all, I am going to close it.

itaispiegel:
Hi everyone,
We're facing the exact same issue in my Redis instance: we keep getting the AOF file corrupted after having power outages.
How come this workaround doesn't fix the problem?
Doesn't running `redis-check-aof --fix` fix the corruption in the file?
I'd be glad if you could elaborate about this 😄

Issue #2827, opened 2022-02-04
Get and expose npm + GitHub data for dependency packages


humphd:
In #2797 we are adding a dependency service, and it currently only includes package names.  Let's add a way to expose both npm and GitHub data for a package.

For npm, we can query the npmjs registry programmatically. For example, you can get the package info for React using https://registry.npmjs.org/react. This will return everything we need in JSON format.

Here are some docs:

- https://github.com/npm/registry/blob/master/docs/REGISTRY-API.md
- https://www.edoardoscibona.com/exploring-the-npm-registry-api
- https://www.npmjs.com/package/query-registry
- https://www.jsdocs.io/package/query-registry

The npm data for a package will also give us GitHub URLs that we can use to query the [GitHub API](https://docs.github.com/en/rest) via [Octokit](https://github.com/octokit/octokit.js).

I don't know if there is any point in doing this for every package, or only on-demand when requested.  Maybe at first, we can expose this data via a REST endpoint in the dependency service: `GET /projects/:name`.

We should cache this data in Redis (available via Satellite), and then we can re-use it for future requests for the same data.

Down the road, once this working, we can decide how to use this data (e.g., find issues to work on).

JerryHue:
I think we can go with a mix of both.

GitHub data is going to be very dynamic. There are some projects that constantly get new issues, as well as close some issues, and so on.

On the other hand, the `npm` data is not that dynamic. Unless there are some extreme circumstances (the package gets completely removed from `npm`), the `npm` package metadata is going to be the same for a long time.

So I would plan it out like this:

During initialization stage, we get the `package` names from `deps.txt`, pull the data from the `npm` on mass (all async to not block the service). Hook some callbacks to pull `npm` data or `GitHub` data (usually on a request for a package).

These callbacks will get in charge of fetching the data and cache it, as well as make sure we are not requesting too many times to `npm` or `GitHub` (gonna have to look deeper into the GitHub API if we can do something like this).

humphd:
I agree, the data is both stable and dynamic, and our fetching and data backend can reflect that.

NOTE: you're describing something very similar to how our RSS feed parser works.  We have a queue in Redis managed via JS with [BullMQ](https://github.com/taskforcesh/bullmq).  It lets us add jobs, and process them over a long period of time using worker processes.

You could easily add another queue to process dependencies/github projects.  We'd enqueue the project, and over time, process (or re-process) it and cache or store the data somehow.

[...]

PR #3126, opened 2022-03-06
Add npm package information to dependencies

JerryHue:
## Issue This PR Addresses

#2827 (this PR does not completely cover the issue, as it only adds the NPM packages)

## Type of Change

## Description

This PR adds the option to get more information of a specific package by pulling such information from the NPM registry.

## Steps to test the PR

1. Pull my changes.
2. Move to `src/api/dependency-discovery`
3. Start the application (`pnpm dev`)
4. Get the whole list of projects through `GET /projects`
5. Get a project's information by using the names given. Use the route `GET /projects/:name` if the project is a regular name, or `GET /projects/:namespace/:name` if the project uses the longer form (e.g. `@senecacdot/satellite`).

## Checklist

- [ ] **Tests**: This PR does not include tests, as they will be included in a future PR.

## Additional notes

While the NPM information is very barebones, we can add more information such as the version, downloads, etc., depending on the needs of the frontend.


src/api/dependency-discovery/src/dependency-list.js
```diff
@@ -1,14 +1,52 @@
...
 
 async function getDependencies() {
-  const dependencies = await readFile(join(cwd(), 'deps.txt'), 'utf8');
+  if (Object.keys(dependenciesGlobal).length === 0) {
+    const dependencyList = await readFile(join(cwd(), 'deps.txt'), 'utf8');
+
+    // The order of the alternatives is important!
+    // The regex engine will favor the first pattern on an alternation
+    // even if the other alternatives are subpatterns
+    for (const dependencyName of dependencyList.split(/\r\n|\n|\r/).filter((line) => line !== '')) {
```
DukeManh, line 15: I think this is enough.
```suggestion
    for (const dependencyName of dependencyList.split(/\r\n?/).filter((line) => !!line)) {
```

src/api/dependency-discovery/src/dependency-list.js
```diff
@@ -1,14 +1,52 @@
...
+const { getPackument } = require('query-registry');
+
+const dependenciesGlobal = {};
 
 async function getDependencies() {
-  const dependencies = await readFile(join(cwd(), 'deps.txt'), 'utf8');
+  if (Object.keys(dependenciesGlobal).length === 0) {
+    const dependencyList = await readFile(join(cwd(), 'deps.txt'), 'utf8');
+
+    // The order of the alternatives is important!
```
DukeManh, line 12: Could you explain what `alternatives`  are you talking about?

src/api/dependency-discovery/src/dependency-list.js
```diff
@@ -1,14 +1,52 @@
...
+  return dependencies[packageName];
+}
 
-  // The order of the alternatives is important!
-  // The regex engine will favor the first pattern on an alternation
-  // even if the other alternatives are subpatterns
-  return dependencies.split(/\r\n|\n|\r/).filter((line) => line !== '');
+async function isPackageDependency(packageName) {
+  const dependencies = await getDependencies();
+  return dependencies.hasOwnProperty(packageName);
```
DukeManh, line 45: Why not
```suggestion
  return !!dependencies[packageName]
```

src/api/dependency-discovery/src/dependency-list.js
```diff
@@ -1,14 +1,52 @@
...
+const dependenciesGlobal = {};
 
 async function getDependencies() {
-  const dependencies = await readFile(join(cwd(), 'deps.txt'), 'utf8');
+  if (Object.keys(dependenciesGlobal).length === 0) {
+    const dependencyList = await readFile(join(cwd(), 'deps.txt'), 'utf8');
+
+    // The order of the alternatives is important!
+    // The regex engine will favor the first pattern on an alternation
+    // even if the other alternatives are subpatterns
```
DukeManh, line 14: Using `forEach` is better here, https://eslint.org/docs/rules/no-restricted-syntax
> iterators/generators require regenerator-runtime, which is too heavyweight for this guide to allow them. Separately, loops should be avoided in favor of array iterations

For some reason, eslint fails to detect this. I'd say eslint is not working at all, I tried adding an unused variable and eslint tests still pass.

DukeManh: cc @cindyledev

DukeManh (changes requested):

src/api/dependency-discovery/src/dependency-list.js
```diff
@@ -1,14 +1,52 @@
 const { readFile } = require('fs/promises');
 const { join } = require('path');
 const { cwd } = require('process');
+const { getPackument } = require('query-registry');
+
+const dependenciesGlobal = {};
```
DukeManh, line 6: I suggest a nit to avoid having a global variable here

```js
const getDependencies = (function () {
  let dependencies = null;
  return async () => {
    if (!dependencies) {
      dependencies = {};
      console.log('reading');
      const dependencyList = await readFile(join(cwd(), 'deps.txt'), 'utf8');

      // The order of the alternatives is important!
      // The regex engine will favor the first pattern on an alternation
      // even if the other alternatives are subpatterns
      dependencyList
        .split(/\r\n|\n|\r/)
        .filter((line) => line !== '')
        .forEach((dependencyName) => {
          dependencies[dependencyName] = null;
        });
    }
    return dependencies;
  };
})();
```
This pattern makes the value of `dependencies` persist, in the meantime hides it from the global.

DukeManh (changes requested):

src/api/dependency-discovery/src/dependency-list.js
JerryHue: I was thinking of using a closure, but I don't know if this is a pattern that everybody is familiar with, and I didn't want to introduce something that would complicate other people's understanding of the code. Otherwise, I agree with this suggestion.

src/api/dependency-discovery/src/dependency-list.js
JerryHue: I meant alternations of the regex: `\r\n` or `\r` or `\n`.

src/api/dependency-discovery/src/dependency-list.js
JerryHue: I believe that macOS uses `\r` for line ending. Is that the case?

src/api/dependency-discovery/src/dependency-list.js
JerryHue: There is a slight difference between `!!dependencies[packageName]` and `hasOwnProperty()`. `hasOwnProperty()` returns true if the string passed is a key of the object, while `!!dependencies[packageName]` returns a truth if `dependencies[packageName]` is truthy. If `dependencies[packageName] === null`, then `!!dependencies[packageName]` would be false, and `hasOwnProperty()` is true.

src/api/dependency-discovery/src/dependency-list.js
DukeManh: `\r\n?` includes '\r` for unix system and `\r\n` for windows.

src/api/dependency-discovery/src/dependency-list.js
DukeManh: Then `\r\n?` should be sufficient right

src/api/dependency-discovery/src/dependency-list.js
DukeManh: I see, you are caching the package info. Still, eslint will complain about this, eslint is not working at the moment.

src/api/dependency-discovery/src/dependency-list.js
DukeManh: It is a common pattern.

src/api/dependency-discovery/src/dependency-list.js
JerryHue: Not really, because for Linux, it is only `\n`, but here we are accepting `\r\n` and `\r`.

JerryHue: I think we should use `\r?\n` instead. From this [stackoverflow answer](https://stackoverflow.com/a/1552782), the line-ending `\r` only occurs on really old Macintosh systems.

src/api/dependency-discovery/src/dependency-list.js
JerryHue: I checked the [available set of lint rules](https://eslint.org/docs/rules/), and none of them really disallow the use of this function. There is only one rule, but it only applies if we use ES2022, which we don't, I believe.

src/api/dependency-discovery/src/dependency-list.js
DukeManh: I think we should include it, for example  in the Typescript repo
https://github.com/microsoft/TypeScript/blob/e64f04bd8c87ece7f800d10d8eb90b7e592907ad/src/harness/util.ts#L35

src/api/dependency-discovery/src/dependency-list.js
DukeManh: It is an interesting rule, for example, if we have a dependency called "hasOwnProperty", when you call `dependencies.hasOwnProperty`, our server will crash because `hasOwnProperty` function is shadowed by the hasOwnProperty key, and it is not a function.

It's this rule https://eslint.org/docs/rules/no-prototype-builtins, valid from ECMAScript 5.1.

src/api/dependency-discovery/src/dependency-list.js
```diff
@@ -1,14 +1,55 @@
...
+    const { id, license, gitRepository } = await getPackument({ name: packageName });
+    dependencies[packageName] = { id, license, gitRepository };
+  }
+
+  return dependencies[packageName];
+}
+
+async function isPackageDependency(packageName) {
+  const dependencies = await getDependencies();
+  return dependencies[packageName] !== undefined;
```
DukeManh, line 48: Your  hasOwnProperty is correct but you want to call it differently.
```suggestion
  return  Object.prototype.hasOwnProperty.call(dependencies, packageName);
```

DukeManh (changes requested):
This is nice, one more nit.

src/api/dependency-discovery/src/router.js
```diff
@@ -1,15 +1,36 @@
...
 
+// For project names that use a full name convention (e.g. @senecacdot/satellite),
+// we will use /projects/:namespace/:name, for projects that have regular name,
+// then we will use /projects/:name.
+// That means that if "name" is undefined, then we hit /projects/:name, instead.
+router.get('/projects/:namespace/:name?', async (req, res, next) => {
+  const { namespace, name } = req.params;
+  const packageName = name ? `${namespace}/${name}` : namespace;
+
+  if (!(await isPackageDependency(packageName))) {
```
humphd, line 23: `isPackageDependency` could throw too, right?  We should wrap that in try/catch as well.

src/api/dependency-discovery/test/dependency.test.js
```diff
@@ -14,9 +14,11 @@ const depsList = [
 ];
 
 jest.mock('../src/dependency-list', () => {
```
humphd, line 16: The way you usually do this with Jest is to add a `__mocks__` directory, and write your overrides in a file there, see https://jestjs.io/docs/manual-mocks

humphd (changes requested):
Looks pretty good.  A few things to fix.

humphd (dismissed):

tpmai22 (dismissed):

src/api/dependency-discovery/src/__mocks__/dependency-list.js
```diff
@@ -0,0 +1,10 @@
+let depsList = [];
+
+module.exports = {
+  __setMockDepList: (newDepsList) => {
+    depsList = newDepsList;
+  },
+  getDependencyList: jest.fn().mockImplementation(() => {
+    return Promise.resolve(depsList);
+  }),
```
TueeNguyen, line 9: This is nice!

DukeManh (dismissed):

src/api/dependency-discovery/src/dependency-list.js
```diff
@@ -1,14 +1,55 @@
...
-  // even if the other alternatives are subpatterns
-  return dependencies.split(/\r\n|\n|\r/).filter((line) => line !== '');
+  return async () => {
+    if (!dependencies) {
+      dependencies = {};
+      const dependencyList = await readFile(join(cwd(), 'deps.txt'), 'utf8');
+
+      dependencyList
+        .split(/\r\n?|\n/g)
+        .filter((line) => !!line)
```
TueeNguyen, line 16: Interesting operator `!!`

src/api/dependency-discovery/src/dependency-list.js
```diff
@@ -1,14 +1,55 @@
...
+        .forEach((name) => {
+          dependencies[name] = null;
```
TueeNguyen, line 18: I think you can add a comment here or on the top of the function explaining the function is returning empty object with keys

src/api/dependency-discovery/src/dependency-list.js
humphd: Another way people do this is `.filter(line => Boolean(line))` or even `.filter(Boolean)`

src/api/dependency-discovery/src/dependency-list.js
JerryHue: Sure, I will add that to clarify.

humphd (approved):

tpmai22 (approved):
Super !

PR #3129, opened 2022-03-06
Opt-out of appendonly mode in redis

### LINKED ISSUE #1365: Avoid Redis database corruption on power failure

JerryHue:
## Issue This PR Addresses

Fixes #1365 

## Type of Change

## Description

The AOF file would get corrupted during a power outage. During recovery, the redis-server would not be able to read this corrupted file, and thus end the process, making the container service unavailable.

Instead, when a power outage occurs, redis shall use the dump.rdb file, which works fine for our current use case.

## Steps to test the PR

1. Pull my changes
2. Wipe out any existing `redis-data` folder.
3. Create a new `redis-data` folder by turning the necessary services for the backend to fill in any posts (`pnpm run services:start` and then `pnpm start`)
4. Shut down all services and backend after filling the database
5. Truncate the `appendonly.aof` file to a considerably shorter length (use `truncate` command in linux and macOS)
6. Run `redis` container only (`pnpm run services:start redis`)
7. Check the logs of the container (`pnpm run services:logs redis`), and you should be able to see redis listening for connections and still alive.

## Checklist

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -305,7 +305,7 @@ services:
   redis:
     image: redis:latest
     container_name: 'redis'
-    command: ['redis-server', '--appendonly', 'yes']
+    command: ['redis-server']
     volumes:
       - ../redis-data:/data
 


humphd (approved):
I think trying rdb is worth an experiment.  The risk is that we lose some data since the last snapshot, but we're moving to using Redis as a cache vs. primary DB so I think it should be OK.  Let's find out!

manekenpix (approved):

Issue #2951, opened 2022-02-16
Build log not showing in `production`


tpmai22:
**What happened**:
Build log not showing on `production` 
![image](https://user-images.githubusercontent.com/43376325/154322896-6b3d8190-72b2-4aa3-8569-142136d88aa3.png)

**What should have happened**:

**How to reproduce it (as precise as possible)**:

**Anything else we need to know?**:
https://telescope.cdot.systems/deploy/status return `null` on all builds 
**Environment**:

- OS:
- Browser:

humphd:
@manekenpix might know if we're running the latest autodployment server in prod, we might not be.

manekenpix:
I think `production` is running the latest version of the `autodeployment` server. The issue is that, at least once a week, there's a power outage in our "server room" and, when the machines come back and the `autodeployment` server is relaunched, it has no build record to serve, and I think that's why the dashboard displays the error message.

humphd:
"Server room" :)

A perfect opportunity to use Supabase and the [Storage API](https://supabase.com/docs/guides/storage)!  We can push build logs and GitHub data about the build into a Supabase storage bucket, and key the name on the git sha.  So maybe for 

```js
// Upload Log
const { data, error } = await supabase.storage
  .from('builds')
  .upload(`public/${sha}/log`, log);

// Upload Info
const { data, error } = await supabase.storage
  .from('builds')
  .upload(`public/${sha}/json`, buildData);
```

Now we can [access these with public URLs](https://supabase.com/docs/guides/storage#accessing-objects).

I'd love to see someone other than @DukeManh or me tackle this one, and we can mentor/review.  How about it, @joelazwar, @rclee91, @TueeNguyen, @sirinoks, @menghif (since you were on the Supabase call today)?

There is another part to this: using these logs in the Dashboard front-end.  We might also want to create a table that keeps track of every build we did: git sha, result (did it build?), links to objects in buckets.

sirinoks:
O_O what even is `sha`

TueeNguyen:
I'm happy to take this.

TueeNguyen:
@sirinoks, it's the id of the commit, it is hashed using SHA (Secure Hash Algorithm). We usually call it sha

DukeManh:
I totally agree, Supabase needs to be maintained by more people. Also, the best way to understand Supabase is to get your hand dirty and do something with it.

@humphd, why not use the database but the storage API?

humphd:
[...]

PR #3137, opened 2022-03-07
Add build log to supabase

TueeNguyen:
## Issue This PR Addresses
- Part of #2951
- I'm adding insert logic first before changing the way `status` gets build log

## Type of Change

### Help needed: 
- I'm not sure how to test it in staging. Local works

### Changes: 
- Add table build_logs to supabase
- Add route /build/:sha to retrieve build log
- Add route /build/add to add build log to supabase
- Add class BuildLog and storage.js to query data
- Add addBuildLog() in check-build.js 
- Expose SUPABASE_URL & SERVICE_ROLE_KEY to `status`

## Steps to test the PR
To test the feature locally, do the following:

- Change `autodeploymentUrl` in `src\api\status\public\js\build-log\check-build-status.js`

```js
const autodeploymentUrl = (path) => `//dev.telescope.cdot.systems/deploy${path}`;  
```

- Allow all `connectSrc` in `src\api\status\src\server.js`

```js
connectSrc: ["'self'", "*"]
```

## Checklist


src/api/status/src/server.js
```diff
@@ -43,7 +43,7 @@ const satelliteOptions = {
           'fonts.googleapis.com',
           'cdn.jsdelivr.net',
         ],
-        connectSrc: ["'self'", '*.fontawesome.com', autodeploymentHost, '*.github.com'],
+        connectSrc: ["'self'", '*.fontawesome.com', autodeploymentHost, '*.github.com', '*'],
```
TueeNguyen, line 46: Will back out, for testing purpose

docker/supabase/volumes/db/init/04-telescope-schema.sql
```diff
@@ -72,3 +80,9 @@ CREATE POLICY feeds_read_policy ON feeds
     FOR SELECT
     USING (true);
 
+ALTER TABLE build_logs ENABLE ROW LEVEL SECURITY;
+
+CREATE POLICY build_logs_policy ON build_logs
+    FOR ALL
+    USING (true);
+
```
TueeNguyen, line 88: How to allow read/write for client with appropriate keys?

src/api/status/public/js/build-log/check-build-status.js
```diff
@@ -3,8 +3,10 @@ import showToast from '../utils/toast.js';
 // Get the proper URL for the autodeployment server based on how this is
 // being loaded. On staging and production, remove the `api.` subdomain.
 // On localhost:1111, use localhost (drop the port).
-const autodeploymentUrl = (path) =>
-  `//${window.location.hostname.replace(/(api\.|\.api)/, '')}/deploy${path}`;
+// const autodeploymentUrl = (path) =>
+//   `//${window.location.hostname.replace(/(api\.|\.api)/, '')}/deploy${path}`;
+
+const autodeploymentUrl = (path) => `//dev.telescope.cdot.systems/deploy${path}`;
```
TueeNguyen, line 9: For testing purpose

src/api/status/public/js/build-log/check-for-build.js
```diff
@@ -4,6 +4,9 @@ import buildHeader from './build-header.js';
 import showToast from '../utils/toast.js';
 
 let build;
+let uploaded = false; //a flag to check if build log is added to supabase
+
+const statusUrl = (path) => `http://localhost:1111${path}`;
```
TueeNguyen, line 9: How to make `status` know which url to use or how to pass `process.env.STATUS_URL` to this public js

src/api/status/public/js/build-log/check-for-build.js
```diff
@@ -49,12 +52,45 @@ export default async function checkForBuild() {
...
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({
+        sha: build.sha,
+        code: build.code,
+        log: log,
+      }),
+    });
+    const addMsg = await addResponse.json();
+    showToast(`${addMsg}`, 'success');
```
TueeNguyen, line 90: Maybe not needed? A console.log is fine

manekenpix:
I took a quick look at the implementation and I think I understand your approach. The front-end fetches the build log from the `autodeployment` server and then sends it to the `status` service to store it in `supabase`? wouldn't it make more sense if, instead of going all the way aroundto through the  front-end, the `autodeployment` server sent the build log directly to the `status` service?

TueeNguyen:
@manekenpix Right, I understand, it could be similar to how `webhook` works. But does `autodeployment` send the build log whenever the client requests `status/build` or does it send at every interval? I don't know much about the `autodeployment` server.

docker/supabase/volumes/db/init/04-telescope-schema.sql
tpmai22: @DukeManh

manekenpix:
> But does `autodeployment` send the build log whenever the client requests `status/build` or does it send at every interval?

@TueeNguyen It sends the build log to any client requesting it. One way to implement this would be that the `status` service could get the build log from the `autodeployment` server right after every deployment, and update `supabase` if necessary (and the front-end would get the log from `status`). There's no need for the `status` service to request a build log in intervals because build logs are generated only when there's a new deployment (and `status` would get the build log at launch). If `status` couldn't get a build log from `autodeployment` because it's down or any other reason, `status` would serve the most recent build log stored in `supabase`.

JerryHue:
I am just pitching this, but what if we use `redis` instead of `supabase` for something like this? Build logs are not something we would like to store for a very long time, so I think `redis` would be a perfect choice for this. After all, we just want to store for 30 days or so. You can keep at most 2 to 4 build logs for production, and 5 to 10 build logs for staging (which can be configured through a configuration variable), and then overwrite the oldest one with the newest one.

src/api/status/public/js/build-log/check-for-build.js
JerryHue: Well, you could use one small trick. Since this is client JS, what if you grab the [URL of the window](https://developer.mozilla.org/en-US/docs/Web/API/Location) and use that? I think that should be fine.

JerryHue: However, if you follow @manekenpix's idea, you might not need to do this.

JerryHue:
> I am just pitching this, but what if we use `redis` instead of `supabase` for something like this? Build logs are not something we would like to store for a very long time, so I think `redis` would be a perfect choice for this. After all, we just want to store for 30 days or so. You can keep at most 2 to 4 build logs for production, and 5 to 10 build logs for staging (which can be configured through a configuration variable), and then overwrite the oldest one with the newest one.

Now that I am rethinking about it, I think we will still need to go with the `supabase` approach, since `redis` is an in-memory database. If we save logs there, they will be taking precious caching space.

TueeNguyen:
> > I am just pitching this, but what if we use `redis` instead of `supabase` for something like this? Build logs are not something we would like to store for a very long time, so I think `redis` would be a perfect choice for this. After all, we just want to store for 30 days or so. You can keep at most 2 to 4 build logs for production, and 5 to 10 build logs for staging (which can be configured through a configuration variable), and then overwrite the oldest one with the newest one.
> 
> Now that I am rethinking about it, I think we will still need to go with the `supabase` approach, since `redis` is an in-memory database. If we save logs there, they will be taking precious caching space.

I wonder if we can create Postgre function to remove the build log after like 10 20 days

humphd:
Using Redis for this is a good idea.  With Redis, we can include a TTL (Time to Live) on the keys, and have them expire after 30 days (or however long we decide).

I also agree that having the various services deal directly with the DB is probably best vs doing everything through REST API calls.  The `autodeployment` server can put build logs directly into Redis, and the `status` service can read from Redis directly to provide its routes.

TueeNguyen:
I can rework this PR in @humphd's suggestion, I think it'll work well, though I don't think it's gonna make it for 2.8, it seems to touch a lot of code

humphd:
Agree, moved to 2.9

manekenpix:
> The autodeployment server can put build logs directly into Redis, and the status service can read from Redis directly to provide its routes.

We should think about this a bit more. The `autodeployment` server is not part of the environment where telescope and all the services are running, so it can't see or communicate with `redis` directly. That's why I suggested that `status` would pull build logs from `autodeployment`.

tpmai22:
Big PR though and great work so far @TueeNguyen I will move it to 2.9

humphd:
> We should think about this a bit more. The `autodeployment` server is not part of the environment where telescope and all the services are running, so it can't see or communicate with `redis` directly. That's why I suggested that `status` would pull build logs from `autodeployment`.

Ah, good point.  I forgot about this

AmasiaNalbandian:

TueeNguyen:
Closing this pull request because we need to refigure out the how to solve this issue, discussion will continue under the issue

PR #3162, opened 2022-03-10
Change stop method to close all db connections

### LINKED ISSUE #3148: Add and expose logic that close all db connections (Redis & Elastic) from Satellite

I want to add a static function, let's say `close()` that closes all db connections (Redis, Elastic) in `Satellite` class. 

The reason is I don't want to create an instance of `Satellite`, I just want to close all db connections leveraging existing code, it would look like this 
```js 
static closeClients() {
  return Promise.all([redis.shutDown(), elastic.shutDown()]);
}
```

https://github.com/Seneca-CDOT/telescope/blob/18a86cec8ef1e0e20e6692d3fc36a1be10333c6b/src/satellite/src/satellite.js#L16

I'm trying to add e2e test to parser service, Jest would hang because of open handles, db connections that aren't close before Jest tries to exit. Refer to #3040

TueeNguyen:
## Issue This PR Addresses
Fixes #3148 

## Type of Change

## Description

- Add shutdown() functions to Satellite.stop()
- Change Elastic mock to reinitialize beforeEach
## Checklist

### DIFF
--- a/src/satellite/src/satellite.js
+++ b/src/satellite/src/satellite.js
@@ -1,5 +1,6 @@
 /* eslint-disable global-require */
 const { createTerminus } = require('@godaddy/terminus');
+const { promisify } = require('util');
 
 const { createApp, createRouter } = require('./app');
 const redis = require('./redis');
@@ -55,9 +56,7 @@ class Satellite {
       onSignal() {
         // Do any async cleanup required to gracefully shutdown. The calls to
         // redis/elastic shutDown() will be no-ops if no connections are open.
-        return Promise.all([this.shutDown(), redis.shutDown(), elastic.shutDown()]).catch((err) =>
-          logger.error({ err }, 'unexpected error while shutting down')
-        );
+        return this.stop();
       },
       logger: (...args) => logger.error(...args),
     });
@@ -66,22 +65,14 @@ class Satellite {
     this.server.listen(port, callback);
   }
 
-  stop(callback) {
-    const self = this;
+  stop() {
+    const { server } = this;
 
-    function finished() {
-      self.server = null;
-      if (typeof callback === 'function') {
-        callback();
-      }
-    }
+    const promisifiedClose = server
+      ? promisify(server.close.bind(server))
+      : () => Promise.resolve();
 
-    if (!this.server) {
-      finished();
-      return;
-    }
-
-    this.server.close(finished);
+    return Promise.all([this.shutDown(), redis.shutDown(), elastic.shutDown(), promisifiedClose()]);
   }
 }
 
--- a/src/satellite/test.js
+++ b/src/satellite/test.js
@@ -73,10 +73,7 @@ describe('Satellite()', () => {
     token = createToken({ sub: 'test-user@email.com' });
   });
 
-  afterEach((done) => {
-    service.stop(done);
-    service = null;
-  });
+  afterEach(() => service.stop());
 
   test('start() should throw if port not defined', () => {
     expect(() => service.start()).toThrow();
@@ -101,7 +98,8 @@ describe('Satellite()', () => {
       service.start(port, async () => {
         const res = await fetch(`${url}/healthcheck`);
         expect(res.ok).toBe(false);
-        service.stop(done);
+        await service.stop();
+        done();
       });
     });
 
@@ -115,7 +113,8 @@ describe('Satellite()', () => {
       service.start(port, async () => {
         const res = await fetch(`${url}/healthcheck`);
         expect(res.ok).toBe(true);
-        service.stop(done);
+        await service.stop();
+        done();
       });
     });
   });
@@ -130,7 +129,8 @@ describe('Satellite()', () => {
         const res = await fetch(`${url}/always-200`);
         expect(res.ok).toBe(true);
         expect(res.headers.get('permissions-policy')).toBe('interest-cohort=()');
-        service.stop(done);
+        await service.stop();
+        done();
       });
     });
 
@@ -142,7 +142,8 @@ describe('Satellite()', () => {
         const res = await fetch(`${url}/always-200`);
         expect(res.ok).toBe(true);
         expect(res.headers.get('permissions-policy')).toBe(null);
-        service.stop(done);
+        await service.stop();
+        done();
       });
     });
   });
@@ -186,7 +187,8 @@ describe('Satellite()', () => {
       expect(res.ok).toBe(true);
       const body = await res.json();
       expect(body).toEqual({ testValue: 'test' });
-      service.stop(done);
+      await service.stop();
+      done();
     });
   });
 
@@ -206,7 +208,8 @@ describe('Satellite()', () => {
       expect(res.ok).toBe(true);
       const body = await res.json();
       expect(body).toEqual({ hello: 'world' });
-      service.stop(done);
+      await service.stop();
+      done();
     });
   });
 
@@ -226,7 +229,8 @@ describe('Satellite()', () => {
       expect(res.ok).toBe(true);
       const body = await res.json();
       expect(body).toEqual({ testValue: 'test' });
-      service.stop(done);
+      await service.stop();
+      done();
     });
   });
 
@@ -252,7 +256,8 @@ describe('Satellite()', () => {
       expect(res.ok).toBe(true);
       const body = await res.json();
       expect(body).toEqual({ testValue: 'parsers-router' });
-      service.stop(done);
+      await service.stop();
+      done();
     });
   });
 
@@ -268,7 +273,7 @@ describe('Satellite()', () => {
       const testRoute = async () => {
         const res = await fetch(`${url}/router/sub-router`);
         expect(res.ok).toBe(true);
-        service.stop(done);
+        done();
       };
 
       service.start(port, () => {
@@ -313,7 +318,8 @@ describe('Satellite()', () => {
       body = await res.json();
       expect(body).toEqual({ hello: 'protected' });
 
-      service.stop(done);
+      await service.stop();
+      done();
     });
   });
 
@@ -381,7 +387,8 @@ describe('Satellite()', () => {
       expect(res.ok).toBe(false);
       expect(res.status).toEqual(401);
 
-      service.stop(done);
+      await service.stop();
+      done();
     });
   });
 
@@ -441,7 +448,8 @@ describe('Satellite()', () => {
       body = await res.json();
       expect(body).toEqual({ hello: 'protected' });
 
-      service.stop(done);
+      await service.stop();
+      done();
     });
   });
 
@@ -483,7 +491,8 @@ describe('Satellite()', () => {
       });
       expect(res.ok).toBe(true);
 
-      service.stop(done);
+      await service.stop();
+      done();
     });
   });
 
@@ -530,7 +539,8 @@ describe('Satellite()', () => {
       body = await res.json();
       expect(body).toEqual({ hello: 'protected' });
 
-      service.stop(done);
+      await service.stop();
+      done();
     });
   });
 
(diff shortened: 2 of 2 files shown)


src/satellite/test.js
```diff
@@ -988,10 +988,14 @@ describe('Elastic()', () => {
...
-    const client = Elastic();
-    const { mock } = client;
+    let client;
+    let mock;
 
-    beforeEach(() => mock.clearAll());
+    beforeEach(() => {
+      client = Elastic();
+      mock = client.mock;
+      mock.clearAll();
```
rclee91, line 997: I'm not sure why we need to make this change for tests here. 
If doing so this way, we won't even need `mock.clearAll()` since a new client and mock is created each time
The original tests were to show what would've happened if all mocks were sharing the same client.

rclee91 (changes requested):

src/satellite/src/satellite.js
```diff
@@ -78,10 +76,12 @@ class Satellite {
 
     if (!this.server) {
       finished();
-      return;
+      return null;
```
humphd, line 79: You're mixing two calling patterns here: callbacks and Promises.  You have to pick one, and I'd suggest using Promises.  This means that every `return` has to be a `Promise`.

You need to modify `stop` like so:

```js
const { promisify } = require('util');
...
stop() {
  const server = this.server;

  // Convert server.close() to return a Promise
  const promisifiedClose = server ? promisify(server.close) : Promise.resolve;

  // Shut down all long-running connections
  return Promise.all([this.shutDown(), redis.shutDown(), elastic.shutDown(), promisifiedClose()]);
}
```

Untested code, but that's the idea.

humphd (changes requested):

src/satellite/src/satellite.js
TueeNguyen: Oh, promisify would help me

tpmai22:
@TueeNguyen will this be ready for 2.8 ?

TueeNguyen:
@tpmai22 Yes

tpmai22:
Super !

TueeNguyen:
I've struggled with this one more than I thought 

I tried to `promisify` the function `server.close` but it returned `reject()` and gave me 
```
TypeError: Cannot read properties of undefined (reading '_handle') 
```
In `test.js`, I changed most of `service.stop(done)` to 
```js
await service.stop(); 
done()
```
I guess, I'll fix other problem with parser test first and come back to this

humphd:
> I've struggled with this one more than I thought
> 
> I tried to `promisify` the function `server.close` but it returned `reject()` and gave me
> 
> ```
> TypeError: Cannot read properties of undefined (reading '_handle') 
> ```
> 
> In `test.js`, I changed most of `service.stop(done)` to
> 
> ```js
> await service.stop(); 
> done()
> ```
> 
> I guess, I'll fix other problem with parser test first and come back to this

You're losing the instance, do this:

```js
const promisifiedClose = server ? promisify(server.close.bind(server)) : Promise.resolve;
```

DukeManh:
let me know if you need more help on this @TueeNguyen.

TueeNguyen:
Thanks @DukeManh, I've just pushed new changes now

src/satellite/test.js
TueeNguyen: Thanks for the explanation

src/satellite/test.js
```diff
@@ -1016,7 +1016,6 @@ describe('Elastic()', () => {
     beforeEach(() => {
       client = Elastic();
       mock = client.mock;
-      mock.clearAll();
```
rclee91: What I was trying to say was, I think it was fine the way it was before. Was there really a need to change it to this way?

src/satellite/src/satellite.js
```diff
@@ -66,22 +65,25 @@ class Satellite {
     this.server.listen(port, callback);
   }
 
-  stop(callback) {
+  stop() {
     const self = this;
```
humphd, line 69: Let's not bother with this, or with `finish` below.  We don't need to set it to `null`.

src/satellite/src/satellite.js
```diff
@@ -66,22 +65,25 @@ class Satellite {
...
+        promisifiedClose()
+          .then(() => {
+            self.server = null;
+            resolve();
+          })
+          .catch((error) => reject(error));
+      });
 
-    this.server.close(finished);
+    return Promise.all([this.shutDown(), redis.shutDown(), elastic.shutDown(), finish()]);
```
humphd, line 86: Change `finish()` to `promisifiedClose()` here.

src/satellite/test.js
```diff
@@ -75,9 +75,8 @@ describe('Satellite()', () => {
     token = createToken({ sub: 'test-user@email.com' });
   });
 
-  afterEach((done) => {
-    service.stop(done);
-    service = null;
+  afterEach(async () => {
```
humphd, line 78: You don't need async/await here, since you're not doing anything with the return value:

```js
afterEach(() => service.stop());
```

src/satellite/test.js
```diff
@@ -103,7 +102,8 @@ describe('Satellite()', () => {
       service.start(port, async () => {
         const res = await fetch(`${url}/healthcheck`);
         expect(res.ok).toBe(false);
-        service.stop(done);
+        await service.stop();
```
humphd, line 105: The other way to do this would be to pass back the Promise this returns to Jest, and it will wait on it:

```js
test('...', async () => {
  ...
  return service.stop();
});
```

src/satellite/test.js
```diff
@@ -988,10 +1010,13 @@ describe('Elastic()', () => {
   });
 
   describe('Tests for mock Elastic()', () => {
-    const client = Elastic();
-    const { mock } = client;
+    let client;
```
humphd, line 1013: Let's revert this to what @rclee91 had before, there's no reason to change.

humphd (changes requested):
I would use Promises vs. `done()` callbacks in these tests, as described in my example.

TueeNguyen:
So I did changes based on your recommendation, it failed 3 tests. Could it be because of some async behavior?
```bash
  ● Satellite() › Default body parsers › should support x-www-form-urlencoded body

    FetchError: request to http://localhost:60234/protected failed, reason: connect ECONNREFUSED 127.0.0.1:60234

      at ClientRequest.<anonymous> (../../node_modules/.pnpm/node-fetch@2.6.7/node_modules/node-fetch/lib/index.js:1491:11)

  ● Elastic() › Tests for mock Elastic() › Should mock an API

    NoLivingConnectionsError: Given the configuration, the ConnectionPool was not able to find a usable Connection for this request.

      at makeRequest (../../node_modules/.pnpm/@elastic+elasticsearch@7.17.0/node_modules/@elastic/elasticsearch/lib/Transport.js:203:43)
      at prepareRequest (../../node_modules/.pnpm/@elastic+elasticsearch@7.17.0/node_modules/@elastic/elasticsearch/lib/Transport.js:463:9)
      at EventEmitter.<anonymous> (../../node_modules/.pnpm/@elastic+elasticsearch@7.17.0/node_modules/@elastic/elasticsearch/lib/Transport.js:485:13)
      at ../../node_modules/.pnpm/@elastic+elasticsearch@7.17.0/node_modules/@elastic/elasticsearch/lib/Transport.js:621:29
      at onBody (../../node_modules/.pnpm/@elastic+elasticsearch@7.17.0/node_modules/@elastic/elasticsearch/lib/Transport.js:376:9)
      at Class.onEnd (../../node_modules/.pnpm/@elastic+elasticsearch@7.17.0/node_modules/@elastic/elasticsearch/lib/Transport.js:291:11)
      at endReadableNT (../../node_modules/.pnpm/readable-stream@2.3.7/node_modules/readable-stream/lib/_stream_readable.js:1010:12)

  ● Elastic() › Tests for mock Elastic() › If an API is not mocked correctly, it should return a 404

    expect(received).toBe(expected) // Object.is equality

    Expected: true
    Received: false

      1034 |         response = await client.cat.health();
      1035 |       } catch (err) {
    > 1036 |         expect(err instanceof errors.ResponseError).toBe(true);
           |                                                     ^
      1037 |         expect(err.body).toStrictEqual({ error: 'Mock not found' });
      1038 |         expect(err.statusCode).toBe(404);
      1039 |       }

      at Object.<anonymous> (test.js:1036:53)
```

src/satellite/test.js
```diff
@@ -75,8 +75,8 @@ describe('Satellite()', () => {
     token = createToken({ sub: 'test-user@email.com' });
   });
 
-  afterEach(async () => {
-    await service.stop();
+  afterEach(() => {
```
humphd, line 78: This is wrong, you need to return the Promise, so Jest waits:

```js
afterEach(() => service.stop())
```

tpmai22:
@DukeManh any idea on this ?

src/satellite/test.js
```diff
@@ -102,12 +102,11 @@ describe('Satellite()', () => {
       service.start(port, async () => {
         const res = await fetch(`${url}/healthcheck`);
         expect(res.ok).toBe(false);
-        await service.stop();
-        done();
+        return service.stop();
```
humphd, line 105: I understand now.  So you should deal with stopping the service in `afterEach` vs here, and go back to your `done()` approach.

In all tests, you are basically doing:

```js
let service;

beforeEach(() => {
  service = createService();
  ...
});

afterEach(() => service.stop())

test('...', async (done) => {
  service.start(..., () => {
    ...
    done();
  });
});
```

humphd (changes requested):

src/satellite/test.js
TueeNguyen: Right, I overlooked those brackets

src/satellite/test.js
TueeNguyen: I don't think function can be `async` and take `done`

humphd: Why do you say that?

TueeNguyen: Actually, it's just test function, Jest just told me this 
```
    Test functions cannot both take a 'done' callback and return something. Either use a 'done' callback, or return a promise.

```

humphd: Move the `async` off of the outer callback function to test:

```js
test('...', (done) => {
  service.start(..., async () => {
    await whatever();
    done;
  });
});
```

src/satellite/test.js
```diff
@@ -270,7 +275,7 @@ describe('Satellite()', () => {
       const testRoute = async () => {
         const res = await fetch(`${url}/router/sub-router`);
         expect(res.ok).toBe(true);
-        service.stop(done);
```
TueeNguyen, line 273: I removed this since `afterEach()` already does `service.stop()`

humphd (approved):

src/satellite/test.js
```diff
@@ -988,10 +1005,14 @@ describe('Elastic()', () => {
   });
 
   describe('Tests for mock Elastic()', () => {
-    const client = Elastic();
-    const { mock } = client;
```
rclee91, line 992: This still, why do we need to change this?

humphd: Ah, missed this, thank you.  @TueeNguyen please revert this bit.

TueeNguyen: @rclee91, `afterEach()` calls `stop()` which has changed to close all Elasticsearch connections, so I want to open the connection again before each mock Elastic test

src/satellite/test.js
rclee91: I see now, unfortunate, we can't showcase mocks using the same client. 
I'll be fixing up some ES mock tests when upgrading ES to 8.0 anyways, I'll try to get tests to the way I want then.

rclee91 (approved):

PR #3190, opened 2022-03-12
feat: Banner image placeholder and fallback background

### LINKED ISSUE #2925: Improve landing page performance
We have an Image service that gives the Telescope landing page a random picture on page load. This is nice but not yet optimal. 
On slow internet connections, these high-quality banner images take a couple of seconds to load and reduce the overall performance.

Moving forward, we should improve the Image service and the front page to: 
 - Return [responsive image ](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images), with various sizes that fit well on different devices
 - Return a low-quality, blurred-out image for initial page load and replace it with a high definition one when it finishes loading

DukeManh:
## Issue This PR Addresses
Fixes #2925

- Add loading banner image background by displaying a tiny 40px wide image (minimal loading time) and transitioning  in the main 2000px wide image (long loading time) when it finishes loading 
- Add fallback gradient banner background
- Fix inconsistent image src, request a specific photo versus a random image chosen by the backend

## Test
- Pull this change locally
- Run server: 
 ```
    cd src/api/image && pnpm dev

    // in another terminal
    cp config/env.staging .env

    // change IMAGE_URL in .env
    # Image Service URL
    IMAGE_URL=http://localhost:4444

   // Run the frontend
   pnpm dev
```
- to see that a gradient background is shown first, followed by a placeholder image, and then the high resolution image.
- If you are not able to see the difference, your internet is probably too fast, good. Go to the Network tab, click `disable cache`, and choose one of the `throttling` options.

## Screenshot

https://user-images.githubusercontent.com/51073515/158033934-5b8e2ca1-6a07-43be-9c0f-ba5c27bb7fa6.mp4

### DIFF
--- a/src/api/image/src/routes/image.js
+++ b/src/api/image/src/routes/image.js
@@ -3,27 +3,28 @@ const { Router, createError } = require('@senecacdot/satellite');
 const { celebrate, Joi, errors, Segments } = require('celebrate');
 
 const { optimize } = require('../lib/image');
-const { getRandomPhotoFilename, getPhotoFilename } = require('../lib/photos');
+const { getRandomPhotoFilename, getPhotoFilename, getPhotoAt } = require('../lib/photos');
 
 const router = Router();
 
 /**
  * Support the following optional query params:
  *
- *  - w: the width to resize the image to. Must be 200-2000. Defaults to 800.
- *  - h: the height to resize the image to. Must be 200-3000. Defaults to height of image at width=800
+ *  - w: the width to resize the image to. Must be 40-2000. Defaults to 800.
+ *  - h: the height to resize the image to. Must be 40-3000. Defaults to height of image at width=800
  *  - t: the image type to render, one of: jpeg, jpg, png, webp. Defaults to jpeg.
  *
- * We also support passing an image name as a param in the URL:
+ * We also support passing an image name or image image index as a param in the URL:
  *
- * - image: should look like look '_ok8uVzL2gI.jpg'. Don't allow filenames like '../../dangerous/path/traversal'.
+ * - index: Can be any positive integer, will return a photo at a index in the circular buffer
+ * - image: should look like '_ok8uVzL2gI.jpg'. Don't allow filenames like '../../dangerous/path/traversal'.
  */
 router.use(
   celebrate({
     [Segments.QUERY]: Joi.object().keys({
       t: Joi.string().valid('jpeg', 'jpg', 'webp', 'png'),
-      w: Joi.number().integer().min(200).max(2000),
-      h: Joi.number().integer().min(200).max(3000),
+      w: Joi.number().integer().min(40).max(2000),
+      h: Joi.number().integer().min(40).max(3000),
     }),
   })
 );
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: src/api/image/src/routes/image.js [hunk 1]
```diff
@@ -22,8 +22,8 @@ router.use(
   celebrate({
```
humphd, line 22: The comment above this line needs to be updated to include the fact that we now allow passing an integer too.

```
* We also support passing an image name as a param in the URL:
 *
 * - image: should look like look '_ok8uVzL2gI.jpg'. Don't allow filenames like '../../dangerous/path/traversal'.
 */
```

src/api/image/src/routes/image.js
```diff
@@ -78,6 +80,13 @@ router.use(
       return;
     }
 
+    // Return a specific photo at an index
+    if (!isNaN(image)) {
+      req.imageFilename = getPhotoAt(image);
```
humphd, line 85: Let's throw away fractional components so we have a pure integer: `getPhotoAt(image|0);`

src/web/src/components/BannerDynamicItems.tsx
```diff
@@ -1,29 +1,44 @@
...
+    backgroundPosition: '50% 0',
+    // fallback banner background
+    background: 'linear-gradient(to right, #021B79, #0575E6)',
   },
 }));
 
-const BannerDynamicText = () => {
+// the placeholder and the main image must refer to the same source
+// we use this to request a consistent image vs a random one
+const index = Math.floor(Math.random() * 100 + 1);
```
humphd, line 22: Should we use a larger upper-bound here so that we can access images beyond 100 when we get there?  Your circular buffer should be fine with something like 1000.

humphd (changes requested):
So cool, I love it.  A few things I notice.

DukeManh:
@humphd, I'm running this again and the images fail to load. Do you know what this error is?

image

humphd:
I'm not 100% sure.  Is that Firefox?  Do you have a Service Worker on localhost:4444 that's interfering somehow (try changing port)?

src/api/image/src/routes/image.js
DukeManh: Thanks, I forgot about this.

DukeManh:
> I'm not 100% sure. Is that Firefox? Do you have a Service Worker on localhost:4444 that's interfering somehow (try changing port)?

It's on any browser I try.

Looks like something was changed on master recently that make the image service have CORS issue. I rebased and have the issue.

I filed #3212

DukeManh:
This is currently blocked by #3212, which makes it hard to test.

src/web/src/components/DynamicImage.tsx
```diff
@@ -32,24 +48,59 @@ const useStyles = makeStyles(() => ({
...
 
-const DynamicImage = () => {
+type DynamicImageProps = {
+  imageURL?: string;
+  placeholderURL?: string;
+  visible?: boolean;
+};
+
+// Define a series of sizes, and let the browser figure out which one to use
+function createSrcset(imageSrc: string) {
```
humphd, line 62: This is a nice touch.

humphd (approved):

rclee91:
I'm wondering if we might need to come to a consensus if we're ignoring Eslint tests until #3228 has been merged. 
Do we approve if all other tests pass

humphd:
@rclee91 I think that's fine, and we can fix things that we catch when the checks are working properly.

rclee91 (approved):

DukeManh:
It's interesting the ESLint recommends `Number.isNaN` instead of `isNaN`, but they behave a bit differently. https://stackoverflow.com/questions/33164725/confusion-between-isnan-and-number-isnan-in-javascript

humphd (dismissed):

humphd (approved):

Issue #3189, opened 2022-03-12
Migrate staging and production to docker.cdot.systems based images


humphd:
Last night we turned on our new Docker Registry at https://docker.cdot.systems.  The [README](https://github.com/Seneca-CDOT/docker.cdot.systems#readme) has info on how it works if you're interested.

We also migrated our first service to use it: our beloved `feed-discovery`.  To do this, we had to make these changes:

1. Create a [workflow to build and deploy](https://github.com/Seneca-CDOT/telescope/blob/master/.github/workflows/deploy-feed-discovery-image.yaml) the service.  We may need to create 1 of these for each of our images, or maybe we can find a way to do a matrix in this file and selectively build/release only the ones that need it.
1. [Remove the `build` section](https://github.com/Seneca-CDOT/telescope/blob/master/docker/docker-compose.yml#L175-L197) from the service definition in `docker/docker-compose.yml`.  Now our base `docker-compose.yml` file defines the service, but not how it gets built/downloaded.
1. [Add the `build` section](https://github.com/Seneca-CDOT/telescope/blob/master/docker/development.yml#L96-L99) to `docker/development.yml`.  This will extend `docker/docker-compose.yml` in development and build from out local `Dockerfile`.
1. [Add an `image` section](https://github.com/Seneca-CDOT/telescope/blob/master/docker/production.yml#L118-L120) to `docker/production.yml`.  This will extend `docker/docker-compose.yml` in production to download the pre-built image from our registry.

We need to do this same dance for all of our Dockerfiles.

Next, I'd like to start using the following Docker tags:

1. `production` - this is the tag we'll `pull` on our production box, for example: `docker.cdot.systems/feed-discovery:production`.  
1. `staging` - this is the tag we'll `pull` on our staging box, for example: `docker.cdot.systems/feed-discovery:staging`.  
1. `latest` - this is the tag we'll `pull` locally or on gitpod if we want to get the most recent build, for example: `docker.cdot.systems/feed-discovery:latest` or just `docker.cdot.systems/feed-discovery`

We'll have to update our `docker/*.yml` files to use environment variable names for the image tags (e.g., `docker.cdot.systems/feed-discovery:${VERSION}` where `VERSION` is `latest` (default), `staging`, or `production` 

Doing this will require a whole bunch of fixes and separate PRs/people.  We should be able to do it in small stages.

humphd:
@rclee91 ran into an issue in our AWS docs where we don't seem to do the right thing:

[...]

PR #3207, opened 2022-03-14
Rework GitHub Actions Workflows for docker.cdot.systems registry

humphd:
Part of #3189.

This reworks our GitHub Actions Workflows, and docker-compose setup, in order to leverage our new Docker Registry at docker.cdot.systems.  Specifically, it does the following:

- Adds `continuous-delivery.yml`, which runs on every `push` to `master`, and builds and pushes our Docker images for staging, publishes any updates to our npm modules (e.g., satellite), and sends a custom deployment webhook to trigger a build on the autodeployment server.
- Renames `node-js-ci.yml` to `continuous-integration.yml`, and only runs it for `pull requests` (i.e., code needs to pass CI before we merge it--we could re-run all these checks, but I don't think we need to if we force CI to pass before merging).
- Adds `deploy-webhook.yml`, which sends a custom webhook to our autodeployment server with the `deploy_tag` to use when building (e.g., `staging` or `production`).
- Adds `docker-build-and-push.yml`, which builds all of our Dockerfiles into images and pushes them to our registry.  It also uses our registry to cache the results, so subsequent builds should be faster.  NOTE: this workflow is used for both staging and production, and the appropriate tag is set in the environment of the calling workflow.
- Adds `pnpm-publish.yml`, which publishes any/all changed npm modules in our monorepo.  You update the `version` in `package.json` and commit that change, and this will do the rest.
- Updates our `docker-compose` to include a `cache_from` key, and specifies the same cached image that CI is using.  This means that any `docker-compose` run that needs to do a build will try to reuse cached layers, which should always be fresh due to our CI using the same cache.  This will work on CI, locally, on GitPod, or AWS.

I've also included a lock file update, which happened when I pulled upstream master.


.github/workflows/deploy-webhook.yml
```diff
@@ -0,0 +1,16 @@
...
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Invoke autodeployment webhook for staging
+        uses: distributhor/workflow-webhook@v2
+        env:
+          webhook_type: 'json-extended'
+          webhook_url: https://dev.telescope.cdot.systems/deploy
+          webhook_secret: ${{ secrets.WEBHOOK_SECRET }}
```
humphd, line 15: @manekenpix I'm going to need to put our `WEBHOOK_SECRET` on staging and production into the GitHub Actions Secrets.  Can you help me when you have a min to figure out what it is?

docker/docker-compose.yml
```diff
@@ -15,6 +15,8 @@ services:
     build:
       context: ../src/api/status
       dockerfile: Dockerfile
+      cache_from:
+        - docker.cdot.systems/status:buildcache
```
cindyledev, line 19: Awesome to see us using build cache now!

Question: since this service doesn't exist in the Docker Registry yet, docker knows to skip this cache_from step when building, right?

.github/workflows/docker-build-and-push.yml
```diff
@@ -0,0 +1,59 @@
...
+    runs-on: ubuntu-latest
+    env:
+      DOCKER_REGISTRY: docker.cdot.systems
+      DOCKER_USERNAME: telescope-ci
+    strategy:
+      matrix:
+        include:
+          # TODO
+          # - context: src/api/Dockerfile
+          #  image: telescope-base
```
Kevan-Y, line 28: This one gonna be un-commented once we change all our services to use the base Image right?

.github/workflows/deploy-webhook.yml
humphd: Secrets now added, thank you @manekenpix!

docker/docker-compose.yml
humphd: I *think* what will happen is that it will just treat this as the cache not existing and build normally.  We'll have to test.

.github/workflows/docker-build-and-push.yml
humphd: I'm not 100% sure how to deal with the build order for this one, so I left it for now to do in a follow-up.

humphd:
I've reworked a few things in this, based on my reading of https://github.community/t/reusable-workflows-secrets-and-environments/203695/14.  Essentially, you have to explicitly pass variables and secrets to shared workflows, and also define these in the shared workflow.  You can't share environment variables or secrets across calling/called workflows.

I think what I have now will work, but it's untested!

humphd:
See also the docs in https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_call, which show examples of using `inputs` and `secrets`.

Kevan-Y (changes requested):
I don't think `pnpm-lock.yaml` file should be here. Don't see any change in our `package.json`

humphd:
> I don't think `pnpm-lock.yaml` file should be here. Don't see any change in our `package.json`

Rebased and pushed to get rid of it.

docker/docker-compose.yml
Kevan-Y: First time seeing `cache_from` in `docker-compose`, so if I understand correctly it will check for layers that can use cache from our registry, and if there is no cache built from our context, right?

Do we also need to do the same thing for feed_discovery?

humphd: Yes, you are correct: it will download vs. re-create layers that already exist in the registry.

NOTE: I already do the `feed-discovery` service here https://github.com/Seneca-CDOT/telescope/pull/3207/files#diff-b020a27fee741b5e10a4d106e6d266ec2bc9a4e694496470529407966142ad35R101

Kevan-Y (approved):

cindyledev (approved):

humphd:
> Looks good, just a few questions.
> 
> * `inputs` are for reading a value from the command line?

No, `inputs`, `outputs`, and `secrets` all describe things you pass into or return from a shared workflow.  A shared workflow is a separate file we run, basically a GitHub Action hosted in your own repository.  Imagine them like a function you are calling, and `inputs` and `secrets` define the arguments you pass to them, and `outputs` the various artifacts that get produced.

See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_callinputs and https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_callsecrets specifically.

> * `secrets` are read from a file stored locally that's different from the `env` file?

No, the `secrets` come from encrypted values in the repo.  Go take a look at https://github.com/Seneca-CDOT/telescope/settings/secrets/actions, we have a bunch.  You can reference these in a workflow via `${{ secrets.SECRET_NAME }}`.  However, when you want to use these in a workflow that you call, you have to explicitly pass the secret to it, otherwise people could steal secrets by trying to use them in linked actions.
 
> And I'm assuming this is not testable until we get the secrets and environment variables set up?

It's not testable, no.  I assume this will break, and then I'll push more fixes.

PR #3211, opened 2022-03-15
Issue 1912: Searchbar Redesign

### LINKED ISSUE #1912: SearchBar re-design
Since I found UI problem with the current searchbar, I suggest a new design for it.

Any suggestion?
Thanks.

AmasiaNalbandian:
## Issue This PR Addresses
Fixes #1912

## Type of Change

## Description
I changed the UI for the searchbar, this will allow us to move on to create more advanced search UI by introducing the advanced search button. For now I have separated the author and post search into their own components. In the next PR I will add the search by title, and dates (one by one!)

## search page light mode wide
chrome_qf2VwTRJVb

chrome_KcrLq2D5f8

## search page light mode narrow
chrome_dOo1rI94du

## search page dark mode wide
chrome_w4iogenluy

## search page dark mode narrow
chrome_ctWQixa3kl

## Steps to test the PR
1. `pnpm i`
2. `cp config/env.staging .env`
3. `pnpm dev`
4. Go to the search
- You can now search posts, and then refine to search WITH authors too. It's not longer one or the other, we can do one or the other OR both. 

## Checklist

### DIFF
--- a/src/web/app/src/components/SearchResults.tsx
+++ b/src/web/app/src/components/SearchResults.tsx
@@ -18,7 +18,7 @@ const useStyles = makeStyles((theme: Theme) =>
     },
     searchResults: {
       padding: 0,
-      width: '100%',
+      width: '81vw',
       justifyContent: 'center',
     },
     errorBackground: {
@@ -57,21 +57,25 @@ const SearchResults = () => {
   const router = useRouter();
   const classes = useStyles();
   const [totalPosts, setTotalPosts] = useState(0);
-  const textParam = Array.isArray(router.query.text)
-    ? router.query.text[0]
-    : router.query.text || '';
-  const filterParam = router.query.filter === 'post' || !router.query.filter ? 'post' : 'author';
+  const postParam = Array.isArray(router.query.post)
+    ? router.query.post[0]
+    : router.query.post || '';
+
+  const authorParam = Array.isArray(router.query.author)
+    ? router.query.author[0]
+    : router.query.author || '';
 
   const prepareUrl = (index: number) =>
-    `${searchServiceUrl}?${filterParam === 'author' ? `author` : `post`}=${encodeURIComponent(
-      textParam
+    `${searchServiceUrl}/?author=${encodeURIComponent(authorParam)}&post=${encodeURIComponent(
+      postParam
     )}&page=${index}`;
 
   // We only bother doing the request if we have something to search for.
-  const shouldFetch = () => textParam.length > 0;
+  const shouldFetch = () => postParam.length > 0 || authorParam.length > 0;
   const { data, size, setSize, error } = useSWRInfinite(
     (index: number) => (shouldFetch() ? prepareUrl(index) : null),
     async (u: string) => {
+      console.log('fetching...');
       const res = await fetch(u);
       const results = await res.json();
 
@@ -81,7 +85,7 @@ const SearchResults = () => {
   );
   const loading = !data && !error;
   // Search result is empty when the the array of posts on the first page is empty
-  const isEmpty = data?.[0]?.length === 0;
+  const isEmpty = !data?.[0]?.length;
   // There no more posts when the last page has no posts
   const isReachingEnd = !data?.[data.length - 1]?.length;
   // Another page is being loaded when size is incremented but data[size - 1] is still undefined
@@ -89,7 +93,7 @@ const SearchResults = () => {
     !isReachingEnd && data && size > 0 && typeof data[size - 1] === 'undefined';
 
   // If there is no posts or if the search bar is empty, then show the search help, otherwise hide it
-  if (!error && (isEmpty || textParam.length === 0)) {
+  if (!error && isEmpty && postParam.length === 0 && authorParam.length === 0) {
     return <SearchHelp />;
   }
   if (error) {
@@ -109,7 +113,7 @@ const SearchResults = () => {
     );
   }
 
-  if (textParam.length && loading) {
+  if ((postParam.length || authorParam.length) && loading) {
     return (
       <Container className={classes.searchResults}>
         <h1 className={classes.spinner}>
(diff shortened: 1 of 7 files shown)


AmasiaNalbandian:
I have removed most of the margin between the input boxes, so it looks like this now: 
image

src/web/src/components/AdvancedSearch.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  const { openDialog } = props;
+
+  return (
+    <>
+      {openDialog && (
+        <div className={classes.root}>
+          <FormControl fullWidth>
+            {/* Author */}
+            <Grid container spacing={2}>
```
TueeNguyen, line 43: The suggestion should make the `label` and the `input` on the the same line
image

```suggestion
            <Grid container justifyContent='center' spacing={2}>
```

Kevan-Y (changes requested):
In dark mode, we don't see the advance search
image

Search error is covering the navbar
image

src/web/src/components/SearchResults.tsx
```diff
@@ -85,7 +90,7 @@ const SearchResults = () => {
     !isReachingEnd && data && size > 0 && typeof data[size - 1] === 'undefined';
 
   // If there is no posts or if the search bar is empty, then show the search help, otherwise hide it
-  if (!error && (isEmpty || textParam.length === 0)) {
+  if (!error && (isEmpty || (post.length === 0 && author.length === 0))) {
```
HyperTHD, line 93: This is causing the search page to immediately search upon a letter being entered which is what we don't want

HyperTHD (changes requested):
Aside from the css issues that've been mentioned, just some big stuff that can cause some significant issues logic wise here

src/web/src/components/SearchResults.tsx
TueeNguyen: I thought it was a feature :)

AmasiaNalbandian:
> In dark mode, we don't see the advance search image
> 
> Search error is covering the navbar image

Thanks for catching these Kevan. 

image

src/web/src/components/AdvancedSearch.tsx
AmasiaNalbandian: I just used MUI input instead 😋

humphd (changes requested):
Screen Shot 2022-03-24 at 11 20 00 AM

A few things I'd like to see change with this, either here or in follow-ups:

- When we have a single input filed (normal search), having the search button embedded in the field itself is a nice touch.  However, when we add more fields, I don't think it make sense, since an advanced search could involve other fields, so it should probably go below the fields, and be embedded in none.
- When we're in a normal search, the top field isn't labelled.  However, in an Advanced search, what is its purpose?  Is it for the body, title, both?  Who knows!  It needs a label, or we need to split it up into separate fields (my preference) so that you can be more specific in saying what you want.
- The alignment of the input fields is not the same, it should be
- The size of the input fields is not the same, it should be
- When you are in normal search and you want to go to an Advanced Search, you click the "Advanced Search" link; that makes sense.  But what does "Advanced Search" mean and do when you're already in Advanced Search?  It should probably get toggled to "Normal Search" or something, so you can go back.

AmasiaNalbandian:

Kevan-Y:
@AmasiaNalbandian is it ready for another round of review?

AmasiaNalbandian:
> @AmasiaNalbandian is it ready for another round of review?

No, I have it working with the new code we put in yesterday but now have to fix UI. Tying it up differently, i have to move some of the components around to address the issues.

tpmai22:
looks good now

src/web/app/src/components/SearchResults.tsx
```diff
@@ -18,7 +18,7 @@ const useStyles = makeStyles((theme: Theme) =>
     },
     searchResults: {
       padding: 0,
-      width: '100%',
+      width: '81vw',
```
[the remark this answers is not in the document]
AmasiaNalbandian: @Kevan-Y I think i'll revert it, but the error banner needs to be smaller then (reason for this change). I should approach it that way. Nice catch again Kevan. Really good testing!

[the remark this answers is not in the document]
Kevan-Y: Instead of having a banner, we can change it to be a similar design when you get a post not found.

Not sure if we should do that in a follow-up issue, since this PR is more specific for advance search. And this issue was there before too.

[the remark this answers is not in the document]
AmasiaNalbandian: Right, however It was easily solved - I think the fix I added works, if we need more enhancement, we can create a new issue I'm ok with that.

TueeNguyen (approved):
Just tested it works. I have one suggestion, we should add `onKeyPressed` for `Enter` key to search.

humphd (approved):
I really like this.

A few things I notice:

- clicking on Advanced then Regular loses whatever is in the Author input.  Not a huge deal, but the post isn't affected, so it's odd.
- I would assume that filling out both post AND author is doing an AND, but it doesn't look like it:

Screen Shot 2022-04-04 at 12 13 51 PM

AmasiaNalbandian:
> I really like this.
> 
> A few things I notice:
> 
> * clicking on Advanced then Regular loses whatever is in the Author input.  Not a huge deal, but the post isn't affected, so it's odd.
> * I would assume that filling out both post AND author is doing an AND, but it doesn't look like it:
> 
> Screen Shot 2022-04-04 at 12 13 51 PM

@humphd 
- yes, I cleared the author field so that we don't end up searching with the last author. I can clear both fields if this feels more natural. 

- The search works on my end, I think it's the fact I need to add an "on key enter" to search. Is that what happened? I clicked the search button and it generated the results. I might open a new issue for this quick fix to not hold back Autocomplete.

TueeNguyen:
Yes @AmasiaNalbandian, no reason to block your work here, it looks good. The `Enter` key feature can be in another PR

Kevan-Y (approved):

AmasiaNalbandian:
OK Thank you everyone, here is the issue for the key-press: 
https://github.com/Seneca-CDOT/telescope/issues/3394


@humphd I believe it will do the AND search and those results come in first, then it will display anything relating to any of those fields. We essentially get more coverage with this method.

Issue #3031, opened 2022-02-23
e2e test timeout


DukeManh:
**What happened**:
One 2e2 test failed on master because the timeout limit exceeds

<img width="1314" alt="image" src="https://user-images.githubusercontent.com/51073515/155426631-62b58593-b201-4d69-b330-dfa48f90cd16.png">

- OS:
- Browser:

cindyledev:
This actually happens quite frequently, I had to re-run the tests for master to be happy :(

humphd:
I suspect that we're maxing out these CI boxes with our many containers.  We should figure out which ones we can not run during tests.

humphd:
We could probably remove status, image, and planet with no problem.  We also don't need the Supabase console.

humphd:
#3017 would help with this too: we could run e2e tests for different parts of the tree in different builders in parallel, and only start the portion of containers they need.

We have a monorepo, but we don't have to run our tests as a whole app.

humphd:
I spent some time looking into this today.  I think there's a race condition with the postgres starting and running migrations, which can leave other services unable to connect to the db.  In my testing, the other supabase containers would crash and restart a lot.

It's hard to pin down, but for example, I can sometimes see the `supabase-auth` container crash:

```
12T20:12:21Z" level=fatal msg="running db migrations: Migrator: problem creating schema migrations: couldn't start a new transaction: could not create new transaction: failed to connect to `host=db user=postgres database=postgres`: dial error (dial tcp 172.22.0.2:5432: connect: connection refused)"
```

I'll also see errors like this in kong:

```
2022/03/12 20:13:52 [error] 27#0: *1309 connect() failed (111: Connection refused) while connecting to upstream, client: 172.23.0.2, server: kong, request: "DELETE /rest/v1/telescope_profiles?id=eq.2b3b2b9ce8 HTTP/1.1", upstream: "http://172.23.0.10:3000/telescope_profiles?id=eq.2b3b2b9ce8", host: "localhost"

2022/03/12 20:13:52 [error] 27#0: *1309 connect() failed (111: Connection refused) while connecting to upstream, client: 172.23.0.2, server: kong, request: "DELETE /rest/v1/telescope_profiles?id=eq.2b3b2b9ce8 HTTP/1.1", upstream: "http://172.23.0.10:3000/telescope_profiles?id=eq.2b3b2b9ce8", host: "localhost"

[...]

PR #3218, opened 2022-03-15
Remove and override restart:unless-stopped for supabase containers

### LINKED ISSUE #2876: Define dev vs. staging/production settings for Supabase containers
We're starting to add the Supabase containers to our Docker network.  Currently, we layer the [docker/supabase/docker-compose.yml](https://github.com/Seneca-CDOT/telescope/blob/master/docker/supabase/docker-compose.yml) file with our existing files via this line https://github.com/Seneca-CDOT/telescope/blob/master/config/env.development#L18 in development (we do something similar in staging and production, for example https://github.com/Seneca-CDOT/telescope/blob/master/config/env.production#L18).

The Supabase containers are all set to `restart: unless-stopped`.  This means that whenever I start Docker Desktop, these containers all restart, whether I'm doing Telescope work or not.  I don't want them to always run, so in development, we should not add `restart: unless-stopped` to each of these.  We should only add this when we do production/staging.

So we need to remove all the `restart: unless-stopped` lines from `docker/supabase/docker-compose.yml`, and create a new file `docker/supabase/production.yml` that lets us override aspects of `docker/supabase/docker-compose.yml`, similar to what we do in https://github.com/Seneca-CDOT/telescope/blob/master/docker/production.yml#L98-L100.

We can also use this separation of dev vs. production to open various ports on the Supabase containers (development) but close them in production.  I'll file another issue on that.

TueeNguyen:
## Issue This PR Addresses
Fixes #2876 

## Type of Change

## Description

Created `docker/supabase/production.yml` to override `restart: unless-stopped` from `docker-compose.yml`
## Checklist

### DIFF
--- a/docker/supabase/docker-compose.yml
+++ b/docker/supabase/docker-compose.yml
@@ -10,7 +10,6 @@ services:
   studio:
     container_name: supabase-studio
     image: supabase/studio:latest
-    restart: unless-stopped
     ports:
       - ${STUDIO_PORT}:3000/tcp
     environment:
@@ -22,7 +21,6 @@ services:
   kong:
     container_name: supabase-kong
     image: kong:2.1
-    restart: unless-stopped
     environment:
       KONG_DATABASE: 'off'
       KONG_DECLARATIVE_CONFIG: /var/lib/kong/kong.yml
@@ -52,7 +50,6 @@ services:
     image: supabase/gotrue:v2.5.22
     depends_on:
       - db
-    restart: unless-stopped
     environment:
       GOTRUE_API_HOST: 0.0.0.0
       GOTRUE_API_PORT: 9999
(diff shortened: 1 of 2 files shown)


TueeNguyen:
In production, we need to hide `studio` behind `portainer` right? Should I do it here or in another issue?

humphd:
I'm nervous about this one, because based on https://github.com/Seneca-CDOT/telescope/issues/3031#issuecomment-1065956273, I wonder if Supabase is relying on this restart behaviour to deal with the load-order crashes that happen while postgres runs migrations on startup.  If we remove this, we may regret it.  We need to think about this in the context of startup/readiness checks.

I'm not sure yet how to deal with all of it.

TueeNguyen:
Closing because supabase Postgre database needs `restart:unless-stopped` to shutdown and start again due to how supabase deals with race condition. This is important for our `sso`'s e2e test

PR #3220, opened 2022-03-15
[satellite]: use Satellite from npm registry

### LINKED ISSUE #3191: Running pnpm with Satellite in the repo breaks tests because of dependency issue
**What happened**:
I am trying to update ElasticSearch to 8.1.0. for Satellite. 
The idea is to get Satellite updated first, do a release, and then have the Telescope Services try out the new Satellite version. 
I updated ES and ES-mock for Satellite in #3187, and the tests for Satellite pass.
However, Search tests are now not passing.  

Updating inside of Satellite shouldn't affect other Services.

Try out #3187

@manekenpix and I spent a lot of time trying to figure this out. 
There is a hack to get this working locally that I'd like to explain first. 

The hack:
1. While working on the branch in #3187, `pnpm clear` to get rid of all the `node_modules` in the root and in all services. If `pnpm clear` doesn't work then manually delete all the `node_modules`
2. Inside of [pnpm-workspace.yaml](https://github.com/Seneca-CDOT/telescope/blob/master/pnpm-workspace.yaml) remove the Satellite related lines, or change the file to below
```
# pnpm workspace definition: https://pnpm.io/pnpm-workspace_yaml
packages:
  # docusaurus web app
  - 'src/docs'
  # web front-end
  - 'src/web'
  # microservices
  - 'src/api/*'
  # autodeployment
  - 'tools/*'
```
3. In root run `pnpm i`
4. head to the Satellite directory, then run `npm i` please note there's no beginning `p`
5. In the Satellite `jest.config` take out `moduleDirectories: ['<rootDir>/src/satellite/node_modules', 'node_modules'],`
6. Voila, both Satellite and Search tests will now work. In root run `pnpm test satellite`, `pnpm test search`, `pnpm test`, and everything should pass. 
7. This is only a local hack, because its not possible to do this in CI

The problem without using the hack is that `pnpm` tries to be very helpful when installing dependencies. After the merge of Satellite to Telescope, Search (as an example) in `pnpm-lock.yaml` looks like this
```
  src/api/search:
    specifiers:
      '@senecacdot/satellite': ^1.26.0
      express-validator: 6.14.0
      nodemon: 2.0.15
    dependencies:
      '@senecacdot/satellite': link:../../satellite
      express-validator: 6.14.0
    devDependencies:
      nodemon: 2.0.15
```
It seems that when running `pnpm install`, `pnpm` links the Services that are using Satellite to the local Satellite, instead of the Satellite package in `node_modules`

So what's happening in the Search tests specifically now is that the test file is trying to use the ElasticSearch-mock that its using right now. But, the `client` in the folders outside of the test file is using the locally updated Satellite. This causes the current test fail because the old ES-mock and the new ElasticSearch aren't compatible. But since I was only updating Satellite, and haven't touched anything else in Telescope, this shouldn't be happening. 

This will not only affect Search, and should affect our tests anytime there will be major updates in Satellite in the future. 
If we want to keep the idea of keeping Satellite as its own app, releasing updates, and then using the new versions in Telescope services, then we'll have to find a way to deal with this issue.

DukeManh:
fix #3191

By default, available packages listed in pnpm-workspaces are linked to node_modules instead of being downloaded from the registry to save space.

This adds an option to make pnpm to use Satellite on npm vs the package.

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -10,3 +10,5 @@ packages:
   - 'src/satellite'
   # autodeployment
   - 'tools/*'
+
+link-workspace-packages: false
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -10,5 +10,3 @@ packages:
   - 'src/satellite'
   # autodeployment
   - 'tools/*'
-
-link-workspace-packages: false
--- a/.npmrc
+++ b/.npmrc
@@ -8,3 +8,5 @@ tag-version-prefix=""
 # Don't use ^ when adding packages, prefix with nothing, see:
 # https://pnpm.io/npmrc#save-prefix
 save-prefix=""
+
+link-workspace-packages=false
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -125,7 +125,7 @@ importers:
       '@babel/preset-env': 7.16.11_@babel+core@7.17.5
       '@babel/preset-react': 7.16.7_@babel+core@7.17.5
       '@babel/preset-typescript': 7.16.7_@babel+core@7.17.5
-      '@senecacdot/eslint-config-telescope': 1.0.0
+      '@senecacdot/eslint-config-telescope': link:tools/eslint
       '@types/jest': 27.4.1
       '@typescript-eslint/eslint-plugin': 4.33.0_2951ba233cd46bb4e0f2f0a3f7fe108e
       '@typescript-eslint/parser': 4.33.0_eslint@7.32.0
@@ -159,7 +159,7 @@ importers:
       '@senecacdot/satellite': 1.27.0
       query-registry: 2.2.0
     devDependencies:
-      '@senecacdot/eslint-config-telescope': 1.0.0
+      '@senecacdot/eslint-config-telescope': link:../../../tools/eslint
       env-cmd: 10.1.0
       eslint: 7.32.0
       nodemon: 2.0.15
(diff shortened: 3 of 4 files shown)


rclee91:
cc @aserputov , @JiaHua-Zou , if you two could take a look and help out with the eslint errors please

pnpm-workspace.yaml
```diff
@@ -10,3 +10,5 @@ packages:
   - 'src/satellite'
   # autodeployment
   - 'tools/*'
+
+link-workspace-packages: false
```
cindyledev, line 14: This change effects both `satellite` and `eslint-config-telescope`. We made many ESLint changes recently for in eslint-config-telescope and the eslint-config-telescope in the npm registry hasn't picked up the new changes yet since we haven't done a release

DukeManh: Actually, I think it makes sense to link eslint because we want eslint config to be reflected right away in subpackages. I link eslint in the new commit

cindyledev:
I pulled this PR and got both satellite and search tests working!

```
PS C:\Users\lecin\Documents\repo\telescope> pnpm test satellite search
> @senecacdot/telescope@2.8.1 test C:\Users\lecin\Documents\repo\telescope
> pnpm jest "satellite" "search"


> @senecacdot/telescope@2.8.1 jest C:\Users\lecin\Documents\repo\telescope
> jest -c jest.config.js -- "satellite" "search"

...

Test Suites: 2 passed, 2 total
Tests:       66 passed, 66 total
Snapshots:   0 total
Time:        5.88 s
```

cindyledev:
So how did you link the eslint stuff to local vs npm registry?

DukeManh:
> So how did you link the eslint stuff to local vs npm registry?

Say you want to use local satellite in src/api/status, you can cd into src/api/status and point to the local eslint package
```
cd src/api/status
pnpm link ../../../tools/eslint
```

DukeManh:
And use `unlink` to undo it

rclee91 (approved):

cindyledev (approved):
I feel like I'm getting gray hairs waiting for e2e tests to finish

humphd:
From Slack, I filed an issue in the microsoft playwright actions repo, and they said that GitHub Actions has been having an outage that's affecting things.  I think this is why things are stalling/failing in weird ways.

PR #3228, opened 2022-03-16
Exclude Satellite from pnpm workspace

DukeManh:
Exclude Satellite from pnpm workspaces.

I think we misunderstood the purpose of `pnpm-workspace.yaml`. The file exists to tell pnpm to link packages from the workspace if the available packages match the declared ranges.

So if we do not want to use the local Satellite, for example, we should not list it in `pnpm-worksapce`.

Better fix of #3191

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -8,5 +8,3 @@ tag-version-prefix=""
 # Don't use ^ when adding packages, prefix with nothing, see:
 # https://pnpm.io/npmrc#save-prefix
 save-prefix=""
-
-link-workspace-packages=false
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -195,7 +195,7 @@ importers:
     devDependencies:
       '@senecacdot/eslint-config-telescope': link:../../../tools/eslint
       del-cli: 4.0.1
-      eslint: registry.npmjs.org/eslint/7.32.0
+      eslint: 7.32.0
       nodemon: 2.0.15
 
   src/api/parser:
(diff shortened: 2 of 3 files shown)


DukeManh:
I updated the description.

rclee91:
I thought we didn't want the services to use local Satellite though

DukeManh:
@rclee91, my bad, I updated the description again, s/do/do not/

rclee91:
In this case, do we still use `pnpm install` to get the right dependencies inside of Satellite?

cindyledev:
I guess we're giving Satellite the same treatment as React Native so they're both living in the same repository as the rest of the services but they functioning like introverted roommates that don't explicitly communicate with the other services unless we want them to by doing this?

<path/to/service>/package.json
```
"@senecadot/satellite: workspace:^"
```

DukeManh:
> In this case, do we still use pnpm install to get the right dependencies inside of Satellite?

yes

rclee91 (approved):

DukeManh:
Wait, I'm wrong.

rclee91:
> > In this case, do we still use pnpm install to get the right dependencies inside of Satellite?
> 
> yes

Sorry, have you tried deleting all the `node_modules` and doing `pnpm install` to see what happens to the `node_modules`? Including the ones in `src/satellite`
Taking out Satellite in the `pnpm workspace` was a part of the "hack" in #3191 , I seem to remember it won't create `node_modules` in `src/satellite`
I could try this again, but my potato laptop takes a long time to test. Just wondering if you had tried.

cindyledev:
We probably have to do the same thing we do with React Native with `postinstall`,

    "postinstall": "cd src/mobile && npm install"

DukeManh:
> We probably have to do the same thing we do with React Native with postinstall,

Yes, probably, either we have to figure out how to link Satellite or do that.
We could use the `workspace:protocol` but it makes Vercel deployment fails here https://github.com/Seneca-CDOT/telescope/runs/5576676664?check_suite_focus=true

DukeManh:
This is probably a bug in pnpm https://github.com/pnpm/pnpm/issues/3561

rclee91:
Is this something that is still needed, even after the merge of #3247 ?

DukeManh:
@rclee91 we don't need this anymore. I'm gonna close it.

Issue #3241, opened 2022-03-17
Reduce memory usage of Docker containers in development


humphd:
<img width="783" alt="Screen Shot 2022-03-17 at 8 56 49 AM" src="https://user-images.githubusercontent.com/427398/158813186-544fc08e-0382-4e25-bfd6-7771decbadd0.png">

```sh
$ docker stats --no-stream
CONTAINER ID   NAME                   CPU %     MEM USAGE / LIMIT     MEM %     NET I/O           BLOCK I/O         PIDS
c6d663dfa922   supabase-storage       2.56%     177MiB / 9.718GiB     1.78%     8.65kB / 13.4kB   1.54MB / 0B       13
d1891a42fdb2   supabase-meta          0.00%     41.05MiB / 9.718GiB   0.41%     2.04kB / 0B       48.9MB / 4.1kB    18
ba15bb8b5c9f   supabase-auth          0.00%     6.441MiB / 9.718GiB   0.06%     12.4kB / 15.9kB   246kB / 0B        9
7a4516e3ae5b   supabase-realtime      1.41%     99.82MiB / 9.718GiB   1.00%     1.08MB / 3.17MB   16.7MB / 3.86MB   37
e750ebfd2fd6   supabase-rest          0.66%     14.61MiB / 9.718GiB   0.15%     365kB / 84.8kB    46.7MB / 0B       28
dc66a31ad328   posts                  0.00%     48.1MiB / 9.718GiB    0.48%     11.1kB / 966B     17.3MB / 0B       12
30210b6bbd54   status                 0.00%     31.34MiB / 9.718GiB   0.31%     6.88kB / 820B     10.6MB / 0B       12
be20e4b21988   sso                    0.00%     34.76MiB / 9.718GiB   0.35%     2.04kB / 0B       10.9MB / 0B       8
e74bda0b386f   rss-bridge             0.02%     14.53MiB / 9.718GiB   0.15%     2.04kB / 0B       32.4MB / 12.3kB   7
6b43677ef3ad   image                  0.00%     30.96MiB / 9.718GiB   0.31%     27.1MB / 1MB      143kB / 25.7MB    12
05a6642e99d7   feed-discovery         0.00%     35.86MiB / 9.718GiB   0.36%     2.04kB / 0B       14.2MB / 0B       8
d80ef1298a9b   search                 0.00%     29.02MiB / 9.718GiB   0.29%     2.04kB / 0B       9.02MB / 0B       8
[...]

PR #3242, opened 2022-03-17
Reduce memory usage for elasticsearch in development

humphd:
This reduces elasticsearch's memory usage in development by 50%:

```sh
$ docker stats --no-stream
CONTAINER ID   NAME            CPU %     MEM USAGE / LIMIT     MEM %     NET I/O        BLOCK I/O    PIDS
96f7cd128f72   elasticsearch   0.89%     567.6MiB / 9.718GiB   5.70%     46MB / 501kB   0B / 182MB   81
```

I don't think we can go lower than this, based on docs I've been reading today.  I've also removed the requirements to not use swap in development, since we don't care about speed as much as memory footprint.

Part of #3241.

### DIFF
--- a/docker/development.yml
+++ b/docker/development.yml
@@ -62,10 +62,11 @@ services:
   elasticsearch:
     ports:
       - '9200:9200'
-    ulimits:
-      memlock:
-        soft: -1
-        hard: -1
+    environment:
+      # Use the minimum possible RAM in development
+      - 'ES_JAVA_OPTS=-Xms256m -Xmx256m'
+      # We only run a single node
+      - discovery.type=single-node
 
   sso:
     environment:
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -329,13 +329,6 @@ services:
   elasticsearch:
     image: docker.elastic.co/elasticsearch/elasticsearch:7.17.1
     container_name: 'elasticsearch'
-    environment:
-      - bootstrap.memory_lock=true
-      - 'ES_JAVA_OPTS=-Xms512m -Xmx512m'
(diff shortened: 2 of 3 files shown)


manekenpix (approved):
```sh
$ docker stats --no-stream
CONTAINER ID   NAME            CPU %     MEM USAGE / LIMIT     MEM %     NET I/O           BLOCK I/O   PIDS
4df133e1fc07   elasticsearch   1.85%     550.8MiB / 15.55GiB   3.46%     45.9MB / 1.66MB   0B / 91MB   75
```

rclee91 (approved):
Running on Windows and WSL, This is my before and after
Before
```terminal
docker stats --no-stream
CONTAINER ID   NAME            CPU %     MEM USAGE / LIMIT     MEM %     NET I/O       BLOCK I/O   PIDS
3c4b88b4391e   elasticsearch   386.95%   804.5MiB / 3.776GiB   20.81%    1.18kB / 0B   0B / 0B     36
```
After:
```terminal
docker stats --no-stream
CONTAINER ID   NAME            CPU %     MEM USAGE / LIMIT     MEM %     NET I/O       BLOCK I/O   PIDS
d9a1431af478   elasticsearch   296.19%   387.8MiB / 3.776GiB   10.03%    1.11kB / 0B   0B / 0B     33
```
It's the memory usage decrease that we're caring about right?

humphd:
Yeah, that's a nice decrease.  It is indeed the memory usage we care about.  Java is a memory hog.

Elasticsearch is really designed to run alone on its own machine, so it is really good at using all available memory.  We don't want that in development.

PR #3247, opened 2022-03-17
Don't link local package

DukeManh:
We've made the decision to not link local packages, both Satellite and Eslint.


humphd (approved):

humphd:
Comment for others reading this PR: we are going to move to using npm registry packages of things like satellite and our eslint config vs. linking in the workspace.  In order to get a new version of these published, update their `version` in the associated `package.json` and `push` to `master`.  It will automatically publish a new version to npm.

TueeNguyen (approved):

cindyledev:
I guess that means we should prioritize getting ESLint fixes in so we can push a release. I'm getting sick of these ESLint errors :(

DukeManh:
@cindyledev, yeah it's not ideal, I'm re-configuring eslint, I think the current config is a bit off

PR #3248, opened 2022-03-17
Wait on postgres to become ready before running sso e2e tests

### LINKED ISSUE #3031: e2e test timeout

humphd:
We discovered that the `postgres` db container that Supabase runs needs a long time to become ready, even after Docker says that the container is up.  If we run our tests during this time, they will fail.  This explains why they sometimes fail and sometimes are OK.

This change will have our e2e tests wait on the postgres db to become actually ready (able to handle a connection).

We will need to leave the postgres port open for local dev, but close it in production for Supabase.

Fixes #3031

### DIFF
--- a/src/api/sso/jest.config.e2e.js
+++ b/src/api/sso/jest.config.e2e.js
@@ -1,14 +1,33 @@
+const { waitForPostgres } = require('@jcoreio/wait-for-postgres');
+
 const baseConfig = require('../../../jest.config.base');
 
-module.exports = {
-  ...baseConfig,
-  rootDir: '../../..',
-  testMatch: ['<rootDir>/src/api/sso/test/e2e/**/*.test.js'],
-  collectCoverageFrom: ['<rootDir>/src/api/sso/src/**/*.js'],
-  // We use jest-playwright to do browser tests, https://github.com/playwright-community/jest-playwright
-  // See the jest-playwright config in the Telescope root ./jest-playwright.config.js
-  preset: 'jest-playwright-preset',
-  // Try reducing the number of tests that can run at once down to 1
-  maxConcurrency: 1,
-  maxWorkers: 1,
+module.exports = async () => {
+  try {
+    console.log('Waiting for Supabase postgres container to become ready...');
+    await waitForPostgres({
+      host: 'localhost',
+      user: 'postgres',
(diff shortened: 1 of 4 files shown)


src/api/sso/jest.config.e2e.js
```diff
@@ -1,14 +1,33 @@
...
-  maxWorkers: 1,
+module.exports = async () => {
+  try {
+    console.log('Waiting for Supabase postgres container to become ready...');
+    await waitForPostgres({
+      host: 'localhost',
+      user: 'postgres',
+      password: process.env.POSTGRES_PASSWORD,
+      database: 'postgres',
+      timeout: 60 * 1000,
```
TueeNguyen, line 13: It should work in 99% of the time!

TueeNguyen (approved):

src/api/sso/jest.config.e2e.js
```diff
@@ -1,14 +1,33 @@
...
-  rootDir: '../../..',
-  testMatch: ['<rootDir>/src/api/sso/test/e2e/**/*.test.js'],
-  collectCoverageFrom: ['<rootDir>/src/api/sso/src/**/*.js'],
-  // We use jest-playwright to do browser tests, https://github.com/playwright-community/jest-playwright
-  // See the jest-playwright config in the Telescope root ./jest-playwright.config.js
-  preset: 'jest-playwright-preset',
-  // Try reducing the number of tests that can run at once down to 1
-  maxConcurrency: 1,
-  maxWorkers: 1,
+module.exports = async () => {
```
TueeNguyen, line 5: Documentation for this config file if anyone wonders https://jestjs.io/docs/configuration

DukeManh (approved):
This should fix it. Thanks @humphd for nailing the issue and coming up with a fix so quickly.

joelazwar (dismissed):

cindyledev (approved):

src/api/sso/jest.config.e2e.js
rclee91: Oh wow! I wasn't aware we could do something like this in jest configs. 
I'll need to catch up on the recording.

rclee91 (approved):

Issue #3230, opened 2022-03-16
e2e tests not using Docker registry cache with docker-compose


humphd:
I just [ran some e2e tests on CI](https://github.com/Seneca-CDOT/telescope/pull/3229/checks) and noticed that `pnpm services:start` isn't using our Docker registry cache.  When we run this, internally we do this: https://github.com/Seneca-CDOT/telescope/blob/master/tools/services.js#L16-L30

That is basically calling `docker-compose ... --build --remove-orphans`.  It looks like `docker` builds vs. `docker-compose` builds use slightly different digests, so they don't seem to be interchangable.

See:

- https://github.com/moby/moby/issues/40126
- https://github.com/docker/build-push-action/issues/493

It would be great if we could use our registry cache for these builds to speed things up.  I'm not 100% clear on the right way to accomplish this.

menghif:
Fixed in #3254

PR #3254, opened 2022-03-18
Use BuildKit with all Docker build commands for cache

humphd:
I want to see if I can force `docker-compose` to use BuildKit and the caching behaviour we want for our image builds.  This might not work, I need to test in CI.

### DIFF
--- a/.github/workflows/e2e-tests-ci.yml
+++ b/.github/workflows/e2e-tests-ci.yml
@@ -10,22 +10,33 @@ jobs:
     strategy:
       matrix:
         node-version: [14.x, 16.x]
+    env:
+      # Whichever Docker containers are needed to pass e2e tests, define them here
+      DOCKER_CONTAINERS: 'sso feed-discovery auth login kong rest meta'
+
     steps:
       - uses: actions/checkout@v2
-      - uses: pnpm/action-setup@v2.0.1
-        with:
-          version: 6.23.5
+
       - uses: actions/setup-python@v2
         with:
-          python-version: '3.x'
+          python-version: 3.x
+
+      - uses: pnpm/action-setup@v2.1.0
+        with:
+          version: 6
+
       - uses: actions/setup-node@v2.5.1
         with:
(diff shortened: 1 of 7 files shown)


humphd:
I've got this working finally.  Here's what it does:

- I've rewritten `.github/workflows/e2e-tests-ci.yml`.  The two most significant changes are that I've limited the Docker containers we start to only those needed to pass the tests (see the `DOCKER_CONTAINERS` environment variable).  Second, I've switched to use `docker compose` (v2) vs. `docker-compose` (v1), since the latter isn't compatible with Docker cache, which we want to use.  The newer `docker compose` uses `docker` vs. python scripts to process the `docker-compose.yml` files.
- I've added a `pnpm:services pull` command, so that we can `pull` images from our registry if we don't want to build.
- I've updated the docs

You can see what this looks like when it runs in this workflow run:

https://github.com/Seneca-CDOT/telescope/runs/5604170384?check_suite_focus=true

This will cut the total running time by more than 50% for the e2e tests, from ~13-18m minutes down to ~7m

NOTE: the action I've re-written will not get run until we merge this (i.e., it will run the current one on `master`).

humphd:
Fixes #3230

JiaHua-Zou (approved):

menghif (approved):
I love the new total duration in CI!

.github/workflows/e2e-tests-ci.yml
```diff
@@ -10,22 +10,33 @@ jobs:
     strategy:
       matrix:
         node-version: [14.x, 16.x]
+    env:
+      # Whichever Docker containers are needed to pass e2e tests, define them here
+      DOCKER_CONTAINERS: 'sso feed-discovery auth login kong rest meta'
```
rclee91, line 15: I am just wondering if we want to add and test Parser e2e tests, if this means we'll have to add more images into this line here?

humphd: Correct.

Issue #3205, opened 2022-03-14
Trigger custom "deploy" webhook in CI to our autodeployment server


humphd:
The way we're going to do our deployment is changing due to our Docker Registry (see #3189).  One thing that this will mean is that the order or a request to have the autodeployment server start a build, and the build/push of our Docker Images, both have to be timed correctly (i.e., we need to build and push all the new Docker Images *before* we start the deploy, which will `pull` them).

Let's modify our autodeployment server to accept a new, custom webhook, and use https://github.com/marketplace/actions/workflow-webhook-action to send such a webhook from our CI/CD workflows when it's time to deploy.

humphd:
I've started this work in #3207.

humphd:
The relevant bit of what I did is https://github.com/Seneca-CDOT/telescope/blob/master/.github/workflows/deploy-webhook.yml#L29.  When we finish doing a docker build and push to our registry, GitHub Actions will `POST` some JSON to our autodeployment server.  It should be a `push` event something like this:

```json
{
    "event": "push",
    "repository": "owner/project",
    "commit": "a636b6f0861bbee98039bf3df66ee13d8fbc9c74",
    "ref": "refs/heads/master",
    "head": "",
    "workflow": "Autodeployment Webhook",
    "data": {
        "deploy_tag": "staging",
    },
    "requestID": "74b1912d19cfe780f1fada4b525777fd",
    ....
}
```

The `data.deploy_tag` tells us which tag to pull and run, one of `staging` or `production`.

PR #3265, opened 2022-03-19
Migrating to use docker.cdot.systems based images

TDDR:
## Issue This PR Addresses

 USAGE: #3189 This is just the first part of the issue.
image

    
    
    
 Still need to do the following in a separate PR :
    
image

## Type of Change

## Description

This is the the first step for issue #3189. Will make separate issues for the remainder?

## Steps to test the PR
This is not testable yet due to needing environment variables passed.

### DIFF
--- a/docker/development.yml
+++ b/docker/development.yml
@@ -12,6 +12,23 @@ version: '3'
 services:
   # API Gateway
   nginx:
+    build:
+      context: ../src/web
+      cache_from:
+        - docker.cdot.systems/nginx:buildcache
+      # next.js needs build-time access to a number of API URL values, forward as ARGs
+      args:
+        # Web front-end URL
+        - WEB_URL=${WEB_URL}
+        # Telescope 1.0 API URL
+        - API_URL=${API_URL}
+        # Telescope 2.0 Microservice URLs
+        - IMAGE_URL=${IMAGE_URL}
+        - SSO_URL=${SSO_URL}
+        - POSTS_URL=${POSTS_URL}
+        - SEARCH_URL=${SEARCH_URL}
+        - FEED_DISCOVERY_URL=${FEED_DISCOVERY_URL}
+        - STATUS_URL=${STATUS_URL}
     volumes:
       - ../config/nginx.conf.development.template:/etc/nginx/nginx.conf
     ports:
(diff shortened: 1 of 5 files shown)


docker/production.yml
```diff
@@ -80,20 +80,24 @@ services:
 
   # status service
   status:
+    image: docker.cdot.systems/status
```
humphd, line 83: All of our deployable images for staging and production use a tag of `:staging` or `:production`, see https://github.com/Seneca-CDOT/telescope/blob/master/.github/workflows/docker-build-and-push.yml#L121.  We need to pick the right tag for our deployment type with each of the `image` definitions in this file.

My suggestion is to do the following:

1. Add a new environment variable near the top of https://github.com/Seneca-CDOT/telescope/blob/master/config/env.staging#L1-L39 named `DOCKER_DEPLOY_TAG=staging`
2. Do something similar and add a new environment variable near the top of https://github.com/Seneca-CDOT/telescope/blob/master/config/env.production#L1-L39 named `DOCKER_DEPLOY_TAG=production` 
3.Add tag to each of the `image` lines you modified in this file that looks like this: image: docker.cdot.systems/image:${DOCKER_DEPLOY_TAG:-latest}.  This says to use the `DOCKER_DEPLOY_TAG` if present, or `latest` if missing.

docker/development.yml
```diff
@@ -12,6 +12,24 @@ version: '3'
 services:
   # API Gateway
   nginx:
+    build:
+      context: ../src/web
+      dockerfile: Dockerfile
```
humphd, line 17: Because we never change the name of the `Dockerfile`, I think this is unnecessary in each of the definitions in this file, see https://docs.docker.com/compose/compose-file/compose-file-v3/#context.

docker/development.yml
```diff
@@ -67,35 +85,78 @@ services:
...
+  planet:
+    build:
+      context: ../src/api/planet
+      dockerfile: Dockerfile
+      cache_from:
+        - docker.cdot.systems/planet:buildcache
+    ports:
+      - '9876:9876'
+    environment:
+      - POSTS_URL=http://posts:5555
```
humphd, line 155: `POSTS_URL=http://posts:${POSTS_PORT:-5555}`

docker/development.yml
```diff
@@ -67,35 +85,78 @@ services:
...
+      context: ../src/api/sso
+      dockerfile: Dockerfile
+      cache_from:
+        - docker.cdot.systems/sso:buildcache
+    environment:
+      # In development and testing, the SSO service needs to contact the Supabase
+      # service directly via Docker vs through the http://localhost/v1/supabase domain.
+      - SUPABASE_URL=http://kong:8000
+    depends_on:
+      - test-web-content
```
humphd, line 123: We also depend on `traefik` and `login` in development.

humphd (changes requested):

Kevan-Y (approved):

Kevan-Y:
Besides the fix @humphd , it looks good.

humphd (approved):
I think this all looks right.  We'll need to test it to know for sure.

Having said that, we will need to land a fix for https://github.com/Seneca-CDOT/telescope/issues/3205 at the same time, so that the autodeployment server waits on the containers to be built and pushed.  I have the webhook and CD stuff on GitHub Actions already done; we just need to teach the autodeployment server to listen for it. @TDDR are you interested in doing that too?

docker/development.yml
```diff
@@ -12,6 +12,23 @@ version: '3'
...
+        - WEB_URL=${WEB_URL}
+        # Telescope 1.0 API URL
+        - API_URL=${API_URL}
+        # Telescope 2.0 Microservice URLs
+        - IMAGE_URL=${IMAGE_URL}
+        - SSO_URL=${SSO_URL}
+        - POSTS_URL=${POSTS_URL}
+        - SEARCH_URL=${SEARCH_URL}
+        - FEED_DISCOVERY_URL=${FEED_DISCOVERY_URL}
+        - STATUS_URL=${STATUS_URL}
```
manekenpix, line 31: This section is common for both development and production, I think it should be in `docker-compose.yml`, otherwise `production.yml` will not get this args.

manekenpix (changes requested):

docker/development.yml
humphd: Except we don't build in production anymore, right?  They should already be embedded in the image.

manekenpix: Oh! I forgot we have [this](https://github.com/Seneca-CDOT/telescope/blob/aecd7cb0e43c00c918504a8d42c05b52218d6c04/.github/workflows/docker-build-and-push.yml#L76) now. I really need to get used this new setting.

manekenpix (approved):

tcvan0707 (approved):

PR #3280, opened 2022-03-22
Remove custom healthcheck of status service

### LINKED ISSUE #3263: New changes in Dockerfile break Dashboard on Staging
**What happened**:
Dashboard is broken on staging.
(Clicking on the "i" on the hero banner)
https://dev.api.telescope.cdot.systems/v1/status/

image

Should show the dashboard

Go here: https://dev.api.telescope.cdot.systems/v1/status/

The docker logs show a healthcheck fail from the `Status` image. (Sometimes it happens to `Posts` too)

```sh
{"level":30,"time":1647656844771,"pid":1,"hostname":"62b3e97e79c2","msg":"Server started on port 1111"}
{"level":50,"time":1647656863276,"pid":1,"hostname":"62b3e97e79c2","msg":"healthcheck failed"
```

I went through the commit history to traceback the commit that may have caused this. 
It seems like it was commit [e1b73fd19100a9d91a458ac538b2f349a6744340](https://github.com/Seneca-CDOT/telescope/commit/e1b73fd19100a9d91a458ac538b2f349a6744340) that had this problem
The commit before this is stable, and the dashboard works. 

Scanning the changes in the commit, it seems like this line here is terminating the process as an error when healthcheck fails.
```docker
HEALTHCHECK --interval=30s --timeout=30s --start-period=5s --retries=3 \
	CMD wget --no-verbose --tries=1 --spider localhost:${PORT}/healthcheck || exit 1
```	
I tested locally changing `exit 1` to `exit 0` and the dashboard showed.
This is only a temporary solution, but I think we should investigate more on `healthcheck`.

Kevan-Y:
## Issue This PR Addresses
Fixes #3263 

## Type of Change

## Description

~~Removed custom healthcheck in status service. `getPackage()` is not needed here. Because we don't have proper try/catch getPackage + with recent change in Dockerfile with Healthcheck. It made the whole services crash.~~

~~I looked at the code, it seems without this (code removed in custom healthcheck) getting the version doesn't have any issue in staging and production.~~

~~For the development side to display the version, we have to run `pnpm start` which start our backend that a API route that return the version and commit for our backend.
https://github.com/Seneca-CDOT/telescope/blob/master/src/backend/web/routes/health.js~~

## Final solution
Removing the custom healthcheck in status services, and in our front-end we do the fetch directly there.

### Development
- `pnpm services:start`
- Go to `localhost:8000`

image

## Checklist


pnpm-lock.yaml
```diff
@@ -343,11 +342,10 @@ importers:
       '@octokit/plugin-retry': 3.0.9
       '@octokit/plugin-throttling': 3.5.2_@octokit+core@3.5.1
       '@popperjs/core': 2.11.4
-      '@senecacdot/satellite': 1.27.0
+      '@senecacdot/satellite': link:../../satellite
```
humphd, line 345: cc @DukeManh, this shouldn't be happening, right?  Has that `.npmrc` change not landed yet?

manekenpix (commented):
I'm not sure if this is a good idea. If I remember correctly, this change was made because the `backend` service is going to disappear once all the microservices are implemented, and we won't be able to get the information we need to display version and commit, that's why `status` was selected as the provider of this information. There's a conversation about injecting this information at build time in the CI, but that would make some parts of the hero banner almost _undebuggable_, so I don't know what the right solution is here.

pnpm-lock.yaml
```diff
@@ -357,7 +355,7 @@ importers:
       xterm-addon-web-links: 0.5.1_xterm@4.18.0
       xterm-addon-webgl: 0.11.4_xterm@4.18.0
     devDependencies:
-      '@senecacdot/eslint-config-telescope': 1.1.0_eslint@7.32.0
+      '@senecacdot/eslint-config-telescope': link:../../../tools/eslint
```
HyperTHD, line 358: Might be the same issue that David mentioned above? I'm not familiar with the latest stuff added since Satellite was recently incorporated

HyperTHD (commented):
Seconding @manekenpix's suggestion that we keep the info for now. I didn't properly incorporate the healthcheck when I wrote it awhile back so my fault for this

Kevan-Y:
How about removing the custom healthcheck in status services, and in our front-end we do the fetch directly there.
https://github.com/Seneca-CDOT/telescope/blob/d8e1f7db2301617dc198c15dbb5072744d907d66/src/web/src/components/Banner.tsx#L145

We can fetch this https://raw.githubusercontent.com/Seneca-CDOT/telescope/master/package.json, which get the data of our package.json.

pnpm-lock.yaml
DukeManh: Looks like Kevan just picked up the change

DukeManh (changes requested):

TueeNguyen (approved):
Learned more nginx stuff from this.

AmasiaNalbandian (approved):

JerryHue (approved):
So, if I understand everything from this PR, in the case of deploying for production, we would need to generate a config that has the specific SHA for that release.

Very good work!

manekenpix (approved):
LGTM :+1: 

Peek 2022-03-22 23-26

PR #3285, opened 2022-03-22
Updating autodeployment webhook to use custom "deploy" tag

### LINKED ISSUE #3189: Migrate staging and production to docker.cdot.systems based images

### LINKED ISSUE #3205: Trigger custom "deploy" webhook in CI to our autodeployment server

TDDR:
## Issue This PR Addresses

USAGE: Fixes #3205 && Fixes #3189 

## Type of Change

## Description

This PR is an attempt to use our new `DOCKER_DEPLOY_TAG` that I made in a [previous PR](https://github.com/Seneca-CDOT/telescope/pull/3265). We want to use it to trigger the proper deploy build from our autodeployment server. 

## Steps to test the PR

This PR will need to be tested in staging

### DIFF
--- a/tools/autodeployment/webhook.js
+++ b/tools/autodeployment/webhook.js
@@ -15,93 +15,16 @@ webhooks.onError((error) => logger.error(error));
 router.post('/', createNodeMiddleware(webhooks, { path: '/' }));
 
 /**
- * GitHub's webhooks send 3 POST requests with different actions whenever a
- * release event takes place:
- *
- *  - created
- *  - published
- *  - released
- *
- * To prevent the autodeployment server from processing the 3 requests
- * (and eventually crashing), we need to ignore everything except 'published'.
- *
- * More information about release events here:
- * https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release
- * The filter combines checking for staging or production build types and also
- * for the right type of action for a release event in production builds
- * @param {string} repo - Repository name (should be 'telescope')
- * @param {string} buildType - Build type: staging or production
- * @param {string} action - Action for a specific release event: created, published or released.
- * @param {string} ref - .Git reference resource.
- * @param {string} mainBranch - Brach used in the repo as main branch.
+ * @otcokit/webhooks .on
+ * https://github.com/octokit/webhooks.js#webhookson
  */
-function requestFilter(repo, buildType, action, ref, mainBranch) {
-  const currentBranch = ref?.split('/').pop();
+webhooks.on('push', (event) => {
+  const githubData = event.payload;
+  const deployTag = githubData.data.deploy_tag;
 
-  return (
-    // Check repo
-    repo === REPO_NAME &&
-    // Check if the reference is the branch used as main or if it is a tag
-    (currentBranch === mainBranch || ref.includes('tags/')) &&
-    // Check the build type and the action
-    (buildType === 'staging' || (buildType === 'production' && action === 'published'))
-  );
-}
-
-/**
- * Extract the commit for this GitHub payload, depending on the event type
- * @param {'push' | 'release'} eventType one of 'push' or 'release'
- * @param {Object} githubData webhook payload from GitHub, see https://github.com/Seneca-CDOT/telescope/settings/hooks
- */
-function getSha(eventType, githubData) {
-  if (eventType === 'push') {
-    // "after": "adf299da6e6e9c3f20208ee6ab61d205a9f70eab",
-    return githubData.after;
+  if (deployTag) {
+    addBuild(deployTag, githubData, githubData.commit);
   }
-  if (eventType === 'release') {
-    // "target_commitish": "1a4594cfc443d307679576d9fdfa402f530a570c",
-    return githubData.release.target_commitish;
-  }
-  logger.warn(`Unknown GitHub event type ${eventType}. Unable to parse git SHA`);
-  return 'unknown';
-}
-
-/**
- * Create a handler for the particular GitHub push event and build type
- * @param {string} buildType - one of `production` or `staging`
- * @param {string} gitHubEvent - the GitHub Push Event name
- */
-function handleEventType(buildType, gitHubEvent) {
-  if (DEPLOY_TYPE !== buildType) {
-    return;
-  }
-
-  /**
-   * @otcokit/webhooks .on
-   * https://github.com/octokit/webhooks.js#webhookson
-   */
-  webhooks.on(gitHubEvent, (event) => {
-    const githubData = event.payload;
-
-    if (
-      requestFilter(
-        githubData.repository.name,
-        buildType,
-        githubData.action,
-        githubData.ref,
-        githubData.repository.master_branch
-      )
-    ) {
-      const sha = getSha(gitHubEvent, githubData);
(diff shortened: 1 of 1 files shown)


tools/autodeployment/webhook.js
```diff
@@ -53,12 +53,12 @@ function requestFilter(repo, buildType, action, ref, mainBranch) {
  * @param {'push' | 'release'} eventType one of 'push' or 'release'
  * @param {Object} githubData webhook payload from GitHub, see https://github.com/Seneca-CDOT/telescope/settings/hooks
  */
-function getSha(eventType, githubData) {
-  if (eventType === 'push') {
+function getSha(deployTag, githubData) {
+  if (deployTag === 'staging') {
```
humphd, line 57: We'll have to see when we test if the payload matches this same style.  I suspect this function will need to change.  NOTE: we should be able to use `.commit` based on https://github.com/distributhor/workflow-webhook, but let's see when we test on staging.

tools/autodeployment/webhook.js
```diff
@@ -67,41 +67,24 @@ function getSha(eventType, githubData) {
...
-
-// Production builds happen when GitHub sends us a `release` event
-handleEventType('production', 'release');
+webhooks.on('push', (event) => {
+  const githubData = event.payload;
 
-// Staging builds happen when GitHub sends us a `push` event
-handleEventType('staging', 'push');
+  if (
+    requestFilter(
```
humphd, line 77: I'm unclear if we will still want to do this or not.

tools/autodeployment/webhook.js
```diff
@@ -67,41 +67,24 @@ function getSha(eventType, githubData) {
...
-      const sha = getSha(gitHubEvent, githubData);
-      addBuild(buildType, githubData, sha);
-    }
-  });
-}
-
-// Production builds happen when GitHub sends us a `release` event
-handleEventType('production', 'release');
+webhooks.on('push', (event) => {
+  const githubData = event.payload;
```
humphd, line 75: Let's cache the deploy tag here, since you use it in multiple places below:

```js
const deployTag = githubData.data.deploy_tag;
```

humphd (changes requested):
This is looking pretty close.

@manekenpix, would you be willing to test this with @TDDR on staging at some point?  I think it will be hard to get right without actually trying it.

tools/autodeployment/webhook.js
manekenpix: I think we need a modified version of this. We need to check if this was a push to `master` or not, otherwise we'll deploy after every push (which we don't want).

manekenpix:
@manekenpix, would you be willing to test this with @TDDR on staging at some point? I think it will be hard to get right without actually trying it.

Yes, @TDDR can ping on slack whenever this is ready for testing.

tools/autodeployment/webhook.js
humphd: I wonder if we can be simpler than this: does it have `deploy_tag`?  If it does, deploy it.  If not, ignore.

TDDR:
I may have oversimplified things, but I figure since we are all just taking an educated guess, may as well start off with the most simplistic approach and add things we need.

tools/autodeployment/webhook.js
```diff
@@ -15,93 +15,16 @@ webhooks.onError((error) => logger.error(error));
...
- * @param {string} ref - .Git reference resource.
- * @param {string} mainBranch - Brach used in the repo as main branch.
+ * @otcokit/webhooks .on
+ * https://github.com/octokit/webhooks.js#webhookson
  */
-function requestFilter(repo, buildType, action, ref, mainBranch) {
-  const currentBranch = ref?.split('/').pop();
+webhooks.on('push', (event) => {
+  const githubData = event.payload;
+  const deployTag = githubData.data.deploy_tag | undefined;
```
humphd, line 23: The `| undefined` isn't necessary

humphd (changes requested):
Agree, simplify!

humphd:
It would mean that we ignore GitHub's webhooks in favour of our own, so we won't have any race condition.

Kevan-Y (approved):

TueeNguyen:
Is the next task is to create a workflow to build and push images with proper tags `staging`/`production`?

tcvan0707 (approved):

TDDR:
> Is the next task is to create a workflow to build and push images with proper tags `staging`/`production`?

That's already been taken care of in https://github.com/Seneca-CDOT/telescope/pull/3207

I believe the registry will be up and running after this last piece, or we might still need to figure out the secrets and inputs?

PR #3295, opened 2022-03-23
Get GitHub issues from dependency repo

### LINKED ISSUE #2827: Get and expose npm + GitHub data for dependency packages

JerryHue:
## Issue This PR Addresses

Fixes #2827 

## Type of Change

## Description

This PR adds the whole GitHub issue thing to collect issues that are considered newcomer friendly.

### A couple of details

This doesn't use Redis, at all. I tried to use Redis at first, but it seems that using it is somewhat redundant when the `githubInformationRequestor` is used. The reason I use this is to minimize the number of requests as possible when there are concurrent requests to the same resource. If we program this naively, when there are two or more requests really close to each other, all of those requests will ask essentially the same information, and thus we would be wasting requests.

I will file a couple of issues to address some stuff that have to be changed:
- use Redis instead of a *requestor*, this would imply some rewriting of the whole service, probably.
- use a GitHub Key for the service to use so that we have an increased rate limit on staging and production.

## Steps to test the PR

1. Pull the changes
2. Go to the `dependency-discovery` folder.
3. Start the service with `pnpm dev`.
4. You should be able to request the service from your browser. For example, `GET https://localhost:10500/github/@babel/core`

## Checklist

### DIFF
--- a/src/api/dependency-discovery/src/github.js
+++ b/src/api/dependency-discovery/src/github.js
@@ -1,0 +1,66 @@
+const { request } = require('@octokit/request');
+
+// Check whether a specific time has reached the GitHub time limit.
+// @param {Date} datetime - a Date object representing a moment in the past
+// @returns {Boolean}
+const hasGitHubExpired = (dateTime) => new Date() - dateTime > 60 * 60 * 1000;
+
+const gitHubInformationRequestor = {};
+
+async function constructGitHubRequest(gitHubUrl) {
+  const labels = ['hacktoberfest', 'good first issue', 'help wanted'];
+
+  const [owner, repo] = new URL(gitHubUrl).pathname.substring(1).split('/');
+
+  const requestPromises = labels.map((label) => {
+    return request('GET /repos/{owner}/{repo}/issues{?assignee,state,labels}', {
+      owner,
+      repo,
+      assignee: 'none',
+      state: 'open',
+      labels: label,
+    });
+  });
+
+  // A single request might look like a response in
+  // https://docs.github.com/en/rest/reference/issues#list-repository-issues
+  const responses = await Promise.all(requestPromises);
+
+  return responses
+    .filter((response) => response.status === 200)
+    .flatMap((response) =>
+      response.data.map((issue) => {
+        return {
+          htmlUrl: issue.html_url,
+          title: issue.title,
+          body: issue.body,
+          createdAt: issue.created_at,
+        };
+      })
+    )
+    .filter((issue, index, array) => array.findIndex((i) => i.htmlUrl === issue.htmlUrl) === index);
+}
+
+function requestGitHubInfo(packageName, gitHubUrl) {
+  const cachedRequest = gitHubInformationRequestor[packageName];
+
+  if (cachedRequest && !hasGitHubExpired(cachedRequest.expireAt)) {
+    return cachedRequest.request;
+  }
+
+  const newRequest = constructGitHubRequest(gitHubUrl);
+
+  const now = new Date();
+  now.setHours(now.getHours() + 1);
+
+  gitHubInformationRequestor[packageName] = {
+    expireAt: now,
+    request: newRequest,
+  };
+
+  return newRequest;
+}
+
+module.exports = {
+  requestGitHubInfo,
+};
--- a/src/api/dependency-discovery/src/router.js
+++ b/src/api/dependency-discovery/src/router.js
@@ -1,5 +1,10 @@
 const { Router } = require('@senecacdot/satellite');
-const { getDependencyList, getNpmPackageInfo, isPackageDependency } = require('./dependency-list');
+const {
+  getDependencyList,
+  getNpmPackageInfo,
+  getGitHubIssues,
+  isPackageDependency,
+} = require('./dependency-list');
 
 const router = Router();
 
@@ -27,6 +32,23 @@ router.get('/projects/:namespace/:name?', async (req, res, next) => {
     } else {
       res.set('Cache-Control', 'max-age=3600');
       res.status(200).json(await getNpmPackageInfo(packageName));
+    }
+  } catch (err) {
+    next(err);
+  }
+});
+
+router.get('/github/:namespace/:name?', async (req, res, next) => {
+  const { namespace, name } = req.params;
+  const packageName = name ? `${namespace}/${name}` : namespace;
+
+  try {
+    if (!(await isPackageDependency(packageName))) {
+      res.status(404).json({ msg: `${packageName} is not a Telescope dependency` });
+      next();
+    } else {
+      res.set('Cache-Control', 'max-age=3600');
+      res.status(200).json(await getGitHubIssues(packageName));
     }
   } catch (err) {
     next(err);
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -155,6 +155,7 @@ importers:
 
   src/api/dependency-discovery:
     specifiers:
+      '@octokit/request': 5.6.3
       '@senecacdot/eslint-config-telescope': 1.1.0
       '@senecacdot/satellite': ^1.27.0
       env-cmd: 10.1.0
@@ -163,6 +164,7 @@ importers:
       query-registry: 2.2.0
       supertest: 6.1.6
     dependencies:
+      '@octokit/request': 5.6.3
       '@senecacdot/satellite': 1.27.0
       query-registry: 2.2.0
       supertest: 6.1.6
--- a/src/api/dependency-discovery/package.json
+++ b/src/api/dependency-discovery/package.json
@@ -21,6 +21,7 @@
   },
   "homepage": "https://github.com/Seneca-CDOT/telescope#readme",
   "dependencies": {
+    "@octokit/request": "5.6.3",
     "@senecacdot/satellite": "^1.27.0",
     "query-registry": "2.2.0",
     "supertest": "6.1.6"
--- a/src/api/dependency-discovery/src/dependency-list.js
+++ b/src/api/dependency-discovery/src/dependency-list.js
@@ -2,6 +2,7 @@ const { readFile } = require('fs/promises');
 const { join } = require('path');
 const { cwd } = require('process');
 const { getPackument } = require('query-registry');
+const { requestGitHubInfo } = require('./github');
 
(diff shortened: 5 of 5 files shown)


src/api/dependency-discovery/src/github.js
```diff
@@ -0,0 +1,60 @@
+const { request } = require('@octokit/request');
+
+const hasGitHubExpired = (dateTime) => Date.now() - dateTime > 60 * 60 * 1000;
+
+const gitHubInformationRequestor = {};
+
+async function constructGitHubRequest(packageName, gitHubUrl) {
+  const labelsToSearchFor = ['hacktoberfest', 'good first issue', 'help wanted'];
```
tcvan0707, line 8: I wonder if we are able to add more labels for this, or the GitHub request only asks for those 3 labels?

src/api/dependency-discovery/src/github.js
```diff
@@ -0,0 +1,60 @@
...
+
+  const [owner, repo] = new URL(gitHubUrl).pathname.substring(1).split('/');
+
+  const requestPromises = labelsToSearchFor.map((label) => {
+    return request('GET /repos/{owner}/{repo}/issues{?assignee,state,labels}', {
+      owner,
+      repo,
+      assignee: 'none',
+      state: 'open',
+      labels: label,
```
DukeManh, line 18: Does GH supports logical operators for parameters, which let you write `label1|label2`?

src/api/dependency-discovery/src/github.js
```diff
@@ -0,0 +1,60 @@
...
+function requestGitHubInfo(packageName, gitHubUrl) {
+  const cachedRequest = gitHubInformationRequestor[packageName];
+
+  if (cachedRequest && !hasGitHubExpired(cachedRequest.expireAt)) {
+    return cachedRequest.request;
+  }
+
+  const newRequest = constructGitHubRequest(packageName, gitHubUrl);
+
+  const now = new Date(Date.now());
```
DukeManh, line 47:
```suggestion
  const now = new Date();
```

src/api/dependency-discovery/src/github.js
```diff
@@ -0,0 +1,60 @@
+const { request } = require('@octokit/request');
+
+const hasGitHubExpired = (dateTime) => Date.now() - dateTime > 60 * 60 * 1000;
```
DukeManh, line 3: `datetime` is a `Date` object, so make sure to subtract the same object type here.

DukeManh (changes requested):

src/api/dependency-discovery/src/github.js
TueeNguyen: We can easily add more but these are most popular beginner friendly labels

TueeNguyen, line 8: `labels` is probably good enough, I think it's good to avoid long variable names if possible

src/api/dependency-discovery/src/github.js
```diff
@@ -0,0 +1,60 @@
...
+      owner,
+      repo,
+      assignee: 'none',
+      state: 'open',
+      labels: label,
+    });
+  });
+
+  const responses = await Promise.all(requestPromises);
+
```
TueeNguyen, line 23: Can you add a comment that links to an example response?

src/api/dependency-discovery/src/router.js
```diff
@@ -33,4 +38,21 @@ router.get('/projects/:namespace/:name?', async (req, res, next) => {
   }
 });
 
+router.get('/github/:namespace/:name?', async (req, res, next) => {
+  const { namespace, name } = req.params;
+  const packageName = name ? `${namespace}/${name}` : namespace;
+
+  try {
+    if (!(await isPackageDependency(packageName))) {
+      res.status(404);
```
TueeNguyen, line 47: Maybe
```suggestion
      res.status(404).json({msg: `${packageName} is not Telescope's dependenies`});
```

TueeNguyen:
Can you supply an example request? I tried `/github/motdotla/dotenv` but didn't work

src/api/dependency-discovery/src/github.js
```diff
@@ -0,0 +1,60 @@
+const { request } = require('@octokit/request');
+
+const hasGitHubExpired = (dateTime) => Date.now() - dateTime > 60 * 60 * 1000;
+
+const gitHubInformationRequestor = {};
+
+async function constructGitHubRequest(packageName, gitHubUrl) {
```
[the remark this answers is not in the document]
JerryHue: I wonder why `eslint` did not warn about this...

src/api/dependency-discovery/src/github.js
JerryHue: Not really. That's why I have to search separately. If you pass all labels at once, GitHub is going to try and find issues that have all of those labels.

That reminds me, I think this might result in some repeated issues. I need to worry about that case, too.

JerryHue:
> Can you supply an example request? I tried `/github/motdotla/dotenv` but didn't work

I think if you are writing it like `/github/@motdotla/dotenv`

src/api/dependency-discovery/src/github.js
```diff
@@ -0,0 +1,66 @@
+const { request } = require('@octokit/request');
+
+// Check whether a specific time has reached the GitHub time limit.
+// @param {Date} datetime - a Date object representing a moment in the past
+// @returns {Boolean}
+const hasGitHubExpired = (dateTime) => new Date() - dateTime > 60 * 60 * 1000;
```
humphd, line 6: A better way to do this is with an LRU cache, for example https://www.npmjs.com/package/lru-cache.  It can handle the TTL stuff and expiring keys.

Can you file a follow-up with an issue to do this in Redis instead of in memory?  That way starting/stopping the containers won't lose the cache (e.g., when we re-deploy).  Redis can handle the TTL for you when you add a key, and we can stuff the entire JSON response into the value.

src/api/dependency-discovery/src/github.js
```diff
@@ -0,0 +1,66 @@
+const { request } = require('@octokit/request');
```
humphd, line 1: Can you file a follow-up to use more of https://github.com/octokit/octokit.js vs. just `request`?  The full https://github.com/octokit/octokit.js does a lot more, including throttling, retries, etc.

src/api/dependency-discovery/src/github.js
```diff
@@ -0,0 +1,66 @@
...
+// @param {Date} datetime - a Date object representing a moment in the past
+// @returns {Boolean}
+const hasGitHubExpired = (dateTime) => new Date() - dateTime > 60 * 60 * 1000;
+
+const gitHubInformationRequestor = {};
+
+async function constructGitHubRequest(gitHubUrl) {
+  const labels = ['hacktoberfest', 'good first issue', 'help wanted'];
+
+  const [owner, repo] = new URL(gitHubUrl).pathname.substring(1).split('/');
```
humphd, line 13: `new URL` can throw.

src/api/dependency-discovery/src/github.js
```diff
@@ -0,0 +1,66 @@
...
+  gitHubInformationRequestor[packageName] = {
+    expireAt: now,
+    request: newRequest,
+  };
+
+  return newRequest;
+}
+
+module.exports = {
+  requestGitHubInfo,
```
humphd, line 65: This doesn't seem to be called anywhere?

humphd (changes requested):

AmasiaNalbandian:

src/api/dependency-discovery/src/github.js
JerryHue: @humphd, would you like me to switch the current implementation to a LRU, or would you prefer to leave it as it is and change it to Redis for 3.0-alpha?

JerryHue: The issue is filed #3323.

src/api/dependency-discovery/src/github.js
JerryHue: Filed #3324

src/api/dependency-discovery/src/github.js
JerryHue: I think you missed this:

https://github.com/Seneca-CDOT/telescope/blob/dba5de07a04870f732211c9f6ca1e2864a5834db/src/api/dependency-discovery/src/dependency-list.js#L55-L59

src/api/dependency-discovery/src/github.js
JerryHue: How do we want to handle this? The exception will bubble up and then will be caught in the `try/catch` found in `router.js`. Is that what we want?

src/api/dependency-discovery/src/github.js
humphd: Perfect.

humphd: Let's leave for now, and do Redis (which is the proper way).

src/api/dependency-discovery/src/github.js
humphd: That's fine, just potentially risky (it depends on callers doing the right thing).

humphd (dismissed):
I think this can go in, and we can do the rest in follow-ups (including tests).

PR #3297, opened 2022-03-23
Add GitHub schema to Postgres

### LINKED ISSUE #3215: Use Postgres to store GitHub data
There has been a lot of discussion regarding storing and indexing GitHub information that appears on posts, and how this information can also be used to leverage the use cases of the `dependency-discovery` issue. For example, in #2831 and #2827, we have mentioned several times this, so I think it is time to lay some foundations.

### Motivation

The reason I would like to store all of this in Postgres has been mentioned in [this comment](https://github.com/Seneca-CDOT/telescope/issues/2831#issuecomment-1059820095) of #2831, but I can summarize it like this:

- Although the GitHub data is data we can extract by analysing the posts, the idea is to save that analysis to avoid repeating the work.
- Postgres is more flexible and general than Elasticsearch.

There is also the fact that we are working with an API that has a limit of requests per hour (that being the GitHub API), so storing part of that information in our side would reduce the amount of requests we would have to make.

### Usage of Postgres

How would Postgres be used across the services? Well, since every service might have a specific use case, the idea is to define a simple, flexible interface in Postgres that all services have to respect to use it.

Part of this interface is, of course, the SQL tables we would have to create. 

I can only speak accurately for the `dependency-discovery`. The way that the `dependency-discovery` service might use the store is keeping the repositories relevant to the dependency list, and caching the issues that users are interested the most (like issues labelled "good first issue").

### Table design

The starting point is three tables:

- GitHub Users (these include organizations),
- GitHub Issues (these include Pull requests),
- GitHub Repositories.

You might see the relationships between all of these:

- A GitHub User can have zero or several Repositories,
- A GitHub repository can have zero or several issues.

### Extended design

As you can see the initial design is very generic, and thus, might need more information depending on the use case. Let's think, for example, the search service.

The search service wants to find the posts depending on the GitHub users mentioned. In this case, we would want to connect the GitHub user we have stored with all of the posts we currently have stored in our tables. We can easily connect these two entities by creating a joint/bridge table.

## Let's discuss

I would like to hear anybody's opinions regarding this. The main problem I currently have is that we may be saving too much information that we could easily get from GitHub. I am aware that the main reason we do this is to avoid asking GitHub several times (and thus reach our request limit too quickly), but we might need to refresh this GitHub information quite frequently. Would we need to program some kind of "update info on demand", and the parts that are less requested we update less frequently?

JerryHue:
## Issue This PR Addresses

Fixes #3215

## Type of Change

## Description

This PR adds a new table schema to Postgres to store the GitHub data we will scrape from the posts.

## Steps to test the PR

1. Pull the changes.
2. Start the services (`pnpm run services:start`).
3. Go to `localhost:8910`.
4. Head over the default project and then find the database tables. There should be tables named `github_users`, `github_repositories`, and `github_issues`.

## Checklist

### DIFF
--- a/docker/supabase/volumes/db/init/05-github-schema.sql
+++ b/docker/supabase/volumes/db/init/05-github-schema.sql
@@ -1,0 +1,36 @@
+CREATE EXTENSION IF NOT EXISTS citext WITH SCHEMA extensions;
+
+CREATE TABLE IF NOT EXISTS github_users (
+  id serial PRIMARY KEY,
+  login citext NOT NULL, -- the user name of the user, known as the `login` field in GitHub API responses
+  type citext check ( type IN ('user', 'organization') ) NOT NULL
+);
+
+CREATE TABLE IF NOT EXISTS github_repositories (
+  id serial PRIMARY KEY,
+  name text NOT NULL,
+  owner integer NOT NULL references github_users ON DELETE CASCADE,
+  UNIQUE (owner, name) -- a user cannot have the same repository, but a repository name can be repeated for many users (e.g. forks)
+);
+
+-- In the GitHub API, pull requests are considered issues, too.
+-- Therefore, we would group both pull requests and issues in a single
+-- table. For more information, read:
+-- https://docs.github.com/en/rest/reference/issues and
+-- https://docs.github.com/en/rest/reference/pulls
+CREATE TABLE IF NOT EXISTS github_issues (
+  id serial PRIMARY KEY,
+  number integer NOT NULL,
+  repo integer NOT NULL references github_repositories ON DELETE CASCADE,
+  type citext check ( type IN ('issue', 'pull_request') ) NOT NULL,
+  UNIQUE (repo, number) -- an issue number is unique across the whole repo, but several repos can have the same number
+);
+
+/* 
+  TODO:
+    * Add policies to restrict the users who can write into these
+      tables (preferably, only the search or parser microservice
+      should write).
+    * Add bridge tables that connect this GitHub information to
+      a posts table that contains the posts themselves.
+ */


docker/supabase/volumes/db/init/05-github-schema.sql
```diff
@@ -0,0 +1,29 @@
+CREATE TABLE IF NOT EXISTS github_users (
+  id serial PRIMARY KEY,
+  username text, -- the user name of the user, known as the `login` field in GitHub API responses
+  user_type text check ( user_type IN ('User', 'Organization') )
```
humphd, line 4: Why use Title Case vs. `user` and `organization`?

docker/supabase/volumes/db/init/05-github-schema.sql
```diff
@@ -0,0 +1,29 @@
...
+);
+
+CREATE TABLE IF NOT EXISTS github_repositories (
+  id serial PRIMARY KEY,
+  repo_name text,
+  owner integer references github_users ON DELETE CASCADE
+);
+
+-- Although pull requests have different features from
+-- regular issues, they can be treated the same for our purposes
```
humphd, line 14: From the [Issues API docs](https://docs.github.com/en/rest/reference/issues), so probably you're right to call these issues:

> Note: GitHub's REST API v3 considers every pull request an issue, but not every issue is a pull request. For this reason, "Issues" endpoints may return both issues and pull requests in the response. You can identify pull requests by the pull_request key. Be aware that the id of a pull request returned from "Issues" endpoints will be an issue id. To find out the pull request id, use the "[List pull requests](https://docs.github.com/rest/reference/pulls#list-pull-requests)" endpoint.

And from the [Pull API docs](https://docs.github.com/en/rest/reference/pulls):

> Every pull request is an issue, but not every issue is a pull request. For this reason, "shared" actions for both features, like manipulating assignees, labels and milestones, are provided within [the Issues API](https://docs.github.com/en/rest/reference/issues).

I think it might be good to expand your comment above to better explain this, as I don't know that everyone would understand this intuitively (calling it `github_issues` will confuse some devs).

docker/supabase/volumes/db/init/05-github-schema.sql
```diff
@@ -0,0 +1,29 @@
...
+CREATE TABLE IF NOT EXISTS github_issues (
+  id serial PRIMARY KEY,
+  issue_number integer,
```
humphd, line 17: `number` since it's already in `github_issues`, so the `issue_` is redundant.

docker/supabase/volumes/db/init/05-github-schema.sql
```diff
@@ -0,0 +1,29 @@
...
+  repo integer references github_repositories ON DELETE CASCADE,
+  issue_state text check ( issue_state IN ('open', 'closed') )
```
humphd, line 19: I don't think this is useful for us to store, as it's not synced.  We can do an API request to determine it.

docker/supabase/volumes/db/init/05-github-schema.sql
```diff
@@ -0,0 +1,29 @@
...
+CREATE TABLE IF NOT EXISTS github_repositories (
+  id serial PRIMARY KEY,
+  repo_name text,
+  owner integer references github_users ON DELETE CASCADE
+);
+
+-- Although pull requests have different features from
+-- regular issues, they can be treated the same for our purposes
+CREATE TABLE IF NOT EXISTS github_issues (
+  id serial PRIMARY KEY,
```
humphd, line 16: Question: will this give us everything we need to query for the item?  For example, a [pull request needs](https://docs.github.com/en/rest/reference/pulls#get-a-pull-request):

```
GET /repos/{owner}/{repo}/pulls/{pull_number}
```

And an [issues needs](https://docs.github.com/en/rest/reference/issues#get-an-issue):

```
GET /repos/{owner}/{repo}/issues/{issue_number}
```

Given these requirements, one thing I notice is that we need to include whether this is an `issue` or `pull request`, since we have to use `.../issues/...` or `.../pulls/...` in the `GET` request.

docker/supabase/volumes/db/init/05-github-schema.sql
```diff
@@ -0,0 +1,29 @@
...
+  issue_number integer,
+  repo integer references github_repositories ON DELETE CASCADE,
+  issue_state text check ( issue_state IN ('open', 'closed') )
+);
+
+/* 
+  TODO:
```
humphd, line 23: File an issue on this also.

humphd (changes requested):

docker/supabase/volumes/db/init/05-github-schema.sql
```diff
@@ -0,0 +1,29 @@
+CREATE TABLE IF NOT EXISTS github_users (
+  id serial PRIMARY KEY,
+  username text, -- the user name of the user, known as the `login` field in GitHub API responses
```
DukeManh, line 3: Can we call name it `login` as well

docker/supabase/volumes/db/init/05-github-schema.sql
TueeNguyen, line 16: We might need to store something like `type text check ( type IN ('issue', 'pull_request')` ) .

docker/supabase/volumes/db/init/05-github-schema.sql
```diff
@@ -0,0 +1,29 @@
...
+CREATE TABLE IF NOT EXISTS github_issues (
+  id serial PRIMARY KEY,
+  issue_number integer,
+  repo integer references github_repositories ON DELETE CASCADE,
+  issue_state text check ( issue_state IN ('open', 'closed') )
+);
+
+/* 
+  TODO:
+    * Add policies to restrict the users who can write into these
```
TueeNguyen, line 24: We can supply our services a secret `service role key`, authorize using Postgre function (not sure how to implement yet)

docker/supabase/volumes/db/init/05-github-schema.sql
JerryHue: Because in the GitHub API the use title case. I can change it to lowercase, but we need to make sure that we always make the types lowercase.

docker/supabase/volumes/db/init/05-github-schema.sql
JerryHue: I tend to avoid using column names that can also be keywords, but I will change it either way.

docker/supabase/volumes/db/init/05-github-schema.sql
JerryHue: Well, we can query just fine from the issues endpoint. However, we won't be able to differentiate them, since we are treating PRs are issues (which is technically fine for the API). I will add an extra column that also records the type.

JerryHue: Yeah, @humphd also brought this point forward.

docker/supabase/volumes/db/init/05-github-schema.sql
JerryHue: I can file a future issue to address this, like how @humphd suggested.

docker/supabase/volumes/db/init/05-github-schema.sql
JerryHue: Filed #3305

humphd (approved):

HyperTHD (dismissed):

docker/supabase/volumes/db/init/05-github-schema.sql
```diff
@@ -0,0 +1,32 @@
+CREATE TABLE IF NOT EXISTS github_users (
+  id serial PRIMARY KEY,
+  login text, -- the user name of the user, known as the `login` field in GitHub API responses
+  type text check ( type IN ('user', 'organization') )
```
DukeManh, line 4: This needs to be `NOT NULL` otherwise the check is not meaningful.

Also, I think it's good to add `NOT NULL` or `DEFAULT` to where it makes sense.

DukeManh (changes requested):

docker/supabase/volumes/db/init/05-github-schema.sql
```diff
@@ -0,0 +1,32 @@
...
+-- In the GitHub API, pull requests are considered issues, too.
+-- Therefore, we would group both pull requests and issues in a single
+-- table. For more information, read:
+-- https://docs.github.com/en/rest/reference/issues and
+-- https://docs.github.com/en/rest/reference/pulls
+CREATE TABLE IF NOT EXISTS github_issues (
+  id serial PRIMARY KEY,
+  number integer,
+  repo integer references github_repositories ON DELETE CASCADE,
+  type text check ( type IN ('issue', 'pull_request') )
```
DukeManh, line 22: Same issue here.

DukeManh (changes requested):

docker/supabase/volumes/db/init/05-github-schema.sql
JerryHue: I don't think there is a safe DEFAULT in this case. I will put a NOT NULL constraint to be explicit

docker/supabase/volumes/db/init/05-github-schema.sql
```diff
@@ -0,0 +1,32 @@
+CREATE TABLE IF NOT EXISTS github_users (
+  id serial PRIMARY KEY,
+  login text, -- the user name of the user, known as the `login` field in GitHub API responses
+  type text check ( type IN ('user', 'organization') ) NOT NULL
+);
+
+CREATE TABLE IF NOT EXISTS github_repositories (
+  id serial PRIMARY KEY,
+  repo_name text,
```
DukeManh, line 9: NOT NULL? A repo without a name doesn't make sense to me.

docker/supabase/volumes/db/init/05-github-schema.sql
```diff
@@ -0,0 +1,32 @@
...
+);
+
+-- In the GitHub API, pull requests are considered issues, too.
+-- Therefore, we would group both pull requests and issues in a single
+-- table. For more information, read:
+-- https://docs.github.com/en/rest/reference/issues and
+-- https://docs.github.com/en/rest/reference/pulls
+CREATE TABLE IF NOT EXISTS github_issues (
+  id serial PRIMARY KEY,
+  number integer,
```
DukeManh, line 20: NOT NULL?

DukeManh (changes requested):

docker/supabase/volumes/db/init/05-github-schema.sql
humphd: Interesting.  If they do it, maybe we should too?  Just seems odd to me.

docker/supabase/volumes/db/init/05-github-schema.sql
JerryHue: So, I added NOT NULL to everything (a repo cannot exist without an owner, and an issue cannot exist without belonging to a repo).

docker/supabase/volumes/db/init/05-github-schema.sql
JerryHue: It is weird, I agree :/
I think we can always make it lowercase with the datatype that @DukeManh used for the emails in the `telescope`-specific schema

humphd (approved):

Issue #3225, opened 2022-03-16
Updating Search Service and tests to implement Autocomplete for Authors


rclee91:
In #1417 , the ES indexer will be modified or a new indexer will be added to include analyzers in order to setup for autocomplete on Authors. 

The Search service will need to be modified to adapt to the new indexing, and corresponding tests will need to be written to reflect the changes.

PR #3301, opened 2022-03-24
Implement Author Autocomplete (legacy backend and Search)

### LINKED ISSUE #1417: Add author suggestions to Elasticsearch indexer backend

If a user starts typing a name, we'd like to be able to present a list of Author completions/suggestions, similar to how Google's search suggestions work.  To do this, we need to add backend support to our [Elasticsearch indexer](https://github.com/Seneca-CDOT/telescope/blob/master/src/backend/utils/indexer.js).

In #1260 we are adding front-end support for autocomplete on an Author search.  
[Here's one good article on possible ways to do this with Elasticsearch](https://medium.com/@taranjeet/elasticsearch-building-autocomplete-functionality-494fcf81a7cf) (there are a few different options we can pick from).

We don't need to send back 1000+ names, but as you type, we could send back a dozen or so, and the more you type the closer we'd get to the name you want.

rclee91:
- Add author autocomplete settings for indexer.js in backend
- refactor Search service to work with new indexing
- create new search query for author autocomplete
- add supertest as a dev-dependency to pass ESLint

## Issue This PR Addresses
Fixes #1417 
also a part of issue #3225 (no new tests are added just yet)

## Type of Change

## Description

- In the legacy `backend`, I've added `settings` and `mapping` for the `posts` index to use [edge n-gram](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-edgengram-tokenizer.html) for author autocomplete
- Because we couldn't specify `type` on index creation, I've refactored the `Search` service and tests to work with the new indexing. The current search query is not affected by this new change. `pnpm test search` would pass all tests.
- To get author autocomplete suggestions, I've created a new query out of `Search`, and ways to test it and see it in action is as follows

## Steps to test the PR

- get on my branch, `pnpm install`
- **\*\*Important\*\*** remove `redis` cache `rm -rf redis-data`, this is so the posts can be re-indexed
- build Docker Images
```terminal
 docker-compose --env-file ./config/env.development up --build redis elasticsearch posts traefik search nginx planet
```
- Wait for a bit for ElasticSearch to connect, then `pnpm start`
- Wait a bit for the posts to be indexed
- Can now test `Search` queries at http://localhost/v1/search
- The original Search queries shouldn't be affected. The new query can be tested at  http://localhost/v1/search/authors/autocomplete/
- When `author` is queried, it will return the number of suggested authors, and an array of suggested authors. As the query gets longer, results will be more precise

image
image
image
image
image

 
## Problems
There is still a lot of tweaking that can be done
- A strict validation for the  http://localhost/v1/search/authors/autocomplete/ isn't yet in place. Currently I am only hijacking the old search validation. If we don't have an `author` query and it passes validation, we'll get a `Error 400`
image
- A query of only one letter won't pass validation, it has to be 2 or more letters. This is part of our rules in `validation.js`
image

## Checklist

- [x] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not (if applicable)variables are added/changed or an explanation of why it does not(if applicable)

### DIFF
--- a/src/api/search/src/bin/validation.js
+++ b/src/api/search/src/bin/validation.js
@@ -44,14 +44,25 @@ const ValidationRules = [
     .bail(),
 ];
 
+const authorRules = [
+  check('author')
+    .exists({ checkFalsy: true })
+    .withMessage('author should exist')
+    .bail()
+    .isLength({ max: 100, min: 1 })
+    .withMessage('invalid author value')
+    .bail(),
+];
+
 /**
  * Validates query by passing rules. The rules are different based on the pathname
- * of the request. If the pathname is '/' it is the basic route.
- * Otherwise, if '/advanced/' it is the advanced search
+ * of the request. If the pathname is '/' it is the advanced search route.
+ * Otherwise, if '/authors/autocomplete, it is the author autocomplete route
  */
 const validateQuery = () => {
   return async (req, res, next) => {
-    await Promise.all(ValidationRules.map((rule) => rule.run(req)));
+    const rules = req.route.path === '/' ? ValidationRules : authorRules;
+    await Promise.all(rules.map((rule) => rule.run(req)));
 
     const result = validationResult(req);
     if (result.isEmpty()) {
--- a/src/api/search/src/routes/query.js
+++ b/src/api/search/src/routes/query.js
@@ -1,12 +1,21 @@
 const { Router, createError } = require('@senecacdot/satellite');
 const { validateQuery } = require('../bin/validation');
-const { search } = require('../search');
+const { search, authorAutocomplete } = require('../search');
 
 const router = Router();
 
 router.get('/', validateQuery, async (req, res, next) => {
   try {
     res.send(await search(req.query));
+  } catch (error) {
+    next(createError(503, error));
+  }
+});
+
+// query for author autocomplete
+router.get('/authors/autocomplete', validateQuery, async (req, res, next) => {
+  try {
+    res.send(await authorAutocomplete(req.query));
   } catch (error) {
     next(createError(503, error));
   }
--- a/src/api/search/src/search.js
+++ b/src/api/search/src/search.js
@@ -4,7 +4,6 @@ const { Elastic } = require('@senecacdot/satellite');
 const client = Elastic();
 
 const index = 'posts';
-const type = 'post';
 
 const calculateFrom = (page, perPage) => {
   const ES_MAX = 10000; // 10K is the upper limit of what ES will return without issue for searches
@@ -94,7 +93,6 @@ const search = async (options) => {
     size: options.perPage,
     _source: ['id'],
     index,
-    type,
     body: results,
   });
 
@@ -103,4 +101,53 @@ const search = async (options) => {
     values: hits.hits.map(({ _id }) => ({ id: _id, url: `${POSTS_URL}/${_id}` })),
   };
 };
-module.exports = { search };
+
+/**
+ * authorAutocomplete allows for querying author autocomplete fields
+ * @param { author } - text to search in author field
+ * @return all the results matching the fields text
+ */
+const authorAutocomplete = async ({ author }) => {
+  const results = {
+    query: {
+      match: {
+        'author.autocomplete': {
+          query: author,
+          operator: 'and',
+        },
+      },
+    },
+    highlight: {
+      fields: {
+        'author.autocomplete': {},
+      },
+    },
+  };
+
+  const {
+    body: { hits },
+  } = await client.search({
+    size: 10000,
+    _source: ['author'],
+    index,
+    body: results,
+  });
+
+  // Filter through all authors and remove duplicates
+  const res = hits.hits.reduce(
+    (acc, { _source }, i) =>
+      acc.find((item) => item.author === _source.author)
+        ? acc
+        : acc.concat({
+            author: _source.author,
+            highlight: hits.hits[i].highlight['author.autocomplete'][0],
+          }),
+    []
+  );
+
+  return {
+    results: res.length,
+    res,
+  };
+};
+module.exports = { search, authorAutocomplete };
--- a/src/backend/utils/indexer.js
+++ b/src/backend/utils/indexer.js
@@ -7,7 +7,86 @@ const { client } = require('../lib/elastic');
 const { logger } = require('./logger');
 
 const index = 'posts';
-const type = 'post';
+
+/*
+  Creates 'posts' index and add settings and mapping for autocomplete if the index doesn't already exist
+  exists API: https://www.elastic.co/guide/en/elasticsearch/reference/master/indices-exists.html
+  create API: https://www.elastic.co/guide/en/elasticsearch/reference/master/indices-create-index.html
+
+  Analysis settings are used to define our custom analyzers and tokenizers
+  analyzer: https://www.elastic.co/guide/en/elasticsearch/reference/current/analyzer.html
+  search_analyzer: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-analyzer.html
+
+  Token filters are built in filters to build customized analyzers
+  'lowercase' changes text to lowercase: https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-lowercase-tokenfilter.html
+  'remove_duplicates' removes duplicate tokens in the same position: https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-remove-duplicates-tokenfilter.html
+  creating a custom analyzer: https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-custom-analyzer.html
+
+  Edge n-gram tokenizer is used for our custom autocomplete tokenizer
+  tokenizer reference: https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-tokenizers.html
+  Edge n-gram tokenizer: https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-edgengram-tokenizer.html
+
+  'min_gram' and 'max_gram' are the minimum and maximum length of characters in a gram.
+  There are some max_gram limitations: https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-edgengram-tokenizer.html#max-gram-limits
+
+  'token_chars' are character classes that are included in a token.
+  Currently set at 'letter' and 'digit'
+
+  Mapping is used to set the custom author.autocomplete field to use our custom analyzers
+  mapping: https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping.html
+  fields: https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-fields.html
+*/
+const setupPostsIndex = async () => {
+  try {
+    const response = await client.indices.exists({ index });
+    // If the index doesn't exist, 404 statusCode is returned
+    if (response.statusCode === 404) {
+      await client.indices.create({
+        index,
+        body: {
+          settings: {
+            analysis: {
+              analyzer: {
+                autocomplete_analyzer: {
+                  tokenizer: 'autocomplete',
+                  filter: ['lowercase', 'remove_duplicates'],
+                },
+                autocomplete_search_analyzer: {
+                  tokenizer: 'lowercase',
+                },
+              },
+              tokenizer: {
+                autocomplete: {
+                  type: 'edge_ngram',
+                  min_gram: 1,
+                  max_gram: 20,
+                  token_chars: ['letter', 'digit'],
+                },
+              },
+            },
+          },
+          mappings: {
+            properties: {
+              author: {
+                type: 'text',
+                fields: {
+                  autocomplete: {
+                    type: 'text',
+                    analyzer: 'autocomplete_analyzer',
+                    search_analyzer: 'autocomplete_search_analyzer',
+                  },
+                },
+                analyzer: 'standard',
+              },
+            },
+          },
+        },
+      });
+    }
+  } catch (error) {
+    logger.error({ error }, `Error setting up ${index} index`);
+  }
+};
 
 /**
  * Indexes the text and id from a post
@@ -22,7 +101,6 @@ const indexPost = async ({ text, id, title, published, author }) => {
   try {
     await client.index({
       index,
-      type,
       id,
       body: {
         text,
@@ -44,7 +122,6 @@ const deletePost = async (postId) => {
   try {
     await client.delete({
       index,
-      type,
       id: postId,
     });
   } catch (error) {
@@ -114,7 +191,6 @@ const search = async (
     size: perPage,
     _source: ['id'],
     index,
-    type,
     body: query,
   });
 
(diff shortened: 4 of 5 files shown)


src/api/search/src/routes/query.js
```diff
@@ -12,4 +12,13 @@ router.get('/', validateQuery, async (req, res, next) => {
   }
 });
 
+// query for autocomplete
+router.get('/complete', validateQuery, async (req, res, next) => {
```
humphd, line 16: I wonder if we should do `/authors/autocomplete` so that we can do `/posts/autocomplete` later etc.?

humphd (commented):
This is amazing!  I've done one read through, and will spend more time on it tomorrow, but it looks fantastic.

We should figure out if we can migrate this to the parser service, since it's landing soon (can happen in follow-up).

src/api/search/src/routes/query.js
rclee91: Sure, I can change it to that, cause its more straightforward.
I also think its also possible to do `multi_match` queries on multiple autocomplete fields in the future too, and we can always modify what is returned from the queries. This is just assumption based off of my research, nothing concrete yet.

rclee91:
I've changed the route name, and I've also added `highlights`
ElasticSearch can help us wrap the searched terms with `<em>...</em>` tags. Which might be useful for the front end. 
So we can decide if we want to keep one of `authors` and `highlights` or both

(I tested without waiting for all feeds to put in)
image

src/api/search/src/routes/query.js
```diff
@@ -12,4 +12,13 @@ router.get('/', validateQuery, async (req, res, next) => {
   }
 });
 
+// query for author autocomplete
+router.get('/authors/autocomplete', validateQuery, async (req, res, next) => {
```
humphd, line 16: I wonder if we need to write another validator, since `validateQuery` is overly broad for what you need here.  That would solve your 2 character limit too.  What about adding `validateAuthor` in https://github.com/Seneca-CDOT/telescope/blob/d8867e9c7f001574ec85630103653dc92a93acfc/src/api/search/src/bin/validation.js.

src/api/search/src/search.js
```diff
@@ -103,4 +101,57 @@ const search = async (options) => {
...
+      fields: {
+        'author.autocomplete': {},
+      },
+    },
+  };
+
+  const {
+    body: { hits },
+  } = await client.search({
+    size: 10000,
```
humphd, line 130: I wonder if there's an upper-limit on what we should return so it's responsive to use in the front-end?  There's probably some guidance on how many results to return for autocomplete queries so that you have useful but small downloads, and it can query and re-render everything fast enough to be useful when typing.

I wonder if we should consider using 10? See https://ux.stackexchange.com/posts/56131/revisions

src/api/search/src/search.js
```diff
@@ -103,4 +101,57 @@ const search = async (options) => {
...
+    _source: ['author'],
+    index,
+    body: results,
+  });
+
+  // Filter through all authors and remove duplicates
```
humphd, line 136: Do we store duplicates?

If so, a faster way to do this is to put the data into a `Set` and then return an Array from that, since Sets don't include duplicates:

```js
const authors = [...new Set(hits.hits)];
```

src/api/search/src/search.js
```diff
@@ -103,4 +101,57 @@ const search = async (options) => {
...
+  const authors = hits.hits.reduce(
+    (acc, { _source }) => (acc.includes(_source.author) ? acc : [...acc, _source.author]),
+    []
+  );
+
+  // Filter through all highlighted authors and remove duplicates
```
humphd, line 142: Didn't you just remove dupes above?

src/backend/utils/indexer.js
```diff
@@ -7,7 +7,58 @@ const { client } = require('../lib/elastic');
...
 
 const index = 'posts';
-const type = 'post';
+
+// creates 'posts' index and add settings and mapping for autocomplete
+const setupPostsIndex = async () => {
+  try {
+    const response = await client.indices.exists({ index });
+    if (response.statusCode === 404) {
+      await client.indices.create({
```
humphd, line 16: Given that you're the only person who understands this code, and we'll have to maintain it, can you add a big comment above this index creation explaining how it works, linking to relevant docs, etc?

humphd (changes requested):

Kevan-Y:
@rclee91 if I put 1 letter in author, does it suppose to return invalid?
`http://localhost/v1/search/authors/autocomplete/?author=k`
image

src/api/search/src/search.js
rclee91: I'll try to answer the "size", "duplicate", and "highlights" conversation here.
Below is what a response from an ES Search request looks like. 
(I am using sample data on Kibana)
```js
{
  "took" : 2,
  "timed_out" : false,
  "_shards" : {
    "total" : 1,
    "successful" : 1,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 3,
      "relation" : "eq"
    },
    "max_score" : 0.28194553,
    "hits" : [
      {
        "_index" : "posts",
        "_id" : "2",
        "_score" : 0.28194553,
        "_source" : {
          "author" : "Dave Humph"
        },
        "highlight" : {
          "author.autocomplete" : [
            "<em>Da</em>ve <em>H</em>umph"
          ]
        }
      },
      {
        "_index" : "posts",
        "_id" : "3",
        "_score" : 0.28194553,
        "_source" : {
          "author" : "Dave Humph"
        },
        "highlight" : {
          "author.autocomplete" : [
            "<em>Da</em>ve <em>H</em>umph"
          ]
        }
      },
      {
        "_index" : "posts",
        "_id" : "1",
        "_score" : 0.24156076,
        "_source" : {
          "author" : "David Humphrey"
        },
        "highlight" : {
          "author.autocomplete" : [
            "<em>Da</em>vid <em>H</em>umphrey"
          ]
        }
      }
    ]
  }
}
```

The "duplicates" are from the posts, cause each author could have more than one post. 
So if we set the `size` to 10, there's a possibility that the first 10 hits from ElasticSearch are all from the same one author.
Right now its set to the maximum amount of hits that it can page through so we make sure we get all `authors` that match the searched query, and then we take out the duplicates. Currently ElasticSearch doesn't have a way of knowing they are "duplicates", because for ES, they are different results from different "posts".

We won't really have this problem if we were autocompleting on another field with more unique text, for example the `title` of each post. 

Currently the `authors` and `highlights` array are from two different properties (`_score` and `highlight`) in the `hits.hits` array. But, since they're technically the same results, just one with `<em>...</em>` tags and one without, I'm assuming we only want to keep one of them then?

I don't think `Set` will work with `hits.hits` since all the `_id` for each post is different, and so will some `_score`. The `_score` is how ElasticSearch determines how relevant the result is to the query.

src/backend/utils/indexer.js
```diff
@@ -7,7 +7,58 @@ const { client } = require('../lib/elastic');
...
+                  autocomplete: {
+                    type: 'text',
+                    analyzer: 'autocomplete_analyzer',
+                    search_analyzer: 'autocomplete_search_analyzer',
+                  },
+                },
+                analyzer: 'standard',
+              },
+            },
+          },
```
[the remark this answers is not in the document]
rclee91: In theory yes. But I haven't done anything in practise to confirm.

src/api/search/src/search.js
```diff
@@ -103,4 +101,57 @@ const search = async (options) => {
...
+  } = await client.search({
+    size: 10000,
+    _source: ['author'],
+    index,
+    body: results,
+  });
+
+  // Filter through all authors and remove duplicates
+  const authors = hits.hits.reduce(
+    (acc, { _source }) => (acc.includes(_source.author) ? acc : [...acc, _source.author]),
```
DukeManh, line 138:
```suggestion
    (acc, { _source }) => (acc.includes(_source.author) ? acc : acc.concat(_source.author),
```

src/api/search/src/search.js
```diff
@@ -103,4 +101,57 @@ const search = async (options) => {
...
+    []
+  );
+
+  // Filter through all highlighted authors and remove duplicates
+  const highlights = hits.hits.reduce(
```
DukeManh, line 143: Since highlight and author are from the same search result, we could merge them into one array instead of using the same index to get the highlight and author of a result. The object can look like this.
```js
[{
author: 'David Li',
highlight: '"<em>David</em> Li"
},
{...}
]
```

DukeManh (changes requested):
Google should consider hiring us.

TueeNguyen (commented):
~~I had to spend some time learning on Edge n gram which was very interesting.
I tried to test this PR, the server ran but I couldn't to get any result back.~~
~~* `http://localhost/v1/search/complete/?author=Da` returns 404~~
~~* `http://localhost/v1/search/authors/autocomplete/?author=Da` returns 404~~

Edit: my backend took a while to index for some reason

Also the query looks strange, should it be `http://localhost/v1/search/autocomplete?author=Da`

DukeManh:
>  tried to test this PR, the server ran but I couldn't to get any result back.

The URL changed to `http://localhost/v1/search/authors/autocomplete/?author=dav`.

Kevan-Y:
> ###

I think it would be better to have `http://localhost/v1/search/authors/autocomplete?author=dav`

src/backend/utils/indexer.js
TueeNguyen: [The documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-edgengram-tokenizer.html#analysis-edgengram-tokenizer) and [this video](https://www.youtube.com/watch?v=FSPwonNOldk&ab_channel=Codetuber)  helped me understand

src/backend/utils/indexer.js
```diff
@@ -7,7 +7,58 @@ const { client } = require('../lib/elastic');
...
+  try {
+    const response = await client.indices.exists({ index });
+    if (response.statusCode === 404) {
+      await client.indices.create({
+        index,
+        body: {
+          settings: {
+            analysis: {
+              analyzer: {
+                autocomplete_analyzer: {
```
TueeNguyen, line 22: This is our custom function not a preserve field. (comment)

src/backend/utils/indexer.js
```diff
@@ -7,7 +7,58 @@ const { client } = require('../lib/elastic');
...
+                  tokenizer: 'autocomplete',
+                  filter: ['lowercase', 'remove_duplicates'],
```
TueeNguyen, line 24: If you already have an analyzer that removes duplicate, do you need a filter in `search.js`? Both should be ok.

src/api/search/src/search.js
rclee91: I'm just curious if there's a reason to use `concat` over what was before. Better readability?

src/backend/utils/indexer.js
rclee91: Its not the same thing. 
This is for removing duplicate `tokens` in the same position when indexing so it doesn't bloat up memory
https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-remove-duplicates-tokenfilter.html

src/api/search/src/search.js
DukeManh: Readability yeah, why use spread syntax when you can use a built-in function. 
I would personal write `acc.push(_source.auth)` to get O(1) 🤓

src/api/search/src/search.js
DukeManh: Btw, should we enable fuzzy query in the `/autocomplete` route?

src/api/search/src/search.js
rclee91: Except I think `acc.push` just returns the new length of the array. To use it we can't use it in one line and would still have to `return acc` somewhere in the callback, overall O(n). 

I guess they're all the same. I'll change to `concat` for better readability.

DukeManh: Correct me if im wrong but arrays are JS objects and are passed by reference. So if you push to `acc` and return it, it's not copied(O(n)) but reused in the next iteration.

Kevan-Y: JS object is passed by reference

src/api/search/src/search.js
rclee91: That depends, how accurate we'd like to be with name searches, or how fuzzy we want it to be. 
I personally don't think we need it for just names.

Kevan-Y:
> http://localhost/v1/search/authors/autocomplete/?author=dav

That interesting how the routing works, `http://localhost/v1/search/authors/autocomplete/?author=dav` and `http://localhost/v1/search/authors/autocomplete?author=dav` are the same

TueeNguyen:
I think whatever is after the `?` counts as query parameter.

src/api/search/src/search.js
Kevan-Y: > Since highlight and author are from the same search result, we could merge them into one array instead of using the same index to get the highlight and author of a result. The object can look like this.
> 
> ```js
> [{
> author: 'David Li',
> highlight: '"<em>David</em> Li"
> },
> {...}
> ]
> ```

Yeah, agree to merge it, I think it would be easier in the front end to use this data if we want the highlight matching the author.

src/api/search/src/search.js
humphd: None of this matters, we're talking about 10, 100, 1000 iterations?  That's not where this code is going to be slow.  What's slow is the network transfer.  Don't optimize the wrong thing.

DukeManh: I agree, it's premature optimization , won't make much difference. We just discussing for the best solution

rclee91:
Added comments
Added new author validation rule
Merged `author` and `highlight`

src/api/search/src/search.js
rclee91: I'd like to get a confirmation on this (please see conversation below)
I don't mind changing it to `10` or something else smaller than `10,000`
I just hope that it is clear that if it is too small we might get fewer author results than what was expected of.

src/api/search/src/bin/validation.js
```diff
@@ -44,14 +44,25 @@ const ValidationRules = [
...
  * Validates query by passing rules. The rules are different based on the pathname
- * of the request. If the pathname is '/' it is the basic route.
- * Otherwise, if '/advanced/' it is the advanced search
+ * of the request. If the pathname is '/' it is the advanced search route.
+ * Otherwise, if '/authors/autocomplete, it is the author autocomplete route
  */
 const validateQuery = () => {
   return async (req, res, next) => {
-    await Promise.all(ValidationRules.map((rule) => rule.run(req)));
+    const rules = req.route.path === '/' ? ValidationRules : authorRules;
```
humphd, line 64: This makes me nervous, but let's go with it for now and fix later.  I'd rather split these routes up vs. having this logic determine what we do.

src/api/search/src/search.js
humphd: Let's begin with this, and reduce it as we learn how it performs.

src/api/search/src/search.js
humphd: Understood, thanks for clarifying and educating me.

humphd (approved):
Let's do it.

Kevan-Y (approved):
LGTM, tested locally it works

Issue #3311, opened 2022-03-24
CI overlooks mismatched pnpm-lock.yaml file


AmasiaNalbandian:
**What happened**:
We have been getting CI to fail the check to see if the pnpm-lock.yaml file is up to date. 

**What should have happened**:
CI should fail every commit that does not have the correct pnpm-lock.yaml file.

**How to reproduce it (as precise as possible)**:
Put up an outdated pnpm-lock.yaml, watch CI pass on Telescope PRs. I think on local branch it's different.

**Anything else we need to know?**:
### Cindy's commit in master - passed CI. 
![chrome_dXEXp9PT3O](https://user-images.githubusercontent.com/77639637/159972465-92f6acc8-ab8b-4041-8e5d-363ca817029d.png)

### Cindy's PR Actions Report
![chrome_EJw6HAUZdf](https://user-images.githubusercontent.com/77639637/159972453-581adf05-641a-40d6-b768-6c594a49e248.png)

### Cindy's Branch on her fork for the same code:
![chrome_Tm67y7xKhQ](https://user-images.githubusercontent.com/77639637/159972429-609ebab1-f78a-40ca-bf18-cad242efe006.png)
**Environment**:

- OS:
- Browser:

JerryHue:
Should we have a single workflow that checks for this and avoids running the rest of the runs, so that we don't have to wait for the rest?

humphd:
Do a test: make a PR that intentionally has the lock-file out of sync, and see what our CI does?

JerryHue:
I will do it now. I think it is urgent we push this, because there are other PRs that add a few dependencies, and we don't want master to start having a stale `pnpm-lock`.

JerryHue:
Also, I think Cindy's CI broke because of an unrelated issue: 
![image](https://user-images.githubusercontent.com/53304516/159974121-409a5bd4-94fb-4879-a88c-131c9e877e38.png)

Seems to be related to Vercel deployment.

JerryHue:
Some interesting results:

- If the `pnpm-lock` file is *stale*, `pnpm install` will fail, and thus CI fails. Here is an example run: 
![image](https://user-images.githubusercontent.com/53304516/159987341-b63041bb-55f0-47f5-a398-fc5f9808d8c8.png)
- If the `pnpm-lock` file is *broken*, `pnpm install` does not fail, and thus CI does not fail either. You can check the run: 
[...]

PR #3312, opened 2022-03-24
Use stale lock-file from release 2.8.3

JerryHue:
This PR is used to trigger the CI.

DO NOT MERGE!!! If you do, Dave will kill me. 😢

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -242,6 +242,7 @@ importers:
       normalize-url: 6.1.0
       rss-parser: 3.12.0
       sanitize-html: 2.5.3
+      set-interval-async: 2.0.3
     devDependencies:
       '@senecacdot/eslint-config-telescope': 1.1.0_eslint@7.32.0
       babel-jest: 27.5.1_@babel+core@7.17.8
@@ -480,7 +481,7 @@ importers:
       nock: 13.2.4
       pretty-quick: 3.1.3_prettier@2.5.1
 
-  src/web/app:
+  src/web:
     specifiers:
       '@emotion/react': 11.8.2
       '@emotion/styled': 11.8.1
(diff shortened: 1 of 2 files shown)


humphd:
```
Lockfile is up-to-date, resolution step is skipped
 ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up-to-date with src/api/parser/package.json
```

It already does what you want.

JerryHue:
@humphd, so if CI fails when the lock-file is stale, then what happened with Tue's PR? Why it did not fail?

humphd:
Who knows?  GitHub Actions is having a bad week, maybe something failed on the infra?  The way we have things, it should work as expect (i.e., error when something isn't updated).

JerryHue:
I think it may fail when the lock-file is broken, not when it is stale

humphd:
```
 WARN  Ignoring broken lockfile at /home/runner/work/telescope/telescope: The lockfile at "/home/runner/work/telescope/telescope/pnpm-lock.yaml" is broken: duplicated mapping key (1041:3)
```

JerryHue:
> I think it may fail when the lock-file is broken, not when it is stale

Sorry, I meant the opposite: I think CI may fail when the lock-file is stale, not when it is broken

> ```
>  WARN  Ignoring broken lockfile at /home/runner/work/telescope/telescope: The lockfile at "/home/runner/work/telescope/telescope/pnpm-lock.yaml" is broken: duplicated mapping key (1041:3)
> ```

Yeah, it only warns about this. `pnpm install` does not care that the lockfile is broken because it repairs it anyway. However, we want the CI to check for real, so I think we should add another job that checks this.

JerryHue:
Very well, now that we know the actual reason, we can submit a PR that addresses this. If we want to discuss more, let's do it in #3311.

PR #3328, opened 2022-03-25
Pass anon_key secret in continuous-delivery.yml workflow

humphd:
We forgot to pass the `SUPBASE_ANON_KEY` secret in the Continuous Delivery workflow, which is required after @joelazwar's supabase front-end changes that just landed.


HyperTHD (approved):

JerryHue (approved):

humphd:
Once this goes green, let's land it next, so the other CD workflows for things we rebase and land afterward will work.

PR #3330, opened 2022-03-25
Move anon_key to inputs vs secrets in CI/CD

humphd:
Today we learned that you can't use `secrets.*` in a build matrix with GitHub Actions, so the `ANON_KEY` has to be an `inputs.*` instead.  We think this is OK, since the anonymous key is being bundled into the front-end (e.g., it's not a secret).

When we change the anon key for production, we need to update this.


AmasiaNalbandian (approved):

JerryHue:
Do we have to put the key right into the `yml` file, instead of being in a `env` variable?

PR #3334, opened 2022-03-25
Include github.sha for GIT_COMMIT in build args for CD

humphd:
The front-end build for nginx needs the `GIT_COMMIT`, but we aren't passing it in with the build args to Docker.  This adds it.

### DIFF
--- a/.github/workflows/docker-build-and-push.yml
+++ b/.github/workflows/docker-build-and-push.yml
@@ -92,6 +92,7 @@ jobs:
               SEARCH_URL=${{ inputs.search_url }}
               FEED_DISCOVERY_URL=${{ inputs.feed_discovery_url }}
               STATUS_URL=${{ inputs.status_url }}
+              GIT_COMMIT=${{ github.sha }}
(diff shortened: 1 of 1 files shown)


Kevan-Y (approved):

.github/workflows/docker-build-and-push.yml
```diff
@@ -92,6 +92,7 @@ jobs:
               SEARCH_URL=${{ inputs.search_url }}
               FEED_DISCOVERY_URL=${{ inputs.feed_discovery_url }}
               STATUS_URL=${{ inputs.status_url }}
+              GIT_COMMIT=${{ github.sha }}
```
JerryHue, line 95: I didn't you could use this! Neato!

Docs: https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

JerryHue (approved):

.github/workflows/docker-build-and-push.yml
humphd: There are so many toys in GitHub Actions, it's amazing.

PR #3336, opened 2022-03-25
Update Dockerfiles for M1

humphd:
I can't build our images anymore on my M1 (ARM) laptop.  This changes things to use a larger node image (with build tools like Python) for installing deps, then copies the deps over to an `-alpine` image that's smaller (no build tools, etc).

I've also removed the manual `tini` stuff, and added `init: true` instead, which does this automatically in Docker engine.

I also fixed the `PORT` stuff, since we might as well use the right port naming.

### DIFF
--- a/src/api/dependency-discovery/Dockerfile
+++ b/src/api/dependency-discovery/Dockerfile
@@ -1,6 +1,6 @@
 ## Base ###########################################################################
 # Set up the base layer
-FROM node:16-alpine3.15 as base
+FROM node:16 as base
 
 RUN npm i -g pnpm
 
@@ -16,18 +16,20 @@ RUN pnpm install --prod
 
 ## Deploy ########################################################################
 # Stage for running our app
-FROM base as deploy
+FROM node:16-alpine3.15 as deploy
+
+WORKDIR /app
 
 COPY --chown=node:node . .
 
 COPY --chown=node:node --from=dependencies /app/node_modules ./node_modules
 
-ENV PORT=10500
+ENV DEPENDENCY_DISCOVERY_PORT=10500
 
 USER node
 
 # Docker Healthcheck
 HEALTHCHECK --interval=30s --timeout=30s --start-period=5s --retries=3 \
-	CMD wget --no-verbose --tries=1 --spider localhost:${PORT}/healthcheck || exit 1
+	CMD wget --no-verbose --tries=1 --spider localhost:${DEPENDENCY_DISCOVERY_PORT}/healthcheck || exit 1
 
 CMD ["node", "src/server.js"]
--- a/src/api/image/Dockerfile
+++ b/src/api/image/Dockerfile
@@ -1,6 +1,6 @@
 ## Base ###########################################################################
 # Set up the base layer
-FROM node:16-alpine3.15 as base
+FROM node:16 as base
 
 RUN npm i -g pnpm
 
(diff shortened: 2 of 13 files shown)


humphd:
I need to fix the `image` service to do a proper build:

```
/node_modules/.pnpm/sharp@0.30.2/node_modules/sharp/lib/sharp.js:31

  throw new Error(help.join('\n'));

  ^


Error: 

Something went wrong installing the "sharp" module


Cannot find module '../build/Release/sharp-linuxmusl-arm64v8.node'

Require stack:

- /node_modules/.pnpm/sharp@0.30.2/node_modules/sharp/lib/sharp.js

- /node_modules/.pnpm/sharp@0.30.2/node_modules/sharp/lib/constructor.js

- /node_modules/.pnpm/sharp@0.30.2/node_modules/sharp/lib/index.js

- /src/lib/image.js

- /src/routes/image.js

- /src/index.js

- /src/server.js


Possible solutions:

- Install with the --verbose flag and look for errors: "npm install --ignore-scripts=false --verbose sharp"

- Install for the current linuxmusl-arm64v8 runtime: "npm install --platform=linuxmusl --arch=arm64v8 sharp"

- Consult the installation documentation: https://sharp.pixelplumbing.com/install

    at Object.<anonymous> (/node_modules/.pnpm/sharp@0.30.2/node_modules/sharp/lib/sharp.js:31:9)

    at Module._compile (node:internal/modules/cjs/loader:1103:14)

    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1157:10)

    at Module.load (node:internal/modules/cjs/loader:981:32)

    at Function.Module._load (node:internal/modules/cjs/loader:822:12)

    at Module.require (node:internal/modules/cjs/loader:1005:19)

    at require (node:internal/modules/cjs/helpers:102:18)

    at Object.<anonymous> (/node_modules/.pnpm/sharp@0.30.2/node_modules/sharp/lib/constructor.js:8:1)

    at Module._compile (node:internal/modules/cjs/loader:1103:14)

    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1157:10)

**Review by Kevan-Y — APPROVED** [review]
LGTM, Tested on x86 system.

**Comment by Kevan-Y** [comment]
@humphd, Ignore my previous approval. I actually got an error.

Steps: `pnpm services:start`

telescope_api_planet
```
node:internal/modules/cjs/loader:936

  throw err;

  ^


Error: Cannot find module '/pnpm'

    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)

    at Function.Module._load (node:internal/modules/cjs/loader:778:27)

    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:77:12)

    at node:internal/main/run_main_module:17:47 {

  code: 'MODULE_NOT_FOUND',

  requireStack: []

}
```

**Review by Kevan-Y — CHANGES_REQUESTED** [review]
For `deploy` stage let's set a `WORKDIR` so our code doesn't go into the root

**Comment by humphd** [comment]
That error is #3347, which we need to fix.

File: src/api/image/Dockerfile [hunk 1]
```diff
@@ -1,6 +1,6 @@
 ## Base ###########################################################################
 # Set up the base layer
-FROM node:16-alpine3.15 as base
+FROM node:16 as base
```
DukeManh, line 3: Could we fix the issue by specifying the architecture on which the images will run?
```suggestion
FROM --platform=linux/amd64 node:16-alpine3.15 as base
```

humphd: If I do this, I'm forced to use `amd64` vs. `arm64` on my Mac.

DukeManh: How is the build time on your M1?

humphd: It's good.  But if I have to emulate twice it's not ideal (Docker is already in Linux and then if I have to add amd64/x86 vs. arm64, that slows things down at runtime).

manekenpix (commented):
Tested on my ARM laptop (Linux). Built successfully, but got [Kevan's error](https://github.com/Seneca-CDOT/telescope/pull/3336#issuecomment-1080757280).

humphd:
Rebased to pick up the fix for the planet `pnpm` Dockerfile bug.

src/api/dependency-discovery/Dockerfile
```diff
@@ -16,18 +16,18 @@ RUN pnpm install --prod
 
 ## Deploy ########################################################################
 # Stage for running our app
-FROM base as deploy
+FROM node:16-alpine3.15 as deploy
```
Kevan-Y, line 20: Add a `WORKDIR <DIRECTORY>`, instead of having everything in the root.
The same thing applies to the rest of Dockerfile.

Kevan-Y (changes requested):

menghif (dismissed):
Tested on M1

src/api/dependency-discovery/Dockerfile
Kevan-Y: So in here https://github.com/humphd/telescope/blob/3f6af21e4dd3a968e315e408c5e60719026aa5c3/src/api/feed-discovery/Dockerfile#L19-L25
The code will be copied into the root folder instead of being in a folder that holds our app.

We can change into 
```dockerfile
WORKDIR /app

COPY --chown=node:node . .

COPY --chown=node:node --from=dependencies /app/node_modules ./node_modules
```
or 
```dockerfile

COPY --chown=node:node . /app

COPY --chown=node:node --from=dependencies /app/node_modules /app/node_modules
```

humphd:
Fixed `WORKDIR` issue, rebased.

Kevan-Y:
> Fixed `WORKDIR` issue, rebased.

@humphd, we have to do with all the other Dockerfile files too.

humphd:
> > Fixed `WORKDIR` issue, rebased.
> 
> @humphd, we have to do with all the other Dockerfile files too.

😢

humphd:
Fixed the rest.

Kevan-Y (approved):

menghif (dismissed):

PR #3361, opened 2022-03-30
docker-compose for Supabase in dev/prod

humphd:
We had a meeting today about things that need to happen for Supabase to get put into production. One of them was to fix how we do volumes for postgres in dev vs. staging/prod.

I've created override files for dev and prod environments, which disable ports and fix the location of persistent volumes we mount.

We need to do the `secrets.env` before we can land this, but I hope @manekenpix and I can do that soon.


TueeNguyen (approved):
Looks ok

Kevan-Y (approved):

humphd:
> We need to do the secrets.env before we can land this, but I hope @manekenpix and I can do that soon.

We can't merge this into master, we don't have secrets up yet.  Please back this out.

PR #3098, opened 2022-03-04
Add studio under our subdomain

Kevan-Y:
## Issue This PR Addresses

fixes #3033 

## Type of Change

- [ ] **Bugfix**: Change which fixes an issue
- [ ] **New Feature**: Change which adds functionality
- [ ] **Documentation Update**: Change which improves documentation
- [ ] **UI**: Change which improves UI

## Description
In https://github.com/Seneca-CDOT/telescope/issues/2964 we setup the DNS for supabase, now we need to use our nginx to re-routed into our subdomain.

## Steps to test the PR
Don't think it's testable on development, this is mostly staging/prod stuff.

## Checklist

- [ ] **Quality**: This PR builds and passes our npm test and works locally
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not (if applicable)
- [ ] **Documentation**: This PR includes updated/added documentation to user exposed functionality or configuration variables are added/changed or an explanation of why it does not(if applicable)

PR #3366, opened 2022-03-30
Rework configs, docker, envs, CI/CD for production secrets

humphd:
This is needed for #3098, #3361 (which @TueeNguyen needs to land manually again).

In this PR, I'm updating our code to use the new secrets @manekenpix and I are adding to staging and production.  I'm going to give the secrets to @manekenpix separate to this PR, for obvious reasons.  Here, I'm doing the following:

1. Update the `anon_key`s we use in staging and production.  These had to be regenerated, since I changed the `JWT_TOKEN` we use.
2. Added a note about how to generate new Supabase keys
3. Removed all uses of `SECRET`, which we don't use anymore (we use `JWT_SECRET`)
4. Removed from env vars that got added to Kong, which @Kevan-Y and I don't think are required.
5. Updated our autodeployment `deploy.sh` script to use two additional env files.  NOTE: these files do **not** live in git, and only exist on the servers.  @manekenpix, let me know if these paths make sense to you (i.e., one level higher than the repo):
    1. `SECRETS_ENV_FILE=../env.production.secrets` for production
    2. `SECRETS_ENV_FILE=../env.staging.secrets` for staging. 

I need to have @manekenpix install the new secrets files on staging/production **before** we land this (looking at you, @TueeNguyen).

### DIFF
--- a/.github/workflows/docker-build-and-push.yml
+++ b/.github/workflows/docker-build-and-push.yml
@@ -55,7 +55,7 @@ on:
       anon_key:
         description: 'The Supabase Anonymous Key'
         required: false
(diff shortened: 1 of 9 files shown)


humphd:
Updated the location of the new secrets env files to be in `../config/*`.  Also updated the docs.

manekenpix (approved):

cindyledev (approved):

PR #3370, opened 2022-03-31
Migrage planet feed list to supabase script

### LINKED ISSUE #3359: Write Postgres Seed script to popluate Feeds table in production and development
The script will use [/tools/migrate/migrate.js](https://github.com/Seneca-CDOT/telescope/blob/master/tools/migrate/migrate.js) and `supabase-js` API to insert feeds to a Table.

Ideally, we only need to run this script once in production. The production Supabase Studio is what we use to manage feeds from now on.

DukeManh:
Fixes #3359.

This modifies the current `Migrate` service to migrate the Planet Feed List to Supabase.

Ideally, we only need to run this once on production and start managing all feeds there.

Once the production database is populated, I think we can use it for development as well because the `feeds` table has RLS rules that allow reading to all users and deleting/updating for feed owners.

### DIFF
--- a/tools/migrate/feed.js
+++ b/tools/migrate/feed.js
@@ -1,10 +1,8 @@
 const fetch = require('node-fetch');
 const jsdom = require('jsdom');
-const fs = require('fs');
 
 const { JSDOM } = jsdom;
 const FEED_URL = 'https://wiki.cdot.senecacollege.ca/wiki/Planet_CDOT_Feed_List';
-const FILE = 'legacy_users.json';
 
 const getWikiText = async (url) => {
   try {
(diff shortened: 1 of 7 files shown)


humphd (approved):

tools/migrate/feed.js
```diff
@@ -38,30 +44,30 @@ const getWikiText = async (url) => {
...
-        // If the URL is valid, continue
+        // eslint-disable-next-line no-new
+        new URL(feed); // If the URL is valid, continue
         [firstName, lastName] = lines[index + 1].replace(/^\s*name\s*=\s*/, '').split(' ');
-        users.push({ firstName, lastName, feed });
+
+        if (!uniqueUrls.has(feed)) {
+          users.push({ firstName, lastName, feed });
+          uniqueUrls.add(feed);
+        }
```
TueeNguyen, line 62: Cool way to avoid duplicates

TueeNguyen (approved):
Tested, it works

humphd (approved):

PR #3372, opened 2022-03-31
Use single env-file in deploy.sh, append secrets instead

humphd:
Apparently we can't use multiple `--env-file` args to `docker-compose`.  This appends the secrets to the env file instead, like we do with a few other things.


TDDR (approved):
Makes sense to me. This is another one that needs to be tested in staging I assume?

manekenpix (approved):

PR #3379, opened 2022-04-01
Fix nginx config for oauth2-proxy

Kevan-Y:
## Description 
In the previous PR https://github.com/Seneca-CDOT/telescope/pull/3378, oauth2-proxy crashed the Nginx, because of `--force-https=true`.

## Changes:
  - Removed `--skip-auth-regex=/forms/*` deprecated and not needed 
  - Removed `--force-https=true` (Not needed, can use our Nginx to put the correct protocol)
  
  ### Based on https://oauth2-proxy.github.io/oauth2-proxy/docs/configuration/tls/
  - Set `--cookie-secure` to true, we want this to be secure 
  - Set `--reverse-proxy` to true, to tell this services is under a proxy
  - Removed `--redirect-url` since it will automatically, using our domain.

### DIFF
--- a/config/nginx.conf.template
+++ b/config/nginx.conf.template
@@ -172,6 +172,7 @@ http {
   }
 
   # Requests to the oauth2-proxy which redirect to supabase studio app once authenticated
+  # Configuration reference: https://oauth2-proxy.github.io/oauth2-proxy/docs/configuration/tls/
   server {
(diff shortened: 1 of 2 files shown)


humphd (approved):
Nice, thanks for researching.

TueeNguyen (approved):

PR #3384, opened 2022-04-02
Update how kong.yml is used to configure Supabase in dev vs. prod

### LINKED ISSUE #3381: Fix anon and service roles keys in kong.yml
We almost have Supabase working in production.  Next problem is that we have incorrect, hard-coded keys at https://github.com/Seneca-CDOT/telescope/blob/master/docker/supabase/volumes/api/kong.yml#L6-L12

We either need to move this file to staging and production and hard-code the proper keys there (probably easiest), or we need to figure out how to overwrite with variables from the env on startup.

@manekenpix if I prepare two files for you with the right keys, would you put them in `../config/kong.yml`?

humphd:
Yesterday @DukeManh, @Kevan-Y and I tried to insert all the feeds into Supabase on staging.  We found that our `service_role_key` wouldn't work, and even our `anon_key` won't pass verification.  The issue is that we are hard-coding the keys into our `kong.yml` config: https://github.com/Seneca-CDOT/telescope/blob/master/docker/supabase/volumes/api/kong.yml#L6-L12

These need to match the keys we use on staging and production.

I've created new `kong.yml` files for staging and production, and sent to @manekenpix to add to the servers.  Once those are in place, we can do this change.

@manekenpix, please confirm that the paths I have used for production match what you do with these files (I think I'm right, but it's untested).

Fixes #3381.

### DIFF
--- a/docker/supabase/supabase-production.yml
+++ b/docker/supabase/supabase-production.yml
@@ -8,3 +8,7 @@ services:
     volumes:
       - ../../../supabase/volumes/db/data:/var/lib/postgresql/data
       - ./supabase/volumes/db/init:/docker-entrypoint-initdb.d
+
+  kong:
+    volumes:
+      - ../../../config/kong.yml:/var/lib/kong/kong.yml
--- a/docker/supabase/supabase-production.yml
+++ b/docker/supabase/supabase-production.yml
@@ -2,13 +2,15 @@
 services:
   storage:
     volumes:
-      - ../../../supabase/volumes/storage:/var/lib/storage
+      - ../../supabase/volumes/storage:/var/lib/storage
 
   db:
     volumes:
-      - ../../../supabase/volumes/db/data:/var/lib/postgresql/data
+      # We keep the actual data in a volume outside of git, so it survives restarts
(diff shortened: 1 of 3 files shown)


docker/supabase/supabase-production.yml
```diff
@@ -8,3 +8,7 @@ services:
     volumes:
       - ../../../supabase/volumes/db/data:/var/lib/postgresql/data
       - ./supabase/volumes/db/init:/docker-entrypoint-initdb.d
+
+  kong:
+    volumes:
+      - ../../../config/kong.yml:/var/lib/kong/kong.yml
```
manekenpix, line 14:
```suggestion
      - ../../config/kong.yml:/var/lib/kong/kong.yml
```
and  `../../../supabase...` should be `../../supabase...`

manekenpix (changes requested):

humphd:
Fixed, thanks for catching.

manekenpix (approved):
LGTM :+1:

DukeManh (approved):

Kevan-Y (approved):

DukeManh:
It's working now https://dev.supabase.telescope.cdot.systems/project/default/editor/16681

PR #3437, opened 2022-04-08
Fix #3435 - Dockerize src/db

### LINKED ISSUE #3435: Containerize db migrations
In #3418, @DukeManh is adding automated database migrations via [Prisma Migrate](https://www.prisma.io/docs/concepts/components/prisma-migrate).  It will allow us to detect when the database schema is out of sync with `master`, and generate a migration to be applied.

Running this requires network access to the Postgres database, which [we have in local development](https://github.com/Seneca-CDOT/telescope/blob/master/docker/supabase/supabase-development.yml#L16), but not in production (i.e., we don't expose it at all).

Accessing the postgres database requires running our code **within** the Docker network.  This means running the code via a container that runs inside our network.  @JerryHue is doing something very similar with postgres backups in #3405.

Unlike the work that @JerryHue is doing, where we can run this automatically and all the time (i.e., we run backups as a `cron` job), migrations are a one-off action.  We won't need them very often.  I'd like to be able to run them via Portainer.  The process would look like this:

1. We create a container in `src/db` that contains all the prisma migrations and code necessary to run them
2. In CI for staging and production, we build and push images to our registry; however, we don't ever use these images in our `docker-compose`
3. When we need to run a migration (e.g., on staging), we ask Portainer to create a new container based on `docker.cdot.systems/migrations:staging` (or whatever we name it).  The same would happen for production (e.g., using`:production`).

This should mean that the migrations are always in sync with what's needed on the particular environment where we need to run them, since they are built from git on each push/tag.

Our first migrations are needed before 3.0 (see #3433), so we should be able to test this process, document it, and fix any bugs.

humphd:
This adds a `Dockerfile` to `src/db` so that we can run the Prisma migrations on staging and production via Portainer in a new container.  This will build and push a new image every time we push to `master` or create a new `tag` for production.

On Portainer, we would create a new container, and specify the correct `DATABASE_URL` environment variable to use for the given environment (staging/production):

Screen Shot 2022-04-08 at 3 39 49 PM

This needs to be tested, but it's a start.

### DIFF
--- a/.github/workflows/docker-build-and-push.yml
+++ b/.github/workflows/docker-build-and-push.yml
@@ -109,6 +109,8 @@ jobs:
             image: search
           - context: src/api/sso
             image: sso
+          - context: src/db
+            image: db-migrations
     steps:
       - uses: actions/checkout@v2
 
--- a/src/db/.dockerignore
+++ b/src/db/.dockerignore
@@ -1,0 +1,2 @@
+node_modules/
+env.*
--- a/src/db/Dockerfile
+++ b/src/db/Dockerfile
@@ -1,0 +1,9 @@
+# NOTE: we assume that the environment variable DATABASE_URL will
+# be defined when a new container is run.  It should take the form:
+# DATABASE_URL=postgresql://postgres:postgress_password@localhost/postgres
+FROM node:16-alpine3.15
+
+WORKDIR /db-migrations
+COPY . .
+RUN npm install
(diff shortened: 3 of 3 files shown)


humphd:
Fixes #3435

DukeManh (approved):

joelazwar (approved):

humphd:
I'll land this and we can give it a shot next week on a call together, then fix bugs we find.

Issue #3350, opened 2022-03-29
Cloud Formation template for Telescope development on AWS


Kevan-Y:
Recently, I looked into a tool called [Terraform](https://www.terraform.io/), an Infrastructure as Code. 
While reading the document to use [vscode ssh](https://github.com/Seneca-CDOT/telescope/blob/master/src/docs/docs/getting-started/vscode-ssh.md), I think it would be a nice plus to have a terraform file to build all the AWS instances, so we can skip the AWS step and only do the ssh connection with vscode.

humphd:
I'd like this, we need it.  I think Cloud Formation might be better, though.  Not because CF is better (I prefer Terraform), but because it will be easier to run in the context of the Learner Lab environment.

I'm going to be teaching Cloud Formation in a few weeks in the cloud course.  You can write them in JSON or YAML, and they look like this:

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/quickref-ec2.html

We would transform @cindyledev's doc into a template that Cloud Formation runs in order to create the actual resources.  It's magic.

Kevan-Y:
@humphd I have a working terraform file in my local, that run [Vscode ssh](https://github.com/Seneca-CDOT/telescope/blob/master/src/docs/docs/getting-started/vscode-ssh.md) steps but missing elastic IP.

humphd:
Ah, nice.  You don't find it clunky to run with the Learner Lab and changing credentials?

Kevan-Y:
> Ah, nice. You don't find it clunky to run with the Learner Lab and changing credentials?

I don't very mind it. It's a few seconds of copying and pasting keys.

I think one thing you need is replace those
```tf
provider "aws" {
  region     = "us-east-1"
  access_key = <ACCESS_TOKEN>
  secret_key = <SECRET_KEY>
  token      =  <TOKEN>
  }
  ```
 If first time you need setup ssh key
 ```tf
 resource "aws_key_pair" "deployer" {
  key_name   = "telescope-ssh-key"
  public_key = <PUBLIC_SSH_KEY>
}
```
Then to run it 
```bash
terraform plan // Checking everything is correct
terraform apply // Apply our change to AWS
```

humphd:
The thing about using Cloud Formation is that we could do it in AWS and not worry about passing around keys, credentials at all.

We can do both, too!

Kevan-Y:
[...]

PR #3455, opened 2022-04-10
Create CloudFormation template

humphd:
Part of #3350.  This adds a CloudFormation Template that will produce the same AWS resources as our manual guide.

### DIFF
--- a/config/ec2-template.yml
+++ b/config/ec2-template.yml
@@ -1,0 +1,89 @@
+AWSTemplateFormatVersion: 2010-09-09
+
+# https://telescope.cdot.systems/docs/getting-started/vscode-ssh
+Description: Telescope EC2 VSCode SSH Development Environment
+
+Parameters:
+  ImageId:
+    Type: AWS::EC2::Image::Id
+    Default: ami-0c02fb55956c7d316
+    Description: EC2 Image ID, defaults to Amazon Linux 2 AMI (HVM) - Kernel 5.10, SSD Volume Type - ami-0c02fb55956c7d316 (64-bit x86)
+
+  KeyName:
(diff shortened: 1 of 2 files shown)


TDDR:
How would one test this out?

rclee91 (approved):
Here's how I tested this
- Have the `ec2-template.yml` ready somewhere on your device
- Boot up AWS learner lab, go to CloudFormation > create stack with new resources
- In `Specify template`  -> `Upload a template file` -> upload the `ec2-template.yml` file
- On the `next` page, give it a stack name, enter your IP
- `Next` -> `Next` -> `Create stack` 
Wait for a bit and everything will be created
image
image
image


Then I think the idea is you can skip a lot of setup instructions and jump directly to [Connect using SSH](https://github.com/Seneca-CDOT/telescope/blob/82f16594caa2d81442accd2a891a3d514f2ff39b/src/web/docusaurus/docs/getting-started/vscode-ssh.md#connect-using-ssh)

I ran into an error, but I think its more of my own local configuration problem, and not really a part of this PR, cause this PR seems to set up the EC2 instance just fine
image

Once you're done using the instance, remember to `DELETE` the stack in CloudFormation, so you don't have to keep paying for the resources. Just use the `ec2-template.yml` file to create the stack again the next time you want to use it.
image

Here is the estimated cost of keeping it running for a whole month, in case anyone is interested.
image

Kevan-Y (approved):
Tested it works well.
image

Kevan-Y:
@rclee91, maybe during ssh setup, you miss some steps?
I want able to connect with ssh on vscode

aserputov:
@humphd should we merge this one ?

aserputov:
@humphd should we merge this one ?

PR #3469, opened 2022-04-12
Add URL hash id to Feeds table

### LINKED ISSUE #3462: Use the hash URL id for feeds in Postgres schema

DukeManh:
Fix #3462

Added migrations to add a URL `hash` column generated by Satellite to the Feeds table.

To add a field to the existing DB, we'll need:

1. First migration to add the field in the table with a default value of `""`
2. Run `to_supabase` script to add hash ids for each row
3. Second migration to drop the default hash id to make it mandatory

### DIFF
--- a/tools/migrate/to_supabase.js
+++ b/tools/migrate/to_supabase.js
@@ -1,4 +1,6 @@
 const { createClient } = require('@supabase/supabase-js');
+const { hash } = require('@senecacdot/satellite');
+const normalizeUrl = require('normalize-url');
 const { parsePlanetFeedList } = require('./feed');
 
 require('dotenv').config();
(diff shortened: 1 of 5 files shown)


tools/migrate/to_supabase.js
```diff
@@ -19,6 +20,7 @@ const { SUPABASE_URL, SERVICE_ROLE_KEY } = process.env;
     url: feed,
     wiki_author_name: `${firstName} ${lastName}`,
     type: 'blog',
+    hash: hash(feed),
```
TueeNguyen, line 23: We should normalize the feed url before hashing so that it's consistent with `feedId` in `Redis`
```js
const normalizeUrl = require('normalize-url');
hash: hash(normalizeUrl(feed))
```

humphd (changes requested):

TueeNguyen (approved):

humphd (approved):

JerryHue (approved):

joelazwar (approved):

PR #3470, opened 2022-04-12
Add TypeScript types to github-url-parser for front-end use

humphd:
I want to use the new github-url-parser module in the front-end, but it needs types defined. This adds type definitions for what I need.

See steps I followed in https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html.

### DIFF
--- a/src/github-url-parser/dist/extract-github-info.d.ts
+++ b/src/github-url-parser/dist/extract-github-info.d.ts
@@ -1,0 +1,9 @@
+declare function _exports(gitHubUrls: any): {
+    repos: any[];
+    issues: any[];
+    pullRequests: any[];
+    commits: any[];
+    users: any[];
+};
+export = _exports;
+//# sourceMappingURL=extract-github-info.d.ts.map
--- a/src/github-url-parser/dist/index.d.ts
+++ b/src/github-url-parser/dist/index.d.ts
@@ -1,0 +1,10 @@
+declare function _exports(htmlString: string, parser?: DOMParser): GitHubInfo;
+export = _exports;
+export type GitHubInfo = {
+    issues: string[];
+    pullRequests: string[];
+    repos: string[];
+    commits: string[];
+    users: string[];
+};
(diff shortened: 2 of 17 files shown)


menghif (approved):

DukeManh (changes requested):

src/github-url-parser/dist/extract-github-info.d.ts
```diff
@@ -0,0 +1,9 @@
+declare function _exports(gitHubUrls: any): {
+    repos: any[];
```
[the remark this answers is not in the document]
humphd: Why on earth would it ignore my typedef?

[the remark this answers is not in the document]
humphd: I fixed it

src/github-url-parser/dist/index.d.ts
```diff
@@ -0,0 +1,10 @@
+declare function _exports(htmlString: string, parser?: DOMParser): GitHubInfo;
```
humphd, line 1: This is the only one that matters.

src/github-url-parser/dist/reserved-names.d.ts
```diff
@@ -0,0 +1,57 @@
+declare const _exports: {
```
DukeManh, line 1: Won't matter much but if you declare the reserved names export in `reserved-names.js` as `/** @type{string[]} */`, it would not export these array function types and save a few bytes.

DukeManh (changes requested):

src/github-url-parser/dist/reserved-names.d.ts
humphd: Good point, fixing

humphd:
Updated, ready to re-review

DukeManh (approved):
_strictly approved_

tcvan0707 (approved):

PR #3477, opened 2022-04-12
Use git sha as token for Turbo Repo (no secrets)

humphd:
New plan: I'm not going to use a secret here, because I can't get it to pass through to the linked action.  The `server-token` just has to be a unique string per-run, so the `github.sha` is perfect.

### DIFF
--- a/.github/workflows/eslint-ci.yml
+++ b/.github/workflows/eslint-ci.yml
@@ -2,10 +2,6 @@ name: ESLint Workflow
 
 on:
   workflow_call:
-    secrets:
-      turbo_server_token:
-        description: 'The Turborepo local server token'
-        required: true
 
 jobs:
   # Make sure eslint passes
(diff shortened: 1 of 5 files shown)


.github/workflows/eslint-ci.yml
```diff
@@ -2,18 +2,15 @@ name: ESLint Workflow
...
 
 jobs:
   # Make sure eslint passes
   eslint:
     name: ESLint Check
     runs-on: ubuntu-latest
     env:
-      SERVER_TOKEN: ${{ secrets.turbo_server_token }}
+      # Use the git sha as a unique token for Turbo Repo
+      SERVER_TOKEN: ${{ github.sha }}
```
menghif, line 13: Since we're using `github.token` directly for `repo-token`, can we do the same with `github.sha`? Without passing it as env

humphd: We can, it's just a way to document it more clearly.

menghif (approved):

PR #3481, opened 2022-04-13
Use turbo repo to cache build artifacts

humphd:
This changes our `build` task to use Turbo Repo to cache build artifacts.  Locally, this means my second `pnpm build` runs in under 1 second:

```
 Tasks:    4 successful, 4 total
Cached:    4 cached, 4 total
  Time:    221ms >>> FULL TURBO
```

I'm following the pattern in https://turborepo.org/docs/features/caching#alter-caching-based-on-environment-variables-and-files to list separate `outputs` for different `build` steps.

As a side note, it would be cool if we can somehow use these cached resources in Actions within a Docker build too.  We'd have to find a way to run the cache server so it was accessible to the Docker engine.

NOTE: this also changed the permissions on the files in `dist/` for the github url parser, which is fine.


menghif (commented):
Why isn't `src/github-url-parser/dist` in gitignore?

humphd:
These are the types we need to export for TypeScript, so we can't ignore.

menghif (approved):
Tested locally. Cache works

tcvan0707 (approved):

PR #3483, opened 2022-04-13
Pass 'SUPABASE_URL' to SSO dev docker

DukeManh:
I'm testing the signup flow using staging supabase. I'd say it works as expected. You can see my test account created on Staging here https://dev.supabase.telescope.cdot.systems/project/default/editor/17293.

I notice `SUPABASE_URL` is hardcoded in SSO docker ENV.

### DIFF
--- a/docker/development.yml
+++ b/docker/development.yml
@@ -125,7 +125,7 @@ services:
     environment:
       # In development and testing, the SSO service needs to contact the Supabase
       # service directly via Docker vs through the http://localhost/v1/supabase domain.
-      - SUPABASE_URL=http://kong:8000
+      - SUPABASE_URL=${SUPABASE_URL:-http://kong:8000}
     depends_on:
       - test-web-content
       - traefik


docker/development.yml
```diff
@@ -125,7 +125,7 @@ services:
     environment:
       # In development and testing, the SSO service needs to contact the Supabase
       # service directly via Docker vs through the http://localhost/v1/supabase domain.
-      - SUPABASE_URL=http://kong:8000
+      - SUPABASE_URL=${SUPABASE_URL}
```
humphd, line 128: Make this:

```yml
- SUPABASE_URL=${SUPABASE_URL:-http://kong:8000}
```

That way you can override it, but it defaults to what we want.

humphd (changes requested):

docker/development.yml
tcvan0707: What does `:-` just before **http** mean?   Can we only specify just `:`?

humphd: Good question. See https://docs.docker.com/compose/compose-file/compose-file-v3/#variable-substitution:

> Both $VARIABLE and ${VARIABLE} syntax are supported. Additionally when using the [2.1 file format](https://docs.docker.com/compose/compose-file/compose-versioning/#version-21), it is possible to provide inline default values using typical shell syntax:
>
>${VARIABLE:-default} evaluates to default if VARIABLE is unset or empty in the environment.
>${VARIABLE-default} evaluates to default only if VARIABLE is unset in the environment.
>Similarly, the following syntax allows you to specify mandatory variables:
>
>${VARIABLE:?err} exits with an error message containing err if VARIABLE is unset or empty in the environment.
>${VARIABLE?err} exits with an error message containing err if VARIABLE is unset in the environment.

DukeManh: Thank you, fixed.

humphd (changes requested):
This is failing due to the usual `localhost` Docker problem, since we're now accepting the value from https://github.com/Seneca-CDOT/telescope/blob/master/config/env.development#L285.  Can you help me understand what this fix is actually trying to accomplish, and why it's needed?

DukeManh:
Because I wanted to test our staging Supabase versus the local one and updating env var doesn't update SSO docker. 

I can close it if we have issues with tests, and change SUPABEBASE_URL in development.yml when I need to.

humphd:
I think this is going to cause more problems than it solves.  We should probably file an issue to make it easy to run your local dev environment against staging (except for various services you want to use locally).

DukeManh:
I agree, closing

PR #3485, opened 2022-04-14
Replaced `octokit/request` with `octokit/octokit` in dependency-discovery

### LINKED ISSUE #3324: Use `octokit/octokit` instead of `octokit/request`
**What would you like to be added**: We are currently using `octokit/request` library to request the GitHub API in the `dependency-discovery` service. Instead of using that library, we could use `octokit/octokit`.

**Why would you like this to be added**: The `octokit/octokit` does more stuff that we can use to simplify some parts of our service.

TDDR:
## Issue This PR Addresses

Fixes #3324 

## Type of Change

## Description

This PR adds the `octokit/octokit` package to the `dependency-discovery` and removes `octokit/request`.

Could also update the usage of `octokit/request` in the [src/web/.../StarField](https://github.com/Seneca-CDOT/telescope/blob/6046273fba4b89ce6281f7a34cafbdec7ef222dc/src/web/app/src/components/StarField.tsx#L3) and the octokit/webhook usage [autodeployment/webhook.js](https://github.com/Seneca-CDOT/telescope/blob/33858ecb2b27f86c59b710a84cc9f14ec51f4376/tools/autodeployment/webhook.js#L2) in a follow up PR?

## Steps to test the PR

Not 100% sure how to test the `dependency-discovery`, but the image properly builds and localhost seems to be working as intended.


JerryHue (dismissed):
There's nothing much to test, since the tests themselves run just fine, which means that the behaviour has kept the same. We would extend what the service does with what the library offers us in a future PR!

src/api/dependency-discovery/src/github.js
```diff
@@ -11,9 +10,10 @@ async function constructGitHubRequest(gitHubUrl) {
   const labels = ['hacktoberfest', 'good first issue', 'help wanted'];
 
   const [owner, repo] = new URL(gitHubUrl).pathname.substring(1).split('/');
+  const octokit = new Octokit();
```
humphd, line 13: Can you add some plugins here like we do in `src/api/status/src/js/octokit.js` for throttling and abuse protection?

humphd (changes requested):

Kevan-Y:
@TDDR, to test it locally:
1. ` cd .\src\api\dependency-discovery\`
2. `pnm dev` or `pnpm start`

Kevan-Y:
@TDDR Are you gonna update the other in a follow up PR?

src/api/dependency-discovery/src/github.js
TDDR: I added the throttling here. I could not see if `octokit-plugin/retry` was included in [`octokit`](https://www.npmjs.com/package/octokit), but the `octokit-plugin/throttling` was in the documentation which leads me to believe `retry` is as well.

Started it all up and it seems to work, not sure how to test `throttling` or `retry` though

TDDR:
> @TDDR Are you gonna update the other in a follow up PR?

I also updated the usage of `octokit/core` and the two plugins to be just [`octokit`](https://www.npmjs.com/package/octokit) in `src/api/status/src/js/octokit.js`

There is one more usage of `octokit/request` in the `Starfield`, but I don't think that needs the extra capabilities of the full `octokit` package

JiaHua-Zou (approved):

humphd (approved):

PR #3488, opened 2022-04-14
Refactor feed-discovery, sso to handle blog, youtube, twitch URLs

humphd:
The sign-up flow needs to be able to handle multiple feed URLs, and different types of feeds (YouTube, Twitch, or regular "blogs").  This PR changes how we process feeds to include `type`, and also to allow working with multiple URLs.  Specifically:

- Adds the `RSS_BRIDGE_URL` to the configs and passes to the `feed-discovery` service so we can use it directly to construct a Twitch feed URL
- Refactors the `feed-discovery` service to work with multiple URLs: we now expect an `Array` vs. a single `blogUrl` in the POST.
- Refactors the `feed-discovery` to special case Twitch and YouTube.  In the Twitch case, we construct a URL to the RSS Bridge container and return that.  In the YouTube case we parse the channel HTML for a feed, and set the type.
- Updated and added tests to the `feed-discovery`
- Fixed the `sso` service to understand how to work with a feed object that looks like: `{ feedUrl, type }` vs. just a URL, and updated tests.
- Updated the front-end web app to allow a space-separated list of URLs to get passed to the `feed-discovery` service

I wrote the code on a call with @tcvan0707 and @menghif, so they can review it.  They will build on this to get the front-end sign-up flow implemented.  To do that, they will do this:

1. The sign-up form will allow entering a list of URLs, for example: `https://my.blog.com/ https://youtube.com/my-channel https://twitch.tv/me`.  Here we have 3 links to "blogs" I want to include
2. The front-end sends this to the `feed-discovery` and we need to fix how we deal with the response: we currently assume it's a list of URLs (strings) but it will be in the form `[{ feedUrl, type}, ...]`.


src/api/feed-discovery/test/router.test.js
```diff
@@ -33,11 +38,146 @@ describe('POST /', () => {
...
+      const invalidUrl = 'invalidlink.com';
+
+      const res = await request(app)
+        .post('/')
+        .set('Authorization', `bearer ${createServiceToken()}`)
+        .send([invalidUrl]);
+      expect(res.status).toBe(400);
+    });
+
+    it('should return 400 if the blog body is not parsable"', async () => {
```
menghif, line 110: There's an extra `"` after parsable.

menghif (approved):
I went through everything again. Looks good to me

src/api/feed-discovery/test/router.test.js
humphd: Fixed.  Thanks for reviewing.

menghif (approved):

src/web/app/src/components/SignUp/Forms/RSSFeeds.tsx
```diff
@@ -158,16 +158,16 @@ const RSSFeeds = connect<{}, SignUpForm>((props) => {
...
         signal: controllerRef.current?.signal,
         method: 'post',
         headers: {
           Authorization: `bearer ${token}`,
           'Content-Type': 'application/json',
         },
-        body: JSON.stringify({
-          blogUrl: values.blogUrl,
-        }),
+        body: JSON.stringify(urls),
```
DukeManh, line 170: The response is `{feedUrl, type}` now so in line 178 you want to map this to an array of URL strings so React can render it.

DukeManh (changes requested):
I get this response from Twich feed 
image

src/web/app/src/components/SignUp/Forms/RSSFeeds.tsx
DukeManh: It's even better to add type to the response so TS can catch this early.

humphd: What does "add type to the response" mean?

DukeManh: I mean `const res: {feedUrls: [{feed:string, type: url}]} = await response.json();`

humphd: Question: if I map this to extract the string URL values, we lose all the types (youtube, blog, twitch).  We probably need a larger fix here to maintain that data.  Should that happen in this PR, or a follow-up?  If this one, I need some guidance so it works with the formik stuff.

DukeManh: The change for displaying the type can be in follow-up, for now, we need to map it to string because React can't display JSON

menghif: I tried to make the change to line 178 like this:

```tsx
setFieldValue(
  'allFeeds',
  res.feedUrls.map((feed: { feedUrl: URL; type: string }) => {
    return feed.feedUrl;
  })
);
```
However I get a 404 when try to submit the data to the database.

humphd: I've added the types and mapped the `feedUrl`.  We need to deal with the `type` in a follow up.

DukeManh (approved):

tcvan0707 (approved):

PR #3489, opened 2022-04-14
Create Terraform template

Kevan-Y:
Part of #3350

Added Terraform to create our AWS resources from the CLI, it's the same resources from our [manual guide](https://github.com/Seneca-CDOT/telescope/blob/master/src/web/docusaurus/docs/getting-started/vscode-ssh.md).

### DIFF
--- a/config/terraform/main.tf
+++ b/config/terraform/main.tf
@@ -1,0 +1,103 @@
+// Ressources providers
+terraform {
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = "~> 3.27"
+    }
+    http = {
+      source  = "hashicorp/http"
+      version = "2.1.0"
+    }
+  }
+
+  required_version = ">= 0.14.9"
+}
+
+// Configure AWS provider
+provider "aws" {
+  region     = var.region
+  access_key = var.access_key
+  secret_key = var.secret_key
+  token      = var.access_token
+}
+
+// Get IP with amazonaws checkip
+data "http" "my_checkip" {
+  url = "https://checkip.amazonaws.com/"
+}
+
+// Declare a local variable and change our my_checkip into ipv4
+locals {
+  my_public_ipv4 = "${chomp(data.http.my_checkip.body)}/32"
+}
+
+// Configure aws_instance (see https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/instance)
+resource "aws_instance" "app_server" {
+  ami                  = "ami-0c02fb55956c7d316"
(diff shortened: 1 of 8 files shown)


sirinoks:
How does one test this?

sirinoks:
How does one test/check this?

Kevan-Y:
@sirinoks I have a docs `src/web/docusaurus/docs/getting-started/terraform.md` in this PR. You can follow this guide.

config/terraform/main.tf
```diff
@@ -0,0 +1,113 @@
...
+    from_port        = 22
+    ipv6_cidr_blocks = []
+    prefix_list_ids  = []
+    protocol         = "tcp"
+    security_groups  = []
+    self             = false
+    to_port          = 22
+    }, {
+    cidr_blocks      = [local.my_public_ipv4]
+    description      = "Open port 3000"
```
humphd, line 88: I don't think we need port 3000.

I kind of wonder if we can do everything via :80 on EC2.

src/web/docusaurus/docs/getting-started/terraform.md
```diff
@@ -0,0 +1,109 @@
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# Setup resources with Terraform
```
humphd, line 4: I think this needs a note about getting an AWS account in the Learner Lab.  It also needs some warning about costs, since this will cost a lot if you start an r5.large and leave it.

humphd (changes requested):

src/web/docusaurus/docs/getting-started/terraform.md
AmasiaNalbandian: I agree, we can leave a note saying that there are two ways to stop the EC2 instance: 

using the aws CLI in learner lab ` aws ec2 stop-instances --instance-ids {instance-id}`
 
or manually go into the EC2 service in AWS learner lab and select the ec2 instance, choose instance dropdown and stop instance.

I'm wondering if this can go under destroy your resources?

config/terraform/main.tf
Kevan-Y: Maybe we can if we create a new config for Nginx. But not sure how we gonna handle the subdomain part. 
Possible change the route to use `domain/api/v1....` would work

Kevan-Y:
- Add a caution in `terrafor.md`.
- Removed port 3000, but left 8000. Probably in the future, we need to check if we can run everything in port 80 in development.

PR #3501, opened 2022-04-16
Add Telescope Developer login account and blog for local dev

humphd:
We need some data to use locally with the parser service, database, login flow, etc.  This adds a few things:

1. A new user that you can login to the `login` service with.  We need to follow this up with an easy way on startup to add this user to the db
2. A "blog" at http://localhost:8888/telescope-developer-blog/index.html with a feed at http://localhost:8888/telescope-developer-blog/feed.xml that we can use to register with the parser.

We need to figure out how to add this content to the database on startup, and also how to deal with the http://localhost issue in docker (e.g., parser won't be able to download a blog from localhost:8888).


TueeNguyen (dismissed):
Not sure about the `saml php` file but the rest looks ok

DukeManh (dismissed):
I think we should have this example account for local dev while figuring out how to automate the steps.

PR #3627, opened 2022-05-30
[React Native] Change default environment file

Kevan-Y:
Our latest mobile app build doesn't have any `env` so it cannot fetch anything.

Fix:
- Change default environment file to use `.env.development`

### DIFF
--- a/src/mobile/app.config.js
+++ b/src/mobile/app.config.js
@@ -1,4 +1,6 @@
-import 'dotenv/config';
+import * as dotenv from 'dotenv';
+
+dotenv.config({ path: './.env.development' });
 
 export default {
   expo: {


manekenpix:
@Kevan-Y  Would it make sense to tell expo to use `.env.development` instead of creating `.env`?

Kevan-Y:
@manekenpix That solution would work too.

manekenpix (approved):

humphd:
It looks like I can only add Secrets to Expo, not ENV VARS like next.js.  Let me know if you want some.

humphd (approved):

Kevan-Y:
Let's merge it and see if that works. I never used expo env after deployment so I'm not sure how it works.

PR #3663, opened 2022-09-28
Get correct display_name for feeds

### LINKED ISSUE #3660: Telescope Sign-Up creates profile and feed in Supabase, but nothing in timeline
I was just talking with Rohan Kaicker in Slack, and there's a problem with Telescope Sign-Up or parser.

A week ago, Rohan did the Sign-Up flow, but there's nothing in the timeline.  I looked in Supabase for prod, and here is what I see:

1. a record was added to `telescope_profiles` with the correct info, including the right GitHub info
2. a record was added to `feeds` with an `id` of `735016f481`

So it looks like things worked on the Sign-Up side.  However, trying to get info about the feed gives a 404:

https://api.telescope.cdot.systems/v1/posts/feeds/735016f481

We need to figure out where the disconnect is happening here: that feed should be queryable.

RC-Lee:
## Issue This PR Addresses
Fixes #3660 

## Type of Change

## Description

I used my old pal `console.log` to see the structure of the feeds that we were pulling from the db.
Turns out there wasn't any `display_name` property with each `feed` object.
```terminal
data structure is: 
 [
  {
    wiki_author_name: 'David Humphrey',
    url: 'http://blog.humphd.org/tag/seneca/rss',
    telescope_profiles: null
  },
  {
    wiki_author_name: null,
    url: 'https://dev.to/feed/rclee',
    telescope_profiles: { display_name: 'Roxanne Lee' }
  }
]
```

## Steps to test the PR
Testing locally is a bit complicated.
1. I changed [src/db/seed/feeds.json](https://github.com/Seneca-CDOT/telescope/blob/master/src/db/seed/feeds.json) to have only 1 feed (for simple testing purposes)
```js
[
  {
    "url": "http://blog.humphd.org/tag/seneca/rss",
    "user_id": null,
    "wiki_author_name": "David Humphrey",
    "html_url": null,
    "type": "blog",
    "invalid": false,
    "flagged": false,
    "id": "a2029d1f3e"
  }
]
```
2. After deleting any redis cache and doing db migration, I went to http://localhost:8910/ ,where the local db console was, to manually add myself a Telescope profile, and linked it to a Feed, making sure I kept `wiki_author_name` as `NULL`

3. Tested to make sure that before the fix my blogs wouldn't show up. (They did not)

4. Tested after the fix, and made sure my blogs showed up (They did)

We won't be able to see the fix on staging cause staging has nothing in the `telescope_profiles` table

## Checklist

### DIFF
--- a/src/api/parser/src/utils/supabase.js
+++ b/src/api/parser/src/utils/supabase.js
@@ -25,7 +25,7 @@ module.exports = {
 
     return data.map((feed) => ({
       // Prefer the a user's display name if present, fallback to wiki name otherwise
-      author: feed.display_name || feed.wiki_author_name,
+      author: feed.telescope_profiles?.display_name || feed.wiki_author_name,
       url: feed.url,
     }));
   },


src/api/parser/src/utils/supabase.js
```diff
@@ -25,7 +25,10 @@ module.exports = {
 
     return data.map((feed) => ({
       // Prefer the a user's display name if present, fallback to wiki name otherwise
-      author: feed.display_name || feed.wiki_author_name,
+      author:
```
humphd, line 28: Let's use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) to simplify this a bit:

```js
author: feed.telescope_profiles?.display_name || feed.wiki_author_name,
```

humphd (changes requested):
Amazing, thank you!

humphd (dismissed):

RC-Lee:
I was wondering if we could wait a few days until hacktoberfest kicks off,
But its also fine if we want to get this in asap.

humphd:
I'm fine to wait 🎃

DukeManh (dismissed):
This is the correct way to get `display_name`, I also see it in `src/api/planet/src/util.js`

https://github.com/Seneca-CDOT/telescope/blob/efff34f3bb983c8e43bdcfdf05602bc471019712/src/api/planet/src/util.js#L88

humphd:
@RC-Lee are you willing to fix the planet code at the same time?

humphd (approved):

PR #3683, opened 2022-10-06
use module aliases vs relative paths

DukeManh:
When the project grows deeper, more `../` are required in the relative import paths to import modules so this configs TS path aliases, that map to certain absolute paths.
 

## Type of Change

## Checklist

### DIFF
--- a/src/web/app/tsconfig.json
+++ b/src/web/app/tsconfig.json
@@ -13,7 +13,17 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "jsx": "preserve",
-    "incremental": true
+    "incremental": true,
+    "baseUrl": "./src",
+    "paths": {
+      "@components/*": ["components/*"],
+      "@hooks/*": ["hooks/*"],
+      "@styles/*": ["styles/*"],
+      "@theme": ["theme/index"],
+      "@translations/*": ["translations/*"],
+      "@config": ["config"],
+      "@interfaces": ["interfaces/index"]
+    }
   },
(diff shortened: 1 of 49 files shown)


RC-Lee:
Search tests will fail while I'm updating ES

humphd (approved):
Nice improvement

src/web/app/tsconfig.json
```diff
@@ -13,7 +13,17 @@
...
+    "baseUrl": "./src",
+    "paths": {
+      "@components/*": ["components/*"],
+      "@hooks/*": ["hooks/*"],
+      "@styles/*": ["styles/*"],
+      "@theme": ["theme/index"],
+      "@translations/*": ["translations/*"],
+      "@config": ["config"],
+      "@interfaces": ["interfaces/index"]
+    }
```
Kevan-Y, line 26: (nit): I personally prefer using like `@/components/*` instead of `@components/*`. It doesn't get confused if we have a package that `@components`

Kevan-Y (approved):

src/web/app/tsconfig.json
cindyledev: I agree with @Kevan-Y here. I also use the `@/components/*` for module aliases

DukeManh:
We'd need to fix #3690 first.

humphd:
@DukeManh do you want this in before tonight (end of Hacktoberfest)?  If so, what does it need?

dadolhay (approved):
This looks good!

The only nit-picky thing that might be considered, that some projects prefer the scoped package inports to be between the "traditional" library imports and the relative imports for readability reasons.

PR #3685, opened 2022-10-06
Deal with missing article guid given as [Object]

### LINKED ISSUE #3648: ERR_INVALID_ARG_TYPE in parser service with data argument
While debugging #3647, I notice the parser service doing this a bunch:

```
{"level":50,"time":1662478418654,"pid":7,"hostname":"0018e3493897","error":{"message":"The \"data\" argument must be of type string or an instance of Buffer, TypedArray, or DataView. Received an instance of Object","code":"ERR_INVALID_ARG_TYPE"},"msg":"Job 673aa2e5a9 failed."}
{"level":30,"time":1662478418671,"pid":7,"hostname":"0018e3493897","msg":"Invalidating feed http://dtheosab.wordpress.com/category/open-source/feed/ for the following reason: The \"data\" argument must be of type string or an instance of Buffer, TypedArray, or DataView. Received an instance of Object"}
```

I note that https://dtheosab.wordpress.com/category/open-source/feed/ does indeed have content, so this is an odd error, maybe in the RSS parser?  Not sure.  Someone should take a look.

RC-Lee:
## Issue This PR Addresses
Fixes #3648 

## Type of Change

## Description
The crash happens here
https://github.com/Seneca-CDOT/telescope/blob/6ad01dc2f13cceca3471ec9671b04ad0f5eccdea/src/api/parser/src/data/post.js#L61
and since we're just throwing an error here
https://github.com/Seneca-CDOT/telescope/blob/6ad01dc2f13cceca3471ec9671b04ad0f5eccdea/src/api/parser/src/feed/processor.js#L108-L111
it fails the job in the queue, which gives us the invalidating feed message.

The problem with this feed is that one of its articles does not return a `guid` as a string (as most other `guid` do), and returns an object
```js
guid: { '$': { isPermaLink: 'false' } }
```

I think I'll treat this as a missing a `guid` like here
https://github.com/Seneca-CDOT/telescope/blob/6ad01dc2f13cceca3471ec9671b04ad0f5eccdea/src/api/parser/src/data/post.js#L125-L126

## Checklist


RC-Lee:
Tests will fail until ES is updated.

src/api/parser/src/data/post.js
```diff
@@ -123,7 +123,8 @@ class Post {
     // link is the URL of the post
     if (!article.link) missing.push('link');
     // guid is the unique identifier of the post
-    if (!article.guid) missing.push('guid');
+    // consider missing guid that is returned as an { '$': { isPermaLink: 'false' } } object.
+    if (!article.guid || article.guid.$) missing.push('guid');
```
humphd, line 127: This is fine.  But follow-up question: could we substitute the `link` here instead, since the `guid` will often be the URL of the article?

Issue #3646, opened 2022-09-06
Help users understand Blog vs. Feed URLs in sign-up


humphd:
I was talking with Stefan today about the Telescope sign-up process, and I think we should do more to educate and clarify the difference between a Blog URL (HTML) and a Feed URL (XML).  The sign-up process should do more to guide people as they do this, and use the opportunity to teach them what it means.

Probably doing something like adding a sentence or two that describes the difference in the sign-up UI would be enough.

mnosov622:
I have some issue, when I sign up, it asks for Blog URL.

When I paste this one https://dev.to/mnosov622/front-end-developer-roadmap-579m it says Unable to discover feeds.

humphd:
OK, you've found a bug. Compare the `<head>` of a particular blog post vs. your blog's root:

- blog post: view-source:https://dev.to/mnosov622/front-end-developer-roadmap-579m
- blog root: view-source:https://dev.to/mnosov622

The latter contains the expected feed URL info (i.e., `<link rel="alternate" type="application/rss+xml" title="DEV Community 👩‍💻👨‍💻 RSS Feed" href="https://dev.to/feed/mnosov622" />`), but the former does not.

If you had pasted in https://dev.to/mnosov622 instead of https://dev.to/mnosov622/front-end-developer-roadmap-579m, it would have worked.

Now, we need to make it *always* work, since people won't necessarily understand the difference between:

- a blog post: https://dev.to/mnosov622/front-end-developer-roadmap-579m
- a blog root: https://dev.to/mnosov622
- a blog's feed: https://dev.to/feed/mnosov622

The code for this lives in https://github.com/Seneca-CDOT/telescope/tree/master/src/api/feed-discovery.  A first step would be to write a test case that fails for this particular case, and then figure out a solution that lets it pass.

mnosov622:
Oh, now I get. Thank you !

humphd:
It's really our bug, not yours.  We should fix this so it works in all cases.

humphd:
A secondary problem we could fix in a separate PR is to deal with the comments feed from WordPress, which we never want:

![image](https://user-images.githubusercontent.com/427398/189464326-3a53d451-de5e-42f5-b0f1-cf6d62b1b814.png)

cychu42:
Hi, I can try to add the description for the difference between a Blog URL (HTML) and a Feed URL (XML) for users to see.

humphd:
Yup, go for it @cychu42

PR #3763, opened 2022-11-16
Add clarification to differentiate between the blog URL and the RSS(feed URL) during sign-up

cychu42:
## Issue This PR Addresses
USAGE: #3646 
This PR addresses one of the issues raised in the issue.

## Type of Change

## Description

image
The issue was that some users experience confusion between a blog URL (blog's root HTML) and select the RSS (blog's feed XML) during sign-up. This PR adds 2 sentences to the prompt to clarify the difference, in `src\web\app\src\components\SignUp\Forms\BlogFeeds.tsx`. 

## Steps to test the PR

Test this by going through the sign-up process to view the Blog Feeds page.
During testing, you can use test account `user1` with password `user1pass`.
You can enter any Github account name to get through the page before this page.

## Checklist

### DIFF
--- a/src/web/app/src/components/SignUp/Forms/BlogFeeds.tsx
+++ b/src/web/app/src/components/SignUp/Forms/BlogFeeds.tsx
@@ -7,7 +7,11 @@ const BlogFeeds = () => {
   return (
     <RSSFeeds
       title="Blog and RSS"
-      prompt="Enter your blog URL and select the RSS you want to use in Telescope"
+      prompt="Enter your blog root URL (HTML) and select the blog's RSS feed URL (XML) you want to use in Telescope, such as:"
+      promptExamples={[
+        'Blog root URL: https://dev.to/user',
+        'RSS feed URL: https://dev.to/feed/user',
+      ]}
       buttonText="Validate"
       helperText="Validate your Blog URL"
       noFeedsSelected="Please validate your blog URL"
--- a/src/web/app/src/components/SignUp/Forms/RSSFeeds.tsx
+++ b/src/web/app/src/components/SignUp/Forms/RSSFeeds.tsx
@@ -136,6 +136,10 @@ const useStyles = makeStyles((theme: Theme) =>
       height: '10px',
       color: '#474747',
     },
+    helpTextExample: {
+      fontSize: '.9em',
+      fontWeight: 'normal',
+    },
   })
 );
 
@@ -144,7 +148,8 @@ type RSSFeedsFormProps = {
     selected: 'blogs' | 'channels';
     discovered: 'allBlogs' | 'allChannels';
   };
-  prompt: string;
+  prompt?: string;
+  promptExamples?: Array<string>;
   title: string;
   buttonText: string;
   helperText: string;
@@ -163,6 +168,7 @@ const RSSFeeds = connect<RSSFeedsFormProps, SignUpForm>(
   ({
     feeds: { selected, discovered },
     prompt,
+    promptExamples,
     title,
     buttonText,
     helperText,
@@ -247,7 +253,13 @@ const RSSFeeds = connect<RSSFeedsFormProps, SignUpForm>(
       <div className={classes.root}>
         <div className={classes.container}>
           <h1 className={classes.blogPageTitle}>{title}</h1>
-          <h2 className={classes.helpText}>{prompt}</h2>
+          <h2 className={classes.helpText}>
+            {prompt}
+            {promptExamples?.map((example) => (
+              <div className={classes.helpTextExample}>{example}</div>
+            ))}
+          </h2>
+
           <div className={classes.infoContainer}>
             <div className={classes.inputsContainer}>
               <TextInput


NeilAn99:
I think it might be nice to provide an example of a blog URL, so users who don't know what HTML and XML are know what to provide.

cychu42:
Maybe, but there are many different sites, so I'm not sure it's practical to list them all or useful to list only one.

NeilAn99:
I think maybe one example would be fine, I think most people use dev.to. But just a simple example, something like:

Blog URL: https://dev.to/user
vs
Blog feed URL: https://dev.to/feed/user

TueeNguyen:
Agree with @NeilAn99, let's add examples of blog url and feed url in 2 separate lines

cychu42:
OK, how about this?
image

src/web/app/src/components/SignUp/Forms/RSSFeeds.tsx
```diff
@@ -143,6 +143,8 @@ type RSSFeedsFormProps = {
     discovered: 'allBlogs' | 'allChannels';
   };
   prompt: string;
+  promptExample1: string;
+  promptExample2: string;
```
batunpc, line 147: Instead of `promptExample1` and `promptExample2`, won't it be better to take an array with prop example?

src/web/app/src/components/SignUp/Forms/ChannelFeeds.tsx
```diff
@@ -8,6 +8,8 @@ const ChannelFeeds = () => {
     <RSSFeeds
       title="YouTube and Twitch"
       prompt="OPTIONAL: Enter your YouTube and/or Twitch channel"
+      promptExample1=""
+      promptExample2=""
```
batunpc, line 12: Here making the prop optional and assigning a default value like empty list , ensure that nothing will break.

src/web/app/src/components/SignUp/Forms/RSSFeeds.tsx
```diff
@@ -245,7 +249,13 @@ const RSSFeeds = connect<RSSFeedsFormProps, SignUpForm>(
       <div className={classes.root}>
         <div className={classes.container}>
           <h1 className={classes.blogPageTitle}>{title}</h1>
-          <h2 className={classes.helpText}>{prompt}</h2>
+          <h2 className={classes.helpText}>
+            {prompt}
+            <br />
+            {promptExample1}
```
batunpc, line 255: Having an array will make these dynamic and reusable

batunpc (changes requested):
@cychu42 Your implementation looks good!
I have left some initial review that we could address

cychu42:
@batunpc Good ideas! I made some changes accordingly.
The only thing I didn't do is assigning a default value, because it doesn't look like the existing code does it for other variables of type `RSSFeedsFormProps` , but I did make the `promptExamples` array optional and make sure nothing would render if there's no value assigned. 
For instance, the next page renders using the same type without any example.
image

batunpc (approved):
Thank you @cychu42 , great job 👍

cychu42:
Thank you all for the reviews. I just rebased and squashed the commits.
I guess I'm now waiting for a reviewer with write access to allow merge to happen.

batunpc (approved):

src/web/app/src/components/SignUp/Forms/RSSFeeds.tsx
```diff
@@ -245,7 +247,16 @@ const RSSFeeds = connect<RSSFeedsFormProps, SignUpForm>(
...
         <div className={classes.container}>
           <h1 className={classes.blogPageTitle}>{title}</h1>
-          <h2 className={classes.helpText}>{prompt}</h2>
+          <h2 className={classes.helpText}>
+            {prompt}
+            {promptExamples?.map((example) => (
+              <>
+                <br />
+                {example}
+              </>
```
manekenpix, line 256:
```suggestion
              <div>{example}</div>
```

manekenpix:
Last minute suggestion:

The text looks a bit bloated and hard to read, what about something more simple? maybe a basic but clear indication of what valid and non valid blog URLs are:

image

It doesn't have to be exactly this ^, but maybe something a bit cleaner would be better for users? It could also include links to blog URL vs feed URL explanations or definitions, or Blog vs Post (some people think a post is a blog).
I'm also ok if the team decides that the changes proposed in this PR are the way to go.

cychu42:
What if we use some shorter text instead:
> Enter your blog root URL (HTML) and select the blog's RSS feed URL (XML) you want to use in Telescope, such as:
> Blog root URL: https://dev.to/user
> RSS feed URL: https://dev.to/feed/user

The 2 examples can be smaller texts like you suggested.

manekenpix:
> What if we use some shorter text instead:
> 
> > Enter your blog root URL (HTML) and select the blog's RSS feed URL (XML) you want to use in Telescope, such as:
> > Blog root URL: https://dev.to/user
> > RSS feed URL: https://dev.to/feed/user
> 
> The 2 examples can be smaller texts like you suggested.

Yeah, I think that'd work.

src/web/app/src/components/SignUp/Forms/BlogFeeds.tsx
```diff
@@ -7,7 +7,9 @@ const BlogFeeds = () => {
   return (
     <RSSFeeds
       title="Blog and RSS"
-      prompt="Enter your blog URL and select the RSS you want to use in Telescope"
+      prompt="Enter your blog URL and select the RSS you want to use in Telescope.
+      A blog URL (HTML) should be the root of your blog itself, and Telescope will find the link(s) of blog feed URL (XML) for you to select, such as:"
```
Myrfion, line 11: Really picky thing, but I think you should add `the` before the blog
```suggestion
      A blog URL (HTML) should be the root of your blog itself, and Telescope will find the link(s) of the blog feed URL (XML) for you to select, such as:"
```

cychu42: Thanks, but it won't apply after the new changes.

cychu42:
OK. I updated it. It now looks like this:
image

For the prompt examples, I used the same font size as the `formControlLabel` class (the RSS URL after validation), which is smaller than the prompt.

manekenpix (approved):
LGTM :+1:

PR #3768, opened 2022-11-17
Allow feed URL to be given during sign up

### LINKED ISSUE #3689: Allow a blog OR feed URL to be given during sign-up
In our sign-up process, sometimes people get confused about the HTML vs. RSS URLs for their blog (see #3646 as well).  If a user gives us a feed URL, we should be smart enough to know that it's a feed URL, and use it.  The feed-discovery service should get updated to allow a feed URL to be given, and simply return it back again (i.e., since it's a feed URL already).

Eakam1007:
## Issue This PR Addresses
Resolves #3689 

## Type of Change

## Description

Currently, only a blog URL can be given during signup. Based on this URL, the feed discovery service returns a list of feed URLs.
This has been updated so a feed URL can also be given. The feed discovery service will check if the given URL is a feed URL, and simply return it back.
This URL is checked by performing a get request and comparing the content type header against a list of valid feed URL content types. The list of valid content types is similar to the one used by the `getFeedURLs` method.

- **Currrent:**
  
  image

- **Updated:**
  image

## Steps to test the PR

- Run telescope locally, including the login sso
- Log in using one of the fake user accounts listed [here](https://telescope.cdot.systems/docs/tools-and-technologies/login)
- When prompted, click on sign up
- Follow the prompts until you get asked for a blog URL
- Enter a blog URL and click on 'Validate Blog'. This should list feed URLs
- Enter a blog feed URL and click on 'Validate Blog'. This should list the same feed URL
- Enter an invalid URL. This should show an error

## Checklist

### DIFF
--- a/src/api/feed-discovery/src/util.js
+++ b/src/api/feed-discovery/src/util.js
@@ -20,6 +20,27 @@ const toTwitchFeedUrl = (twitchChannelUrl) => {
   throw new Error('not a Twitch URL');
 };
 
+const isFeedUrl = async (url) => {
+  try {
+    const { statusCode, headers } = await got(url);
+    const contentType = headers['content-type'];
+    const validContentTypes = [
+      'application/xml',
+      'application/rss+xml',
+      'application/atom+xml',
+      'application/x.atom+xml',
+      'application/x-atom+xml',
+      'application/json',
+      'application/json+oembed',
+      'application/xml+oembed',
+    ];
+
+    return statusCode === 200 && validContentTypes.some((ct) => contentType.includes(ct));
+  } catch (err) {
+    return false;
+  }
+};
+
 const getBlogBody = async (blogUrl) => {
   try {
     logger.debug({ blogUrl }, 'Getting blog body');
@@ -101,3 +122,4 @@ module.exports.getBlogBody = getBlogBody;
 module.exports.getFeedUrls = getFeedUrls;
 module.exports.isTwitchUrl = isTwitchUrl;
 module.exports.toTwitchFeedUrl = toTwitchFeedUrl;
+module.exports.isFeedUrl = isFeedUrl;
--- a/src/api/feed-discovery/src/middleware.js
+++ b/src/api/feed-discovery/src/middleware.js
@@ -1,6 +1,6 @@
 const { logger, createError } = require('@senecacdot/satellite');
 
-const { isTwitchUrl, toTwitchFeedUrl, getBlogBody, getFeedUrls } = require('./util');
+const { isTwitchUrl, toTwitchFeedUrl, isFeedUrl, getBlogBody, getFeedUrls } = require('./util');
 
 // A middleware to ensure we get an array
 module.exports.checkForArray = () => {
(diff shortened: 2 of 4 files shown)


src/api/feed-discovery/src/util.js
```diff
@@ -20,6 +20,27 @@ const toTwitchFeedUrl = (twitchChannelUrl) => {
...
+  try {
+    const { statusCode, headers } = await got(url);
+    const contentType = headers['content-type'];
+    const validContentTypes = [
+      'application/xml',
+      'application/rss+xml',
+      'application/atom+xml',
+      'application/x.atom+xml',
+      'application/x-atom+xml',
+      'application/json',
```
humphd, line 33: I'm not sure that the `rss-parser` module we use (see https://github.com/rbren/rss-parser) is capable of parsing all of these types.  We might want to limit it to valid Atom + RSS only.

humphd (changes requested):
Let's add a test to https://github.com/Seneca-CDOT/telescope/blob/master/src/api/feed-discovery/test/router.test.js as well.

Question: if we allow using a feed URL, does that mean we'll break the back-end, since it thinks the feed URL is the blog URL (HTML) when we add the user?

Eakam1007:
> Question: if we allow using a feed URL, does that mean we'll break the back-end, since it thinks the feed URL is the blog URL (HTML) when we add the user?

Oh, I did not think to look into that. I will check what effect it will have.
I suspect it will only affect the places where the blog URL is linked such as when clicking on profiles.

alexsam29:
> Question: if we allow using a feed URL, does that mean we'll break the back-end, since it thinks the feed URL is the blog URL (HTML) when we add the user?

I'm reviewing the code right now and I tried to look at all the places in the backend that uses `blogUrl` to find any places where the backend might break. There were only two files that I could find: `util.js` and `middleware.js`.

In `util.js` the only function that uses `blogUrl` is [getBlogBody](https://github.com/eakam1007/telescope/blob/f25806464b88e7c004f96c12215fa0b21789bd67/src/api/feed-discovery/src/util.js#L44). But that should not be affected by the use of a feed URL since the only other place that calls `getBlogBody` is `discoverFeedUrls` in `middleware.js`. However, if a feed URL is used that function should not be called because the changes on [line 46](https://github.com/eakam1007/telescope/blob/f25806464b88e7c004f96c12215fa0b21789bd67/src/api/feed-discovery/src/middleware.js#L46) of this PR should prevent that.

In `middleware.js`, blogUrl is used to create a new profile in the function [createNewProfile](https://github.com/eakam1007/telescope/blob/f25806464b88e7c004f96c12215fa0b21789bd67/src/api/sso/src/middleware.js#L103).  But I don't believe using a feed URL for `blogUrl` will break anything since this is just used to store user information in the database.

Eakam1007:
Thanks for the info @alexsam29! 
Following up on it, I checked the registration process and found that the blog URl is saved as the html_url field in the feeds model. I could not find anywhere where that field is being used.
I signed up locally with a feed URL and I did not notice any difference. Since the blog feed contains a link to the (html) blog url, clicking on author profiles redirects them correctly.
So, I don't think allowing a feed URl should break the back-end.

humphd (approved):
Great job on this.

PR #3769, opened 2022-11-19
Add search and dependencies to banner buttons

manekenpix:
## Description

This PR adds `search` and `dependencies` to the hero space (top right corner). This way, it'll match what we have in our `navbar` and users won't have to scroll down to force the `navbar` to show up and have access to `search` and/or `dependencies`.

Logged in user:
image

Logged out user:
image

## Steps to test the PR

- Checkout the branch used for this PR
- Run `docker-compose --env-file config/env.development up  --build`
- Assert that the 2 new elements added to the banner section match the style and behaviour of the other elements in the same section.

## Checklist

### DIFF
--- a/src/web/app/src/components/BannerButtons.tsx
+++ b/src/web/app/src/components/BannerButtons.tsx
@@ -1,6 +1,7 @@
 import Link from 'next/link';
 import { useRouter } from 'next/router';
-import { makeStyles, Tooltip, withStyles, Zoom } from '@material-ui/core';
+import { makeStyles, Tooltip, withStyles, Zoom, IconButton, Theme } from '@material-ui/core';
+import SearchIcon from '@material-ui/icons/Search';
 import Button from '@material-ui/core/Button';
 import clsx from 'clsx';
 import useAuth from '../hooks/use-auth';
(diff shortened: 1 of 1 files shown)


SerpentBytes:
That's a nice addition.

alexsam29 (commented):
Hi, just a little nitpick feedback. I feel it would be better to switch around `Dependencies` with `About Us`. Just so it matches the order in the `navbar` where `About` is above `Dependencies`; for consistency.

image

mnosov622 (commented):
Do you think it's better to highlight sign in button, so it will differ from other links ?

image

humphd (dismissed):

manekenpix:
>Hi, just a little nitpick feedback. I feel it would be better to switch around Dependencies with About Us. Just so it matches the order in the navbar where About is above Dependencies; for consistency.

@alexsam29 I don't know, I think `search` and `dependencies` should go together, it makes more sense to me somehow, if anything I think we could change the order in the `navbar`.

> Do you think it's better to highlight sign in button, so it will differ from other links ?

@mnosov622 Maybe, I'm not sure, we can try playing with that in a follow-up PR.

humphd (approved):

PR #3789, opened 2023-01-24
Changing Cookie to be sameSite none

### LINKED ISSUE #3567: Improve error handling of auth callback
Sometimes when I try to sign in on production, I get this:

Screen Shot 2022-04-27 at 8 16 55 AM

I suspect my token has expired.

There's no value in giving a 400 here.  We should redirect to do the full sign-in flow instead.

STR:

1. Sign-up for your account on Telescope (prod)
1. Sign-in
2. Leave your browser tab open over night so that the token can expire (I'm not sure how long this will take)
3. Come back to the tab and refresh, you should be logged out
4. Click sign-in

sfrunza13:
## Issue This PR Addresses

Fixes #3567

## Type of Change

## Description

There is a 400 we sometimes get because the cookie that is sent to us with the post binding from Azure AD is sometimes blocked since it is from a different origin/site. Enabling sameSite 'none' should hopefully fix this. By default sameSite is set to 'Lax' which sometimes block incoming cookies from other sites such as the request AD makes to our callback.

This is all outlined nicely in the following document David found https://web.dev/samesite-cookie-recipes/#unsafe-requests-across-sites. 

>This pattern is used for sites that may redirect the user out to a remote service to perform some operation before returning, for example redirecting to a third-party identity provider. Before the user leaves the site, a cookie is set containing a single use token with the expectation that this token can be checked on the returning request to mitigate [Cross Site Request Forgery (CSRF)](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html) attacks. If that returning request comes via POST then it will be necessary to mark the cookies as `SameSite=None; Secure`.

## Checklist

### DIFF
--- a/src/api/sso/src/index.js
+++ b/src/api/sso/src/index.js
@@ -18,6 +18,7 @@ const service = new Satellite({
         secret: process.env.JWT_SECRET || `telescope-has-many-secrets-${Date.now()}!`,
         resave: false,
         saveUninitialized: false,
+        cookie: { secure: true, sameSite: 'none' },
       })
     );
     app.use(passport.initialize());
--- a/src/api/sso/src/index.js
+++ b/src/api/sso/src/index.js
@@ -18,7 +18,7 @@ const service = new Satellite({
         secret: process.env.JWT_SECRET || `telescope-has-many-secrets-${Date.now()}!`,
         resave: false,
         saveUninitialized: false,
-        cookie: { secure: true, sameSite: 'none' },
+        cookie: process.env.NODE_ENV === 'production' && { secure: true, sameSite: 'none' },
       })
     );
     app.use(passport.initialize());


humphd (approved):
This looks good, but this PR is missing all the context from our discussion last night, including links to research we did.  Could you add that so we have a record?

src/api/sso/src/index.js
```diff
@@ -18,6 +18,7 @@ const service = new Satellite({
         secret: process.env.JWT_SECRET || `telescope-has-many-secrets-${Date.now()}!`,
         resave: false,
         saveUninitialized: false,
+        cookie: { secure: true, sameSite: 'none' },
```
humphd, line 21:
```suggestion
        cookie: process.env.NODE_ENV === 'production' && { secure: true, sameSite: 'none' },
```

humphd (changes requested):

src/api/sso/src/index.js
humphd: How about we only do this in production?

humphd (approved):

PR #3793, opened 2023-01-25
set trust for first proxy when env is prod

### LINKED ISSUE #3792: Trusting the proxy

We could try to add trust for proxy in case this makes a difference with the session cookie.secure set to true since we are using nginx.
```
var app = express()
app.set('trust proxy', 1) // trust first proxy
app.use(session({
...
```

In the [express-session](https://www.npmjs.com/package/express-session) docs they claim this trust should be set if cookie.secure is set to true. This is a part of an ongoing effort to get cookies to send within the post binding part of the Azure AD authentication which I broke with #3789. Could be contributing/considered along with the latest issue #3791.

sfrunza13:
## Issue This PR Addresses
Fixes #3792 

## Type of Change

## Description
I am trying to find a way to make the new cookie settings work, perhaps this might be a step in the right direction.

## Checklist

- [x] **Quality**: This PR builds and passes our npm test and works locally (but it only changes anything in prod so idk if this means anything)

### DIFF
--- a/src/satellite/src/app.js
+++ b/src/satellite/src/app.js
@@ -10,7 +10,7 @@ const { errorHandler } = require('./middleware');
 function createApp(router, options = {}) {
   const app = express();
 
-  if (process.env.NODE_ENV === 'production') app.set('trust proxy', 1);
+  if (process.env.NODE_ENV === 'production') app.set('trust proxy', 2);
 
   app.use(pinoHttp({ logger }));
 
--- a/src/satellite/src/app.js
+++ b/src/satellite/src/app.js
@@ -10,6 +10,8 @@ const { errorHandler } = require('./middleware');
 function createApp(router, options = {}) {
   const app = express();
 
+  if (process.env.NODE_ENV === 'production') app.set('trust proxy', 1);
+
   app.use(pinoHttp({ logger }));
 
   // Allow disabling or passing options to helmet


humphd (changes requested):
From the [docs](https://expressjs.com/en/guide/behind-proxies.html):

> Number: use the address that is at most n number of hops away from the Express application. req.socket.remoteAddress is the first hop, and the rest are looked for in the X-Forwarded-For header from right to left. A value of 0 means that the first untrusted address would be req.socket.remoteAddress, i.e. there is no reverse proxy.When using this setting, it is important to ensure there are not multiple, different-length paths to the Express application such that the client can be less than the configured number of hops away, otherwise it may be possible for the client to provide any value.

We are behind 2 proxies:

1. nginx
2. traefik

I wonder if this should be 2?  I'm not sure.

sfrunza13:
I suppose you're right, it should probably be 2

humphd (approved):
I think this is good.  @manekenpix, do you have any concerns?

manekenpix:
> I think this is good. @manekenpix, do you have any concerns?

@humphd LGTM :+1:

sfrunza13:
Can I squash and merge?

manekenpix:
@sfrunza13 I think this is good to go.

humphd:
@sfrunza13 yeah, go for it

Issue #3794, opened 2023-01-26
SSO broken


sfrunza13:
**What would you like to be added**:

Specify origin and credential true for cors to see if cookies will send.

**Why would you like this to be added**:

Might make the secure option that I added for cookies work

PR #3795, opened 2023-01-26
#3794

sfrunza13:
## Issue This PR Addresses
#3794 

## Type of Change

## Checklist


humphd:
@sfrunza13 this branch has a whole bunch of stuff in it (commits) not related to your fix.  Can you squash/rebase on latest master?

src/api/sso/src/index.js
```diff
@@ -8,6 +8,7 @@ const routes = require('./routes');
 require('./authentication');
 
 const service = new Satellite({
+  cors: { origin: 'https://telescope.cdot.systems', credentials: true },
```
humphd, line 11: The trouble we'll have here is that the origin is different depending on how you run the app: dev, staging, production.

We store this value in an environment variable named `WEB_URL`.  It's different on every deployment.  So here you should use `process.env.WEB_URL`.

However, we don't seem to expose this to our SSO container, see https://github.com/Seneca-CDOT/telescope/blob/master/docker/docker-compose.yml#L74-L96.  You need to add it there.

And in our local dev env file https://github.com/Seneca-CDOT/telescope/blob/master/src/api/sso/env.local, which should mimic what we have in https://github.com/Seneca-CDOT/telescope/blob/master/config/env.development#L31.

There might be other things we need to make this work, but that's a start.

humphd (changes requested):
Great job getting the commits fixed, I know that wasn't fun.

src/api/sso/src/index.js
sfrunza13: I will try this soon, is it ok if this takes a day or two? I am backed up a bit on other school work.

humphd: Yes, leave this for a while, it's not important in the next week or so.

